PR #166, opened 2018-08-08
fix: jsii-pacmak dependencies

eladb:
Private modules are only needed as `devDependencies`.


eladb:
Covered in #167

PR #177, opened 2018-08-14
feat(jsii): Further normalize assembly outputs

### LINKED ISSUE #60: Java "step" builders are not future-compatible
Brought up by @kiiadi - our beautiful Java "step" builder, which enforces required fields before optional fields will cause backwards compatibility issues when a field is changed from required to optional (which is not supposed to be a breaking change).

Furthermore, the current approach will also cause a breaking change if the order of properties change across revisions, which is also unacceptable.

RomainMuller:
Sort member arrays (`properties` and `methods`), to produce fully sorted assemblies:
1. `static` members
2. `immutable` members (only relevant for `properties`)
3. non-`optional` members (only relevant for `properties`)
4. lexicographically sorted

Also, stop emitting step builders for interfaces in Java (they break due to the property order change the above causes). Conveniently, this fixes #60.


packages/jsii-pacmak/test/expected.jsii-calc-lib/java/src/main/java/software/amazon/jsii/tests/calculator/lib/MyFirstStruct.java
```diff
@@ -19,152 +13,151 @@
...
-            public MyFirstStruct build() {
-                MyFirstStruct result = this.instance;
-                this.instance = new Jsii$Pojo();
-                return result;
-            }
+        /**
+         * Builds the configured instance.
+         * @return a new instance of {@link MyFirstStruct}
+         * @throws NullPointerException if any required attribute was not provided
+         */
```
eladb: Cute

packages/jsii-pacmak/test/expected.jsii-calc-lib/java/src/main/java/software/amazon/jsii/tests/calculator/lib/StructWithOnlyOptionals.java
```diff
@@ -16,112 +18,140 @@
...
         return new Builder();
     }
 
     /**
-     * A fluent builder class for {@link StructWithOnlyOptionals}.
+     * A builder for {@link StructWithOnlyOptionals}
      */
-    public static final class Builder {
-        private Jsii$Pojo instance = new Jsii$Pojo();
+    final class Builder {
```
eladb: Shouldn’t it be static as well?

packages/codemaker/package.json
```diff
@@ -5,8 +5,8 @@
   "main": "lib/index.js",
   "types": "lib/index.d.ts",
   "scripts": {
-    "build": "tsc",
-    "watch": "tsc -w",
+    "build": "tsc -b",
```
eladb: What’s `-b`?

packages/jsii-pacmak/test/expected.jsii-calc-lib/java/src/main/java/software/amazon/jsii/tests/calculator/lib/StructWithOnlyOptionals.java
RomainMuller: It's a member of an `interface`, so it is inherently static.

packages/codemaker/package.json
RomainMuller: In full, `--build`. It's the new compiler mode that supports project references.

packages/jsii/tsconfig.json
```diff
@@ -48,10 +48,19 @@
...
+    "**/*.ts"
+  ],
   "exclude": [
-    "test/**/neg.*.ts"
+    "**/*.d.ts",
+    "node_modules",
+    "test"
+  ],
+  "references": [
+    { "path": "../jsii-spec" }
```
eladb: I don't like this one bit. Why are we referencing outside of the module tree?

packages/jsii-pacmak/tsconfig.json
```diff
@@ -48,7 +48,19 @@
...
+    "composite": true                         /* Allow this project to be referenced by others. */
+  },
+  "include": [
+    "**/*.ts"
+  ],
+  "exclude": [
+    "**/*.d.ts",
+    "node_modules"
+  ],
+  "references": [
```
eladb: What happens when someone needs to add a dependency? Do they have to remember to add it here?

packages/jsii/tsconfig.json
RomainMuller: Well, this particular feature has been added in `typescript` *specifically* to improve the  developer experience around monorepos and composite projects... If you're not into the `../` reference, the best alternative is to reference via `./node_modules/jsii-spec` instead?

packages/jsii-pacmak/tsconfig.json
RomainMuller: It should work without being added in `references`. If it's a dependency from the same monorepo (aka you have the `.ts` files available, not just the `.d.ts`), then you'll get sub-optimal compile times.

packages/jsii/tsconfig.json
eladb: `./node_modules` sounds much better for sure. But what happens if the reference directory doesn't have .ts files (i.e. it's not a mono-repo or we split those modules into separate repos)

packages/jsii-pacmak/tsconfig.json
eladb: got it... did you see any improvement?

RomainMuller: It feels faster, but it's a small project so it's hard to tell for sure...

packages/jsii/tsconfig.json
RomainMuller: If there is no `tsconfig.json` in the pointed to directory, `typescript` will not like it. When the projects are "untangled", they need those project references edited out.

packages/jsii-pacmak/tsconfig.json
eladb: TBH feels like a redundant smartassness to me... More configuration, more ways for things to go wrong, harder to understand what's going on... If you are keen on this, be my guest, but in my view if there is no clear benefit, KISS!

RomainMuller: Yeah, so I'll be generating project references for JSII assemblies' managed `tsconfig.json`, but I'm removing the support for the `jsii` monorepo packages (it doesn't work when using references *in-tree*, and the error message... isn't helpful).

eladb: Not sure I understand and it's worrying me. Let's talk about it tomorrow?

RomainMuller: We can talk about this over the PR that I'll submit a bit later that'll create those automatically. No need to make a fuss now :)

eladb: I've already started crying 😢

PR #182, opened 2018-08-16
BREAKING CHANGE: Reduce duplication of information in jsii assemblies

RomainMuller:
* The `namespace` field no longer is prefixed with the `assembly` name
  + As a consequence, it is often blank, and thus skipped from the output (smaller files 🎉)
* There is no `subtypes` property anymore (it used to contain the list of nested types)
* There is no `parenttype` property anymore (if the namespace is a type name, then the type is nested)
* Introduced a new `overrides` field on `Method` and `Property`, indicating the `NamedTypeReference` of the ancestor that defines the member
  + Used to inject `@Override` annotations in Java code
  + Can be leveraged by other code generators that "need to know" when a method overrides (I'm looking at you, `C#`!)

### DIFF
--- a/packages/jsii-dotnet-generator/src/Amazon.JSII.Generator.UnitTests/Class/ClassGeneratorTests.cs
+++ b/packages/jsii-dotnet-generator/src/Amazon.JSII.Generator.UnitTests/Class/ClassGeneratorTests.cs
@@ -12,11 +12,11 @@ namespace Amazon.JSII.Generator.UnitTests.Class
 
         string Render(ClassType classType)
         {
-            Symbols.MapTypeToPackage("myFqn", "myPackage");
-            Symbols.MapNamespace("myNamespace", "MyNamespace");
+            Symbols.MapTypeToPackage("myFqn", classType.Assembly);
+            Symbols.MapNamespace(classType.QualifiedNamespace, "MyNamespace");
             Symbols.MapTypeName("myFqn", "MyClass", JsonModel.Spec.TypeKind.Class);
 
-            ClassGenerator generator = new ClassGenerator("myPackage", classType, Symbols, Namespaces);
+            ClassGenerator generator = new ClassGenerator(classType.Assembly, classType, Symbols, Namespaces);
 
             SyntaxTree syntaxTree = generator.CreateSyntaxTree();
             return syntaxTree.ToString();
@@ -27,11 +27,10 @@ namespace Amazon.JSII.Generator.UnitTests.Class
         {
             ClassType classType = new ClassType
             (
-                "myFqn",
-                "myPackage",
-                "myClass",
-                "myNamespace",
-                false,
+                fullyQualifiedName: "myFqn",
+                assembly: "myPackage",
+                name: "myClass",
+                isAbstract: false,
                 initializer: new Method(true, false, false)
             );
 
@@ -63,11 +62,10 @@ namespace Amazon.JSII.Generator.UnitTests.Class
         {
             ClassType classType = new ClassType
             (
-                "myFqn",
-                "myPackage",
-                "myClass",
-                "myNamespace",
-                true,
+                fullyQualifiedName: "myFqn",
+                assembly: "myPackage",
+                name: "myClass",
+                isAbstract: true,
                 initializer: new Method(true, false, false)
             );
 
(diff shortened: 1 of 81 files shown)


packages/jsii-dotnet-generator/src/Amazon.JSII.Generator.UnitTests/Class/ClassGeneratorTests.cs
```diff
@@ -100,11 +98,10 @@ public void IncludesDocs()
             // Docs are not currently generated as part of the C# code.
             ClassType classType = new ClassType
             (
-                "myFqn",
-                "myPackage",
-                "myClass",
-                "myNamespace",
-                false,
+                fullyQualifiedName: "myFqn",
```
eladb: Much nicer

packages/jsii-pacmak/lib/targets/java.ts
```diff
@@ -531,7 +533,7 @@ class JavaGenerator extends Generator {
         this.code.line(`${access} final static ${propType} ${propName};`);
     }
 
-    private emitProperty(cls: spec.Type, prop: spec.Property, includeGetter = true, overrides = false) {
+    private emitProperty(cls: spec.Type, prop: spec.Property, includeGetter = true, overrides: boolean = !!prop.overrides) {
```
eladb: If you can obtain this from `prop`, why do you need it in the signature? Also I am shocked that you can refer to another argument here...

packages/jsii-pacmak/lib/targets/java.ts
```diff
@@ -1049,18 +1056,21 @@ class JavaGenerator extends Generator {
...
         const [mod, ...name] = fqn.split('.');
         const depMod = this.findModule(mod);
-        const javaPackage = depMod.targets && depMod.targets.java && depMod.targets.java.package;
-        if (!javaPackage) { throw new Error(`The module ${mod} does not have a java.package setting`); }
-
-        // since this type was needed for some reason when we generate this code, we want to make
-        // sure it's module is included as a direct dependency, even if it's not defined as a direct
-        // dependency of this assembly in jsii. this can happen, for example, when we generate
-        // interface proxies and builders which "flatten" the hirarchy.
+        // Make sure any type that's used from a dependency is listed as a direct dependency of the POM.
```
eladb: Instead of "that's used" I would write "which is explicitly referenced"

packages/jsii/lib/compiler.ts
```diff
@@ -178,7 +178,18 @@ export async function compileSources(entrypoint: string,
 
     if (rootModule) {
         await processModule(rootModule, []);
-        addTypeInfo(assm, types);
+        for (const type of types.sort((l, r) => l.fqn.localeCompare(r.fqn))) {
```
rix0rrr: Can you not put this in a method? All other calls in this spot are method calls.

packages/jsii/lib/util.ts
```diff
@@ -8,6 +8,9 @@ export function filterEmpty(_key: any, value: any): any {
     if (typeof value === 'object' && Object.keys(value).length === 0) {
         return undefined;
     }
+    if (!value) {
```
eladb: are we sure we want to omit empty strings, `false` and `0`?

packages/jsii/lib/compiler.ts
```diff
@@ -178,7 +178,18 @@ export async function compileSources(entrypoint: string,
...
+            assm.types = assm.types || {};
+            if (type.namespace) {
+                const ns = `${type.assembly}.${type.namespace}`;
```
eladb: Can this be implemented as a validation function like the ones below?

packages/jsii/lib/compiler.ts
RomainMuller: That's definitely an option. It felt heavy-handed, but I can "restore" `addTypeInfo` and just reduce what it does to just what's needed :)

packages/jsii-pacmak/lib/targets/java.ts
RomainMuller: So I can obtain it from `prop` but in the case where we generate the proxy class for an interface, I want to output `@Override` for everything (since it's supposed to implement the interface). In this case, the `prop` won't tell me what to do.

And yes, you can refer to another argument, because the default value for JS arguments is syntax sugar for prefixing the code block with `if (arg == null) { arg = default; }`. So other arguments are known (but you sometimes need be careful that default assignments are de-sugared in declaration order).

packages/jsii-pacmak/lib/targets/java.ts
RomainMuller: Makes sense. `used` is very broad.

packages/jsii/lib/util.ts
RomainMuller: So:
* There is no `number` field in the assembly specification, so there shouldn't be any `0`.
* All strings that can possibly be blank are also nullable (`namespace`, `Method#name` - since constructor are nullable).
* All `boolean` are also nullable. Also, `undefined == false`, so this one is really a no-brainer.

packages/jsii/lib/compiler.ts
RomainMuller: @rix0rrr  came up with the same suggestion. Great minds think alike?

PR #188, opened 2018-08-17
feat(jsii): Re-implemented jsii to support --watch and produce better error reporting

RomainMuller:
Re-wrote JSII using the TypeScript compiler services API, integrating with the `ts.ProgramBuilder` classes so that incremental build is possible. Additonally, error messages leverage `TypeScript`'s built-in diagnostic formatter, and enable the rendering of code excerpts with error messages.

The error message generation can be further improved by maintaining a map of entity (type, method, property, parameter, ...) to node, which could then be used by the `Validator` class to emit error messages that are tied to the source code. This was omitted for now as the change is already substantial enough.

I have successfully built the [AWS CDK](https://github.com/awslabs/aws-cdk) using this enhanced compiler. Also, it produces almost no changes in the regression tests' output (only change is that it is now able to carry `enum` member documentation).


eladb:
* Rebase
* Fix title of the PR

packages/jsii/lib/compiler.ts
```diff
@@ -1,1573 +1,178 @@
...
+    target: ts.ScriptTarget.ES2018
+};
+
+const LOG = log4js.getLogger('jsii/compiler');
+export const DIAGNOSTICS = 'diagnostics';
+
+export interface CompilerOptions {
+    /** The information about the project to be built */
+    projectInfo: ProjectInfo;
+    /** Whether the compiler should watch for changes or juts compile once */
```
eladb: juts => just

packages/jsii/lib/compiler.ts
```diff
@@ -1,1573 +1,178 @@
...
-            target.docs = target.docs || {};
-            target.docs.comment = comment;
-        }
-
-        function normalize(tagName: string): string {
-            switch (tagName) {
-                // JavaDoc is very particular about it being @return.
-                case 'returns': return 'return';
-                default: return tagName;
+    public async emit(...files: string[]): Promise<EmitResult | never> {
```
eladb: What's the use case for `files`?

packages/jsii/lib/compiler.ts
```diff
@@ -1,1573 +1,178 @@
...
-                targets: jsii.targets,
-                dependencies: jsii.dependencies
-            };
-        }
-
-        // add all types to lookup table.
-        if (jsii.types) {
-            for (const fqn of Object.keys(jsii.types)) {
-                lookup.set(fqn, jsii.types[fqn]);
+const SOURCE_DIRS = new Set(['bin', 'lib', 'test']);
```
eladb: this should probably be configurable (at some point). issue?

packages/jsii/lib/assembler.ts
```diff
@@ -0,0 +1,879 @@
...
+    }
+
+    /**
+     * Attempt emitting the JSII assembly for the program.
+     *
+     * @return the result of the assembly emission.
+     */
+    public async emit(): Promise<EmitResult> {
+        this._diagnostics = [];
+        if (!this.projectInfo.description) {
```
eladb: Wouldn't it make more sense to perform these validations in `ProjectInfo`?

packages/jsii/lib/assembler.ts
```diff
@@ -0,0 +1,879 @@
...
+                             ts.DiagnosticCategory.Suggestion,
+                            'There is not "README.md" file. It is recommended to have one.');
+        }
+
+        this._types = {};
+        this._deferred = [];
+        const mainFile = path.resolve(this.projectInfo.projectRoot, this.projectInfo.types.replace(/\.d\.ts(x?)$/, '.ts$1'));
+        for (const sourceFile of this.program.getSourceFiles().filter(f => !f.isDeclarationFile)) {
+            if (sourceFile.fileName !== mainFile) { continue; }
+            if (LOG.isTraceEnabled()) {
```
eladb: Why do you need to check if trace is enabled? Isn't that just a log level?

packages/jsii/lib/assembler.ts
```diff
@@ -0,0 +1,879 @@
...
+            dependencies: _toDependencies(this.projectInfo.dependencies),
+            bundled: this.projectInfo.bundleDependencies,
+            types: this._types,
+            targets: this.projectInfo.targets,
+            readme,
+            fingerprint: '<TBD>'
+        };
+
+        const validator = new Validator(this.projectInfo, assembly);
+        const validationResult = await validator.emit();
```
eladb: What's the value in making the validator an emitter? What does it emit really? What's the polymorphic benefit?

packages/jsii/lib/assembler.ts
```diff
@@ -0,0 +1,879 @@
...
+        if (!validationResult.emitSkipped) {
```
eladb: I usually prefer the special-case not to be the success path

packages/jsii/lib/assembler.ts
```diff
@@ -0,0 +1,879 @@
...
+            const assemblyPath = path.join(this.projectInfo.projectRoot, '.jsii');
+            LOG.trace(`Emitting assembly: ${colors.blue(assemblyPath)}`);
+            await fs.writeJson(assemblyPath, _fingerprint(assembly), { replacer: utils.filterEmpty, spaces: 2 });
+        }
+
+        // Clearing ``this._types`` to allow contents to be garbage-collected.
+        delete this._types;
```
eladb: Maybe put in `finally` to avoid "remembering" to do this

packages/jsii/lib/assembler.ts
```diff
@@ -0,0 +1,879 @@
...
+        } else if (ts.isModuleDeclaration(node)) {
+            const moduleType = this._typeChecker.getTypeAtLocation(node);
+            if (LOG.isTraceEnabled()) {
+                LOG.trace(`Entering namespace: ${colors.cyan([...namePrefix, moduleType.symbol.name].join('.'))}`);
+            }
+            const allTypes = new Array<spec.Type>();
+            for (const prop of moduleType.getProperties()) {
+                allTypes.push(...await this._visitNode(prop.valueDeclaration, namePrefix.concat(node.name.getText())));
+            }
+            if (LOG.isTraceEnabled()) {
```
eladb: Why do you need to check if trace is enabled?

packages/jsii/lib/compiler.ts
RomainMuller: It's used when running the tests against the negatives (the standard behaviour otherwise globs all the negatives, which isn't what we want). There was an other way around (basically, copy each file to a temporary work-dir with a `package.json` file to run the tests against), but it was a lot more work.

Adding a comment on `@param` to reflect that.

packages/jsii/lib/compiler.ts
RomainMuller: #203

packages/jsii/lib/assembler.ts
RomainMuller: I decided to have `ProjectInfo` remain rather un-opinionated, especially w/r/t optionals. Could make it also emit `diagnostics`, I suppose. I'm going to hold off on this one for now though.

packages/jsii/lib/assembler.ts
RomainMuller: It's a log level. The check is to avoid taking on possibly expensive side-effects (this string interpolates with color & all).

packages/jsii/lib/assembler.ts
RomainMuller: The validator actually further annotates the type specification (for example, it tags the`overridden` members).

packages/jsii/lib/assembler.ts
RomainMuller: Yeah, but it actually forces me to write more code (that is possibly less readable).

packages/jsii/lib/assembler.ts
RomainMuller: Yeah - Actually why wasn't it there already? o_O

packages/jsii/lib/assembler.ts
RomainMuller: Again, its a way to make compilation faster by not building large interpolated strings if we're not going to use them.

packages/jsii/lib/assembler.ts
eladb: Indeed sounds very expensive. Sometimes those loggers allow you to pass in a function which is lazy evaluated only if the level is enabled...

packages/jsii/lib/assembler.ts
eladb: That sounds wrong.

packages/jsii/lib/assembler.ts
RomainMuller: Yeah but that actually makes the syntax of logging heavier. That's a bit annoying.

packages/jsii/lib/assembler.ts
RomainMuller: Yah I think ultimately I want to merge those into the `Assembler` class entirely. Separating made it easier for me to wrap my head around the two classes of concerns I had to deal with here:
1. Understanding the type model that I obtain from the `tsc` API
2. Make sure we generate valid type definitions for JSII standards

But we can "fail faster" if we merge...

packages/jsii/lib/assembler.ts
eladb: wrapping every trace call with an "if" statement is also heavy...

PR #191, opened 2018-08-21
Fix namespace conversion for scoped packages, and output .NET artifacts to dist/dotnet/ rather than just dist/.

mpiroc:
1. `jsii-pacmak`, when searching for local dependencies, looks in `dist/dotnet/`. But it places .NET build output directly in `dist/`. This change fixes that.
2. `jsii-dotnet-generator` previously considered each type's namespace to *implicitly* begin with the assembly's root namespace. At some point, the `jsii` compiler changed to explicitly include the assembly name in each type's namespace. This change fixes that.
3. As a side effect of <span>#</span>2, some generated folders began with `@`, causing the C# compiler to interpret them as response (`.rsp`) files. The fix for <span>#</span>2 also fixes this issue.


mpiroc:
Closing this, as the namespace fix is not necessary--I was testing against `.jsii` files generated by an older version of `jsii`. I will create a new PR with just the output path change.

PR #205, opened 2018-08-29
Add NuGet strong-name/publishing metadata to .NET packages

mpiroc:
Add all relevant metadata needed to sign and publish `Amazon.JSII.JsonModel`, `Amazon.JSII.Generator`, and `Amazon.JSII.Runtime` to [nuget.org](https://www.nuget.org/).

This metadata includes the path to a `key.snk`. This file will be retrieved by a script in a separate pull request that I'll send out shortly. If `key.snk` is not present, it will be ignored and .NET assemblies will not be strong-named.


packages/jsii-dotnet-generator/Directory.Build.props.t.js
```diff
@@ -1,10 +1,9 @@
-const path = require('path');
-const version = require('./package.json').version.replace(/\+.+$/, ''); // omit "+build" postfix;
+const version = require('./package.json').version.replace(/\+.+$/, ''); // omit "+build" suffix
```
eladb: Why do we need to publish the code generator to NuGet? Isn't it just bundled as part of `jsii-pacmak`? I don't see a use case for someone to use it directly...

packages/jsii-dotnet-jsonmodel/NuGet.Metadata.props.t.js
```diff
@@ -0,0 +1,17 @@
+const package = require('./package.json');
+
+process.stdout.write(`<Project>
+  <PropertyGroup>
+    <GeneratePackageOnBuild>True</GeneratePackageOnBuild>
+    <PackageOutputPath>..\\..\\bin\\$(Configuration)\\NuGet\\</PackageOutputPath>
```
eladb: Will the backslash be a problem on UNIX systems?

mpiroc: Nope, tested on WSL and `superchain`.

packages/jsii-dotnet-generator/Directory.Build.props.t.js
mpiroc: It isn't necessary, I just included it for completeness. Will remove.

Issue #175, opened 2018-08-14
Java interface proxies do not respect optional arguments in methods


eladb:
They should convert optional members into method overloads.

Repro:

```ts
export interface InterfaceWithOptional {
  myMethod(arg1: string, optionalArg2: number?);
}
```

Package to java and the generated interface will include two methods (one with a single argument and one with two) but the proxy will only contain a single method. Compilation will fail.

RomainMuller:
This will likely be true for constructors, as well.

eladb:
Constructors [are covered](https://github.com/awslabs/jsii/blob/master/packages/jsii-pacmak/lib/generator.ts#L180) by [`generateOverloadsForMethodsWithOptionals`](https://github.com/awslabs/jsii/blob/46f17714f22fde54f4add209141daddd2dadf0ad/packages/jsii-pacmak/lib/generator.ts#L27)

PR #208, opened 2018-08-31
fix: Missing types in JSII assembly, invalid Java code, confusing docs

RomainMuller:
* The new `jsii` would omit interfaces defined within namespaces because of the way namespaces were processed (using members instead of listing all exports).
* Some `jsii` type validations could, in some rare cases, happen too early, attempting to dereference types that hadn't been processed yet.
* The `jsii-pacmak` generator for Java could generate property names that were reserved words (e.g: `assert`).
* the `jsii-pacmak` generator for Sphinx would generate confusing (or incorrect) type documentation for entities of array types, and particularly so for arrays of unions.
* Fix for #175: interface proxies do not respect optional method arguments

Should probably add (I don't believe I'll be able to do this before I go on holidays 😅):
+ [x] Test surface in `jsii-calc` or its dependencies that exercise the interface-in-namespace.
+ [ ] Test surface in `jsii-calc` or its dependencies that exercise reserved words of various languages.


packages/jsii-pacmak/lib/targets/java.ts
```diff
@@ -754,20 +754,20 @@ class JavaGenerator extends Generator {
         for (const prop of props) {
             if (prop.optional) { this.code.line(JSR305_NULLABLE); }
             // tslint:disable-next-line:max-line-length
-            this.code.line(`private${prop.immutable ? ' final' : ''} ${prop.fieldJavaType} ${prop.fieldName} = ${_validateIfNonOptional(`_${prop.fieldName}`, prop)};`);
+            this.code.line(`private${prop.immutable ? ' final' : ''} ${prop.fieldJavaType} $${prop.fieldName} = ${_validateIfNonOptional(`_${prop.fieldName}`, prop)};`);
```
eladb: I think `fieldName` should include the "$" prefix so you don't have to repeat this.
Also, I would use an underscore, which I believe is more idiomatic, no?

packages/jsii/lib/assembler.ts
```diff
@@ -391,18 +391,20 @@ export class Assembler implements Emitter {
...
-                    `Base type of ${jsiiType.fqn} (${jsiiType.base.fqn}) is not a class`);
-            }
+            this._defer(() => {
+                const baseType = this._dereference(jsiiType.base!);
+                if (!baseType) {
+                    this._diagnostic(type.symbol.valueDeclaration,
+                                    ts.DiagnosticCategory.Error,
+                                    `Unable to resolve type ${jsiiType.base!.fqn} (base type of ${jsiiType.fqn})`);
+                } else if (spec.isClassType(baseType)) {
+                    jsiiType.initializer = baseType.initializer;
```
eladb: I am concerned about `defer`. For example, the fact that you only assign the initializer here looks like a potential race condition. How do I know what order these `defer` statements actually execution? Now `initializer` is a moving target!

I am okay with using `defer` for validation (maybe should be renamed to `validate`), but it seems very fragile for mutations of the model

PR #216, opened 2018-09-07
Force generated assemblies to reference jsii-only dependencies.

mpiroc:
When a NuGet package is restored, all of its listed dependencies are restored (whether or not they are actually used). But when a .NET assembly is built, if it does not reference any .NET types from the dependency, the compiled .NET assembly will not reference the dependency .NET assembly.

At runtime, `jsii-dotnet-runtime` loads JSII dependencies by looking through the .NET assembly's references. If the dependency does not appear in references, it won't be found.

This breaks the following scenario:

* A JSII assembly `my-package` depends on `my-dependency`.
* In the TypeScript source, `my-package` requires symbols from `my-dependency`. Therefore the NuGet package `MyPackage` lists `MyDependency` as a dependency.
* However, the generated .NET source for `MyPackage` does not reference any symbols from `MyDependency`.

At runtime, the first time a type from `MyPackage` is used, `jsii-dotnet-runtime` will search for `MyPackage`'s dependencies, and ask `jsii-runtime` to load each of them (i.e. `my-dependency`) before attempting to load `my-package`. But due to the missing assembly reference, it won't be able to find `my-dependency`. So when the javascript code for `my-package` attempts to `require('my-dependency')`, it will fail.

To work around this issue, for each package, we generate a type `MyNamespace.Internal.DependencyResolution.Anchor` that references the `Anchor` from each of its dependencies. This way *all* listed dependencies will show up in the .NET assembly's references, avoiding this issue.

### DIFF
--- a/packages/jsii-dotnet-generator/src/Amazon.JSII.Generator.UnitTests/AssemblyGeneratorTests.cs
+++ b/packages/jsii-dotnet-generator/src/Amazon.JSII.Generator.UnitTests/AssemblyGeneratorTests.cs
@@ -29,6 +29,17 @@ namespace Amazon.JSII.Generator.UnitTests
             return $"{Path.Combine(GetPackageOutputRoot(dotnetPackage), dotnetAssembly)}.csproj";
         }
 
+        static string GetAnchorFilePath(string dotnetPackage, string dotnetNamespace)
+        {
+            string path = GetPackageOutputRoot(dotnetPackage);
+
+            foreach (string token in dotnetNamespace.Split('.')) {
+                path = Path.Combine(path, token);
+            }
+
+            return Path.Combine(path, "Internal", "DependencyResolution", "Anchor.cs");
+        }
+
         static string GetTypeFilePath(string dotnetPackage, string dotnetNamespace, string dotnetType)
         {
             string directory = Path.Combine(GetPackageOutputRoot(dotnetPackage), Path.Combine(dotnetNamespace.Split('.')));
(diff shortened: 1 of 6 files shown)


packages/jsii-dotnet-generator/src/Amazon.JSII.Generator.UnitTests/AssemblyGeneratorTests.cs
```diff
@@ -352,6 +363,115 @@ public void CreatesProjectFileWithDependencies()
             file.Received().WriteAllText(projectFilePath, Arg.Do<string>(actual => PlatformIndependentEqual(expected, actual)));
         }
 
+        [Fact(DisplayName = Prefix + nameof(CreatesAnchorFile))]
+        public void CreatesAnchorFile()
```
costleya: _ CreateAnchorFile

packages/jsii-dotnet-generator/src/Amazon.JSII.Generator.UnitTests/AssemblyGeneratorTests.cs
```diff
@@ -29,6 +29,17 @@ static string GetProjectFilePath(string dotnetPackage, string dotnetAssembly)
...
+            string path = GetPackageOutputRoot(dotnetPackage);
+
+            foreach (string token in dotnetNamespace.Split('.')) {
+                path = Path.Combine(path, token);
+            }
+
+            return Path.Combine(path, "Internal", "DependencyResolution", "Anchor.cs");
+        }
+
         static string GetTypeFilePath(string dotnetPackage, string dotnetNamespace, string dotnetType)
```
costleya: _ private (be obvious)

packages/jsii-dotnet-generator/src/Amazon.JSII.Generator.UnitTests/AssemblyGeneratorTests.cs
costleya: Verb is actually correct for a test.

mpiroc: This method doesn't create an anchor file, it's a test method. So the title means `"Verify that the assembly generator creates the anchor file."`
`won'tfix`

packages/jsii-dotnet-generator/src/Amazon.JSII.Generator.UnitTests/AssemblyGeneratorTests.cs
mpiroc: I'm find with adding explicit access modifiers where they are currently implied, but currently the codebase omits all redundant access modifiers. If we change this, we should change it across the whole codebase.

Created #217 to track this.

PR #222, opened 2018-09-12
Fix #221: improve diagnostics when dereference fails

### LINKED ISSUE #221: Improve diagnostics when a type could not be referenced
In some cases, when a type could not be referenced, the diagnostic message doesn't indicate that (the `undefined` return value will still cause an error, but we also want to know that we couldn't deref the type)

eladb:
In some cases, when a type could not be referenced, the diagnostic 
message doesn't indicate that (the undefined return value will still 
cause an error, but we also want to know that we couldn't deref the 
type)

### DIFF
--- a/packages/jsii/lib/assembler.ts
+++ b/packages/jsii/lib/assembler.ts
@@ -321,10 +321,11 @@ export class Assembler implements Emitter {
                 continue;
             }
             this._defer(() => {
-                if (!spec.isClassType(this._dereference(ref, base.symbol.valueDeclaration))) {
+                const deref = this._dereference(ref, base.symbol.valueDeclaration);
+                if (deref && !spec.isClassType(deref)) {
                     this._diagnostic(base.symbol.valueDeclaration,
                                     ts.DiagnosticCategory.Error,
-                                    `Base type of ${jsiiType.fqn} is not a class or cannot be dereferenced (${spec.describeTypeReference(ref)})`);
+                                    `Base type of ${jsiiType.fqn} is not a class (${spec.describeTypeReference(ref)})`);
                 }
             });
             jsiiType.base = ref;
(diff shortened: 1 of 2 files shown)


packages/jsii/lib/assembler.ts
```diff
@@ -314,10 +321,10 @@ export class Assembler implements Emitter {
                 continue;
             }
             this._defer(() => {
-                if (!spec.isClassType(this._dereference(ref))) {
+                if (!spec.isClassType(this._dereference(ref, base.symbol.valueDeclaration))) {
                     this._diagnostic(base.symbol.valueDeclaration,
                                     ts.DiagnosticCategory.Error,
-                                    `Base type of ${jsiiType.fqn} is not a class (${spec.describeTypeReference(ref)})`);
+                                    `Base type of ${jsiiType.fqn} is not a class or cannot be dereferenced (${spec.describeTypeReference(ref)})`);
```
rix0rrr: Make this 'is not exported', that's more in line with what people are thinking about when writing code.

packages/jsii/lib/assembler.ts
```diff
@@ -499,12 +506,14 @@ export class Assembler implements Emitter {
...
-                if (!spec.isInterfaceType(this._dereference(ref))) {
-                    const baseType = this._dereference(ref);
+                const baseType = this._dereference(ref, base.symbol.valueDeclaration);
+                if (!spec.isInterfaceType(baseType)) {
                     if (baseType) {
+                        // tslint:disable:max-line-length
                         this._diagnostic(base.symbol.valueDeclaration,
                                         ts.DiagnosticCategory.Error,
-                                        `Base type of ${jsiiType.fqn} is not an interface (${baseType.kind} ${spec.describeTypeReference(ref)})`);
+                                        `Base type of ${jsiiType.fqn} is not an interface or cannot be dereferenced (${baseType.kind} ${spec.describeTypeReference(ref)})`);
```
rix0rrr: Same here.

packages/jsii/lib/assembler.ts
rix0rrr: Better yet, suppress the second error message by checking that result of _dereference is not `undefined`.

eladb: ✔️

packages/jsii/lib/assembler.ts
eladb: ✔️

eladb:
@rix0rrr fixed

PR #234, opened 2018-09-17
fix(jsii): Defaulted parameters were not rendered as optional

### LINKED ISSUE #233: Java: method overloads are not generated (regression)
Oddly, method overloads are no longer generated (see CDK `Bucket` constructor, and `Bucket.grantRead` examples)

RomainMuller:
Parameters with a default value were not represented as optional in the assembly
documents due to an oversight when re-writing `jsii`. There was also no coverage
in the test corpus for this use-case, so I've added some.

Fixes #233

### DIFF
--- a/packages/jsii-calc/lib/compliance.ts
+++ b/packages/jsii-calc/lib/compliance.ts
@@ -240,6 +240,10 @@ export class RuntimeTypeChecking {
     public methodWithOptionalArguments(arg1: number, arg2: string, arg3?: Date) {
         arg1; arg2; arg3;
     }
+
+    public methodWithDefaultedArguments(arg1: number = 2, arg2: string, arg3: Date = new Date()) {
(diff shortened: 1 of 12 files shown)


packages/jsii-calc/lib/compliance.ts
```diff
@@ -240,6 +240,10 @@ export class RuntimeTypeChecking {
     public methodWithOptionalArguments(arg1: number, arg2: string, arg3?: Date) {
         arg1; arg2; arg3;
     }
+
+    public methodWithDefaultedArguments(arg1: number = 2, arg2: string, arg3: Date = new Date()) {
```
eladb: Can you add a test to verify this for ctors as well?

RomainMuller: Done!

PR #238, opened 2018-09-18
feat: Document overriden/inherited members

### LINKED ISSUE #196: Documentation doesn't mention inherited members
The documentation generated by `jsii-pacmak` does not include references to inherited members that were not overridden. This makes the documentation appear a little confusing and makes discovery of a full API surface more complicated.

For example, the `s3.Bucket` class inherits most of it's interesting members from `s3.BucketRef`, but this is not entirely obvious (particularly to a newcomer). The documentation should include inherited members as part of a class' documentation in one way or another.

RomainMuller:
Add a documentation line outlining the parent declaration of overridden
members as well as inherited members that are not overridden locally.
This should make browsing the documentation a lot nicer.

Inherited and overridden statements in the documentation always refer to
the fully qualified name of the super statement.

Fixes #196

### DIFF
--- a/packages/jsii-pacmak/test/expected.jsii-calc-base/sphinx/_scope_jsii-calc-base.rst
+++ b/packages/jsii-pacmak/test/expected.jsii-calc-base/sphinx/_scope_jsii-calc-base.rst
@@ -196,3 +196,10 @@ BaseProps (interface)
       :type: string *(abstract)*
 
 
+   .. py:attribute:: foo
+
+      *Inherited from* :py:attr:`@scope/jsii-calc-base-of-base.VeryBaseProps.foo`
+
+      :type: :py:class:`@scope/jsii-calc-base-of-base.Very`\  *(abstract)*
+
+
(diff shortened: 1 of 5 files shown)


packages/jsii-pacmak/test/expected.jsii-calc-base/sphinx/_scope_jsii-calc-base.rst
```diff
@@ -196,3 +196,10 @@ BaseProps (interface)
       :type: string *(abstract)*
 
 
+   .. py:attribute:: foo
+
+      *Inherited from* :py:attr:`@scope/jsii-calc-base-of-base.VeryBaseProps.foo`
```
rix0rrr: Is it not more correct to say "inherited from <class>" instead of "inherited from <method>" ?

RomainMuller: But... but they're the same string...

rix0rrr: Oh well. I wrote something between `<` and `>` but got eaten by the MarkDown rendered. Shoulda guessed.

I meant to say, "inherited from $CLASS" seems more correct than "inherited from $METHOD".

Fuuuuu--- that gif is distracting!

RomainMuller: That's what I kinda guessed when I saw your first comment. I wanted the link to be to the parent class' method/property, and not to the class itself... But what I can do is set a custom label on the link, so it'll be the best of both.

PR #241, opened 2018-09-21
fix: dotnet abstract classes should have proxy implementations

### LINKED ISSUE #223: Unable to instantiate object when return type is abstract (.NET)
Clone of #220 for .NET

costleya:
Abstract classes have can implementations that are not shared through jsii. When trying to instantiate one for these types as a return value, the runtime crashes. This change creates proxy class types for abstract classes. These types can be instantiated, and have implementations for any methods/properties marked as abstract.

Fixes #223

### DIFF
--- a/packages/jsii-dotnet-generator/src/Amazon.JSII.Generator.UnitTests/Class/AbstractClassProxyGeneratorTests.cs
+++ b/packages/jsii-dotnet-generator/src/Amazon.JSII.Generator.UnitTests/Class/AbstractClassProxyGeneratorTests.cs
@@ -1,0 +1,188 @@
+using Amazon.JSII.Generator.Class;
+using Amazon.JSII.JsonModel.Spec;
+using Xunit;
+
+namespace Amazon.JSII.Generator.UnitTests.Class
+{
+    public class AbstractClassProxyGeneratorTests : GeneratorTestBase
+    {
+        private const string Prefix = nameof(Generator) + "." + nameof(AbstractClassProxyGenerator) + ".";
+
+        private string Render(ClassType classType)
+        {
+            Symbols.MapTypeToPackage("myFqn", classType.Assembly);
+            Symbols.MapNamespace(classType.QualifiedNamespace, "MyNamespace");
+            Symbols.MapTypeName("myFqn", "MyClass", TypeKind.Class);
+
+            var generator = new AbstractClassProxyGenerator(classType.Assembly, classType, Symbols, Namespaces);
+
+            var syntaxTree = generator.CreateSyntaxTree();
+            return syntaxTree.ToString();
+        }
+
+        [Fact(DisplayName = Prefix + nameof(IncludesAttribute))]
+        public void IncludesAttribute()
+        {
+            var classType = new ClassType
+            (
+                "myFqn",
+                "myPackage",
+                "myClass",
+                true,
+                initializer: new Method(true, false, false)
+            );
+
+            var actual = Render(classType);
+            var expected =
+                @"namespace MyNamespace
+{
+    [JsiiInterfaceProxy(typeof(MyClass), ""myFqn"")]
+    internal class MyClassProxy : MyClass
+    {
+        private MyClassProxy(ByRefValue reference): base(reference)
+        {
+        }
+    }
+}";
+            Assert.Equal(expected, actual, ignoreLineEndingDifferences: true);
+        }
+
+        [Fact(DisplayName = Prefix + nameof(IncludesDocs))]
+        public void IncludesDocs()
+        {
+            // Docs are not currently generated as part of the C# code.
+            ClassType classType = new ClassType
+            (
+                fullyQualifiedName: "myFqn",
+                assembly: "myPackage",
+                name: "myClass",
+                isAbstract: true,
+                initializer: new Method(true, false, false),
+                docs: new Docs {{"foo", "bar"}}
+            );
+
+            string actual = Render(classType);
+            string expected =
+                @"namespace MyNamespace
+{
+    /// <remarks>foo: bar</remarks>
+    [JsiiInterfaceProxy(typeof(MyClass), ""myFqn"")]
+    internal class MyClassProxy : MyClass
+    {
+        private MyClassProxy(ByRefValue reference): base(reference)
+        {
+        }
+    }
+}";
+            Assert.Equal(expected, actual, ignoreLineEndingDifferences: true);
+        }
+
+        [Fact(DisplayName = Prefix + nameof(IncludesProperties))]
+        public void IncludesProperties()
+        {
+            ClassType classType = new ClassType
+            (
+                fullyQualifiedName: "myFqn",
+                assembly: "myPackage",
+                name: "myClass",
+                isAbstract: true,
+                initializer: new Method(true, false, false),
+                properties: new[]
+                {
+                    new Property
+                    (
+                        name: "myProp",
+                        type: new TypeReference("myPropTypeFqn"),
+                        isImmutable: false,
+                        isAbstract: true,
+                        isProtected: false
+                    ),
+                    new Property
+                    (
+                        name: "notMyProp",
+                        type: new TypeReference("myPropTypeFqn"),
+                        isImmutable: false,
+                        isAbstract: false,
+                        isProtected: false
+                    )
+                }
+            );
+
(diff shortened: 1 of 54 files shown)


eladb:
Please make sure you add the conformance test [returnAbstract](
https://github.com/awslabs/jsii/blob/04cb6bb77d2f7f2a77dfe6a071c138028ec9e0e4/packages/jsii-java-runtime-test/project/src/test/java/software/amazon/jsii/testing/ComplianceTest.java#L826)

It would be nice if we could find a way to enforce that all language bindings implement all the conformance tests...

packages/jsii-dotnet-generator/src/Amazon.JSII.Generator.UnitTests/Class/AbstractClassProxyGeneratorTests.cs
```diff
@@ -0,0 +1,188 @@
...
+            var syntaxTree = generator.CreateSyntaxTree();
+            return syntaxTree.ToString();
+        }
+
+        [Fact(DisplayName = Prefix + nameof(IncludesAttribute))]
+        public void IncludesAttribute()
+        {
+            var classType = new ClassType
+            (
+                "myFqn",
```
mpiroc: I've found that these tests are much less brittle when I use named parameters for all api/spec types. This has several benefits:

1. Less of a false positive or false negative when the api/spec changes.
2. Less work to update tests when the spec changes.
3. Robust against changes to the ordering of parameters, especially adjacent parameters of the same type.

packages/jsii-dotnet-generator/src/Amazon.JSII.Generator/TypeProxyGeneratorBase.cs
```diff
@@ -0,0 +1,101 @@
...
+            {
+                var typeOfExpression = SF.TypeOfExpression(Symbols.GetNameSyntax(Type));
+                var fullyQualifiedNameLiteral = SF.Literal(Type.FullyQualifiedName);
+
+                return SF.List(new[]
+                {
+                    SF.AttributeList(SF.SeparatedList(new[]
+                    {
+                        SF.Attribute(
+                            SF.ParseName("JsiiInterfaceProxy"),
```
mpiroc: If you're re-using `JsiiInterfaceProxyAttribute` for abstract classes, it needs a better name. Maybe something like `JsiiTypeProxy`?

packages/jsii-dotnet-generator/src/Amazon.JSII.Generator/TypeProxyGeneratorBase.cs
```diff
@@ -0,0 +1,101 @@
...
+                (
+                    SyntaxKind.BaseConstructorInitializer,
+                    SF.ParseArgumentList("(reference)")
+                ),
+                SF.Block(),
+                null
+            );
+        }
+
+        private bool IsImplementingInterface => Type is InterfaceType;
```
mpiroc: Nit: Property name is confusing, and no need to add a property for this. Just inline `if (Type is InterfaceType)` in `CreateType.CreateBaseList`.

packages/jsii-dotnet-runtime/src/Amazon.JSII.Runtime/Services/ReferenceMap.cs
```diff
@@ -46,8 +48,9 @@ ConstructorInfo GetByRefConstructor()
                 Type type = _types.GetClassType(byRefValue.FullyQualifiedName);
                 if (type == null)
```
mpiroc: Just make this:

```
if (type == null || type.IsAbstract) {
    ...
}
```

And remove the check later in the method.

packages/jsii-dotnet-runtime/src/Amazon.JSII.Runtime/Services/ReferenceMap.cs
```diff
@@ -58,8 +61,14 @@ ConstructorInfo GetByRefConstructor()
 
                 BindingFlags constructorFlags = BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic;
 
-                return type.GetConstructor(constructorFlags, null, new[] { typeof(ByRefValue) }, null);
+                // Get proxy class implementation for abstract types.
+                if (type.IsClass && type.IsAbstract)
```
mpiroc: We already know that the type is a `class`, because `GetClassType` did not return null.

packages/jsii-dotnet-generator/src/Amazon.JSII.Generator/TypeProxyGeneratorBase.cs
```diff
@@ -0,0 +1,101 @@
...
+
+        protected virtual IEnumerable<MemberDeclarationSyntax> CreateConstructors()
+        {
+            yield return SF.ConstructorDeclaration
+            (
+                SF.List<AttributeListSyntax>(),
+
+                // Only Amazon.JSII.Runtime should create interface proxies,
+                // so we make the constructor private.
+                SF.TokenList(SF.Token(SyntaxKind.PrivateKeyword)),
```
mpiroc: Should we make proxies `sealed` as well?

packages/jsii-dotnet-generator/src/Amazon.JSII.Generator/Class/AbstractClassProxyGenerator.cs
```diff
@@ -0,0 +1,164 @@
...
+                            methods));
+                    }
+                }
+
+                return methods;
+            }
+
+            // Only get the first declaration encountered, and keep it if it is abstract.
+            return GetAllMethodsRecurse(type, Enumerable.Empty<Method>())
+                .GroupBy(m => (m.Name,
```
mpiroc: What's the point of this `GroupBy`? Can you add a comment explaining?

costleya: I can add a comment.

packages/jsii-dotnet-generator/src/Amazon.JSII.Generator/TypeProxyGeneratorBase.cs
costleya: I was following the style some other classes were using. I can make this change for now though.

packages/jsii-dotnet-runtime/src/Amazon.JSII.Runtime/Services/ReferenceMap.cs
costleya: I agree that we already know it is a class, since if it was an interface it was already converted to a proxy. Will remove redundant check.

packages/jsii-dotnet-runtime/src/Amazon.JSII.Runtime/Services/ReferenceMap.cs
costleya: Good catch. Will change.

packages/jsii-dotnet-generator/src/Amazon.JSII.Generator/TypeProxyGeneratorBase.cs
costleya: Sure

packages/jsii-dotnet-generator/src/Amazon.JSII.Generator.UnitTests/Class/AbstractClassProxyGeneratorTests.cs
costleya: Sounds good, but this is a good bit of work somewhat unrelated to this change. I will create a new issue to track this. https://github.com/awslabs/jsii/issues/242

packages/jsii-dotnet-generator/src/Amazon.JSII.Generator/Class/AbstractClassProxyGenerator.cs
costleya: Actually, I already did. The list contains ALL methods and properties encountered, in the order encountered. An abstract class can have concrete implementations. Therefore, we only generate methods/properties if the first member encountered is unimplemented.

costleya:
@eladb can you approve this and merge when you get the chance if there are no more issues?

indikaudagedara:
Has this already been released as nuget?

PR #247, opened 2018-09-26
fix(jsii): better usage reporting of private types

rix0rrr:
Anonymous types that are deduced by TypeScript get reported as:

    Cannot use private type __object in exported declarations

But the usage location would not be reported, because the field
'valueDeclaration' of the Symbol would be empty. Looking into the
.declarations member does give a declaration in that case, so
we now use that as a fallback so we can show exactly where that
type is referenced.


rix0rrr:
Just tried, negative testing does not include testing that an error location is reported.

PR #249, opened 2018-09-27
fix(kernel): Return object literals as references

### LINKED ISSUE #248: Kernel returns copy of object literals, preventing updates from applying
As reported in awslabs/aws-cdk#774, the `jsii` kernel returns copies of object literals when coalescing them to a particular named type (typically an interface). This makes updates to interface-typed properties not reflect to the object they were obtained from.

A CDK example is when reading `options` from an `@aws-cdk/cdk.Resource` instance, and attempting to update it.

RomainMuller:
Use the javascript `Proxy` class to coalesce object literals to an
interface type, allowing them to be returned by reference instead of by
value.

Introduces a test in the `jsii-kernel` to demonstrate the feature works
as intended.

Fixes #248
Fixes awslabs/aws-cdk#774

### DIFF
--- a/packages/jsii-kernel/lib/kernel.ts
+++ b/packages/jsii-kernel/lib/kernel.ts
@@ -14,6 +14,7 @@ import { TOKEN_DATE, TOKEN_ENUM, TOKEN_REF } from './api';
  */
 const OBJID_PROP = '$__jsii__objid__$';
 const FQN_PROP = '$__jsii__fqn__$';
+const PROXIES_PROP = '$__jsii__proxies__$';
 
 /**
  * A special FQN that can be used to create empty javascript objects.
@@ -923,12 +924,17 @@ export class Kernel {
         // so the client receives a real object.
         if (typeof(v) === 'object' && targetType && spec.isNamedTypeReference(targetType)) {
             this._debug('coalescing to', targetType);
-            const newObjRef = this._create({ fqn: targetType.fqn });
-            const newObj = this._findObject(newObjRef);
-            for (const k of Object.keys(v)) {
-                newObj[k] = v[k];
-            }
-            return newObjRef;
+            const proxies = v[PROXIES_PROP] = v[PROXIES_PROP] || {};
+            if (proxies[targetType.fqn]) { return proxies[targetType.fqn]; }
+            const proxy = new Proxy(v, {
+                has(target: any, propertyKey: string) {
+                    return propertyKey in target || propertyKey === FQN_PROP;
+                },
+                get(target: any, propertyKey: string) {
+                    return propertyKey === FQN_PROP ? targetType.fqn : target[propertyKey];
+                }
+            });
+            return this._createObjref(proxy, targetType.fqn);
         }
 
         // date (https://stackoverflow.com/a/643827/737957)
--- a/packages/jsii-kernel/lib/kernel.ts
+++ b/packages/jsii-kernel/lib/kernel.ts
@@ -930,10 +930,17 @@ export class Kernel {
         // so the client receives a real object.
         if (typeof(v) === 'object' && targetType && spec.isNamedTypeReference(targetType)) {
             this._debug('coalescing to', targetType);
+            /*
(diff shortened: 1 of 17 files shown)


packages/jsii-kernel/lib/kernel.ts
```diff
@@ -923,12 +924,17 @@ export class Kernel {
...
-            const newObj = this._findObject(newObjRef);
-            for (const k of Object.keys(v)) {
-                newObj[k] = v[k];
-            }
-            return newObjRef;
+            const proxies = v[PROXIES_PROP] = v[PROXIES_PROP] || {};
+            if (proxies[targetType.fqn]) { return proxies[targetType.fqn]; }
+            const proxy = new Proxy(v, {
+                has(target: any, propertyKey: string) {
+                    return propertyKey in target || propertyKey === FQN_PROP;
```
eladb: What about `OBJID_PROP`?

packages/jsii-kernel/lib/kernel.ts
```diff
@@ -923,12 +924,17 @@ export class Kernel {
         // so the client receives a real object.
         if (typeof(v) === 'object' && targetType && spec.isNamedTypeReference(targetType)) {
             this._debug('coalescing to', targetType);
-            const newObjRef = this._create({ fqn: targetType.fqn });
-            const newObj = this._findObject(newObjRef);
-            for (const k of Object.keys(v)) {
-                newObj[k] = v[k];
-            }
-            return newObjRef;
+            const proxies = v[PROXIES_PROP] = v[PROXIES_PROP] || {};
```
eladb: Why not just call `this._createObjref(target, targetType.fqn)`. Wouldn't that achieve the same goal?

rix0rrr:
Does this apply to all objects? What about objects we're passing "as data" in data-typed interfaces?

packages/jsii-kernel/lib/kernel.ts
RomainMuller: The same object literal may be returned under different type identities, so you cannot issue an object proxy that has only one type & expect it to "always work".

RomainMuller:
@rix0rrr I don't reckon there is any such thing as *as data* per se. Data-types are still returned as object references as far as I know.

packages/jsii-kernel/lib/kernel.ts
eladb: Got it, so the behavior of `_createObjref` doesn't make sense in that respect because it adds FQN_PROP and OBJID_PROP to the target, and as you say, the same object can represent two different FQNs. I am wondering if in this case it should still represent the same OBJID, which also contains the FQN...

RomainMuller: What FQN, though? I like the idea of being able to return a single OBJID for the object literal (because it would make it trivial to implement identity checks, or even de-duplication on the front-ends)... However, consider the following (valid) typescript:

```ts
export class TrickyType {
  private readonly _magic: TypeA & TypeB = {};

  public get magicA(): TypeA {
    return this._magic;
  }

  public get magicB(): TypeB {
    return this._magic;
  }
}

export interface TypeA {
  foo?: string;
}

export interface TypeB {
  bar?: string;
}
```

Now, if you want `magicA` and `magicB` to return the same object reference, which of `TypeA` and `TypeB` do you tag in the `ObjRef`, knowing front-end languages will use this to determine which class to make a proxy instance for?

RomainMuller: The path I've taken here has a weaker guarantee that if you call `magicA` several times, you'll get the same object reference each time... But `magicB` will be another. We can make it be a single reference, however it'd require (way) more significant changes (basically, the `ObjRef` shape will need to change so the type information is modeled separately).

eladb: I am okay with that. The current implementation returned a different object reference in every call, so that's definitely an improvement. Worth documenting in code though

RomainMuller: I've added comments... let me know if not clear enough (or too verbose? :D)

PR #254, opened 2018-10-03
 fix(kernel + java): skip overrides of private methods/properties

### LINKED ISSUE #244: java/kernel: overrides should not be created for private methods/properties
Brought up in: awslabs/aws-cdk#757

When jsii objects are created, property/method overrides should not be defined for private members, both on the native side and on the javascript side.

eladb:
jsii-kernel will ignore override requests for methods/properties that
resolve on the object but are not defined in the public API of the
type.

java runtime will not request overrides for methods/properties that
use "private" accessibility.

Fixes #244

### DIFF
--- a/packages/jsii-java-runtime-test/project/src/test/java/software/amazon/jsii/testing/ComplianceTest.java
+++ b/packages/jsii-java-runtime-test/project/src/test/java/software/amazon/jsii/testing/ComplianceTest.java
@@ -11,6 +11,7 @@ import software.amazon.jsii.tests.calculator.AsyncVirtualMethods;
 import software.amazon.jsii.tests.calculator.Calculator;
 import software.amazon.jsii.tests.calculator.CalculatorProps;
 import software.amazon.jsii.tests.calculator.DerivedStruct;
+import software.amazon.jsii.tests.calculator.DoNotOverridePrivates;
 import software.amazon.jsii.tests.calculator.DoubleTrouble;
 import software.amazon.jsii.tests.calculator.GiveMeStructs;
 import software.amazon.jsii.tests.calculator.IFriendlier;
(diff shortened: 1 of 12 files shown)


packages/jsii-java-runtime-test/project/src/test/java/software/amazon/jsii/testing/ComplianceTest.java
```diff
@@ -837,6 +838,73 @@ public void returnAbstract() {
...
+            }
+        };
+
+        assertEquals("privateProperty", obj.privatePropertyValue());
+    }
+
+    @Test
+    public void doNotOverridePrivates_property_getter_private() {
+        DoNotOverridePrivates obj = new DoNotOverridePrivates() {
+            private String getPrivateProperty() {
```
RomainMuller: How 'bout the setters?

packages/jsii-kernel/lib/kernel.ts
```diff
@@ -521,10 +531,16 @@ export class Kernel {
...
     }
 
-    private _applyPropertyOverride(obj: any, objref: api.ObjRef, override: api.Override) {
+    private _applyPropertyOverride(obj: any, objref: api.ObjRef, override: api.Override, propInfo?: spec.Property) {
         const self = this;
         const propertyName = override.property!;
 
+        // if this is a private property (i.e. doesn't have `propInfo` the object has a key)
+        if (!propInfo && propertyName in obj) {
+            this._debug(`Skipping override of private property ${propertyName}`);
```
RomainMuller: Shouldn't this `throw`? I can't help but think this is the symptom of a bug...

packages/jsii-kernel/lib/kernel.ts
```diff
@@ -570,6 +586,13 @@ export class Kernel {
         const self = this;
         const methodName = override.method!;
 
+        // If this is a private method (doesn't have methodInfo, key resolves on the object), we
+        // are going to skip the override.
+        if (!methodInfo && obj[methodName]) {
+            this._debug(`Skipping override of private method ${methodName}`);
```
RomainMuller: Same here...

packages/jsii-kernel/lib/kernel.ts
```diff
@@ -596,6 +619,8 @@ export class Kernel {
                     });
                 }
             });
+        } else if (!methodInfo && Object.keys(obj).includes(methodName)) {
+            throw new Error('boom');
```
RomainMuller: Congratulations - you earned yourself a distracting gif of confusing!

packages/jsii-kernel/lib/kernel.ts
eladb: Not necessarily, there could be case where the user would define a public method in a derived native class with the same name as a private method in a base javascript class. When the native jsii client instantiates the javascript object, it doesn't know that this method is private on the javascript side, do it will pass it as an override (you could argue that the jsii client should only pass overrides for known public methods, but currently this is not the behavior, at least in java).

packages/jsii-kernel/lib/kernel.ts
eladb: Same response.

packages/jsii-kernel/lib/kernel.ts
eladb: Oops. Debug code.

packages/jsii-java-runtime-test/project/src/test/java/software/amazon/jsii/testing/ComplianceTest.java
eladb: Added

eladb:
@RomainMuller let me know if this looks okay

packages/jsii-kernel/lib/kernel.ts
RomainMuller: That makes sense to me. I think the front-end code should be a little more educated about the type model it was generated for... So I reckon this should be a `WARN` and not a `DEBUG`.

PR #262, opened 2018-10-10
chore: add Java test on autoprops from private constructor

rix0rrr:


eladb:
Any chance you could update the [.NET test](ationTests/ComplianceTests.cs#L812) to use the static method, and make sure they have the same name?

PR #265, opened 2018-10-16
fix(jsii): use base interfaces for 'datatype' property

### LINKED ISSUE #264: jsii: interfaces incorrectly marked as datatypes
In the following case, the 

```ts
interface IBaseInterface {
  readonly value: string;
  doSomething(): void;
}

interface IDerivedInterface extends IBaseInterface {
  readonly otherValue: string;
}
```

The interface `IDerivedInterface` is incorrectly marked as `{datatype: true}`, because the interface consists only of properties. Correct behavior would have been to take the parent interfaces into account as well while making this assessment.

This leads the Java code generator to emit a Builder that contains duplicate fields (and setters) and ignoring the methods in the upstream classes (encountered while making an `IRole` interface):

```
@aws-cdk/aws-iam: [ERROR] /tmp/jsii-pacmak-codeUbHgJB/src/main/java/software/amazon/awscdk/services/iam/IRole.java:[26,69] variable _principal is already defined in class software.amazon.awscdk.services.iam.IRole.Builder
@aws-cdk/aws-iam: [ERROR] /tmp/jsii-pacmak-codeUbHgJB/src/main/java/software/amazon/awscdk/services/iam/IRole.java:[52,24] method withPrincipal(software.amazon.awscdk.services.iam.PolicyPrincipal) is already defined in class software.amazon.awscdk.services.iam.IRole.Builder
@aws-cdk/aws-iam: [ERROR] /tmp/jsii-pacmak-codeUbHgJB/src/main/java/software/amazon/awscdk/services/iam/IRole.java:[76,83] variable $principal is already defined in class <anonymous software.amazon.awscdk.services.iam.IRole$Builder$1>
@aws-cdk/aws-iam: [ERROR] /tmp/jsii-pacmak-codeUbHgJB/src/main/java/software/amazon/awscdk/services/iam/IRole.java:[90,76] method getPrincipal() is already defined in class <anonymous software.amazon.awscdk.services.iam.IRole$Builder$1>
@aws-cdk/aws-iam: [ERROR] /tmp/jsii-pacmak-codeUbHgJB/src/main/java/software/amazon/awscdk/services/iam/IRole.java:[73,32] <anonymous software.amazon.awscdk.services.iam.IRole$Builder$1> is not abstract and does not override abstract method attachManagedPolicy(java.lang.String) in software.amazon.awscdk.services.iam.IPrincipal
```

rix0rrr:
If an interface inherits from a non-datatype interface, it should no
longer be classified as a datatype interface itself.

Making this change requires that information about the base classes
has already been determined, so I introduced an ordering mechanism
for 'deferred's.

Fixes #264.

### DIFF
--- a/packages/jsii/lib/assembler.ts
+++ b/packages/jsii/lib/assembler.ts
@@ -11,6 +11,7 @@ import literate = require('./literate');
 import { ProjectInfo } from './project-info';
 import utils = require('./utils');
 import { Validator } from './validator';
+import { NamedTypeReference, isInterfaceType } from 'jsii-spec';
 
 // tslint:disable:no-var-requires Modules without TypeScript definitions
 const sortJson = require('sort-json');
@@ -23,7 +24,7 @@ const LOG = log4js.getLogger('jsii/assembler');
  */
 export class Assembler implements Emitter {
     private _diagnostics = new Array<Diagnostic>();
-    private _deferred = new Array<() => void>();
+    private _deferred = new Array<DeferredRecord>();
     private _types: { [fqn: string]: spec.Type };
 
     /**
@@ -75,9 +76,8 @@ export class Assembler implements Emitter {
                 await this._visitNode(node.declarations[0]);
             }
         }
-        for (const deferred of this._deferred) {
-            deferred();
-        }
+
+        this.callDeferredsInOrder();
 
         // Skip emitting if any diagnostic message is an error
         if (this._diagnostics.find(diag => diag.category === ts.DiagnosticCategory.Error) != null) {
@@ -87,9 +87,8 @@ export class Assembler implements Emitter {
             try {
                 return { diagnostics: this._diagnostics, emitSkipped: true };
             } finally {
-                // Clearing ``this._diagnostics`` and ``this._deferred`` to allow contents to be garbage-collected.
+                // Clearing ``this._diagnostics`` to allow contents to be garbage-collected.
                 delete this._diagnostics;
-                delete this._deferred;
             }
         }
 
@@ -128,9 +127,8 @@ export class Assembler implements Emitter {
             // Clearing ``this._types`` to allow contents to be garbage-collected.
             delete this._types;
 
-            // Clearing ``this._diagnostics`` and ``this._deferred`` to allow contents to be garbage-collected.
+            // Clearing ``this._diagnostics`` to allow contents to be garbage-collected.
             delete this._diagnostics;
-            delete this._deferred;
         }
 
         async function _loadReadme(this: Assembler) {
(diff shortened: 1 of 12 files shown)


packages/jsii/lib/assembler.ts
```diff
@@ -913,3 +983,9 @@ function _toDependencies(assemblies: ReadonlyArray<spec.Assembly>): { [name: str
     }
     return result;
 }
+
+interface DeferredRecord {
+    fqn: string;
+    depFqns: string[];
```
RomainMuller: Some doc comments would be welcome here... Is "dep" for "depended" or "depending" ?

packages/jsii/lib/assembler.ts
```diff
@@ -148,11 +146,38 @@ export class Assembler implements Emitter {
...
+     * Defer a callback until a (set of) types are available
+     *
+     * This is a helper function around _defer() which encapsulates the _dereference
+     * action (which is basically the majority use case for _defer anyway).
+     *
+     * Will not invoke the function with any 'undefined's; an error will already have been emitted in
+     * that case anyway.
+     */
+    // tslint:disable-next-line:max-line-length
+    private _deferUntilTypeAvailable(fqn: string, baseTypes: NamedTypeReference[], referencingNode: ts.Node, cb: (...xs: spec.Type[]) => void) {
```
RomainMuller: Should it not be `_deferUntilType+s+Available`?
I would also like it if the doc had `@param`s.

packages/jsii/lib/assembler.ts
```diff
@@ -148,11 +146,38 @@ export class Assembler implements Emitter {
...
      * have been discovered yet, and verifying properties about them.
      *
+     * @param fqn FQN of the current type.
+     * @param deps List of FQNs of types this callback depends on. All deferreds for all
      * @param cb the function to be called in a deferred way. It will be bound with ``this``, so it can depend on using
      *           ``this``.
      */
-    private _defer(cb: () => void) {
-        this._deferred.push(cb.bind(this));
+    private _defer(fqn: string, depFqns: string[], cb: () => void) {
```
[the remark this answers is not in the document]
rix0rrr: I don't agree that it's sketchy--it's running some analysis in a guaranteed order. I was debating whether to rewrite the whole assembler into two phases, first collect all types and relationships between types, then do the regular processing in order. 

Part of the processing doesn't require the base types though, so I ultimately decided it would be a lower-impact change to leave the standalone analysis in the first pass and only do the analysis that requires base types in the second pass--and the "deferred" mechanism could be extended easily enough to accomodate the ordering required for that second pass.

Now, whether or not it should be inlined is another case. In principle the ordered analysis method is more general than it's being used for right now. It doesn't NEED to work on `NamedTypeReferences` that are `dereferenced`... just so happens that all places where it's being used use that pattern, so I extracted that out again. I don't think it hurts to show there are two layers to it.

PR #272, opened 2018-10-22
fix: require distinct argument and property names

### LINKED ISSUE #268: Reject Props arguments that have a property with the same name as a function parameter
Given a signature like:

```typescript

export interface MyProps {
    fooBar: string;
}

export class MyClass {
    public doAThing(fooBar: string, props: MyProps) {
    }
}
```

That works fine in TypeScript, but in languages where we want to "lift" a final props parameter into arguments to the function itself (like for example, keyword arguments in Python), this will fail because it will generate code like:

```python
class MyClass:
    def do_a_thing(foo_bar: str, *, foo_bar):
        ...
```

Which is an error, because there is a collision on the "real" function parameter names, and the lifted parameter names, and you can't have two parameters with the same name.

Ideally the JSII compiler would reject this method signature, and require one of the two names to change. Alternatively, if there were a way to either opt-in or opt-out of the lifting of ``MyProps`` into keyword arguments, that could also allow people a way to resolve this without requiring them to choose different names, they would just not lift that particular data type into keyword arguments.

rix0rrr:
Restrict the usage of datatype parameters in functions, so that
languages with keyword argument support can inline datatype parameters
into the function declaration. This means it cannot share any field
names with remaining function argument names.

Fixes #268.


eladb:
Can we also annotate the method in the jsii spec that keyword arguments can be expanded?

packages/jsii/lib/assembler.ts
```diff
@@ -628,6 +630,29 @@ export class Assembler implements Emitter {
 
         this._visitDocumentation(symbol, method);
 
+        // If the last parameter is a datatype, verify that it does not share any field names with
+        // other function arguments, so that it can be turned into keyword arguments by jsii frontends
+        // that support such.
+        const lastParamTypeRef = apply(last(parameters), x => x.type);
+        const lastParamSymbol = last(signature.getParameters());
+        if (lastParamTypeRef && spec.isNamedTypeReference(lastParamTypeRef)) {
+            this._deferUntilTypesAvailable(symbol.name, [lastParamTypeRef], lastParamSymbol!.declarations[0], (lastParamType) => {
```
eladb: I am worried about this programming model. I really liked your idea of phases as a safer and more deterministic alternative.

PR #291, opened 2018-11-02
fix(runtime/dotnet): Correct a number of type mappings

### LINKED ISSUE #290: .NET fails mapping objects to interface types when returned from callbacks
When a method that is modeled to return an interface type is overridden in .NET, the callback handler fails to map the object to the JSII type, producing the `Could not infer JSII type for .NET type` error.

This is because the callback code attempts to map the type using the `PrimitiveType.Any` reference type instead of what is modeled on the overridden function.

This was identified as the root cause of awslabs/aws-cdk#1027.

RomainMuller:
The `jsii` compiler incorrectly mapped types homonym with a standard
type (`String`, `Number`, ...) to the primitive type on usage sites.
This commit corrects this behavior and adjusts the C# tests accordingly.

Additionally, when the return type of a callback was an interface, the
C# runtime was unable to determine the correct JSII type to use, despite
it has the information on the type of the callback method available. The
resolution behavior has also been fixes, as well as a couple of other
glitches that became apparent as a result of the `Number` type starting
to be correctly represented as the `@scope/jsii-calc-lib.Number` type.

Fixes #290
Fixes awslabs/aws-cdk#1027

### DIFF
--- a/packages/jsii-dotnet-runtime-test/test/Amazon.JSII.Runtime.IntegrationTests/ComplianceTests.cs
+++ b/packages/jsii-dotnet-runtime-test/test/Amazon.JSII.Runtime.IntegrationTests/ComplianceTests.cs
@@ -83,10 +83,10 @@ namespace Amazon.JSII.Runtime.IntegrationTests
             Assert.Equal("World", types.ArrayProperty[1]);
 
             // map
-            IDictionary<string, double> map = new Dictionary<string, double>();
-            map["Foo"] = 123;
+            IDictionary<string, Number> map = new Dictionary<string, Number>();
+            map["Foo"] = new Number(123);
             types.MapProperty = map;
-            Assert.Equal((double) 123, types.MapProperty["Foo"]);
+            Assert.Equal((double) 123, types.MapProperty["Foo"].Value);
         }
 
         [Fact(DisplayName = Prefix + nameof(DynamicTypes))]
@@ -816,6 +816,43 @@ namespace Amazon.JSII.Runtime.IntegrationTests
             obj.ReadWriteString = "Hello";
 
             Assert.Equal("Hello", obj.ReadOnlyString);
+        }
+
+        [Fact(DisplayName = Prefix + nameof(TestReturnInterfaceFromOverride))]
+        public void TestReturnInterfaceFromOverride()
+        {
+            var n = 1337;
+            var obj = new OverrideReturnsObject();
+            var arg = new NumberReturner(n);
+            Assert.Equal(4 * n, obj.Test(arg));
+        }
+
+        class NumberReturner : DeputyBase, IIReturnsNumber
+        {
+            public NumberReturner(double number)
+            {
+                NumberProp = new Number(number);
+            }
+
+            [JsiiProperty("numberProp", "{\"fqn\":\"@scope/jsii-calc-lib.Number\"}", true)]
+            public Number NumberProp { get; }
+
+            [JsiiMethod("obtainNumber", "{\"fqn\":\"@scope/jsii-calc-lib.IDoublable\"}", "[]",true)]
+            public IIDoublable ObtainNumber()
+            {
+                return new Doublable(this.NumberProp);
+            }
+
+            class Doublable : DeputyBase, IIDoublable
+            {
+                public Doublable(Number number)
+                {
+                    this.DoubleValue = number.DoubleValue;
+                }
+
+                [JsiiProperty("doubleValue","{\"primitive\":\"number\"}",true)]
+                public Double DoubleValue { get; }
+            }
         }
 
         class MulTen : Multiply
--- a/packages/jsii-dotnet-runtime/src/Amazon.JSII.Runtime/CallbackExtensions.cs
+++ b/packages/jsii-dotnet-runtime/src/Amazon.JSII.Runtime/CallbackExtensions.cs
@@ -15,13 +15,12 @@ namespace Amazon.JSII.Runtime
         {
             try
             {
-                TypeReference returnType;
-                object frameworkResult = callback.InvokeCallbackCore(referenceMap, out returnType);
+                CallbackResult frameworkResult = callback.InvokeCallbackCore(referenceMap);
 
                 converter.TryConvert(
-                    returnType ?? new TypeReference(primitive: PrimitiveType.Any),
+                    frameworkResult.Type,
                     referenceMap,
-                    frameworkResult,
+                    frameworkResult.Value,
                     out object result
                 );
 
@@ -42,29 +41,28 @@ namespace Amazon.JSII.Runtime
             }
         }
 
-        static object InvokeCallbackCore(this Callback callback, IReferenceMap referenceMap, out TypeReference returnType)
+        static CallbackResult InvokeCallbackCore(this Callback callback, IReferenceMap referenceMap)
         {
             if (callback.Invoke != null)
             {
-                return InvokeMethod(callback.Invoke, referenceMap, out returnType);
+                return InvokeMethod(callback.Invoke, referenceMap);
             }
 
             if (callback.Get != null)
             {
-                return InvokeGetter(callback.Get, referenceMap, out returnType);
+                return InvokeGetter(callback.Get, referenceMap);
             }
 
             if (callback.Set != null)
             {
                 InvokeSetter(callback.Set, referenceMap);
-                returnType = null;
                 return null;
             }
 
             throw new ArgumentException("Callback does not specificy a method, getter, or setter to invoke");
         }
 
-        static object InvokeMethod(InvokeRequest request, IReferenceMap referenceMap, out TypeReference returnType)
+        static CallbackResult InvokeMethod(InvokeRequest request, IReferenceMap referenceMap)
         {
             request = request ?? throw new ArgumentNullException(nameof(request));
             DeputyBase deputy = referenceMap.GetOrCreateNativeReference(request.ObjectReference);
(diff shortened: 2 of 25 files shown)


packages/jsii-dotnet-runtime/src/Amazon.JSII.Runtime/jsii-runtime/jsii-runtime.js
```diff
@@ -0,0 +1,2 @@
...(null,0);let c=0;const h=n.alloc(512),u=(n,r)=>{if(n)return s(n);if((c+=r)<512&&r)return o.read(t,h,c,h.length-c,a+c,u);if(0===a&&31===h[0]&&139===h[1])return s(new Error("cannot append to compressed archives"));if(c<512)return s(null,a);const f=new l(h);if(!f.cksumValid)return s(null,a);const d=512*Math.ceil(f.size/512);return a+d+512>i?s(null,a):(a+=d+512)>=i?s(null,a):(e.mtimeCache&&e.mtimeCache.set(f.path,f.mtime),c=0,void o.read(t,h,0,512,a,u))};o.read(t,h,0,512,a,u)})(f,h.size,(n,o)=>{if(n)return s(n);const c=new a.WriteStream(e.file,{fd:f,start:o});r.pipe(c),c.on("error",s),c.on("close",i),m(r,t)})});o.open(e.file,c,h)});return i?c.then(i,i):c},p=(e,t)=>{t.forEach(t=>{"@"===t.charAt(0)?c({file:h.resolve(e.cwd,t.substr(1)),sync:!0,noResume:!0,onentry:t=>e.add(t)}):e.add(t)}),e.end()},m=(e,t)=>{for(;t.length;){const i=t.shift();if("@"===i.charAt(0))return c({file:h.resolve(e.cwd,i.substr(1)),noResume:!0,onentry:t=>e.add(t)}).then(i=>m(e,t));e.add(i)}e.end()}},function(e,t,i){"use strict";const n=i(12),r=(i(14).EventEmitter,i(18)),s=i(2),o=i(11),a=i(0),c=i(93),h=(c.sync,i(37)),l=Symbol("onEntry"),u=Symbol("checkFs"),f=Symbol("isReusable"),d=Symbol("makeFs"),p=Symbol("file"),m=S
```
eladb: I don't think this file is supposed to be checked in

packages/jsii-dotnet-runtime/src/Amazon.JSII.Runtime/jsii-runtime/jsii-runtime.js.map
```diff
@@ -0,0 +1 @@
...AAzB,GACA,SAAAqY,EAAAC,EAAAtY,GACA,GAAAA,GAAA,iBAAAA,EAEA,GAAAA,EAAAuY,KAAA,CACA,IAAAC,EAAAT,EAAApZ,QAAA2Z,EAAAtY,EAAAuY,MACAJ,EAAAK,GAAAL,EAAAK,GAAAL,EAAAK,GAAA,QAFA,CAKA,IAAAC,EAAAzY,EAAAG,GAAA4X,EAAApZ,QAAA2Z,EAAAtY,EAAAG,IAAAmY,EACA,GAAAG,EAAA,CAGA,GADAA,EAAAlU,QAAA,SAAAkU,GAAA,KACAP,EAAAO,GAAA,CACA,IAAAT,EAAApT,kBAAAsT,EAAAO,GAAAzY,GACA,UAAAe,MAAA,WAAAf,EAAA,8CAEA,OAAAkY,EAAAO,GAEAP,EAAAO,GAAAzY,EAEA,KAAAyY,IAAAxa,OAAA,KACAia,EAAAO,EAAAC,UAAA,EAAAD,EAAAxa,OAAA,IAAA+B,GAGA2Y,EAAAF,EAAA,SAAAzY,EAAA4Y,iBAAA1X,MAAAlB,EAAA4Y,OAAA5Y,EAAA4Y,QACAD,EAAAF,EAAA,WAAAzY,EAAA6Y,mBAAA3X,MAAAlB,EAAA6Y,SAAA7Y,EAAA6Y,UACAR,EAAAI,EAAA,mBAAAzY,EAAA8Y,iBACAC,EAAAN,EAAA,cAAAzY,EAAAgZ,YACAX,EAAAI,EAAA,wBAAAzY,EAAAiZ,sBACAF,EAAAN,EAAA,eAAAzY,EAAAkZ,aACAH,EAAAN,EAAA,qBAAAzY,EAAAmZ,mBACAJ,EAAAN,EAAA,gBAAAzY,EAAAoZ,cACAT,EAAAF,EAAA,YAAAzY,EAAAqZ,UACAV,EAAAF,EAAA,SAAAzY,EAAAsZ,OACAX,EAAAF,EAAA,SAAAzY,EAAAuZ,OACAZ,EAAAF,EAAA,SAAAzY,EAAAwZ,OACAnB,EAAAI,EAAA,OAAAzY,EAAAyZ,MAEA,SAAAd,EAAAL,EAAA5W,GACA,GAAAA,aAAAR,MACA,QAAAxJ,EAAA,EAAgBA,EAAAgK,EAAAzD,OAAkBvG,IAClC2gB,EAAAC,EAAA,IAAA5gB,EAAAgK,EAAAhK,IAGA,SAAAqhB,EAAAT,EAAA5W,GACA,GAAAA,GAAA,iBAAAA,EACA,QAAAlI,KAAAkI,EACA2W,EAAAC,EAAA,IAAA9e,EAAAkI,EAAAlI,IAI
```
eladb: I don't think this file is supposed to be checked in

packages/jsii-pacmak/test/expected.jsii-calc-lib/dotnet/Amazon.JSII.Tests.CalculatorPackageId.LibPackageId/.jsii
```diff
@@ -92,6 +92,25 @@
       ],
```
eladb: I don't think key.snk is supposed to be checked in, is it?

packages/jsii-dotnet-runtime/src/Amazon.JSII.Runtime/jsii-runtime/jsii-runtime.js
RomainMuller: No it is not. What the...

packages/jsii-pacmak/test/expected.jsii-calc-lib/dotnet/Amazon.JSII.Tests.CalculatorPackageId.LibPackageId/.jsii
RomainMuller: No it is not, and I definitely remember *REMOVING* it from the commit several times, and it is even in my global ignores list. WHAT THE ACTUAL FUNK, GIT?

packages/jsii-dotnet-runtime/src/Amazon.JSII.Runtime/CallbackExtensions.cs
```diff
@@ -41,28 +42,29 @@ public static object InvokeCallback(this Callback callback, IReferenceMap refere
             }
         }
 
-        static object InvokeCallbackCore(this Callback callback, IReferenceMap referenceMap)
+        static object InvokeCallbackCore(this Callback callback, IReferenceMap referenceMap, out TypeReference returnType)
```
costleya: I would prefer to have a concrete structure of { type, value } than using out variables. This makes code hard to follow.

packages/jsii-dotnet-runtime/src/Amazon.JSII.Runtime/Services/Converters/FrameworkToJsiiConverter.cs
```diff
@@ -251,6 +251,10 @@ protected override bool TryConvertMap(IReferenceMap referenceMap, TypeReference
                     return false;
                 }
 
+                if (convertedElement != null && !(convertedElement is String) && !convertedElement.GetType().IsPrimitive)
```
costleya: _ This can be cleaner, I feel. But doing so requires for extensive refactoring, so whatever.

packages/jsii-dotnet-runtime-test/test/Amazon.JSII.Runtime.IntegrationTests/ComplianceTests.cs
```diff
@@ -83,10 +83,10 @@ public void CollectionTypes()
             Assert.Equal("World", types.ArrayProperty[1]);
 
             // map
-            IDictionary<string, double> map = new Dictionary<string, double>();
-            map["Foo"] = 123;
+            IDictionary<string, Number> map = new Dictionary<string, Number>();
+            map["Foo"] = new Number(123);
```
costleya: How common is "Number" going to be? If users have to start using objects instead of numbers across strongly typed languages, that sucks.

mpiroc: `Number` is a type in `jsii-calc-lib`, it's not built in to JSII.

RomainMuller: Correct. And the fact it is named like so allowed me to discover no fewer than 3 bugs last week, so I think it's a good thing :)

packages/jsii-dotnet-runtime/src/Amazon.JSII.Runtime/Services/Converters/FrameworkToJsiiConverter.cs
RomainMuller: Yup. This felt slightly dirty, but indeed, I couldn't afford that kind of time right now.

packages/jsii-dotnet-runtime/src/Amazon.JSII.Runtime/CallbackExtensions.cs
RomainMuller: I was adopting the surrounding style here, but yeah, adding a `CallbackResult` type.

PR #297, opened 2018-11-07
fix(jsii-runtime): treat "null" as "undefined"

### LINKED ISSUE #282: .NET: null property object support
From @costleya – Requiring code to pass around empty StackProps() param is confusing.  This is a bug in the .NET binding generation code.

eladb:
Since most languages do not have a distinction between "null" and
"undefined", jsii will effectively convert any "null" value passed into
an argument, a property or inside an object to "undefined".

Adds a compliance test to Java and .NET called "NullShouldBeTreatedAsUndefined".

Fixes awslabs/aws-cdk#157
Fixes #282

### DIFF
--- a/packages/jsii-java-runtime-test/project/src/test/java/software/amazon/jsii/testing/ComplianceTest.java
+++ b/packages/jsii-java-runtime-test/project/src/test/java/software/amazon/jsii/testing/ComplianceTest.java
@@ -25,6 +25,8 @@ import software.amazon.jsii.tests.calculator.JSObjectLiteralToNativeClass;
 import software.amazon.jsii.tests.calculator.Multiply;
 import software.amazon.jsii.tests.calculator.Negate;
 import software.amazon.jsii.tests.calculator.NodeStandardLibrary;
+import software.amazon.jsii.tests.calculator.NullShouldBeTreatedAsUndefined;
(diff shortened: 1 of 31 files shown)


packages/jsii-java-runtime-test/project/src/test/java/software/amazon/jsii/testing/ComplianceTest.java
```diff
@@ -927,6 +929,18 @@ public void classWithPrivateConstructorAndAutomaticProperties() {
         assertEquals("Hello", obj.getReadOnlyString());
     }
 
+    @Test
+    public void nullShouldBeTreatedAsUndefined() {
+        NullShouldBeTreatedAsUndefined obj = new NullShouldBeTreatedAsUndefined("hello", null);
```
RomainMuller: I'm not even sure the `Jackson` configuration we have would include the `null`s in the generated JSON. But I reckon the intent is still validated, right?

RomainMuller: The POJOs get serialized to JSON before they're sent to the kernel, right? That JSON :)

eladb: Got it. null is serialized to "null"

PR #303, opened 2018-11-08
fix(docs): improve docs rendering

### LINKED ISSUE #298: sphinx: Default values aren't shown in docs
I'm noticing that the "@default" flag in the code comments is not translating into a default value shown in the docs.  From the docs, it reads as though there is no default value, which is confusing.

For example, default value for containerPort (80) in LoadBalancedFargateService:

https://github.com/awslabs/aws-cdk/blob/master/packages/%40aws-cdk/aws-ecs/lib/load-balanced-fargate-service.ts#L62

Is not shown in the docs:

https://awslabs.github.io/aws-cdk/refs/_aws-cdk_aws-ecs.html#@aws-cdk/aws-ecs.LoadBalancedFargateServiceProps.containerPort

### LINKED ISSUE #299: sphinx: multiple lines in jsdocs are emitted as one single statement

### LINKED ISSUE #300: sphinx: optional properties or arguments are described with "or undefined" instead of "(optional)"

### LINKED ISSUE #301: jsii: it is possible to define a method/ctor with non-optional after an optional
Probably a typescript bug, but it is now possible to define something like this:

```ts
function foo(arg1 = 'hello', arg2: string) {
}
```

This means that `arg1` is optional and `arg2` is required, which doesn't work.

### LINKED ISSUE #302: sphinx: no need to indicate "abstract" for interface properties

eladb:
Fixes #301: reject non-optional arguments after optional arguments
Fixes #298: show default values for optional properties
Fixes #302: omit "abstract" from interface properties (they are by definition abstract)
Fixes #300: indicate optionals with "optional" instead of "or undefined"
Fixes #299: render multiple jsdoc lines as multiple lines in sphinx

### DIFF
--- a/packages/jsii-pacmak/lib/targets/sphinx.ts
+++ b/packages/jsii-pacmak/lib/targets/sphinx.ts
@@ -254,12 +254,12 @@ class SphinxDocsGenerator extends Generator {
     //
     // Properties
 
-    protected onProperty(_cls: spec.ClassType, prop: spec.Property) {
-        this.renderProperty(prop);
+    protected onProperty(cls: spec.ClassType, prop: spec.Property) {
+        this.renderProperty(cls, prop);
     }
 
-    protected onStaticProperty(_cls: spec.ClassType, prop: spec.Property) {
-        this.renderProperty(prop);
+    protected onStaticProperty(cls: spec.ClassType, prop: spec.Property) {
+        this.renderProperty(cls, prop);
     }
 
     //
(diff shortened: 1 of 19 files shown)


packages/jsii-pacmak/lib/targets/sphinx.ts
```diff
@@ -651,9 +651,14 @@ class SphinxDocsGenerator extends Generator {
             this.code.line();
         }
         const readonly = prop.immutable ? ' *(readonly)*' : '';
-        const abs = prop.abstract ? ' *(abstract)*' : '';
+        const abs = (parent.kind !== spec.TypeKind.Interface &&  prop.abstract) ? ' *(abstract)*' : '';
```
rix0rrr: I wonder about the value of showing "abstract" at all.

Does it help consumers of a class know that a method is abstract? Or do they simply care about what it would do if they were to call it?

eladb: I think they do if they are looking to extend a class.

PR #305, opened 2018-11-09
feat(jsii-dotnet-runtime): Improve .NET Performance

### LINKED ISSUE #304: Improve .NET performance
The .NET runtime is much slower than Java. Investigate possible performance improvements.

costleya:
The .NET runtime performace has been enhanced by adding an Assembly Load cache to JsiiTypeAttributeBase to avoid expensive repeated
use of reflection and file system access.

Compliance tests now share a client to speed up execution and avoid problems with cache usage.

.NET integration tests now hard-reference the .NET runtime project, and has been added to the runtime solution. There is no disadvantage to doing this,
and it makes development easier.

Resolves #304


costleya:
Rough metrics, a CI/CD pipeline (CodeCommit/CodeBuild/CodePipeline):
Before: 10.535 seconds
After: 0.932 seconds

rix0rrr:
yay

packages/jsii-dotnet-runtime-test/test/Amazon.JSII.Runtime.IntegrationTests/Amazon.JSII.Runtime.IntegrationTests.csproj
```diff
@@ -7,7 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Amazon.JSII.Tests.CalculatorPackageId" Version="$(JsiiVersion)" />
```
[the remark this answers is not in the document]
costleya: Oh, my editor was tricky. Good catch.

packages/jsii-dotnet-runtime/src/Amazon.JSII.Runtime.sln
```diff
@@ -14,6 +14,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "UnitTests", "UnitTests", "{
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Amazon.JSII.Runtime.UnitTests", "Amazon.JSII.Runtime.UnitTests\Amazon.JSII.Runtime.UnitTests.csproj", "{96CC0C0B-1D90-448F-9BFC-07CE93D2CE29}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Amazon.JSII.Runtime.IntegrationTests", "..\..\jsii-dotnet-runtime-test\test\Amazon.JSII.Runtime.IntegrationTests\Amazon.JSII.Runtime.IntegrationTests.csproj", "{7BD15A18-BE3A-4729-9B8C-570BF214C4CE}"
```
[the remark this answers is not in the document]
costleya: Yes, I mentioned this in the commit. This makes development on the runtime easier, and as long as they are in the same repository, I see no reason not to have them in the same solution.

packages/jsii-dotnet-runtime/src/Amazon.JSII.Runtime/Deputy/JsiiTypeAttributeBase.cs
```diff
@@ -17,27 +20,32 @@ protected JsiiTypeAttributeBase(Type nativeType, string fullyQualifiedName)
...
+                ProcessedAssemblies.Add(GetAssemblyKey(assembly));
                 return;
             }
 
-            foreach (Assembly dependency in dependencies)
+            foreach (var referencedAssembly in assembly.GetReferencedAssemblies())
             {
-                Load(dependency);
+                var loadedReference = Assembly.Load(referencedAssembly);
+                Load(loadedReference);
```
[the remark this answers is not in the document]
costleya: I agree. And this change could probably be moved up to the Client level (this attribute is doing WAY too much). The client also has a record of what jsii assemblies have been read - these two can probably be merged. That can be a refactor though - making a small surgical change here to get this improvement out to customers.

packages/jsii-dotnet-runtime/src/Amazon.JSII.Runtime/Deputy/JsiiTypeAttributeBase.cs
```diff
@@ -17,27 +20,32 @@ protected JsiiTypeAttributeBase(Type nativeType, string fullyQualifiedName)
             Load(nativeType.Assembly);
         }
 
-        void Load(Assembly assembly)
+        private static void Load(Assembly assembly)
         {
-            IEnumerable<Assembly> dependencies = assembly.GetReferencedAssemblies()
-                .Select(assemblyName => Assembly.Load(assemblyName));
+            if (ProcessedAssemblies.Contains(GetAssemblyKey(assembly)))
```
mpiroc: While `ConcurrentBag` is thread-safe, this check isn't and can result in duplicates loads. But if `JsiiTypeAttributeBase.Load` is idempotent and the performance impact of the occasional duplicate load isn't significant, this is good enough. Can you add a comment that mentions the idempotency requirement?

If thread-safety isn't a requirement, there's no need to use `ConcurrentBag`.

packages/jsii-dotnet-runtime/src/Amazon.JSII.Runtime/Deputy/JsiiTypeAttributeBase.cs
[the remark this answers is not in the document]
mpiroc: As a quick fix for the potential confusion, how about just qualifying the symbol? i.e. `JsiiTypeAttributeBase.Load(loadedReference)`

packages/jsii-dotnet-runtime/src/Amazon.JSII.Runtime/Deputy/JsiiTypeAttributeBase.cs
```diff
@@ -46,8 +54,13 @@ void Load(Assembly assembly)
 
             IClient client = serviceProvider.GetRequiredService<IClient>();
             client.LoadPackage(attribute.Name, attribute.Version, tarballPath);
+
+            ProcessedAssemblies.Add(GetAssemblyKey(assembly));
+
```
mpiroc: Nit: Extra newline

packages/jsii-dotnet-runtime/src/Amazon.JSII.Runtime/Deputy/JsiiTypeAttributeBase.cs
```diff
@@ -1,14 +1,17 @@
...
 using System.Linq;
 using System.Reflection;
+using Amazon.JSII.Runtime.Services;
+using Microsoft.Extensions.DependencyInjection;
 
 namespace Amazon.JSII.Runtime.Deputy
 {
     public abstract class JsiiTypeAttributeBase : Attribute
     {
+        private static readonly ConcurrentBag<string> ProcessedAssemblies =
```
mpiroc: Nit: Are we still attempting to follow [the CoreFX style guide](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md)?

> We use \_camelCase for internal and private fields and use readonly where possible. Prefix internal and private instance fields with \_, static fields with s\_ and thread static fields with t\_. When used on static fields, readonly should come after static (e.g. static readonly not readonly static). Public fields should be used sparingly and should use PascalCasing with no prefix when used.

I don't like the Hungarian notation, we can omit that. But PascalCase is reserved for public fields.

packages/jsii-dotnet-runtime/src/Amazon.JSII.Runtime/Deputy/JsiiTypeAttributeBase.cs
```diff
@@ -17,27 +20,32 @@ protected JsiiTypeAttributeBase(Type nativeType, string fullyQualifiedName)
...
+                var loadedReference = Assembly.Load(referencedAssembly);
+                Load(loadedReference);
             }
 
             // find the .tgz resource
             var tarballResourceName = assembly.GetManifestResourceNames().FirstOrDefault(name => name.EndsWith(".tgz"));
             if (tarballResourceName == null)
             {
-                throw new JsiiException("Cannot find embedded tarball resource in assembly " + assembly.GetName(), null);
+                throw new JsiiException("Cannot find embedded tarball resource in assembly " + assembly.GetName(),
```
mpiroc: IIRC `JsiiException` is reserved for exceptions that come directly from the JSII runtime. Use a standard exception type or roll your own instead.

costleya: This was not a change, just a newline added due to a reformat. I will change it in a future PR.

https://github.com/awslabs/jsii/issues/308

packages/jsii-dotnet-runtime/src/Amazon.JSII.Runtime/Deputy/JsiiTypeAttributeBase.cs
costleya: We are still attempting this. But since it is a static readonly, it follows -kinda- the same rules as a const, so it's PascalCase. If it was not readonly, it would be camelCase.

I could change my mind though. I'm following the guidance of ReSharper here.

packages/jsii-dotnet-runtime/src/Amazon.JSII.Runtime/Deputy/JsiiTypeAttributeBase.cs
costleya: If each call to js is an independent call, it's possible that someone could write a multithreaded application that works.

I'm not saying they should. But if it doesn't work, I don't want a cache to be the reason.

I don't think there has been a concrete stance on multithreading/async calls. The CDK people are still talking about options as well.

packages/jsii-dotnet-runtime/src/Amazon.JSII.Runtime/Deputy/JsiiTypeAttributeBase.cs
```diff
@@ -17,27 +23,32 @@ protected JsiiTypeAttributeBase(Type nativeType, string fullyQualifiedName)
...
+            if (ProcessedAssemblies.Contains(GetAssemblyKey(assembly)))
+            {
+                return;
+            }
 
-            JsiiAssemblyAttribute attribute = assembly.GetCustomAttribute<JsiiAssemblyAttribute>();
+            var attribute = assembly.GetCustomAttribute<JsiiAssemblyAttribute>();
             if (attribute == null)
             {
+                ProcessedAssemblies.Add(GetAssemblyKey(assembly));
```
eladb: You can move this to line 32 so there's a single code path that adds to the map

costleya: This map add is a short-circuit for non-jsii assemblies. It makes it so each assembly is reflected upon only once.

PR #311, opened 2018-11-12
v0.8.0

eladb:
* Bump to v0.8.0, see CHANGELOG
* Automatically update "peerDependencies" in "bump.sh"


rix0rrr:
> Bump to v0.8.0, see CHANGELOG

Oy oy, this is not according to release process :P

rix0rrr:
Wait why are we bumping MV? I don't see any breaking changes in the changelog?

eladb:
> Wait why are we bumping MV? I don't see any breaking changes in the changelog?

It's a good question. That's what lerna determined. Let me check.

> Oy oy, this is not according to release process :P

How strongly do you feel about pasting the changelog into the commit message? I feel it's a redundant step (and haven't seen it in other projects). Should we change the MCM?

rix0rrr:
> How strongly do you feel about pasting the changelog into the commit message?

Good question. I do feel it's sorta nice when you're looking back through the history, but I don't do that all that often so meh. I'm fine changing the CM.

Question: where does the GitHub Releases publisher get the changelog to publish? From the commit or from the changelog? Because I *do* want the change log there.

eladb:
> Question: where does the GitHub Releases publisher get the changelog to publish? From the commit or from the changelog? Because I do want the change log there.

From the changelog of course

eladb:
> Wait why are we bumping MV? I don't see any breaking changes in the changelog?

From lerna logs, it seems like it doesn't respect pre-1.0 versioning. It detects that the overall repo has a "minor" bump, and that's what it does (1.7.8=>1.8.0 is minor)...

Let me see if that's something I can fix somehow. Otherwise we won't be able to use Lerna for automated releases pre-1.0 and that's unacceptable.

eladb:
See https://github.com/awslabs/aws-cdk/issues/272 for details. Basically, we can't use "lerna publish" for pre 1.0 - I will update to 0.7.8 manually for now and we'll explore as we move to modular versioning and automatic releases.

eladb:
Closing in favor of https://github.com/awslabs/jsii/pull/new/benisrae/v0.7.9

PR #328, opened 2018-12-05
feat(jsii-reflect): library for exploring jsii type systems

eladb:
__jsii-reflect__ is a library for strong-type reflection of jsii type systems. It creates an in-memory object graph from .jsii module specifications and exposes a rich API for exploring the type system.

__jsii-tree__ is a command-line utility bundled with jsii-reflect which prints an ascii tree of a jsii type system.

__oo-ascii-tree__ is an object-oriented library for producing ascii trees. It supports multi-line nodes and some other cool features. 

See [jsii-reflect README](https://github.com/awslabs/jsii/blob/38ddc1db82fa4c334b38b9d6fc2ec2b2c6b4a0fd/packages/jsii-reflect/README.md) and [oo-ascii-tree README](https://github.com/awslabs/jsii/blob/38ddc1db82fa4c334b38b9d6fc2ec2b2c6b4a0fd/packages/oo-ascii-tree/README.md) for more details.


packages/jsii-reflect/README.md
```diff
@@ -0,0 +1,189 @@
...
+typesystem.load('jsii-calc-base/.jsii');
+typesystem.load('jsii-calc-base-of-base/.jsii');
+```
+
+Or, you can point `load` to an npm module's directory, in which case the loader
+will transitively load all dependent modules:
+
+```ts
+// will automatically bring in all dependencies
+typesystem.load('jsii-calc');
```
rix0rrr: If I'm doing this and I want to do linting on the module I'm loading, is there a way for me to know the "primary" assembly?

packages/jsii-reflect/lib/assembly.ts
```diff
@@ -0,0 +1,197 @@
...
+   */
+  public get homepage(): string {
+    return this.spec.homepage;
+  }
+
+  /**
+   * The module repository, maps to "repository" from package.json
+   * This is required since some package managers (like Maven) require it.
+   */
+  public get repository() {
```
rix0rrr: No type annotation

packages/jsii-reflect/lib/interface.ts
```diff
@@ -0,0 +1,81 @@
...
+        out.push(...base.getMethods(inherited));
+      }
+    }
+    if (this.interfaceSpec.methods) {
+      out.push(...this.interfaceSpec.methods.map(m => new Method(this.system, this.assembly, this, m)));
+    }
+    return out;
+  }
+
+  public getProperties(inherited = false) {
```
rix0rrr: Why have both `properties` and `getProperties()`? I mean, I can see it looking at the code, but the distinction will not be immediately obvious to consumers.

I'd prefer one of:

* Just one `properties(inherited)` method.

* Make `getProperties()` private and expose `properties`, `ownProperties` properties.

packages/jsii-reflect/lib/tree.ts
```diff
@@ -0,0 +1,278 @@
...
+   * Output with ANSI colors
+   * @default true
+   */
+  colors?: boolean;
+}
+
+/**
+ * Visualizes a `TypeSystem` as an ASCII tree.
+ */
+export class TypeSystemTree extends AsciiTree {
```
rix0rrr: Why inheritance? Why `is-a` type system tree an ascii tree?

Couldn't `visualizeTypeSystemTree()` be an operation that produces an AsciiTree?

packages/jsii-reflect/test/jsii-tree.test.ts
```diff
@@ -0,0 +1,66 @@
...
+async function jsiiTreeTest(expected: string, ...args: string[]) {
+  expect(await jsiiTree(...args)).toEqual(await readExpected(expected));
+}
+
+async function jsiiTree(...args: string[]) {
+  const command = [
+    path.join(__dirname, '..', 'bin', 'jsii-tree'),
+    args.join(' '),
+    '--no-colors',
+    path.dirname(require.resolve('jsii-calc/package.json'))
```
rix0rrr: Please no! Not another test suite that will break when we change something about jsii and we add a compliance test :(
 
At the very least respect the UPDATE_DIFF flag or we're going to have pain on the next jsii change.

packages/oo-ascii-tree/README.md
```diff
@@ -0,0 +1,166 @@
...
+ ├─┬ child2
+ │ ├── grandchild1
+ │ └── grandchild2
+ └── child3
+```
+
+## Advanced Usage
+
+You can also subclass `AsciiTree` to encapsulate some model. The following
+example declares a `TitleNode` which formats a title with "====" underline and a
```
rix0rrr: This is good, but there's no actual need to subclass right?

packages/jsii-reflect/README.md
eladb: Good question. I'll add something.

eladb: Added `filesystem.roots`

packages/jsii-reflect/lib/assembly.ts
eladb: It's inferred from the spec: `{ repository: string, url: string }`

packages/jsii-reflect/lib/method.ts
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    return this.spec.name;
+  }
+
+  /**
+   * The return type of the method (undefined if void or initializer)
+   */
+  public get returns(): TypeReference {
+    return new TypeReference(this.system, this.spec.returns);
```
RomainMuller: Feels like this could (should?) be cached, maybe.

packages/jsii-reflect/lib/interface.ts
eladb: I tried to stay 1:1 with the spec in terms of names, so `properties` will return the non-inherited properties while `getProperties(true)` will also return the inherited ones. It's okay to have more than one way to do something :-)

Added some docs

packages/jsii-reflect/lib/tree.ts
```diff
@@ -0,0 +1,278 @@
...
+  constructor(typesys: TypeSystem, options: TypeSystemTreeOptions = { }) {
+    super();
+
+    if (options.showAll) {
```
RomainMuller: As mentioned before, this feels unnecessary here, but okay, why not.

packages/jsii-reflect/lib/tree.ts
eladb: It is an is-a relationship, a TypeSystemTree is an AsciiTree. I can rename it to `TypeSystemAsciiTree` if it will be clearer.

I didn't want a dependency between `TypeSystem` and the `AsciiTree`, so those are decoupled class hierarchies. I originally thought to extract `jsii-tree` to a separate module, but for pragmatic reasons decided to just bundle it together with `jsii-reflect`.

packages/jsii-reflect/test/jsii-tree.test.ts
eladb: Almost everything in the jsii repo depends on these calc libs for testing. I am not sure what the problem is. 

Obvsiouly you are right about UPADTE_DIFF.

packages/oo-ascii-tree/README.md
eladb: It's a useful paradigm for certain use cases. Not required.

packages/jsii-reflect/lib/tree.ts
```diff
@@ -0,0 +1,278 @@
...
+      options.dependencies = true;
+      options.inheritance = true;
+      options.members = true;
+      options.signatures = true;
+      options.types = true;
+    }
+
+    const shouldUseColors = options.colors === undefined ? true : options.colors;
+    const colorsEnabled = colors.enabled;
+    try {
```
RomainMuller: Would wrap the try-finally in a helper function (but I guess it's also good as-is). Reason I'm saying is because I was initially confused by `shouldUseColors` and `colorsEnabled`, and had to read down almost 20 lines to understand why `colorsEnabled` exists.

packages/jsii-reflect/package.json
```diff
@@ -0,0 +1,42 @@
...
+    "build": "tsc",
+    "watch": "tsc -w",
+    "test": "jest",
+    "package": "package-js"
+  },
+  "dependencies": {
+    "fs-extra": "^7.0.1",
+    "jsii-spec": "^0.7.11",
+    "yargs": "^12.0.5",
+    "colors": "^1.3.2",
```
RomainMuller: Order :)

packages/oo-ascii-tree/README.md
```diff
@@ -0,0 +1,166 @@
+# oo-ascii-tree
+
+Renders ASCII trees from an object-oriented object graph.
```
RomainMuller: Is there anything really OO graph-specific here? I feel like this is underselling it to a niche when it's actually able to be a general purpose "make me a tree view of this"

packages/jsii-reflect/test/jsii-tree.test.ts
eladb: Added diff tests

packages/jsii-reflect/README.md
```diff
@@ -0,0 +1,189 @@
...
+
+jsii-tree is a command-line tool which can be used to print an ASCII tree of
+a jsii type system:
+
+```console
+$ jsii-tree jsii-calc
+assemblies
+ ├─┬ jsii-calc
+ │ └─┬ types
+ │   ├── AbstractClass class
```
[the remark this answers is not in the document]
eladb: Done

packages/jsii-reflect/bin/jsii-tree.ts
```diff
@@ -0,0 +1,42 @@
...
+    .usage('$0 <JSII-FILE | MODULE-DIR...>', 'Prints an ASCII tree representation of a jsii type system.', args => args
+      .positional('JSII-FILE', { desc: 'path to a .jsii file to load, all dependency .jsii files must be explicitly supplied' })
+      .positional('MODULE_DIR', { desc: 'path to an jsii npm module directory, all jsii dependencies will be loaded transitively' }))
+    .option('types', { type: 'boolean', alias: 't', desc: 'show types', default: false })
+    .option('members', { type: 'boolean', alias: 'm', desc: 'show type members', default: false })
+    .option('inheritance', { type: 'boolean', alias: 'i', desc: 'show base classes and implemented interfaces', default: false })
+    .option('dependencies', { type: 'boolean', alias: 'd', desc: 'show assembly dependencies', default: false })
+    .option('signatures', { type: 'boolean', alias: 's', desc: 'show method and property signatures', default: false })
+    .option('colors', { type: 'boolean', desc: 'enable/disable ANSI colors in output', default: true })
+    .option('all', { type: 'boolean', alias: 'a', desc: 'show all details', default: false })
```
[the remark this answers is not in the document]
eladb: It seems like `implies` will _require_ that the implicated key will be defined, it won't automatically define it... Weird naming... https://github.com/yargs/yargs/blob/master/docs/api.md#impliesx-y

Sorted.

packages/jsii-reflect/bin/jsii-tree.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  const typesys = new TypeSystem();
+
+  for (const fileOrDirectory of options.jsiiFile) {
+    await typesys.load(fileOrDirectory);
+  }
+
+  const tst = new TypeSystemTree(typesys, {
+    dependencies: options.dependencies || options.all,
+    types: options.types || options.all || options.members || options.inheritance,
```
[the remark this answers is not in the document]
eladb: Sadly, that's not what `implies` does: https://github.com/yargs/yargs/blob/master/docs/api.md#impliesx-y

packages/jsii-reflect/lib/assembly.ts
```diff
@@ -0,0 +1,197 @@
...
+  /**
+   * The version of the spec schema
+   */
+  public get schema(): jsii.SchemaVersion {
+    return this.spec.schema;
+  }
+
+  /**
+   * The name of the assembly
+   * @minLength 1
```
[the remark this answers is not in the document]
eladb: Copy & pasted from the spec

packages/jsii-reflect/lib/class.ts
```diff
@@ -0,0 +1,116 @@
...
+      out.push(this.base);
+      out.push(...this.base.getAncestors());
+    }
+    return out;
+  }
+
+  public getProperties(inherited = false) {
+    const out = new Array<Property>();
+    if (inherited && this.base) {
+      out.push(...this.base.getProperties(inherited));
```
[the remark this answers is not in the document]
eladb: I believe it will include them since they will be explicitly called out in the class's API, or did I misunderstood you?

packages/jsii-reflect/lib/class.ts
```diff
@@ -0,0 +1,116 @@
...
+  public getAncestors() {
+    const out = new Array<ClassType>();
+    if (this.base) {
+      out.push(this.base);
+      out.push(...this.base.getAncestors());
+    }
+    return out;
+  }
+
+  public getProperties(inherited = false) {
```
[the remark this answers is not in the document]
eladb: Added

packages/jsii-reflect/lib/index.ts
```diff
@@ -0,0 +1,15 @@
+export * from './assembly';
+export * from './class';
+export * from './dependency';
+export * from './docs';
+export * from './docs';
```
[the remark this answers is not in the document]
eladb: Fixed

packages/jsii-reflect/lib/index.ts
```diff
@@ -0,0 +1,15 @@
...
+export * from './enum';
+export * from './interface';
+export * from './method';
+export * from './overridable';
+export * from './parameter';
+export * from './property';
+export * from './type';
+export * from './type-ref';
+export * from './type-system';
+export * from './tree';
```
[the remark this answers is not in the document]
eladb: Fixed

packages/jsii-reflect/lib/interface.ts
```diff
@@ -0,0 +1,81 @@
...
+
+export class InterfaceType extends Type {
+  constructor(
+    public system: TypeSystem,
+    public assembly: Assembly,
+    private interfaceSpec: jsii.InterfaceType) {
+    super(system, assembly, interfaceSpec);
+  }
+
+  public get kind(): jsii.TypeKind {
```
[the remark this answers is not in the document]
eladb: Yap.

packages/jsii-reflect/lib/interface.ts
```diff
@@ -0,0 +1,81 @@
...
+   * List of methods.
+   */
+  public get methods(): Method[] {
+    return this.getMethods();
+  }
+
+  /**
+   * List of properties.
+   */
+  get properties(): Property[] {
```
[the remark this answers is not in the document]
eladb: Done

packages/jsii-reflect/lib/interface.ts
```diff
@@ -0,0 +1,81 @@
...
+   * have idiomatic ways to allow defining concrete instances such interfaces.
+   * For example, in Java, the generator will produce a PoJo and a builder
+   * which will allow users to create a concrete object with data which
+   * adheres to this interface.
+   */
+  public get datatype(): boolean {
+    return !!this.datatype;
+  }
+
+  public getMethods(inherited = false) {
```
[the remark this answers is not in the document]
eladb: I tried to keep the surface area of the spec, but still wanted to allow people to get the inherited list.

packages/jsii-reflect/lib/interface.ts
```diff
@@ -0,0 +1,81 @@
...
+      for (const base of this.interfaces) {
+        out.push(...base.getProperties(inherited));
+      }
+    }
+    if (this.interfaceSpec.properties) {
+      out.push(...this.interfaceSpec.properties.map(p => new Property(this.system, this.assembly, this, p)));
+    }
+    return out;
+  }
+}
```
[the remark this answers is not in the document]
eladb: Done

packages/jsii-reflect/lib/method.ts
```diff
@@ -0,0 +1,97 @@
...
+
+export class Method implements Documentable, Overridable {
+  constructor(
+    public readonly system: TypeSystem,
+    public readonly assembly: Assembly,
+    public readonly parentType: Type,
+    private readonly spec: jsii.Method) { }
+
+  public toString() {
+    return `method:${this.parentType.fqn}.${this.name}`;
```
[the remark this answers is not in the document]
eladb: But the signature is not included...

packages/jsii-reflect/lib/method.ts
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  /**
+   * The name of the method. Undefined if this method is a initializer.
```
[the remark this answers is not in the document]
eladb: Done

packages/jsii-reflect/lib/method.ts
eladb: Since everything is immutable, I'll leave caching for a later stage. There are many opportunities for caching all across this library since it's all lazy.

packages/jsii-reflect/lib/tree.ts
eladb: This is for the library use-case, not the command line tool.

packages/jsii-reflect/lib/tree.ts
eladb: done

packages/jsii-reflect/package.json
eladb: I'll pass. I don't see tons of value in maintaining alphabetical order in these lists. It's more work than benefit IMHO

packages/oo-ascii-tree/README.md
eladb: It's just that there's already an **ascii-tree** npm module and this one is different in the sense that it uses classes to represent the nodes (the other one uses a string syntax as input). Nothing interesting.

PR #346, opened 2019-01-11
fix(kernel): Improve tagged type of wire values

### LINKED ISSUE #345: kernel: return type for unexported types
Someone reported a failing case over email, where we were returning a `Construct` to Java, which should have been a particular L1 class.

Key insight: the type we construct a proxy for depends on the requested type. Compliance test:

```ts
export class PublicClass {
    hello(): void;
}

export interface IPublicInterface {
    bye(): void;
}

export class InbetweenClass extends PublicClass {
}

class PrivateClass extends InbetweenClass implements IPublicInterface {
    public hello() {
        process.stdout.write('Hello\n');
    }

    public bye() {
        process.stdout.write('Bye\n');
    }
}

export class Constructors {
    public static makeClass(): PublicClass {
        return new PrivateClass();
    }

    public static makeInterface(): IPublicInterface {
        return new PrivateClass();
    }
}
```

Given the same object instance:

* `Constructors.makeClass()` should return a proxy object for `InbetweenClass`.
* `Constructors.makeInterface()` should return a proxy object for `IPublicInterface`.

Given:

* The actual type (`PrivateClass`)
* The requested type (`PublicClass` vs. `IPublicInterface`)
* The inheritance hierarchy

We should be returning the most specific type on the inheritance path between actual and requested that is itself `exported`.

RomainMuller:
When passing references across the JSII language boundary, the static
return type of a method is used instead of the runtime type of the
object if they are not the same, even if they are compatible. This
causes issues that make it impossible to up-cast from methods such as
`IConstruct.findChild` that return a super-class that is expected to be
up-casted to it's known dynamic type.

Fixes #345

---

TODO:
- [x] Add compliance test

### DIFF
--- a/packages/jsii-calc/lib/compliance.ts
+++ b/packages/jsii-calc/lib/compliance.ts
@@ -1234,7 +1234,10 @@ export class JsiiAgent {
     }
 };
 
-// https://github.com/awslabs/jsii/issues/345
+// Ensure the JSII kernel tags instances with the "most appropriate" FQN type label, so that runtimes are able to
+// correctly choose the implementation proxy that should be used. Failure to do so could cause situations where userland
+// needs to up-cast an instance to an incompatible type, which certain runtimes (such as Java) will prevent.
+// @See https://github.com/awslabs/jsii/issues/345
 export class PublicClass {
     public hello(): void {}
 }
--- a/packages/jsii-calc/lib/compliance.ts
+++ b/packages/jsii-calc/lib/compliance.ts
@@ -1233,3 +1233,23 @@ export class JsiiAgent {
         return process.env.JSII_AGENT;
     }
 };
+
+// https://github.com/awslabs/jsii/issues/345
+export class PublicClass {
+    public hello(): void {}
+}
+export interface IPublicInterface {
+    bye(): void;
+}
+export class InbetweenClass extends PublicClass {}
+class PrivateClass extends InbetweenClass implements IPublicInterface {
+    public bye(): void {}
+}
+export class Constructors {
+    public static makeClass(): PublicClass {
+        return new PrivateClass();
+    }
+    public static makeInterface(): IPublicInterface {
+        return new PrivateClass();
+    }
+}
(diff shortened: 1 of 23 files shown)


rix0rrr:
I think we need a test

rix0rrr:
Also don't like how this doesn't cover the intermediate class case. It's not that much more work and covers us for future fuckups.

RomainMuller:
But... does this not cover the "more correct behavior"? I don't see the difference... o_O

RomainMuller:
Added the compliance tests (kernel, Java, C#).

packages/jsii-calc/lib/compliance.ts
```diff
@@ -1232,4 +1232,24 @@ export class JsiiAgent {
     public static get jsiiAgent(): string | undefined {
         return process.env.JSII_AGENT;
     }
-};
\ No newline at end of file
+};
+
+// https://github.com/awslabs/jsii/issues/345
```
eladb: Add a short explanation here as well

packages/jsii-java-runtime-test/project/src/test/java/software/amazon/jsii/testing/ComplianceTest.java
```diff
@@ -968,6 +973,15 @@ public void receiveInstanceOfPrivateClass() {
         assertTrue(new ReturnsPrivateImplementationOfInterface().getPrivateImplementation().getSuccess());
     }
 
+    @Test
+    public void objRefsAreLabelledUsingWithTheMostCorrectType() {
+        final PublicClass classRef = Constructors.makeClass();
+        final IPublicInterface ifaceRef = Constructors.makeInterface();
+
+        assertTrue(classRef instanceof InbetweenClass);
```
rix0rrr: I don't see how this can be true. But if it works, then it works.

RomainMuller: This property being true is literally the whole point of this change.... 🤔

packages/jsii-calc/lib/compliance.ts
RomainMuller: Sure!

PR #348, opened 2019-01-26
feat(jsii): support multiple class declaration sites

rix0rrr:
Classes can have members added to them in multiple declarations.
Combine all sites into a single class model.


packages/jsii/lib/assembler.ts
```diff
@@ -406,20 +406,29 @@ export class Assembler implements Emitter {
...
-            } else if (ts.isPropertyDeclaration(memberDecl)
-                    || ts.isPropertySignature(memberDecl)
-                    || ts.isAccessor(memberDecl)) {
-                await this._visitProperty(member, jsiiType);
-            } else {
-                this._diagnostic(memberDecl,
-                                 ts.DiagnosticCategory.Warning,
-                                 `Ignoring un-handled ${ts.SyntaxKind[memberDecl.kind]} member`);
+        if (!type.isClass()) {
+            throw new Error('Oh no');
```
RomainMuller: Poor error message 🤣

rix0rrr: The alternative is "this shouldn't happen" :)

PR #351, opened 2019-02-01
fix: remove use of private API

### LINKED ISSUE #350: Get rid of private API use
To parse `tsconfig.json`, the compiler needs a `ParseConfigHost`, which contains functions to inspect the filesystem. Unlike a `CompilerHost`, there is no public factory function to create one.

There is a private function to turn an existing `CompilerHost` into a `ParseConfigHost` which we call today, but we've gotten some signal that this function may be removed in an upcoming version, so we're going to have to implement our own `ParseConfigHost` soon.

https://github.com/awslabs/jsii/blob/4ecf28ceaaf130bd1622765468bc64a43fbb5efd/packages/jsii/lib/compiler.ts#L262

rix0rrr:
Don't use the upstream private API function to turn a `CompilerHost` into a
`ParseConfigHost` anymore; instead, I copied the implementation, which is
simple enough.

Fixes #350.

### DIFF
--- a/packages/jsii/lib/compiler.ts
+++ b/packages/jsii/lib/compiler.ts
@@ -325,6 +325,8 @@ function nodeJsCompatibleSearchPaths(dir: string): string[] {
 }
 
 function parseConfigHostFromCompilerHost(host: ts.CompilerHost): ts.ParseConfigHost {
+    // Copied from upstream
+    // https://github.com/Microsoft/TypeScript/blob/9e05abcfd3f8bb3d6775144ede807daceab2e321/src/compiler/program.ts#L3105
     return {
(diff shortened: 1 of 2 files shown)


packages/jsii/lib/compiler.ts
```diff
@@ -323,3 +323,18 @@ function nodeJsCompatibleSearchPaths(dir: string): string[] {
 
     return ret;
 }
+
+function parseConfigHostFromCompilerHost(host: ts.CompilerHost): ts.ParseConfigHost {
```
RomainMuller: Can you add a link to the original implementation? I mean a permalink to the particular location... If it ever needs to be updated, that'll probably help track down where the new version of it lives...

PR #360, opened 2019-02-13
fix(jsii): detect double interface member declarations

### LINKED ISSUE #340: jsii: do not allow interfaces to redefine members (dotnet does not support it)
Repro:

```ts
intetface A {
  foo: number;
}

interface B extends A {
  foo: number;
}
```

This is legal in typescript but .NET will fail due to `hides inherited member. Use the new keyword if hiding was intended`

rix0rrr:
In C# it's prohibited to declare an interface member
that also exists in an inherited interface. Have jsii
detect this illegal pattern.

Strictly speaking, this only true if we don't overload,
but since we don't support overloading anyway we just
check on the member names, not the types.

Fixes #340.

### DIFF
--- a/packages/jsii/lib/assembler.ts
+++ b/packages/jsii/lib/assembler.ts
@@ -624,6 +624,27 @@ export class Assembler implements Emitter {
             }
         });
 
+        // Check that no interface declares a member that's already declared
+        // in a base type (not allowed in C#).
+        const memberNames = interfaceMemberNames(jsiiType);
+        const checkNoIntersection = (...bases: spec.Type[]) => {
+            for (const base of bases) {
+                if (!spec.isInterfaceType(base)) { continue; }
+
+                const baseMembers = interfaceMemberNames(base);
+                for (const memberName of memberNames) {
+                    if (baseMembers.includes(memberName)) {
+                        this._diagnostic(type.symbol.declarations[0],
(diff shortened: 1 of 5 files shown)


packages/jsii/lib/assembler.ts
```diff
@@ -624,6 +624,27 @@ export class Assembler implements Emitter {
             }
         });
 
+        // Check that no interface declares a member that's already declared
+        // in a base type (not allowed in C#).
```
costleya: C# allows this. https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/new-modifier

rix0rrr: Well this is the error I get:
```
@aws-cdk/aws-iam: IRoleProxy.cs(113,21): error CS0111: Type 'IRoleProxy' already defines a member called 'AddToPolicy' with the same parameter types [/tmp/jsii-pacmak-codec20iQw/Amazon.CDK.AWS.IAM/Amazon.CDK.AWS.IAM.csproj]
```

rix0rrr: Yeah I guess 'new' will do it. I don't get the point of it though, and our C# generator doesn't do the necessary analysis to add the modifier, so I think disallowing in jsii is the easiest  way forward.

PR #366, opened 2019-02-22
feat(jsii-reflect) Add TypeSystem.hasAssembly

RomainMuller:
It can be useful to determine whether an assembly is present or not in a `TypeSystem`, for example when trying to find classes that extend a particular base -- if the assembly that base belongs to is not loaded, then it is guaranteed that no class can extend it. This allows avoiding to resort to an exception-driven control flow.

Implements some of #367


packages/jsii-reflect/lib/type-system.ts
```diff
@@ -102,6 +102,10 @@ export class TypeSystem {
     return asm;
   }
 
+  public hasAssembly(name: string): boolean {
```
eladb: Maybe `includesAssembly` in the spirit of `Array.includes`?

PR #372, opened 2019-03-07
fix(jsii): prohibit exported const enums

rix0rrr:
'const enums' are inlined at the usage site by TypeScript and so the
generated type will not be in the JavaScript source code in the
assembly, even though the declaration will be there.

This leads to "symbol not found" errors upon trying to load it.

https://www.typescriptlang.org/docs/handbook/enums.html#const-enums

Fixes awslabs/aws-cdk#1969

### DIFF
--- a/packages/jsii/lib/assembler.ts
+++ b/packages/jsii/lib/assembler.ts
@@ -482,6 +482,15 @@ export class Assembler implements Emitter {
             LOG.trace(`Processing enum: ${colors.gray(namespace.join('.'))}.${colors.cyan(type.symbol.name)}`);
         }
 
+        const decl = type.symbol.valueDeclaration;
+        const flags = ts.getCombinedModifierFlags(decl);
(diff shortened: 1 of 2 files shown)


packages/jsii/lib/assembler.ts
```diff
@@ -482,6 +482,15 @@ export class Assembler implements Emitter {
...
         }
 
+        const decl = type.symbol.valueDeclaration;
+        const flags = ts.getCombinedModifierFlags(decl);
+        // tslint:disable-next-line:no-bitwise
+        if (flags & ts.ModifierFlags.Const) {
+            this._diagnostic(decl,
+                             ts.DiagnosticCategory.Error,
+                            `Exported enum cannot be declared 'const'`);
+        }
```
RomainMuller: Should we return from there? So we don't include it in the `.jsii` assembly we'd create?

rix0rrr: No.

PR #376, opened 2019-03-12
feat: pass data types (structs) by-value instead of by-ref

### LINKED ISSUE #375: Values that are not set in Java builders should be treated as unset in JavaScript
See awslabs/aws-cdk#965

eladb:
Data types ("structs") are defined through a TypeScript interface that only contains properties and doesn't begin with an `I`. This change also adds a requirement that all properties (shallow) are marked `readonly`.

This allows passing around structs by-value safely because any consuming code would not expect to be able to write to the object.

Python already passes structs by value, especially in the context of when a struct is used as keyword arguments. This use case will also exist in Ruby.

This is also the general perception around data that's passed around in most programming languages.

To enforce this, the jsii compiler will now fail if a struct includes properties that are not marked `readonly`.

Both the Java and .NET generators have been modified to generate builders which allow users to construct structs by serializing them as JSON hashes instead of passed down by reference.

Existing compliance tests in all languages have been fixed.

This change fixes awslabs/aws-cdk#965 and fixes #375 by erasing any `null`s or `undefined` values from passed in objects, so they do not appear to have been defined at all. Added a compliance test called **erase-unset-data-values** to verify.

BREAKING CHANGE: all properties in interfaces which represent data types must be marked as `readonly`. Otherwise, jsii compilation will fail.

---


packages/jsii-pacmak/test/expected.jsii-calc-base/java/src/main/java/software/amazon/jsii/tests/calculator/base/BaseProps.java
```diff
@@ -68,6 +68,14 @@ public void setFoo(final software.amazon.jsii.tests.calculator.baseofbase.Very v
                     this.$foo = java.util.Objects.requireNonNull(value, "foo is required");
                 }
 
+                public com.fasterxml.jackson.databind.JsonNode $jsii$toJson() {
```
RomainMuller: But this is a mutable object... If passed by value, then "set" operations will not propagate back to the JVM, will they?

eladb: You are right, but practically we (the CDK) always treats these data types as immutable, and I would like to "officially" define in the jsii spec that data types are immutable. 

Sadly, it's really hard to express this in typescript today, but I am following up on https://github.com/Microsoft/TypeScript/pull/29510 to see if maybe we can enforce this at the typescript level by requiring that data types will be marked as "const".

I am wondering if perhaps for now we can change the jsii compiler to automatically define all properties of data types as readonly, so that generated language proxies can assume that and this won't be "incorrect".,

What do you think?

RomainMuller: Well... the `const` assertion described in Microsoft/TypeScript#29510 doesn't seem to apply to interface definitions (which is what we're looking at here) but to object literals. I guess what'd make this less gruesome would be if we could define a `readonly interface Foo { /*...*/ }` instead of having to mark all the individual fields as `readonly`.

The problem if we make all data type fields `readonly` in the JSII assembly (only) is that you can still perform mutations on the `TypeScript` and `Javascript` languages (because you don't go through a proxy there), meaning you are actually still able to write code that won't behave the way it was intended from other languages. This dissonance between language behaviors bothers me a lot, similar to how I am bothered by how `Token` in TypeScript can be passed a function, but not in any other language.

To a certain extent what is happening here if we do NOT mark fields `readonly` is we're effectively looking to achieve *by value* versus *by reference*. I think what you are hinting at is you would like to be able to specify a parameter as `const props: FooProps` (or `props: const FooProps`?) and this is how you'd signal the data type can be passed in *by value* (for it turns immutable on the way into the function). If I understand this right, then I am (very) supportive. But that still makes this PR rely on good intentions that we canot (easily) verify are applied.

eladb: Okay, here's what I am going to do (following online conversation with @rix0rrr and @RomainMuller):

Add enforcement in the jsii compiler (initially as a warning) to require that all properties defined on a `datatype` are marked with `readonly` (those properties will always be marked "readonly" in the jsii manifest even if the source did not have readonly).

RomainMuller: I approve, despite the cumbersomeness of the situation in the current state of things (hopefully we get syntax sugar in `tsc` via something like Microsoft/TypeScript#21152.

costleya:
I'm trying to get a sense of what is different for .NET developers. You mention:

> Both the Java and .NET generators have been modified to generate "builders" which allow users to construct these data types by serialize as JSON hashes instead of passed down by reference.

What does this look like? Are the conversions between .NET and JSII still transparent?

eladb:
> What does this look like? Are the conversions between .NET and JSII still transparent?

Yes, they are still transparent. The main change is that now when an object that implements a data type interface is passed down to jsii, it will be serialized as a json map instead of allocated as an anonymous reference.

eladb:
Copy: @nnir

eladb:
That's going to be fun!

rix0rrr:
Any chance it's a small effort to update the C# generator to emit (an overload with?) keyword arguments?

And maybe `null` defaults?

I know it should be a different change, but while you're in here anyway...

eladb:
Still waiting for an approval from:
- Python (@garnaat or @dstufft)
- .NET (@costleya)

dstufft:
That sounds fine to me!

PR #377, opened 2019-03-12
Abstract proxy classes need to inherent from parent's proxies

dstufft:
In the JSII, we generate a proxy class for any Abstract classes. This proxy class inherits from the abstract class, and does nothing but provide implementations of any abstract methods/properties that just call into the JSII. However, this only worked for abstract methods defined on this specific abstract class, not any parent abstract classes.

This PR resolves that issue, by building up a list of all abstract classes in an abstract class's inheritance tree, and then has the proxy depend on it. 

Because the actual name of the abstract class is not part of a module's public interface, this PR adds a ``jsii.proxy_for`` function, which takes an abstract class and returns the proxy class for it. 

#### Before:

```python
class Value(scope.jsii_calc_base.Base, metaclass=jsii.JSIIAbstractClass, jsii_type="@scope/jsii-calc-lib.Value"):
    @staticmethod
    def __jsii_proxy_class__():
        return _ValueProxy

    @property
    @jsii.member(jsii_name="value")
    @abc.abstractmethod
    def value(self) -> jsii.Number:
        ...

    ...


class _ValueProxy(Value):
    @property
    @jsii.member(jsii_name="value")
    def value(self) -> jsii.Number:
        return jsii.get(self, "value")
```

#### After:

```python
class Value(scope.jsii_calc_base.Base, metaclass=jsii.JSIIAbstractClass, jsii_type="@scope/jsii-calc-lib.Value"):
    @staticmethod
    def __jsii_proxy_class__():
        return _ValueProxy

    @property
    @jsii.member(jsii_name="value")
    @abc.abstractmethod
    def value(self) -> jsii.Number:
        ...

    ...


class _ValueProxy(Value, jsii.proxy_for(scope.jsii_calc_base.Base)):
    @property
    @jsii.member(jsii_name="value")
    def value(self) -> jsii.Number:
        return jsii.get(self, "value")
```

### DIFF
--- a/packages/jsii-pacmak/lib/targets/python.ts
+++ b/packages/jsii-pacmak/lib/targets/python.ts
@@ -682,20 +682,23 @@ class TypedDictProperty implements PythonBase {
 interface ClassOpts extends PythonTypeOpts {
     abstract?: boolean;
     interfaces?: spec.NamedTypeReference[];
+    abstractBases?: spec.ClassType[];
 }
 
 class Class extends BasePythonClassType {
 
     private abstract: boolean;
+    private abstractBases: spec.ClassType[];
     private interfaces: spec.NamedTypeReference[];
 
     constructor(name: string, fqn: string, opts: ClassOpts) {
         super(name, fqn, opts);
 
-        const { abstract = false, interfaces = [] } = opts;
+        const { abstract = false, interfaces = [], abstractBases = [] } = opts;
 
         this.abstract = abstract;
         this.interfaces = interfaces;
+        this.abstractBases = abstractBases;
     }
 
     public dependsOn(resolver: TypeResolver): PythonType[] {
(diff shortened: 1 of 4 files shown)


packages/jsii-pacmak/lib/targets/python.ts
```diff
@@ -1580,4 +1590,22 @@ class PythonGenerator extends Generator {
...
     }
+
+    private getAbstractBases(cls: spec.ClassType): spec.ClassType[] {
+        const abstractBases: spec.ClassType[] = [];
+
+        if (cls.base !== undefined) {
+            const base = this.findType(cls.base.fqn);
+
+            if (!spec.isClassType(base)) {
+                throw new Error("Class inheritence that isn't a class?");
```
garnaat: This isn't a super-useful error message.  Under what circumstances would we end up here?

dstufft: Theoretically never I think. We would end up here if a class inherited from something that wasn't a class, like a primitive type or an interface or something, which shouldn't be possible AFAIK. However the types technically allow it so it's possible that a bug or something causes the JSII gets into a state where it has a class inheriting from something that isn't a class.

This basically exists for two reasons:

* ``findType`` doesn't return a ``ClassType``, it returns a ``Type`` and using ``spec.IsClassType`` like this informs the compiler that base is a ``ClassType`` if it gets past this if statement.
* Emulates an ``assert`` statement to verify that some assumption of the world holds true, but that should never *not* be true.

garnaat: Okay, understood.

PR #387, opened 2019-03-18
Implement Async Function Support

dstufft:
Implements async function support, exposing it the same as a sync function in the Python language.

This is basically done, but will need to coordinate with sync callbacks to make sure we're not doing anything drastically different for the bits that are going to be shared.

### DIFF
--- a/packages/jsii-python-runtime/src/jsii/_kernel/__init__.py
+++ b/packages/jsii-python-runtime/src/jsii/_kernel/__init__.py
@@ -15,8 +15,12 @@ from jsii._kernel.types import JSClass, Referenceable
 from jsii._kernel.types import (
     EnumRef,
     LoadRequest,
+    BeginRequest,
+    CallbacksRequest,
     CreateRequest,
+    CompleteRequest,
     DeleteRequest,
+    EndRequest,
     GetRequest,
     InvokeRequest,
     SetRequest,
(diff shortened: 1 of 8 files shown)


packages/jsii-python-runtime/src/jsii/_kernel/__init__.py
```diff
@@ -36,6 +40,12 @@ class Object:
     __jsii_type__ = "Object"
 
 
+def _handle_callback(kernel, callback):
```
dstufft: Note, this doesn't actually handle everything, this is one of the things that we need to coordinate with the sync callbacks with over who wants to "own" doing the complete implementation of this.

packages/jsii-python-runtime/tests/test_compliance.py
```diff
@@ -49,8 +49,7 @@
 #       Tests as closely as possible to make keeping them in sync easier.
 
 # These map distinct reasons for failures, so we an easily find them.
-xfail_async = pytest.mark.xfail(reason="Implement async methods", strict=True)
-xfail_callbacks = pytest.mark.xfail(reason="Implement callback support", strict=True)
+xfail_callbacks = pytest.mark.skip(reason="Implement callback support")
```
dstufft: I had to switch this to a skip, because running these tests and failing on them left the process in a bad state.

packages/jsii-python-runtime/src/jsii/_kernel/__init__.py
garnaat: Do you just want to merge this version and then I can do a PR with the updated version?

dstufft: That works for me.

Issue #287, opened 2018-10-30
Ensure Data Types have a "pure" inheritance tree


dstufft:
Data Types and "Normal" interfaces need to have distinct inheritance trees for the Python generated code to function. IOW, a data type can't inherent from a non data type, and a non data type can't inherit from a data type.

https://github.com/awslabs/jsii/pull/271 should solve the case where a data type cannot inherit from a non data type, however it's still the case that a non data type can inherit from a data type. IOW this currently is accepted:

```typescript

export interface SomeProps {
    foo: str
}

export interface ISomething extends SomeProps {
    bar(): void;
}
```

However, that could would cause an error in Python, because a "normal" interface cannot inherit from a data type, so we should reject that in the JSII compiler.

rix0rrr:
This should already be the case. I'm fairly sure I added checks to this effect. No?

rix0rrr:
Can't seem to find a test for it though ;(

dstufft:
My memory is that the checks only existed for things that start with ``I``, but that things that don't start with ``I`` could still exhibit this problem.

RomainMuller:
Existing check (`jsii` already rejects):
- [X] An interface extending another one which is _not_ a struct, _cannot_ be a struct.

Missing checks (`jsii` will happily proceed):
- [ ] A class may not implement a struct interface
- [ ] A regular interface may not implement a struct

PR #390, opened 2019-03-18
feat: internal accessibility

### LINKED ISSUE #388: strip elements marked with @internal jsdoc tag
In accordance with typescript's [--strip-internal](http://www.typescriptlang.org/docs/handbook/compiler-options.html), we should enable this feature by default and omit any declarations marked `@internal` from the jsii manifest so they don't appear in the API (similar to underscore prefix).

eladb:
Respect the `@internal` jsdoc tag on types and members. Any type of member marked internal will not be exposed in the API of the module and in .d.ts. files (through `--strip-internal`).

Members (properties/methods) that are marked as `@internal` must also have an underscore prefix (i.e. `_myMethod`) and vice versa in order to ensure that implementers or subclasses won't modify  accessibility.

If a class implements (or an interface extends) an internal or private interface, that interface is erased from the implementer's API. So it is possible to use internal interfaces as bases, while still maintaining API integrity (related to #287).

Fixes #388

BREAKING CHANGE: member names that begin with underscore now must be marked as "@internal" in their jsdocs, which will cause them to disappear from type declaration files and jsii APIs.

---


packages/jsii-calc/lib/compliance.ts
```diff
@@ -1335,3 +1335,34 @@ export class Constructors {
...
     }
 }
+
+// internal can be used to represent members that can only be accessed from the current module
+export class StripInternal {
+    public youSeeMe = 'hello';
+
+    /**
+     * This is an internal thing
+     * @internal
```
RomainMuller: Are we doing this as a replacement for names starting with a `_`? Additional way to achieve the same behavior?

packages/jsii-calc/lib/compliance.ts
```diff
@@ -1335,3 +1335,34 @@ export class Constructors {
...
+ * @internal
+ */
+export class InternalClass {
+    public iAmNotHere = 'yes';
+}
+
+/**
+ * @internal
+ */
+export interface IInternalInterface { }
```
RomainMuller: What happens when I implement this `IInternalInterface` from a non-`@internal` type? I feel there's a host of missing test cases here that I suspect will blow up because the changes in the compiler itself are relatively minor in comparison to what I had expected to find...

RomainMuller:
Other question that occurred to me now... is `@internal` on a method inherited by overrides or not? If not, the case for making a single method on an `interface` internal is really awkward, since that'd have to be repeated for every single implementation of it...

eladb:
> Other question that occurred to me now... is @internal on a method inherited by overrides or not? If not, the case for making a single method on an interface internal is really awkward, since that'd have to be repeated for every single implementation of it...

That's a good question. Let me check

eladb:
So the answer is that you have to explicitly define @internal on all inherited members, or otherwise the typescript compiler will make the method visible. Does it make sense to enforce this with jsii? I guess it does, no?

packages/jsii/test/negatives/neg.internal-underscore.3.ts
```diff
@@ -0,0 +1,6 @@
+///!MATCH_ERROR: the name of members marked as @internal must begin with an underscore
+
+export interface IMyInterface {
+  /** @internal */
+  methodWithInternalButNoUnderscore: string;
```
rix0rrr: There is no difference between this test and neg 1.

eladb: Was supposed to be a class. Will fix. Thanks

packages/jsii-calc/lib/compliance.ts
eladb: Addressed.

packages/jsii/test/negatives/neg.internal-underscore.3.ts
eladb: Fixed.

packages/jsii/lib/assembler.ts
```diff
@@ -479,11 +505,47 @@ export class Assembler implements Emitter {
...
-    private async _visitEnum(type: ts.Type, namespace: string[]): Promise<spec.EnumType> {
+    /**
+     * @returns true if this member is internal and should be omitted from the type manifest
+     */
+    private _isPrivateOrInternal(symbol: ts.Symbol, validateDeclaration?: ts.Declaration): boolean {
+        const hasInternalJsDocTag = _hasInternalJsDocTag(symbol);
+        const hasUnderscorePrefix = symbol.name !== '__constructor' && symbol.name.startsWith('_');
+
+        if (_isPrivate(symbol)) {
+            LOG.trace(`skipping ${symbol.name} because it is marked "private"`);
```
RomainMuller: That trace here might be misleading - this is just returning a boolean, you cannot presume what is done with it.

packages/jsii/lib/assembler.ts
```diff
@@ -569,13 +635,21 @@ export class Assembler implements Emitter {
         };
 
         for (const base of (type.getBaseTypes() || [])) {
+            // if this interface is internal/private, we erase it from the API definition
+            if (this._isPrivateOrInternal(base.symbol)) {
+                LOG.trace(`erasing hidden interface ${base.symbol.name}`);
```
RomainMuller: You'll want to mention what type you're erasing a hidden base from.

packages/jsii/lib/assembler.ts
```diff
@@ -479,11 +505,47 @@ export class Assembler implements Emitter {
...
+
+        if (!hasInternalJsDocTag && !hasUnderscorePrefix) {
+            return false;
+        }
+
+        // we only validate if we have a declaration
+        if (validateDeclaration) {
+            if (!hasUnderscorePrefix) {
+                this._diagnostic(validateDeclaration, ts.DiagnosticCategory.Error,
+                    `${symbol.name}: the name of members marked as @internal must begin with an underscore`);
```
RomainMuller: I don't know if you need to mention `symbol.name` here, since you're attaching the diagnostic to the declaration.

packages/jsii/lib/assembler.ts
```diff
@@ -384,27 +390,35 @@ export class Assembler implements Emitter {
                 this._diagnostic(clause, ts.DiagnosticCategory.Error, `Ignoring ${ts.SyntaxKind[clause.token]} heritage clause`);
                 continue;
             }
+
             for (const expression of clause.types) {
                 const iface = this._typeChecker.getTypeFromTypeNode(expression);
+
+                // if this interface is internal/private, we erase it from the API definition
+                if (this._isPrivateOrInternal(iface.symbol)) {
+                    LOG.trace(`erasing hidden interface ${iface.symbol.name}`);
```
RomainMuller: Erased from where?

packages/jsii/lib/assembler.ts
```diff
@@ -337,11 +337,15 @@ export class Assembler implements Emitter {
         return [jsiiType];
     }
 
-    private async _visitClass(type: ts.Type, namespace: string[]): Promise<spec.ClassType> {
+    private async _visitClass(type: ts.Type, namespace: string[]): Promise<spec.ClassType | undefined> {
```
RomainMuller: I don't think we're taking care of the `@internal` base-class case here. In such cases, you need to look at the base's base until there is none left.

packages/jsii/lib/assembler.ts
```diff
@@ -384,27 +390,35 @@ export class Assembler implements Emitter {
...
                 continue;
             }
+
             for (const expression of clause.types) {
                 const iface = this._typeChecker.getTypeFromTypeNode(expression);
+
+                // if this interface is internal/private, we erase it from the API definition
+                if (this._isPrivateOrInternal(iface.symbol)) {
+                    LOG.trace(`erasing hidden interface ${iface.symbol.name}`);
+                    continue;
```
RomainMuller: When the interface is `@internal`, then you want to check if that interface has non-`@internal` super-interfaces and register those instead (because interface implementation/extension is transitive). This is also true for `_visitInterface`.

packages/jsii/lib/assembler.ts
eladb: Fixed.

packages/jsii/lib/assembler.ts
eladb: From my experience with the errors this was helpful

packages/jsii/lib/assembler.ts
eladb: added

packages/jsii/lib/assembler.ts
eladb: Good point. Shoot.

packages/jsii/lib/assembler.ts
eladb: I believe we do. I am adding a test to that extent. The error will be "Unable to resolve referenced type 'jsii.InternalBaseClass'. Missing export?"

PR #392, opened 2019-03-19
fix: copy non-hidden bases when erasing hidden interfaces

eladb:
When a type inherits (extends/implements) from a hidden (private/internal)
interface, and that interface extends a non-hidden interface, we need to 
copy the non-hidden interface to the consuming type so that it can be polymorphically
used.

Follow up on #390

Co-authored-by: RomainMuller <rmuller@amazon.com>


packages/jsii/test/negatives/neg.implements-class.ts
```diff
@@ -1,4 +1,4 @@
-///!MATCH_ERROR: Implements clause of jsii.TryingToImplementClass uses jsii.NotAnInterface as an interface
+///!MATCH_ERROR: Inheritence clause of jsii.TryingToImplementClass uses jsii.NotAnInterface as an interface
```
RomainMuller: I reckon this should be `Inheritance` 😞 (Obv. this comes from the error message :D)

eladb: fixed

PR #395, opened 2019-03-20
fix(runtime): Passing 'this' to a callback from constructor

RomainMuller:
In Javascript, a constructor is allowed to pass references to `this` to
method calls (effectively passing a partially initialized instance).
When done, the kernel will assign an Object ID to `this` on the spot in
order to pass the reference through the JSII boundary. However, the
`create` API before this change would have attempted to re-allocate an
Object ID to the object (which is illegal and causes a crash).

In addition, callbacks from at least Java and .NET runtimes could not
receive JSII object references, for they would not turn them into native
objects before passing them to the implementation, resulting in a class
conversion error.


packages/jsii-java-runtime/project/src/main/java/software/amazon/jsii/JsiiEngine.java
```diff
@@ -362,7 +363,28 @@ private JsonNode invokeCallbackSet(final SetRequest req) {
...
-        return OM.valueToTree(invokeMethod(obj, method, req.getArgs().toArray()));
+        return OM.valueToTree(invokeMethod(obj,
+                                           method,
+                                           req.getArgs()
+                                              .stream()
+                                              .map(this::fromKernel)
+                                              .toArray()));
+    }
+
+    private Object fromKernel(final Object object) {
```
eladb: Can you add some commentary explaining what this method is doing

eladb: This seems like a duplicate of [`JsiiObjectMapper.treeToValue`](https://github.com/awslabs/jsii/blob/dd07831003f542ec67b6c64d369271ff0e0b1b44/packages/jsii-java-runtime/project/src/main/java/software/amazon/jsii/JsiiObjectMapper.java#L96). I am assuming .NET would have a similar method to deserialize from jsii.

I am also wondering if there are other cases (such as enums, dates) that would fail in this situation.

packages/jsii-pacmak/test/expected.jsii-calc/dotnet/Amazon.JSII.Tests.CalculatorPackageId/Amazon/JSII/Tests/CalculatorNamespace/PartiallyInitializedThisConsumerProxy.cs
```diff
@@ -0,0 +1,18 @@
+using Amazon.JSII.Runtime.Deputy;
+
+namespace Amazon.JSII.Tests.CalculatorNamespace
+{
+    [JsiiTypeProxy(typeof(PartiallyInitializedThisConsumer), "jsii-calc.PartiallyInitializedThisConsumer")]
```
costleya: If this is an interface that customers are actually supposed to implement, and not simply a pointer to implementations that were not exported, then proxies should not be generated for this interface. We might need to add an extra JSII flag to denote this difference.

RomainMuller: That's an abstract class, not an interface, actually. And I don't see why the JS-land couldn't send an instance of that out to the .NET land, so the .NET land would still need the proxy/deputy.

RomainMuller:
I have a further change to improve the new compliance tests' feature coverage (include a date and an enum value), however this fell on a bug that @rix0rrr fixed in #401, so I will wait until that was merged before I update the PR with that.

packages/jsii-java-runtime/project/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
```diff
@@ -0,0 +1 @@
+mock-maker-inline
```
eladb: what is this?

packages/jsii-python-runtime/tests/test_compliance.py
```diff
@@ -895,3 +897,14 @@ def test_eraseUnsetDataValues():
     }
     assert EraseUndefinedHashValues.does_key_exist(opts, "option1")
     assert not EraseUndefinedHashValues.does_key_exist(opts, "option2")
+
+
+@xfail_callbacks
```
eladb: issue

eladb:
I meant "nothing interesting to say". I love this change and that we are using Jackson like the Java gods intended

packages/jsii-calc/package.json
```diff
@@ -36,11 +36,13 @@
   ],
   "dependencies": {
     "@scope/jsii-calc-base": "^0.8.0",
+    "@scope/jsii-calc-base-of-base": "^0.8.0",
```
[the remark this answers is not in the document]
RomainMuller: It suppresses a JSII warning because some of the types in there are re-exported... So it needs to be a peer, too... And peer deps should also be modeled in dependencies (or should they?)

packages/jsii-dotnet-runtime/src/Amazon.JSII.Runtime/CallbackExtensions.cs
```diff
@@ -117,7 +119,20 @@ static void InvokeSetter(SetRequest request, IReferenceMap referenceMap)
                 throw new InvalidOperationException($"Received callback for {deputy.GetType().Name}.{request.Property} setter, but this property does not have a setter");
             }
 
-            methodInfo.Invoke(deputy, new object[] { request.Value });
+            methodInfo.Invoke(deputy, new object[] { FromKernel(request.Value, referenceMap) });
+        }
+
+        private static object FromKernel(object obj, IReferenceMap referenceMap)
```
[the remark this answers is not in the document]
RomainMuller: Of course! I thought I had done this...

packages/jsii-java-runtime/project/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
RomainMuller: This is so I can use `Mockito.mock` on `final` classes.

packages/jsii-python-runtime/tests/test_compliance.py
RomainMuller: I believe @garnaat is on top of this and I don't want to introduce an unnecessary merge conflict on his way... ¯\_(シ)_/¯

PR #401, opened 2019-03-26
fix(kernel): make type serialization explicit and recursive

rix0rrr:
Serialize and deserialize types according to their declared static type,
and add validation on the runtime types matching the declared types.

This is in contrast to previously, when we mostly only used the runtime
types to determine what to do, and hardly any validation was done. The
runtime types used to be able to freely disagree with the declared
types, and we put a lot of burden on the JSII runtimes at the other
end of the wire.

Fix tests that used to exercise the API with invalid arguments.

Remove Proxy objects since they only existed to prevent accidentally
overwriting certain keys via the jsii interface, and Symbols serve
the same purpose (but simpler).

Fixes awslabs/aws-cdk#1981.


packages/jsii-kernel/lib/serialization.ts
```diff
@@ -0,0 +1,622 @@
+
+ // tslint:disable:max-line-length
+/**
+ * Handling of types in JSII
+ *
+ * Types will be serialized according to the following table:
+ *
+ *                         ┬───────────────────────────────────────────────────────────────────────────────────────────────┐
```
eladb: Very cool

packages/jsii-kernel/lib/serialization.ts
```diff
@@ -0,0 +1,622 @@
...
+    },
+    deserialize(value) {
+      return value;
+    },
+  },
+
+  // ----------------------------------------------------------------------
+  [SerializationClass.Enum]: {
+    serialize(value, type, host): WireEnum | undefined {
+      host.debug('hullo');
```
eladb: remove (not really important).

packages/jsii-kernel/lib/serialization.ts
```diff
@@ -0,0 +1,622 @@
...
+    deserialize(value, type, host) {
+      if (nullAndOk(value, type)) { return undefined; }
+
+      const mapType = type as spec.CollectionTypeReference;
+      return mapValues(value, v => host.recurse(v, mapType.collection.elementtype));
+    },
+  },
+
+  // ----------------------------------------------------------------------
+  [SerializationClass.Struct]: {
```
eladb: go structs

packages/jsii-kernel/lib/serialization.ts
```diff
@@ -0,0 +1,622 @@
...
+      if (nullAndOk(value, type)) { return undefined; }
+
+      if (typeof value !== 'object' || value == null) {
+        throw new Error(`Expected object, got ${JSON.stringify(value)}`);
+      }
+
+      // This looks odd, but if an object was originally passed in as a by-ref
+      // class, and it happens to conform to a datatype interface we say we're
+      // returning, return the actual object instead of the serialized value.
+      // NOTE: Not entirely sure yet whether this is a bug masquerading as a
```
eladb: I think it's okay for the kernel to allow passing in structs by reference. It should Just Work I guess

RomainMuller: Yeah I think it can be backwards-compatible.

PR #403, opened 2019-03-26
Misc Cleanups for the Python Runtime

dstufft:
- Adds the License information for the `jsii` library.
- Add the Description/README.md information to the Python metadata, so it will show up on PyPI.
- Add some Classifiers to tag the JSII with some basic metadata about Python support, project lifecycle, target audience, etc.

### DIFF
--- a/packages/jsii-python-runtime/.gitignore
+++ b/packages/jsii-python-runtime/.gitignore
@@ -10,3 +10,5 @@ dist/
 
 __pycache__
 .mypy_cache
+
+README.md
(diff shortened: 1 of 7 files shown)


packages/jsii-python-runtime/.gitignore
```diff
@@ -10,3 +10,5 @@ dist/
 
 __pycache__
 .mypy_cache
+
+README.md
```
eladb: Why ignore?

dstufft: Because it's copied over from the top level directory and thus is a generated file in this directory.

PR #407, opened 2019-03-28
feat(python): Add support for synchronous callbacks

### LINKED ISSUE #405: Python: Ensure callback arguments handled correctly
As seen in #395, there was an issue in the Java and .NET runtimes (ref #404) around parsing out callback arguments into the correct native object.

The Python runtime should at least implement the same compliance test as is introduced in #395.

garnaat:
This PR adds support for synchronous callbacks in Python as well as a number of other small fixes.

Fixes #405

---


packages/jsii-python-runtime/src/jsii/_kernel/__init__.py
```diff
@@ -120,6 +123,39 @@ def _make_reference_for_native(kernel, d):
...
+        method = getattr(obj, callback.cookie)
+        return method(*callback.invoke.args)
+    elif callback.get:
+        obj = _reference_map.resolve_id(callback.get.objref.ref)
+        return getattr(obj, callback.cookie)
+    elif callback.set:
+        obj = _reference_map.resolve_id(callback.set.objref.ref)
+        return setattr(obj, callback.cookie, callback.set.value)
+    else:
+        raise AttributeError
```
dstufft: Do we want this to be a ``JSIIError`` or something? Or at the very least have a descriptive message? Ideally we should *never* hit this unless the JSII adds another type of callback, so it's basically just asserting a known assumption about the world.

packages/jsii-python-runtime/src/jsii/_kernel/providers/process.py
```diff
@@ -244,7 +256,9 @@ def _jsii_runtime(self):
         return paths[0]
 
     def _next_message(self) -> Mapping[Any, Any]:
-        return json.loads(self._process.stdout.readline(), object_hook=ohook)
+        # return json.loads(self._process.stdout.readline(), object_hook=ohook)
```
dstufft: Probably need to just delete the line of code here? Or Uncomment it and delete the next two lines.

packages/jsii-python-runtime/src/jsii/_kernel/providers/process.py
```diff
@@ -289,13 +303,15 @@ def send(
         # Send our data, ensure that it is framed with a trailing \n
         self._process.stdin.write(b"%b\n" % (data,))
         self._process.stdin.flush()
-
+
```
dstufft: Probably want to strip this whitespace, a well as the other bit of whitespace added.

packages/jsii-python-runtime/tests/test_compliance.py
```diff
@@ -899,10 +886,9 @@ def test_eraseUnsetDataValues():
     assert not EraseUndefinedHashValues.does_key_exist(opts, "option2")
 
 
-@xfail_callbacks
 def test_objectIdDoesNotGetReallocatedWhenTheConstructorPassesThisOut():
     class PartiallyInitializedThisConsumerImpl(PartiallyInitializedThisConsumer):
-        def consume_partially_initialized_this(self):
+        def consume_partially_initialized_this(self, obj, dt, en):
```
RomainMuller: Would be nice if you can also assert that `dt` and `en` have the correct values (like the Java implementation does).

eladb:
👯

PR #408, opened 2019-03-28
fix(python): Lift the entire data class hierarchy

dstufft:
They Python code generation was only lifting the properties of the specified data class, but it wasn't lifting the properties of all of the data classes in the inheritance tree.

We have to do that, so we'll go ahead and walk our entire interface hierarchy and build up a flat list of all of the properties, and use that list to generate our lifted properties.

---

### DIFF
--- a/packages/jsii-pacmak/lib/targets/python.ts
+++ b/packages/jsii-pacmak/lib/targets/python.ts
@@ -363,14 +363,15 @@ abstract class BaseMethod implements PythonBase {
         if (this.liftedProp !== undefined) {
             // Remove our last item.
             pythonParams.pop();
+            const liftedProperties = this.getLiftedProperties(resolver);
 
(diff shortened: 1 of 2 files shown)


packages/jsii-pacmak/lib/targets/python.ts
```diff
@@ -476,6 +477,29 @@ abstract class BaseMethod implements PythonBase {
...
+
+        const stack = [this.liftedProp];
+        let current = stack.shift();
+        while (current !== undefined) {
+            // Add any interfaces that this interface depends on, to the list.
+            if (current.interfaces !== undefined) {
+                stack.push(...current.interfaces.map(ifc => resolver.dereference(ifc) as spec.InterfaceType));
+            }
+
+            // Add all of the properties of this interface to our list of properties.
```
eladb: Recursion always reads easier to me, but that's just preference

PR #416, opened 2019-04-01
feat(kernel): Normalize empty structs to undefined

### LINKED ISSUE #411: Kernel: empty structs should be turned to undefined
Keyword argument lifting makes it impossible to differentiate an argument not passed from an argument passed with all attributes `undefined`, hence we need to treat `{}` as `undefined` when passing empty structs to the JavaScript code.

RomainMuller:
Upon deserializing objects from a "native" front-end (Java, .NET, ...),
the kernel now interprets an empty value for a struct type (aka data
type) as `undefined`, since front-ends that perform keyword lifting such
as Python will not be able to distinguish between "no argument passed"
and "an empty argument passed".

Fixes #411

### DIFF
--- a/packages/jsii-calc/lib/compliance.ts
+++ b/packages/jsii-calc/lib/compliance.ts
@@ -1587,7 +1587,7 @@ export class ConsumersOfThisCrazyTypeSystem {
 
 //
 // Ensure the JSII kernel can pass "this" out to JSII remotes from within the constructor (this is dirty, but possible)
-///
+//
 export abstract class PartiallyInitializedThisConsumer {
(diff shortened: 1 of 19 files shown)


packages/jsii-calc/lib/compliance.ts
```diff
@@ -1600,3 +1600,19 @@ export class ConstructorPassesThisOut {
         }
     }
 }
+
+//
+// Consumes a possibly empty struct and verifies it is turned to undefined when passed
+//
```
eladb: Add a reference to an issue

eladb:
Shall we get this merged?

PR #421, opened 2019-04-02
feat(jsii): Enforce use of peerDependencies

### LINKED ISSUE #361: Enforce peer dependencies
JSII modules should enforce that any dependency they take is also a peer dependency, as this is the least common denominator of all the runtimes that would be supported (aka it's the "easiest" way to manage dependencies).

RomainMuller:
Any direct dependency to another jsii module must be recorded as a
peerDependency as well, guaranteeing a consistent behavior of
dependencies across all possible jsii runtimes.

BREAKING CHANGE: All direct dependencies must be duplicated in
                 peerDependencies unless they are in bundledDependencies.

Fixes #361


packages/jsii/lib/project-info.ts
```diff
@@ -38,23 +38,55 @@ export interface ProjectInfo {
     readonly projectReferences?: boolean;
 }
 
-export async function loadProjectInfo(projectRoot: string): Promise<ProjectInfo> {
-    const pkg = require(path.join(projectRoot, 'package.json'));
+export async function loadProjectInfo(projectRoot: string, fixPeerDependencies: boolean): Promise<ProjectInfo> {
```
eladb: use keyword arguments for `fixPeerDeps`

PR #425, opened 2019-04-03
feat(jsii): Erase un-exported base classes instead of prohibiting those

### LINKED ISSUE #417: Erase non-exported base classes
If an exported class extends a non-exported class, it should be possible to erase the non-exported class from the public API by copying it's interface to the derived class. This is allowed in TypeScript and there is no reason it won't be allowed in jsii.

The compiler should respect any public bases of the erased class and copy those as well to the derived class.

This is already supported for interfaces.

Repro:

```ts
export class PublicBaseOfBase {
  public foo() { }
}

class PrivateBase extends PublicBaseOfBase {
  public goo() { }
}

export class Derived extends PrivateBase {
  public noo() { }
}
```

After erasure, the API of `Derived` should be:

```ts
class Derived extends PublicBaseOfBase {
  public goo() { }
  public noo() { }
}
```

RomainMuller:
When an exported class extends an un-exported class, merge the declarations
of the un-exported base into the exported one, and replace the base class
relationship with the base's own base (recursively until an exported base is
found, or no further base class is declared - in which case the exported class
will not have a base class at all).

Declarations from the erased base class(es) will be ignored if there is another
declaration with the same name and type in the exported type (or another
erased base class that is closer to the exported type in the inheritance chain).

Fixes #417


packages/jsii/lib/assembler.ts
```diff
@@ -467,14 +479,18 @@ export class Assembler implements Emitter {
       throw new Error('Oh no');
     }
 
-    for (const decl of type.symbol.declarations) {
+    const allDeclarations: Array<{ decl: ts.Declaration, type: ts.InterfaceType | ts.BaseType }>
+      = type.symbol.declarations.map(decl => ({ decl, type }));
+    // Considering erased bases' declarations, too, so they are "blended in"
+    erasedBases.forEach(base => allDeclarations.push(...base.symbol.declarations.map(decl => ({ decl, type: base }))));
```
eladb: Can you convert to a for-loop? Makes it much easier for me to read :-)... In general, would appreciate if the code would be a little less condensed (I am old...)

packages/jsii/lib/assembler.ts
```diff
@@ -417,12 +417,24 @@ export class Assembler implements Emitter {
...
+
+    const erasedBases = new Array<ts.BaseType>();
+    for (let base of (type.getBaseTypes() || [])) {
       if (jsiiType.base) {
         this._diagnostic(base.symbol.valueDeclaration, ts.DiagnosticCategory.Error, `Found multiple base types for ${jsiiType.fqn}`);
         continue;
       }
 
+      // tslint:disable-next-line: no-bitwise
+      while (base && (ts.getCombinedModifierFlags(base.symbol.valueDeclaration) & ts.ModifierFlags.Export) === 0) {
```
eladb: A little comment explaining what this is doing.

packages/jsii/lib/assembler.ts
```diff
@@ -501,7 +517,9 @@ export class Assembler implements Emitter {
       }
     }
 
-    const constructor = type.symbol.members && type.symbol.members.get(ts.InternalSymbolName.Constructor);
+    const constructor = (type.symbol.members && type.symbol.members.get(ts.InternalSymbolName.Constructor))
+      // If no local constructor, look for one in the erased bases
+      || (erasedBases.map(base => base.symbol.members && base.symbol.members.get(ts.InternalSymbolName.Constructor)).find(ctor => ctor != null));
```
eladb: style: expand to multiple line - each operator in the chain in a separate line will go a long way for readability

packages/jsii/lib/assembler.ts
RomainMuller: I value vertical context, which is why I don't put a lot of vertical whitespace instinctively...

eladb: I value both. That’s why I like empty lines

packages/jsii/lib/assembler.ts
RomainMuller: Actually improved this even further by factoring the "complex" expression caused by the lack of a JS "safe traversal" operator (🧐) into a separate method. This looks dope now!

RomainMuller:
Holding off from merging this until #415 lands, so as to avoid creating unnecessary merge conflicts.

PR #429, opened 2019-04-04
feat(jsii): record source locations in assembly

rix0rrr:
Add the source locations of symbols to the JSII assembly, and give jsii-reflect some routines to query those, so that we'll be able to generate source location hyperlinks in the API reference.


.vscode/tasks.json
```diff
@@ -0,0 +1,15 @@
+{
```
RomainMuller: Please no!

packages/jsii-reflect/lib/method.ts
```diff
@@ -109,4 +110,18 @@ export class Method implements Documentable, Overridable, TypeMember {
 
   public isMethod(): this is Method { return true; }
   public isProperty(): this is Property { return false; }
+
+  /**
+   * Return the location in the module
+   */
+  public get moduleLocation(): SourceLocation | undefined {
```
RomainMuller: The name suggests it'll return the location **of** the module, not **in** the module. I don't really like how this is named `moduleLocation` anyway, I'd rather have this named `sourceLocation`, really.

packages/jsii-reflect/lib/source.ts
```diff
@@ -0,0 +1,49 @@
+import { Assembly } from "./assembly";
+
+export interface SourceLocation {
+  filename: string;
+  line: number;
+}
+
+interface Locatable {
+  readonly assembly: Assembly;
+  readonly moduleLocation?: SourceLocation;
```
RomainMuller: Would name `soureLocation`, or maybe even just `location`.

packages/jsii-reflect/lib/source.ts
```diff
@@ -0,0 +1,49 @@
+import { Assembly } from "./assembly";
+
+export interface SourceLocation {
+  filename: string;
+  line: number;
+}
+
+interface Locatable {
```
RomainMuller: Lacks documentation.

packages/jsii-reflect/lib/source.ts
```diff
@@ -0,0 +1,49 @@
+import { Assembly } from "./assembly";
+
+export interface SourceLocation {
```
RomainMuller: Lacks documentation.

packages/jsii-reflect/lib/source.ts
```diff
@@ -0,0 +1,49 @@
...
+ * (Currently only supports GitHub URLs)
+ */
+export function repositoryUrl(item: Locatable, ref: string = 'master'): string | undefined {
+  const loc = repositoryLocation(item);
+  if (!loc) { return undefined; }
+
+  const repo = item.assembly.repository;
+  if (!repo.url.startsWith('https://') || repo.url.indexOf("github.com") === -1 || !repo.url.endsWith('.git')) { return undefined; }
+
+  // Turn https://github.com/awslabs/aws-cdk.git ->  https://github.com/awslabs/aws-cdk/blob/REF/filename#L<number>
```
RomainMuller: Would be interesting to look into making this commit or tag-specific, so you get to the exact version of the file that corresponds to the doc you're looking at.

packages/jsii-reflect/package.json
```diff
@@ -28,6 +28,7 @@
     "jest": "^24.1.0",
     "jsii-build-tools": "^0.8.2",
     "jsii-calc": "^0.8.2",
+    "jsii": "^0.8.2",
     "ts-jest": "^24.0.0",
     "typescript": "^3.3.3333"
```
RomainMuller: If you depend on `jsii`, you shouldn't depend on `typescript` anymore...

packages/jsii-spec/lib/spec.ts
```diff
@@ -100,6 +100,15 @@ export interface Assembly extends Documentable {
...
     readme?: { markdown: string };
+
+    /**
+     * Source directory of the module root, relative to the repository root.
+     *
+     * If undefined or the empty string, no source location information is
+     * available. If the module root is equal to the root of the repository,
+     * the value is '.'.
+     */
+    repositoryLocation?: string;
```
RomainMuller: `repositoryPath`? `pathInRepository`?

packages/jsii-spec/lib/spec.ts
```diff
@@ -432,6 +456,11 @@ export interface Property extends Documentable, Overridable {
      * Implies `static` and `immutable`.
      */
     const?: boolean;
+
+    /**
+     * Where in the module this definition was found
+     */
+    moduleLocation?: SourceLocation;
```
RomainMuller: `sourceLocation?: SourceLocation`.

Also, since this is present on a number of things, why not introduce a `SourceLocatable` interface and extend that?

packages/jsii/lib/assembler.ts
```diff
@@ -345,6 +348,15 @@ export class Assembler implements Emitter {
     return [jsiiType];
   }
 
+  private declarationLocation(node: ts.Declaration): spec.SourceLocation {
+    const file = node.getSourceFile();
+    const line = ts.getLineAndCharacterOfPosition(file, node.getStart()).line;
+    return {
+      filename: path.relative(this.projectInfo.projectRoot, path.resolve(file.fileName)),
```
RomainMuller: Is the `path.resolve` necessary here?

packages/jsii/test/test.enums.ts
```diff
@@ -38,6 +39,7 @@ export = {
       fqn: 'testpkg.Foo',
       kind: 'enum',
       members: [{ name: 'Bar' }, { name: 'Baz' }],
```
RomainMuller: Would be nice if the members were also locatable.... 😇 (But ... like not important at all).

packages/jsii-reflect/lib/method.ts
rix0rrr: I originally had it as `sourceLocation`, but I want the name to be clear about what it's relative to. In this case, it's a module-relative location.

It becomes relevant later when we're combining it with the assembly path to get a repository-relative location.

packages/jsii-reflect/lib/source.ts
rix0rrr: As does most everything in `jsii-reflect`. But mmkay.

packages/jsii-reflect/lib/source.ts
rix0rrr: Yes, would be nice. I suppose the only solution for that is embedding git commits in the assembly as well, but then the question becomes how we're going to get those in there.

packages/jsii-reflect/package.json
rix0rrr: Strongly disagree. Just because dependencies exist somewhere down the dependency tree, doesn't mean I should just invisibly use them. You should always declare your direct dependencies, it's the only way to do proper dependency management.

If we were to rewrite `jsii` to Java, `jsii-reflect` would break for no good reason.

If we were to need to depend on a different version of `typescript` between these 2 packages, `jsii-reflect` would break for no good reason.

packages/jsii-spec/lib/spec.ts
rix0rrr: I guess. Doesn't seem so valuable but I can do that.

packages/jsii/lib/assembler.ts
rix0rrr: I don't know. I added it for safety. Guess I should put one around `projectRoot` too.

packages/jsii-reflect/lib/method.ts
RomainMuller: I still think `moduleLocation` is misleading, _especially_ considering your argument.
You could name that `sourceCoordinates` (to me, "coordinates" feels more "relative to some referential") than "location")

packages/jsii-reflect/lib/source.ts
RomainMuller: I guess code generation could provide context (e.g: git commit-ish, which is a fancy way to mean commit SHA, tag, branch, ...) that is then passed there & used? Can be done later, though (maybe create issue).

packages/jsii-reflect/package.json
RomainMuller: Why do you depend on `jsii` then? If it's not used instead of `tic`, then I don't think it should be depended on at all...

packages/jsii-spec/lib/spec.ts
RomainMuller: Well for one, if you want to rename those... There's only one place... And it allows stuff that uses that to be less tied to the list of specific types that have the `sourceLocation` property.

packages/jsii/lib/assembler.ts
RomainMuller: Was asking mostly for curiosity :)

rix0rrr: I tried it out and it's actually not necessary!

packages/jsii-reflect/package.json
rix0rrr: We depend on `jsii` because the unit tests depend on jsii.

packages/jsii-pacmak/test/expected.jsii-calc-lib/dotnet/Amazon.JSII.Tests.CalculatorPackageId.LibPackageId/.jsii
```diff
@@ -364,6 +444,10 @@
             "summary": "The value."
           },
           "immutable": true,
+          "locationInModule": {
```
eladb: Wouldn't it be sufficient to use a single line with `:` to separate file/line? Just thinking of how much global energy this will preserve.

packages/jsii-reflect/lib/property.ts
```diff
@@ -83,4 +83,18 @@ export class Property implements Documentable, Overridable, TypeMember {
...
+   * Return the location in the module
+   */
+  public get locationInModule(): SourceLocation | undefined {
+    return this.spec.locationInModule;
+  }
+
+  /**
+   * Return the location in the repository
+   */
+  public get locationInRepository(): SourceLocation | undefined {
```
eladb: maybe a base class for Property/Method that reuses docs/location and other stuff

packages/jsii-reflect/package.json
eladb: I am with Rico.

packages/jsii-pacmak/test/expected.jsii-calc-lib/dotnet/Amazon.JSII.Tests.CalculatorPackageId.LibPackageId/.jsii
rix0rrr: Aren't you the king of Worse Is Better ;)

packages/jsii-reflect/lib/property.ts
rix0rrr: Probably a good idea. The next person after me can do it, I already made one for `ClassType/InterfaceType`.

packages/jsii-pacmak/test/expected.jsii-calc-lib/dotnet/Amazon.JSII.Tests.CalculatorPackageId.LibPackageId/.jsii
eladb: Always. That's what I strive for. For example, stringified tokens.

PR #430, opened 2019-04-04
fix(jsii): check that static and nonstatic members don't share a name

### LINKED ISSUE #427: Check that there are not static and non-static members with the same name
It's definitely not allowed in Java, but probably is allowed in TypeScript, so jsii should guard against it.

rix0rrr:
This is not allowed in Java, and leads to compiler warnings in C#.

Fixes #427.


packages/jsii/test/negatives/neg.static-member-mixing.ts
```diff
@@ -0,0 +1,17 @@
+///!MATCH_ERROR: non-static member 'funFunction' of class 'Sub' conflicts with static member in ancestor 'SuperDuper'
```
RomainMuller: How about the case where I try to declare both on the same class?

PR #431, opened 2019-04-04
feat(jsii): check that referenced @param's exist

### LINKED ISSUE #422: Verify that @param blocks refer to existent argument names
Should be done by jsii

rix0rrr:
Verify integrity of the documentation by forcing the parameter names
referred to in @param declarations to actually exist.

Fixes #422.


packages/jsii/lib/assembler.ts
```diff
@@ -645,18 +645,10 @@ export class Assembler implements Emitter {
   }
 
   /**
-   * Register documentations on a ``spec.Documentable`` entry.
-   *
-   * @param sym       the symbol holding the JSDoc information
```
RomainMuller: It is sad that you're dropping rich documentation and making it less nice, while you're fixing the problems with it...

packages/jsii/lib/assembler.ts
```diff
@@ -669,6 +661,20 @@ export class Assembler implements Emitter {
...
 
+  /**
+   * Check that all parameters the doc block refers to with a @param declaration actually exist
+   */
+  private _validateReferencedDocParams(method: spec.Method, methodSym: ts.Symbol) {
+    const params = getReferencedDocParams(methodSym);
+    const actualNames = new Set((method.parameters || []).map(p => p.name));
+    for (const param of params) {
+      if (!actualNames.has(param)) {
+        this._diagnostic(methodSym.valueDeclaration, ts.DiagnosticCategory.Error,
```
RomainMuller: Could this be a warning instead? I mean - is there any reason to actually fail on this?

rix0rrr: It could, but I'm wary. In my experience, warnings don't get fixed. But sure, I'll throw in a warning.

packages/jsii/lib/assembler.ts
rix0rrr: Didn't feel like it was adding a lot of value. If the parameter is declared as `sym: Symbol` and the documentation basically says "this is the symbol", then I don't feel like the documentation is adding a lot of information.

Similarly for the return value. Declaration says: `function(...): Docs`, I don't feel the need to add `@returns the docs`.

packages/jsii/lib/assembler.ts
RomainMuller: We can turn this into an error in a later version if you like (using. warning first gives some notice period, which is great for CDK releases :D)

RomainMuller: And - feel free to record this in an issue so we don't forget!

PR #432, opened 2019-04-04
feat(jsii-spec): Model parameter optionality

### LINKED ISSUE #296: Move the "optional" attribute from TypeReference to Parameter
The specification models the `optional` attribute of parameters on their `TypeReference` instead of directly on the parameter. This creates a confusion between `null`able and `optional`.

The specification needs to be modified so optional parameters carry an `optional` attribute, while `TypeReferences` carry a `nullable` attribute.

Related: #230 #284 #295

### LINKED ISSUE #414: JSII doesnt genereate nullable enum in .Net
The following code in the .net CDK

```
  [JsiiByValue]
  public class SubnetSelection : ISubnetSelection
  {
    [JsiiProperty("subnetName", "{\"primitive\":\"string\",\"optional\":true}", true)]
    public string SubnetName { get; set; }

    [JsiiProperty("subnetType", "{\"fqn\":\"@aws-cdk/aws-ec2.SubnetType\",\"optional\":true}", true)]
    public SubnetType SubnetType { get; set; }
  }
```
doesnt follow the nullability in vpc-ref.ts

```
export interface SubnetSelection {
  /**
   * Place the instances in the subnets of the given type
   *
   * At most one of `subnetType` and `subnetName` can be supplied.
   *
   * @default SubnetType.Private
   */
  readonly subnetType?: SubnetType;

  /**
   * Place the instances in the subnets with the given name
   *
   * (This is the name supplied in subnetConfiguration).
   *
   * At most one of `subnetType` and `subnetName` can be supplied.
   *
   * @default name
   */
  readonly subnetName?: string;
}
```

Which makes .Net code throw `Only one of subnetType and subnetName can be supplied` everytime a SubnetName is used, even if no value is defined for SubnetType

https://github.com/awslabs/aws-cdk/blob/master/packages/@aws-cdk/aws-ec2/lib/vpc-ref.ts

RomainMuller:
Method `Parameters`, `Properties` and method return types
now carry an `optional` flag that indicates whether they are
optional or required, and the `TypeReference#optional` field
was removed.

BREAKING CHANGE: JSII assemblies generated by older versions of the tool
will fail loading with this new version, and vice-versa. Re-compile your
projects in order to fix this.

Fixes #296
Fixes #414

---


RomainMuller:
Note that I decided against modeling *optional properties*, because it is essentially meaningless... A property can be considered optional as long as it's type is `nullable`.

packages/jsii-kernel/lib/kernel.ts
```diff
@@ -717,15 +717,15 @@ export class Kernel {
...
                 if (params.length <= i) { return; } // No vararg was provided
                 for (let j = i ; j < params.length ; j++) {
-                    if (params[j] === undefined) {
+                    if (!param.type.nullable && params[j] === undefined) {
                         // tslint:disable-next-line:max-line-length
                         throw new Error(`Unexpected 'undefined' value at index ${j - i} of variadic argument '${param.name}' of type '${this._formatTypeRef(param.type)}'`);
                     }
                 }
-            } else if (!param.type.optional && arg === undefined) {
+            } else if (!param.type.nullable && arg === undefined) {
```
rix0rrr: This should be `param.optional` I think. This code is doing argument matching, not type matching.

packages/jsii-kernel/lib/serialization.ts
```diff
@@ -483,7 +483,7 @@ export function serializationType(typeRef: CompleteTypeReference, lookup: TypeLo
     // Propagate the top-level 'optional' field to each individual subtype
```
rix0rrr: Miss comment.

packages/jsii-pacmak/lib/targets/python.ts
```diff
@@ -374,12 +360,12 @@ abstract class BaseMethod implements PythonBase {
                 for (const prop of liftedProperties) {
                     const paramName = toPythonParameterName(prop.name);
                     const paramType = resolver.resolve(prop.type, { forwardReferences: false });
-                    const paramDefault = prop.type.optional ? "=None" : "";
+                    const paramDefault = prop.type.nullable ? "=None" : "";
```
rix0rrr: This is not doing method arguments, right? A non-optional nullable type shouldn't have a default.

packages/jsii-spec/lib/spec.ts
```diff
@@ -451,10 +451,32 @@ export interface Parameter extends Documentable {
...
+    modifier?: ParameterModifier;
+}
+
+/**
+ * Modifiers for a parameter.
+ */
+export enum ParameterModifier {
+    /**
+     * Indicates that the parameter may be omitted. When a parameter has the
+     * `Optional` midufuer, the following must also be true:
```
rix0rrr: fun typo :)

packages/jsii/lib/assembler.ts
```diff
@@ -1175,6 +1162,28 @@ function _fingerprint(assembly: spec.Assembly): spec.Assembly {
   return { ...assembly, fingerprint };
 }
 
+/**
+ * Removes the `Optional` modifier from parameters that have it, but are not followed only with
+ * other `Optional` or `Variadic` parameters. This is done after parsing out all parameters as it
+ * is otherwise difficult to assert whether a parameter with a default value is `Optional` or it it
+ * merely has a `nullable` `#type`.
+ */
+function _removeInvalidOptionalModifiers(parameters?: spec.Parameter[]) {
```
rix0rrr: I see, is this the magic sauce that makes us backwards compatible.

Are we going to try looking at the declaration at all? Or too risky? (`name?: type` vs `name: type | undefined` ?)

packages/jsii/lib/assembler.ts
```diff
@@ -1175,6 +1162,28 @@ function _fingerprint(assembly: spec.Assembly): spec.Assembly {
...
+  }
+
+  for (const param of parameters.filter(_cannotBeOptional)) {
+    delete param.modifier;
+  }
+
+  function _cannotBeOptional(param: spec.Parameter, index: number, array: spec.Parameter[]): boolean {
+    return spec.isOptional(param)
+      // Checks whether there is one non-optional/variadic parameter after this one in the array
+      && array.find((p, i) => i > index && !spec.isOptional(p) && !spec.isVariadic(p)) != null;
```
rix0rrr: Can be `array.some()`

rix0rrr:
It slightly concerns me that there are breaking changes in the assembly model. Do we know for a fact users will always use the new kernel to load the new assemblies?

Don't we have a `version` in the assembly metadata? Did you bump that, and does the kernel actually check the version it's expecting?

packages/jsii-spec/lib/spec.ts
RomainMuller: Oh there be many midufüren around here ...

packages/jsii/lib/assembler.ts
RomainMuller: Ugh I never find it because I keep thinking about the ruby name... `.any?`

RomainMuller:
@rix0rrr new revision should have addressed all of your comments as well, and sets the Schema version of `jsii/0.9.0` (which I probably should flip to `0.10.0`). I wonder if I should get this to be expressive of a minimum version requirement, instead of looking like a fixed version... (can be a "for later" thing as well, but I guess if anyone has ideas...)

packages/jsii-calc-base-of-base/test/assembly.jsii
```diff
@@ -16,7 +16,7 @@
     "type": "git",
     "url": "https://github.com/awslabs/jsii.git"
   },
-  "schema": "jsii/1.0",
+  "schema": "jsii/0.9.0",
```
rix0rrr: Waitaminit! I don't like how this is version number is going backwards, and we already have a `jsiiVersion` field that encodes the JSII version. Why can't we just discretely bump the schema versions 1 -> 2 -> 3 etc? Saves us having to do the whole "what's the next release version" dance, too.

packages/jsii-spec/lib/spec.ts
```diff
@@ -7,7 +7,7 @@ export interface Assembly extends Documentable {
     /**
      * The version of the spec schema
      */
-    schema: SchemaVersion.V1_0;
+    schema: SchemaVersion.LATEST;
```
rix0rrr: I think if you're going to do this, it needs a version check in the jsii-kernel and jsii-reflect as well.

packages/jsii-spec/lib/spec.ts
```diff
@@ -330,29 +330,38 @@ export enum PrimitiveType {
...
+     *
+     * @default false
      */
     promise?: boolean;
 }
 
+/**
+ * A reference to a type (primitive, collection or fqn).
+ */
+export type TypeReference = NamedTypeReference | PrimitiveTypeReference | CollectionTypeReference | UnionTypeReference;
```
rix0rrr: I support this!

packages/jsii-spec/lib/spec.ts
```diff
@@ -330,29 +330,38 @@ export enum PrimitiveType {
 }
 
 /**
- * A reference to a type (primitive, collection or fqn).
+ * Represents some instance of a type, for example as the return value of a method, a parameter, or a property.
  */
-export type TypeReference = TypeReferenceBase & (NamedTypeReference | PrimitiveTypeReference | CollectionTypeReference | UnionTypeReference);
+export interface TypeInstance<T extends TypeReference> {
```
rix0rrr: I think I'm going to say that this interface ony makes sense for return types, and I think I object to its use in other places.

packages/jsii-spec/lib/spec.ts
```diff
@@ -390,7 +399,7 @@ export interface CollectionTypeReference extends TypeReferenceBase {
         /**
          * The type of an element (map keys are always strings).
          */
-        elementtype: TypeReference;
+        elementtype: TypeInstance<TypeReference>;
```
rix0rrr: It does not make sense for either array or map elements to be of type `Promise<T>`, or optional.

packages/jsii-spec/lib/spec.ts
```diff
@@ -409,7 +418,7 @@ export interface UnionTypeReference extends TypeReferenceBase {
          * All the possible types (including the primary type).
          * @minItems 2
          */
-        types: TypeReference[];
+        types: Array<TypeInstance<TypeReference>>;
```
rix0rrr: It does not make sense for a union type to be something like `number | string?` or `number | Promise<string>`.

packages/jsii-spec/lib/spec.ts
```diff
@@ -484,13 +493,17 @@ export interface Parameter extends Documentable {
     name: string;
 
     /**
-     * The type of the parameter.
+     * The specification for the value of this parameter (includes it's declared type and whether it is optional).
      */
-    type: TypeReference;
+    value: TypeInstance<TypeReference>;
```
rix0rrr: A parameter can never be a `Promise<T>`. It *can* be optional, however.

packages/jsii-spec/lib/spec.ts
```diff
@@ -508,7 +521,7 @@ export interface Method extends Documentable, Overridable, SourceLocatable {
     /**
      * The return type of the method (undefined if void or initializer)
      */
-    returns?: TypeReference;
+    returns?: TypeInstance<TypeReference>;
```
rix0rrr: This is the only place where `TypeInstance` makes sense to me.

Although I think `promise?: boolean` makes more sense as an attribute of `Method` to me (`async?: boolean`) than as an attribute of the return type.

packages/jsii-spec/lib/spec.ts
```diff
@@ -440,9 +449,9 @@ export interface Property extends Documentable, Overridable, SourceLocatable {
     name: string;
 
     /**
-     * The type of the property.
+     * The specification for the property's value (including it's declared type and whether it is optional).
      */
-    type: TypeReference;
+    value: TypeInstance<TypeReference>;
```
rix0rrr: Properties cannot be `promise`. They *can* be *optional* however.

rix0rrr:
I think I'd prefer a modeling like this:

```ts
interface OptionalValue {
    optional?: boolean;
    type: TypeReference;
}

interface Method {
    async?: boolean;
    returns?: OptionalValue;
    // ...
}

interface Property extends OptionalValue {
    // ...
}

interface Collection {
    element: TypeReference;
    // ...
}

```

packages/jsii-spec/lib/spec.ts
```diff
@@ -324,40 +412,53 @@ export enum PrimitiveType {
...
  */
-export type TypeReference = TypeReferenceBase & (NamedTypeReference | PrimitiveTypeReference | CollectionTypeReference | UnionTypeReference);
+export type TypeReference = NamedTypeReference | PrimitiveTypeReference | CollectionTypeReference | UnionTypeReference;
 
 /**
- * Common attributes of a TypeReference.
+ * The standard representation of the `any` type (includes optionality marker).
  */
-export interface TypeReferenceBase {
+export const CANONICAL_ANY: Readonly<PrimitiveTypeReference> = { optional: true, primitive: PrimitiveType.Any };
```
eladb: Why does `any` imply `optional`? It implies nullable, but that's a different thing.

packages/jsii-spec/lib/spec.ts
```diff
@@ -324,40 +412,53 @@ export enum PrimitiveType {
...
+
+/**
+ * Qualifiers for a type reference.
+ */
+interface QualifiedTypeReference {
```
eladb: Maybe `TypeReferenceQualifiers`, because this is not a type reference.

packages/jsii-pacmak/test/expected.jsii-calc/python/src/jsii_calc/__init__.py
```diff
@@ -58,7 +58,7 @@ def __init__(self) -> None:
         jsii.create(AllTypes, self, [])
 
     @jsii.member(jsii_name="anyIn")
-    def any_in(self, inp: typing.Any) -> None:
+    def any_in(self, inp: typing.Any=None) -> None:
```
eladb: That's wrong... any does not imply optionality

packages/jsii/lib/assembler.ts
```diff
@@ -555,17 +557,17 @@ export class Assembler implements Emitter {
 
       // tslint:disable-next-line:no-bitwise
       if ((ts.getCombinedModifierFlags(ctorDeclaration) & ts.ModifierFlags.Private) === 0) {
-        jsiiType.initializer = { initializer: true };
+        jsiiType.initializer = {};
```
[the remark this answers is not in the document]
RomainMuller: You have an initializer, not a method. Methods have a name (guaranteed). Initializers don't (guaranteed).

packages/jsii/lib/assembler.ts
```diff
@@ -204,22 +203,26 @@ export class Assembler implements Emitter {
    *
    * @returns the de-referenced type, if it was found, otherwise ``undefined``.
    */
-  private _dereference(ref: spec.NamedTypeReference, referencingNode: ts.Node | null): spec.Type | undefined {
-    const [assm, ] = ref.fqn.split('.');
+  private _dereference(ref: string | spec.NamedTypeReference, referencingNode: ts.Node | null): spec.Type | undefined {
```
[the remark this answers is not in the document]
RomainMuller: Oh boy. That alias even exists! I can just use it :)

packages/jsii-spec/lib/spec.ts
RomainMuller: I guess I think of optional as allowing "no value" through, and any also allows "no value" though. But I guess I see your point still...

packages/jsii-calc-base-of-base/test/assembly.jsii
```diff
@@ -35,15 +35,14 @@
...
+    "sphinx": {}
   },
   "types": {
     "@scope/jsii-calc-base-of-base.Very": {
       "assembly": "@scope/jsii-calc-base-of-base",
       "fqn": "@scope/jsii-calc-base-of-base.Very",
-      "initializer": {
-        "initializer": true
-      },
+      "initializer": {},
```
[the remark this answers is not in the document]
RomainMuller: Empty object here is the default initializer (no parameters, public). I thought of also modeling that as "true" or the literal string "default" instead... but the empty object actually describes what I want, based on its default values... so I figured I wouldn't overthink it.

RomainMuller:
@rix0rrr well then you seem in vivid disagreement with @eladb and I am certainly not re-doing this whole thing a fourth time unless we are 100% certain the 4th is the right one and it has significant benefits over this... this is a full day of painful delving into C# failing tests each time...

packages/jsii-calc-base-of-base/test/assembly.jsii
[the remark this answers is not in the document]
eladb: I understand where this is coming from, but still don't like this as a general modeling principle. Can we do better?

eladb:
@RomainMuller what disagreement?

packages/jsii-spec/lib/spec.ts
RomainMuller: Both validate the schema, so the version check is in there.

RomainMuller:
@eladb isn't what @rix0rrr "would prefer" a minor variation over the previous iteration , which I would then literally have aster a whole day redoing all over again?

RomainMuller:
I want to be able to express:
- optional result
- optional parameter (as in can be undefined)
- optional value in collection (list/map)

packages/jsii-spec/lib/spec.ts
RomainMuller: I can totes declare a `public readonly iSwearItWillHaveAValueAtSomePoint: Promise<string>`. If that's an API I *like*? No. But it can be made (and JSII will happily generate a `promise: true` for that type reference.

packages/jsii-spec/lib/spec.ts
RomainMuller:
```ts
// In fact, totally possible:
async function impossible(arg: Promise<Foo>): Promise<Bar>;
```

packages/jsii-spec/lib/spec.ts
RomainMuller: And yet, this is valid TypeScript, and thus could be authored in a TypeScript signature that JSII would totally parse out in that way

packages/jsii-spec/lib/spec.ts
RomainMuller: So an array may never contain `null`? Or it'll contain "surprise" `null`s?

packages/jsii-calc-base-of-base/test/assembly.jsii
RomainMuller: I don't really care. I kinda felt it was nice to correlate that with the `jsii` version so you can, from an error (like "Whoosh the version is `jsii/0.9.0`, but I expected to find `jsii/0.42.12`), determine what you need to do. I however would only update that version number if/when the spec gets a breaking change (so it acts as a "minimum required version" of sorts.

packages/jsii-spec/lib/spec.ts
rix0rrr: But that is something else though! Feels like now we're expressing nullability, not optionality.

packages/jsii-spec/lib/spec.ts
rix0rrr: Nobody says we can't reject that in jsii. Will the kernel handle it correctly?

packages/jsii-spec/lib/spec.ts
rix0rrr: Sure. What I meant to say is that the kernel does not have adequate provisions to handle this, so we shouldn't allow it either.

packages/jsii-spec/lib/spec.ts
rix0rrr: See my previous comment, I'm not sure the kernel will handle that situation properly.

eladb:
Is it possible to decouple these discussions? We are in alignment on optionals, so let’s wrap up this PR, and follow up with a discussion and a separate PR on nullables. I don’t see a good reason to couple those together, right?

RomainMuller:
Additional change discussions will happen in #440

rix0rrr:
> Is it possible to decouple these discussions?

I don't think so. We currently have one flag which we interpret as "optional" in a value context (property, argument) and as "nullable" in a type context (return values, `Array<Nullable<string>>`).

I would suggest we either make a decision on what properties we want from our type system and encode those, or we leave the current situation which is occasionally confusing but serviceable enough.

There is only one thing we cannot model today, which is a required, nullable parameter.

packages/jsii-spec/lib/spec.ts
```diff
@@ -210,34 +239,46 @@ export interface Docs {
...
+     * one.
+     *
+     * @default none
      */
     returns?: string;
 
     /**
      * Whether the API item is beta/experimental quality
+     *
+     * @default Stability.Experimental
```
eladb: inherit from parent (and if no parent - stable)

PR #454, opened 2019-04-12
fix(jsii): flatten out dependency list

### LINKED ISSUE #453: jsii assemblies contain redundant dependency information
Looking at the start of the dependency section of the `@aws-cdk/app-delivery` assembly, it looks like this:

```
  "dependencies": {
    "@aws-cdk/aws-cloudformation": {
      "dependencies": {
        "@aws-cdk/aws-iam": {
          "dependencies": {
            "@aws-cdk/cdk": {
              "dependencies": {
                "@aws-cdk/cx-api": {
                  "peer": true,
                  "targets": {
                    "dotnet": {
                      "assemblyOriginatorKeyFile": "../../key.snk",
                      "namespace": "Amazon.CDK.CXAPI",
                      "packageId": "Amazon.CDK.CXAPI",
                      "signAssembly": true
                    },
                    "java": {
                      "maven": {
                        "artifactId": "cdk-cx-api",
                        "groupId": "software.amazon.awscdk"
                      },
                      "package": "software.amazon.awscdk.cxapi"
                    },
                    "js": {
                      "npm": "@aws-cdk/cx-api"
                    },
                    "python": {
                      "distName": "aws-cdk.cx-api",
                      "module": "aws_cdk.cx_api"
                    }
                  },
                  "version": "0.28.0"
                }
              },
              "peer": true,
              "targets": {
                "dotnet": {
                  "assemblyOriginatorKeyFile": "../../key.snk",
                  "namespace": "Amazon.CDK",
                  "packageId": "Amazon.CDK",
                  "signAssembly": true
```

* Why does it contain transitive dependencies?
* Why do the dependencies have information on the targets *and their build instructions* (`../../key.snk`???) in there?

For funzies... in a package with highly complex dependencies such as `@aws-cdk/app-delivery`, removing the `dependencies` node from the assembly reduces the file size from

    8.2 MB

to

    18 kB (!!!!)

rix0rrr:
Stop deeply nesting dependency packages and their dependencies. 

Emphatically go to direct-level dependencies only, no need to list transitive dependencies.

The original reason for including all transitive dependencies was so
that the .NET generator could have a full view of transitively inherited
interfaces, so it can generate full proxies.

Added a test interface to verify that the build succeeds, and it does,
so apparently we don't need them anymore.

@eladb, am I missing anything? This seems too simple to be true. Or did we
just get rid of a vestigial data structure?

Fixes #453.


rix0rrr:
Well the build obviously disagrees with what my commit message says but locally it's passing for some reason.

PR #461, opened 2019-04-16
fix(jsii-reflect): don't load same assembly multiple times

rix0rrr:
deCDK tests were calling `loadModule()` for every package.
`loadModule()` did have some recursion avoidance INSIDE one call,
but not across multiple load calls on the same type system.

This brings down the deCDK tests from 80s to 10s.

Add an option to disable validation, which would bring the full CDK
typesystem loading down from 10s to 600ms (validation is enabled
by default).

Some more type exposure updates so that we can update awslint
and friends in the CDK repository to take advantage of the new
jsii model.

### DIFF
--- a/packages/jsii-reflect/lib/type-system.ts
+++ b/packages/jsii-reflect/lib/type-system.ts
@@ -38,15 +38,15 @@ export class TypeSystem {
    * @param fileOrDirectory A .jsii file path or a module directory
    * @param validate Whether or not to validate the assembly while loading it.
    */
-  public async load(fileOrDirectory: string, validate = true) {
+  public async load(fileOrDirectory: string, options: { validate?: boolean } = {}) {
     if ((await stat(fileOrDirectory)).isDirectory()) {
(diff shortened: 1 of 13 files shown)


packages/jsii-reflect/lib/type-system.ts
```diff
@@ -32,18 +32,21 @@ export class TypeSystem {
...
    * If `fileOrDirectory` is a directory, it will be treated as a jsii npm module.
    *
+   * Not validating makes the difference between loading assemblies with lots
+   * of dependencies (such as app-delivery) in 90ms vs 3500ms.
+   *
    * @param fileOrDirectory A .jsii file path or a module directory
+   * @param validate Whether or not to validate the assembly while loading it.
    */
-  public async load(fileOrDirectory: string) {
+  public async load(fileOrDirectory: string, validate = true) {
```
eladb: Prefer keyword arguments: `{ validate: true }`

PR #462, opened 2019-04-16
v0.10.0

rix0rrr:
See CHANGELOG


RomainMuller:
Needs to get in the patch from #465. It might not need a separate entry in `CHANGELOG` though, as this is a fix over a fix that is already documented here and hadn't been released yet.

PR #466, opened 2019-04-17
fix(dotnet): Correctly generate "optional" markers

RomainMuller:
Method parameters were missing the "optional" marker due to the logic
that was initially intended to prune only documentation from the JSON
data attached to the method attribute.


eladb:
Does it make sense to add a compliance test as well?

RomainMuller:
@eladb I've added #467 for that :)

Issue #473, opened 2019-04-19
Python: information around struct base classes is lost


rix0rrr:
Structs are implemented in Python using `mypy_extensions.TypedDict`.

Unfortunately, this class uses a metaclass that completely destroys class inheritance information:

```
    def test_inheritance_maintained(self):
        """Check that for JSII struct types the inheritance relationship is maintained."""
        bases = inspect.getmro(jsii_calc.DerivedStruct)
        print(jsii_calc.DerivedStruct.__annotations__)
        print(bases)
```

The `__annotations__` contain all annotations from base classes, but the `getmro()` returns:

```
(<class 'importlib._bootstrap.DerivedStruct'>, <class 'dict'>, <class 'object'>)
```

Even though the actual declaration of `DerivedStruct` looks like this:

```python
class _DerivedStruct(scope.jsii_calc_lib.MyFirstStruct, jsii.compat.TypedDict, total=False):
    anotherOptional: typing.Mapping[str,scope.jsii_calc_lib.Value]
    """This is optional."""
    optionalAny: typing.Any
    optionalArray: typing.List[str]

@jsii.data_type(jsii_type="jsii-calc.DerivedStruct")
class DerivedStruct(_DerivedStruct):
```

So we are missing the bases `_DerivedStruct` and `MyFirstStruct` (and everything above).

Without these bases, we can't generate docs for those members.

rix0rrr:
The way I've fixed this now is actually not great, because it's not going to be useful to IDEs.

We should just generate a complete "Attributes" block in the class docstring instead. This can be done later.

PR #482, opened 2019-04-23
fix(python): maintain inheritance chain for structs

rix0rrr:
Because structs all inherit from TypedDict, and TypedDict erases the
inheritance chain of structs, we have to maintain a copy of the
inheritance hierarchy on the class objects, for later use during
doc generation.

This addresses (part of) #473.


packages/jsii-pacmak/lib/targets/python.ts
```diff
@@ -655,14 +656,15 @@ class TypedDict extends BasePythonClassType {
...
+            code.line(`@jsii.data_type_optionals(jsii_struct_bases=[${baseInterfaces.join(', ')}])`);
             code.openBlock(`class _${this.name}(${classParams.concat(["total=False"]).join(", ")})`);
             for (const member of optionalMembers) {
                 member.emit(code, resolver);
             }
             code.closeBlock();
 
             // Now we'll emit the mandatory members.
-            code.line(`@jsii.data_type(jsii_type="${this.fqn}")`);
+            code.line(`@jsii.data_type(jsii_type="${this.fqn}", jsii_struct_bases=[_${this.name}])`);
```
RomainMuller: What's the `_` for here?

rix0rrr: Every struct gets compiled into 2 classes:

```python
class _MyStruct(...):
  pass

class MyStruct(_MyStruct):
  pass
```

One holds the required fields, the other one the optional fields.

PR #499, opened 2019-05-07
chore: allow compiling only the JS packages

rix0rrr:

### DIFF
--- a/packages/jsii-pacmak/generate.sh
+++ b/packages/jsii-pacmak/generate.sh
@@ -2,9 +2,11 @@
 set -euo pipefail
 
 # embed the dotnet code generator program as a resource
-dng="lib/targets/dotnet-generator"
-mkdir -p ${dng}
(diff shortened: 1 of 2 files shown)


packages/jsii-pacmak/generate.sh
```diff
@@ -2,9 +2,11 @@
 set -euo pipefail
 
 # embed the dotnet code generator program as a resource
-dng="lib/targets/dotnet-generator"
-mkdir -p ${dng}
-rsync -av node_modules/jsii-dotnet-generator/cli/publish/ ${dng}/
+if ! ${NO_DOTNET:-false}; then
```
eladb: What about Java/Python?

rix0rrr: The generators for those are written in TypeScript, so they can be compiled (just not used to compile Java/Python packages).

This is about compiling the jsii packages themselves.

rix0rrr:
Guess this will never get merged

PR #504, opened 2019-05-17
fix(jsii-diff): catch exception if type disappeared from other assembly

rix0rrr:
Make jsii-diff not crash in the following case:

- Assembly A depends on Assembly B, and they both update simultaneously.
- Assembly A' uses a return type T' freshly introduced in Assembly B'.

When doing the supertype check, the compatibility check of A -> A',
the lookup of T' inside B would lead to an exception, which would
crash the tool.


rix0rrr:
Not tested because adding tests on this is hard

PR #512, opened 2019-06-03
feat(jsii-spec): Add optional metadata field

RomainMuller:
This can be used to track free-form maintainer data, for use in systems that
are not part of the standard JSII toolchain. For example, it can be used to
encode configuration behavior for proprietary tools that process assemblies.


eladb:
Why is "awsdocs" a jsii target if it's not a jsii target?

Issue #440, opened 2019-04-08
JSII Assembly Spec - V2


RomainMuller:
This issue is to collect improvement ideas for a V2 of the JSII Assembly Specification.

- Model methods/properties as `{ [name: string]: T }` instead of `T[]` to statically guarantee name uniqueness
- Model method overrides on the method itself (in essence, an override is an alternate parameter lists)
- Rename `array` to `list`, and consider modeling immutability
- Support `set`
- Make structs first class citizens (decouple from `interface`)
- distinguish “interface”, “enum”, “class” and “struct” instead of TypeRef.fqn
- Parse docs into an AST instead of a "markdown string"
- Denote `stability` on entities instead of their `docs` attribute

PR #515, opened 2019-06-03
feat: Register module-level stability

RomainMuller:
Adds the necessary code to detect and forward module-level stability
indications to the assembly, so it can be used by downstream languages.
This effectively removes the `markdown` attribute from the `Assembly` in
favor of using the `docs` attribute.

Also, renders a `package-info.json` file for packages that have
documentation (a `README.md` file, deprecation notice, or stability
indication).

Implements awslabs/cdk-ops#367


packages/jsii-spec/lib/spec.ts
```diff
@@ -134,13 +134,6 @@ export interface Assembly extends Documentable {
      * @default none
      */
     types?: { [fqn: string]: Type };
-
```
eladb: No breaking changes to the jsii spec at this time

packages/jsii-calc/test/assembly.jsii
```diff
@@ -182,6 +182,11 @@
     }
   },
   "description": "A simple calcuator built on JSII.",
+  "docs": {
+    "remarks": "![API Stability: experimental](https://img.shields.io/badge/API%20Stability-experimental-important.svg)\n\nThis library is used to demonstrate and test the features of JSII\n\n## Sphinx\n\nThis file will be incorporated into the sphinx documentation.\n\nIf this file starts with an \"H1\" line (in our case `# jsii Calculator`), this\nheading will be used as the Sphinx topic name. Otherwise, the name of the module\n(`jsii-calc`) will be used instead.\n\n\n\n\n",
+    "stability": "experimental",
```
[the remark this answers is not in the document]
RomainMuller: I like it consistent... If it's a top-level thing, then it should be for *everything* (incl. types, members, ...). Would you agree with that?

packages/jsii-pacmak/lib/targets/sphinx.ts
```diff
@@ -714,14 +714,14 @@ class SphinxDocsGenerator extends Generator {
      * @returns header: the contents of the header (or undefined)
      */
     private emitReadme(assm: spec.Assembly): { readmeFile?: string, readmeHeader?: string } {
-        if (!assm.readme) {
+        if (!assm.docs || !assm.docs.summary) {
```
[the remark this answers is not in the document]
RomainMuller: Not that I want to clean up, really just that `assm.docs` is currently 100% unused and by the current schema is where `stability` is hosted... So I figured I would avoid "effectively" adding a new field by replacing the `readme` instead.

packages/jsii-pacmak/lib/targets/java.ts
```diff
@@ -346,6 +349,39 @@ class JavaGenerator extends Generator {
...
+            this.code.line(' *');
+        }
+        if (mod.docs.deprecated) {
+            this.code.line(` * @deprecated ${mod.docs.deprecated}`);
+        } else if (mod.docs.stability) {
+            this.code.line(` * @stability ${mod.docs.stability}`);
+        }
+        this.code.line(' */');
+        if (mod.docs.deprecated) {
+            this.code.line('@Deprecated');
```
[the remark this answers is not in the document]
RomainMuller: The deprecation message lives in the comment block... The [`@Deprecated`](https://docs.oracle.com/javase/7/docs/api/java/lang/Deprecated.html) annotation takes no arguments.

packages/jsii-spec/lib/spec.ts
RomainMuller: But why though?

packages/jsii-calc/package.json
```diff
@@ -5,6 +5,7 @@
   "main": "lib/index.js",
   "types": "lib/index.d.ts",
   "private": true,
+  "stability": "experimental",
```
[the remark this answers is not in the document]
RomainMuller: I am removing the "deprecated" banner enforcement from `jsii` in favor of the next layer (the enforcement felt heavy-handed and too prescriptive).

packages/jsii-pacmak/lib/targets/java.ts
```diff
@@ -346,6 +349,39 @@ class JavaGenerator extends Generator {
...
+            }
+            if (mod.docs.remarks) {
+                mdown.push(mod.docs.remarks);
+            }
+            for (const line of md2html(mdown.join('\n')).split('\n')) {
+                this.code.line(` * ${line}`);
+            }
+            this.code.line(' *');
+        }
+        if (mod.docs.deprecated) {
```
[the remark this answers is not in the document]
RomainMuller: Unfortunately, this attribute *already* exists on `docs` and moving it out would be a breaking change. I made a note in #440 to address in V2.

packages/jsii-pacmak/lib/targets/sphinx.ts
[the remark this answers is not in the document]
RomainMuller: Moved the `readme` back to it's own attribute.

packages/jsii-pacmak/lib/targets/java.ts
```diff
@@ -346,6 +349,32 @@ class JavaGenerator extends Generator {
...
+            this.code.line(` * @deprecated ${mod.docs.deprecated}`);
+        } else if (mod.docs.stability) {
+            this.code.line(` * @stability ${mod.docs.stability}`);
+        }
+        this.code.line(' */');
+        if (mod.docs.deprecated) {
+            this.code.line('@Deprecated');
```
eladb: Can you put the message here?

packages/jsii-pacmak/lib/targets/java.ts
```diff
@@ -441,6 +470,9 @@ class JavaGenerator extends Generator {
                                 configuration: {
                                     failOnError: false,
                                     show: 'protected',
+                                    sourceFileExcludes: {
+                                        exclude: ['**/$Module.java']
```
eladb: Nice touch! add a comment explaining why

packages/jsii/lib/assembler.ts
```diff
@@ -143,12 +145,23 @@ export class Assembler implements Emitter {
 
     async function _loadReadme(this: Assembler) {
```
eladb: why are those not simple private methods?

packages/jsii/lib/project-info.ts
```diff
@@ -105,6 +107,8 @@ export async function loadProjectInfo(projectRoot: string, { fixPeerDependencies
 
         name: _required(pkg.name, 'The "package.json" file must specify the "name" attribute'),
         version: _required(pkg.version, 'The "package.json" file must specify the "version" attribute'),
+        deprecated: pkg.deprecated,
+        stability: _validateStability(pkg.stability, pkg.deprecated),
```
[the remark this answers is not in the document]
RomainMuller: This follows the naming pattern of other `_validate` functions in the same file. Not renaming them all.

packages/jsii/lib/assembler.ts
```diff
@@ -61,11 +61,12 @@ export class Assembler implements Emitter {
...
     }
     const readme = await _loadReadme.call(this);
-    if (!readme) {
+    if (readme == null) {
       this._diagnostic(null,
         ts.DiagnosticCategory.Suggestion,
-        'There is not "README.md" file. It is recommended to have one.');
+        'There is no "README.md" file. It is recommended to have one.');
     }
+    const docs = _loadDocs.call(this);
```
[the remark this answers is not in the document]
RomainMuller: `_loadDocs` is not actually an instance method... it's nested under that scope... But it needs to have `this` available so it needs to be explicitly bound.

packages/jsii-pacmak/lib/targets/java.ts
RomainMuller: The message is present two lines above in the `@deprecated` Javadoc field. This one is the annotation, which - again - is not parameterized.

packages/jsii/lib/assembler.ts
RomainMuller: The prototype of this class is already... quite heavily polluted. I did not want to add (more) clutter.

packages/jsii/lib/assembler.ts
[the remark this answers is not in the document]
eladb: Got it. thanks

packages/jsii-calc/lib/compliance.ts
```diff
@@ -1682,3 +1681,35 @@ export abstract class VoidCallback {
     }
     protected abstract overrideMe(): void;
 }
+
```
eladb: what about @stability? It also seems like there's also a test that has @stable in there... Are those two different things?

packages/jsii-calc/lib/compliance.ts
```diff
@@ -1682,3 +1681,35 @@ export abstract class VoidCallback {
...
+/**
+ * This tests code generation of deprecation markers
+ *
+ * @deprecated without replacement
```
eladb: We also need a test to verify deprecation at the package level

packages/jsii-pacmak/test/expected.jsii-calc/java/src/main/java/software/amazon/jsii/tests/calculator/DeprecatedClass.java
```diff
@@ -0,0 +1,65 @@
+package software.amazon.jsii.tests.calculator;
+
+/**
+ * This tests code generation of deprecation markers.
+ * 
+ * @deprecated without replacement
```
eladb: Where is @stability?

packages/jsii-pacmak/test/expected.jsii-calc/python/src/jsii_calc/__init__.py
```diff
@@ -550,6 +547,53 @@ def arg2(self) -> typing.Optional[str]:
         return jsii.get(self, "arg2")
 
 
+class DeprecatedClass(metaclass=jsii.JSIIMeta, jsii_type="jsii-calc.DeprecatedClass"):
+    """This tests code generation of deprecation markers.
+
+    Deprecated:
```
eladb: Missing "Stability" information

packages/jsii-pacmak/test/expected.jsii-calc/java/src/main/java/software/amazon/jsii/tests/calculator/DeprecatedClass.java
RomainMuller: I find it silly to tag both `@stability deprecated` and `@deprecated` (and also `@Deprecated`). This is the "Java-canonical" way to denote deprecation.

RomainMuller: I've worked around this itch in a clever (very java-y) way :P

packages/jsii-calc/lib/stability.ts
```diff
@@ -0,0 +1,104 @@
...
+  /** @stable */
+  OptionB
+}
+
+/** @deprecated for the show */
+export interface DeprecatedStruct {
+  /** @deprecated for the show */
+  readonly readonlyProperty: string;
+}
+/** @deprecated for the show */
```
eladb: For testability, I would make the deprecation strings different for each element type so we can verify they are rendered properly.

packages/jsii-dotnet-generator/src/Amazon.JSII.Generator/DocComment/DocCommentGeneratorBase.cs
```diff
@@ -53,7 +53,6 @@ protected IEnumerable<XmlNodeSyntax> GetRemarksNodes()
             List<string> remarks = new List<string>();
             if (!String.IsNullOrEmpty(Documentable.Docs.Remarks)) { remarks.Add(Documentable.Docs.Remarks); }
             if (!String.IsNullOrEmpty(Documentable.Docs.Default)) { remarks.Add($"default: {Documentable.Docs.Default}"); }
-            if (!String.IsNullOrEmpty(Documentable.Docs.Deprecated)) { remarks.Add($"deprecated: {Documentable.Docs.Deprecated}"); }
```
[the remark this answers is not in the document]
RomainMuller: The deprecation is already documented in the `[Obsolete("message")]` annotation, and I did not want to duplicate that... Would you rather I keep both?

packages/jsii-calc/lib/stability.ts
RomainMuller: Can do!

packages/jsii-dotnet-generator/src/Amazon.JSII.Generator/DocComment/DocCommentGeneratorBase.cs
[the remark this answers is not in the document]
eladb: No, just wanted to ensure it's not a miss

PR #517, opened 2019-06-04
First draft of a .NET code generator handling calc-base-of-base, calc…

rix0rrr:
Reopening because previous merge (https://github.com/awslabs/jsii/pull/503) failed, branch was merged into itself. 

-----

his is the first draft of a .NET code generator handling calc-base-of-base, calc-base and calc-lib

The generator is split in 3 files:

- dotnet.ts : the target class containing the bulk of the generator code.

- dotnet/filegenerator.ts: a file generator class which generates miscellaneous .NET required files (the csproj and assembly info file)

- dotnet/nameutils.ts: Various naming utility classes

This generator produces a 1:1 output with the calc-base-of-base, calc-base and calc-lib jsii models.

There is a few TODOs left as they are identified work required later on when implement calc

This is not supposed to build in the current status, as we still need to update the jsii-diff package.


assyadh:
Putting this PR on hold, for an incoming new revision

assyadh:
Updated the PR to include the full new expected files. This generator is mergeable to master.

I will work on another set of changes to REMOVE the old generator from the codebase. This will be for a later 'cleanup' PR.

rix0rrr:
> I will work on another set of changes to REMOVE the old generator from the codebase. This will be for a later 'cleanup' PR.

Unsure what this means. Is it still being exercised? 

* If not, we should be able to remove it today, right? 
* If it is, what is the current behavior then?

By the way, needs a merge and regen from `master`, apparently

assyadh:
> > I will work on another set of changes to REMOVE the old generator from the codebase. This will be for a later 'cleanup' PR.
> 
> Unsure what this means. Is it still being exercised?
> 
> * If not, we should be able to remove it today, right?
> * If it is, what is the current behavior then?
> 
> By the way, needs a merge and regen from `master`, apparently

I'll work on removing all the old generator code. My intention was to get two different PRs so that it's not as big.

assyadh:
Closing in favor of a new one including the latest changes to the old generator (stability ...)

PR #539, opened 2019-06-17
fix(python): parameter names in docstrings should be snake_case, not camelCase

### LINKED ISSUE #538: python: parameter names in docstrings are camelCase not snake_case
Currently the docstring generated for parameters of a method uses camelCase even though the actual parameter names are snake_case, e.g.:

```python
    def add_port_mappings(self, *, container_port: jsii.Number, host_port: typing.Optional[jsii.Number]=None, protocol: typing.Optional["Protocol"]=None) -> None:
        """Add one or more port mappings to this container.

        Arguments:
            portMappings: -
            containerPort: Port inside the container.
            hostPort: Port on the host. In AwsVpc or Host networking mode, leave this out or set it to the same value as containerPort. In Bridge networking mode, leave this out or set it to non-reserved non-ephemeral port.
            protocol: Protocol. Default: Tcp

        Stability:
            experimental
        """
```

garnaat:
Fixes #538.

### DIFF
--- a/packages/jsii-pacmak/test/expected.jsii-calc/python/src/jsii_calc/__init__.py
+++ b/packages/jsii-pacmak/test/expected.jsii-calc/python/src/jsii_calc/__init__.py
@@ -414,7 +414,7 @@ class AllowedMethodNames(metaclass=jsii.JSIIMeta, jsii_type="jsii-calc.AllowedMe
         """getXxx() is not allowed (see negatives), but getXxx(a, ...) is okay.
 
         Arguments:
-            withParam: -
+            with_param: -
 
         Stability:
             experimental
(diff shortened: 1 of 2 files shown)


packages/jsii-pacmak/test/expected.jsii-calc/python/src/jsii_calc/__init__.py
```diff
@@ -1367,15 +1367,15 @@ def derived_to_first(self, *, another_required: datetime.datetime, bool: bool, n
...
-            anotherRequired: 
+            another_required: 
             bool: 
-            nonPrimitive: An example of a non primitive property.
-            anotherOptional: This is optional.
-            optionalAny: 
-            optionalArray: 
+            non_primitive: An example of a non primitive property.
+            another_optional: This is optional.
+            optional_any:
```
eladb: What do we think about the params without text? how does this render? Should we add some placeholder text?

garnaat: Well, the current code returns a ``-`` character if there are not docs for the argument but that causes things to render funny, as in https://docs.aws.amazon.com/cdk/api/latest/python/aws_cdk.aws_ecs/Ec2Service.html#aws_cdk.aws_ecs.Ec2Service.add_placement_constraints.  I think that should be changed, either to a blank character, empty string, or some other default string.

eladb: Not sure I follow... So this is okay?

PR #540, opened 2019-06-17
New .NET Generator with new expected files

assyadh:
Hi,

This contains a new generator for .NET written in typescript with the new expected files
This also removes of the old generator codebase.

Fixes https://github.com/awslabs/jsii/issues/463
Fixes https://github.com/awslabs/jsii/issues/528


RomainMuller:
This is currently broken due to lingering references to the old packages here and there.

assyadh:
> This is currently broken due to lingering references to the old packages here and there.

Got it. Was missing a packages.json update. it builds now.

packages/jsii-pacmak/test/expected.jsii-calc/dotnet/Amazon.JSII.Tests.CalculatorPackageId/Amazon/JSII/Tests/CalculatorNamespace/ClassThatImplementsTheInternalInterface.cs
```diff
@@ -2,7 +2,9 @@
 
 namespace Amazon.JSII.Tests.CalculatorNamespace
 {
-    /// <remarks>stability: Experimental</remarks>
+    /// <remarks>
+    /// stability: Experimental
+    /// </remarks>
     [JsiiClass(nativeType: typeof(ClassThatImplementsTheInternalInterface), fullyQualifiedName: "jsii-calc.ClassThatImplementsTheInternalInterface")]
     public class ClassThatImplementsTheInternalInterface : DeputyBase, IINonInternalInterface
```
costleya: I see we are implicitly implementing the interfaces. What happens when we implement 2 interfaces with the same method signatures? Can't we always explicitly implement?

packages/jsii-pacmak/lib/targets/dotnet/dotnetgenerator.ts
```diff
@@ -272,9 +272,16 @@ export class DotNetGenerator extends Generator {
             this.code.line();
         }
 
+        if (initializer) {
```
costleya: would making the null check in the "emitIfNecessary" be cleaner?

packages/jsii-pacmak/lib/targets/dotnet/dotnetgenerator.ts
```diff
@@ -272,9 +272,16 @@ export class DotNetGenerator extends Generator {
...
+            this.dotnetRuntimeGenerator.emitDeprecatedAttributeIfNecessary(initializer);
+        }
         this.code.openBlock(`protected ${className}(ByRefValue reference): base(reference)`);
```
costleya: Can't properties and methods also be "deprecated" instead of just at the class level?

assyadh: It's handled when emitting the attributes.

Here is an example for methods: https://github.com/awslabs/jsii/blob/677645dc734dc5f2abb059c05eb295aa939524eb/packages/jsii-pacmak/lib/targets/dotnet/dotnetruntimegenerator.ts#L74

packages/jsii-pacmak/test/expected.jsii-calc/dotnet/Amazon.JSII.Tests.CalculatorPackageId/Amazon/JSII/Tests/CalculatorNamespace/ClassThatImplementsTheInternalInterface.cs
assyadh: You mean on the inherited/implemented classe/interfaces? 

Sure, I can make a change to just always put the fully qualified type.

packages/jsii-pacmak/lib/targets/dotnet/dotnetgenerator.ts
assyadh: Sure. will move the check around.

packages/jsii-pacmak/test/expected.jsii-calc/dotnet/Amazon.JSII.Tests.CalculatorPackageId/Amazon/JSII/Tests/CalculatorNamespace/ClassThatImplementsTheInternalInterface.cs
costleya: Just for reference: https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/interfaces/explicit-interface-implementation

assyadh: Actually, it looks like the work may be more involved, I don't want to delay this PR (Which the goal is to be a 1:1 against the OLD generator) so I created this issue to address the problem:

https://github.com/awslabs/jsii/issues/560

PR #541, opened 2019-06-17
feat(jsii): enforce enum names to be UPPER_CASE

shivlaks:


packages/jsii/lib/validator.ts
```diff
@@ -63,13 +64,26 @@ function _defaultValidations(): ValidationFunction[] {
 
     function _typeNamesMustUsePascalCase(_: Validator, assembly: spec.Assembly, diagnostic: DiagnosticEmitter) {
         for (const type of _allTypes(assembly)) {
+            if (spec.isEnumType(type)) { continue; }
```
eladb: The enum /type/ should be pascal case, just members should be capital

packages/jsii/test/negatives/neg.enum-name.2.ts
```diff
@@ -1,4 +1,4 @@
-///!MATCH_ERROR: Type names must use PascalCase: My_Enum
+///!MATCH_ERROR: Enum names must use TRUMP_CASE: My_Enum
```
eladb: Nope! Just the member names should be TRUMP_CASE

packages/jsii/lib/validator.ts
shivlaks: ah i see, totally misread that - so enum values then?
i'll get on it.

packages/jsii/test/negatives/neg.enum-name.2.ts
shivlaks: I think my Java disposition's are showing - enum names vs enum values. I'll get it updated!

packages/jsii/lib/validator.ts
eladb: Enum __members__:

```ts
export enum MyEnum {
  MEMBER_NUMBER_ONE,
  MEMBER_FOO,
  MEMBER_XOO
}
```

shivlaks: updated

packages/jsii-dotnet-runtime-test/test/Amazon.JSII.Runtime.IntegrationTests/ComplianceTests.cs
```diff
@@ -260,21 +260,21 @@ public void GetAndSetEnumValues()
...
         [Fact(DisplayName = Prefix + nameof(EnumFromScopedModule))]
         public void UseEnumFromScopedModule()
         {
             ReferenceEnumFromScopedPackage obj = new ReferenceEnumFromScopedPackage();
-            Assert.Equal(EnumFromScopedModule.Value2, obj.Foo);
-            obj.Foo = EnumFromScopedModule.Value1;
-            Assert.Equal(EnumFromScopedModule.Value1, obj.LoadFoo());
-            obj.SaveFoo(EnumFromScopedModule.Value2);
-            Assert.Equal(EnumFromScopedModule.Value2, obj.Foo);
+            Assert.Equal(EnumFromScopedModule.VALUE2, obj.Foo);
```
eladb: wait... what's the convention for .NET? Do we want .NET to also use all-caps?
Same question for Python.... @garnaat is it idiomatic to use all caps for enum values?

eladb: - .NET seems to prefer PascalCase.
- Python looks like ALL_CAPS are good.

So basically, the .NET code generator needs to covert the case for .NET.

shivlaks: that's a good question. I thought we wanted to avoid dropping into the language specific mangling... let me check.

shivlaks: I think Pascal casing for both names and values are the idiomatic thing for .NET

eladb: No. We wanted to normalize the typescript/jsii source to CAPITALS but then, jsii is all about language-specific idiomacy.

shivlaks: Sounds good. I'll make the changes to the generation of .NET so we can change enum values into PascalCase.

shivlaks: we decided we would make the changes for .NET generation to use enum members in `PascalCase` outside of this PR. 

Should I create an issue for it before resolving this conversation?

packages/jsii/lib/validator.ts
```diff
@@ -70,6 +71,19 @@ function _defaultValidations(): ValidationFunction[] {
...
     }
 
+    function _enumMembersMustUserUpperSnakeCase(_: Validator, assembly: spec.Assembly, diagnostic: DiagnosticEmitter) {
+        for (const type of _allTypes(assembly)) {
+            if (!spec.isEnumType(type)) { continue; }
+
+            for (const member of type.members) {
+                if (member.name && member.name !== Case.constant(member.name)) {
+                    diagnostic(ts.DiagnosticCategory.Error,
+                               `Enum members must use TRUMP_CASE: ${member.name}`);
```
eladb: I wonder if "ALL_CAPS" is a more suitable name

shivlaks: I actually like it more, will make the change before pushing

packages/jsii-dotnet-runtime-test/test/Amazon.JSII.Runtime.IntegrationTests/ComplianceTests.cs
eladb: Sure, "right click, create issue"

packages/jsii/lib/validator.ts
shivlaks: Done.

PR #552, opened 2019-06-25
feat(pacmak): support adding a suffix to Java package version

RomainMuller:
Allows setting a `jsii.targets.java.maven.versionSuffix` key that
defines a string to be appended at the end of the version number of the
Maven package. If the suffix starts with `.` or `-`, it will be appended
to the version number directly (`0.12.0` with suffix `.dp` will render
`0.12.0.dp`); but otherwise, it'll be prepended with a `-` before it is
concatenated with the version number (`0.12.0` with suffix `dp` will
render `0.12.0-dp`).

---


packages/jsii-pacmak/lib/targets/java.ts
```diff
@@ -492,6 +493,22 @@ class JavaGenerator extends Generator {
...
+        /**
+         * Combines a version number with an optional suffix. If the suffix starts with '-' or '.', it will be
+         * concatenated as-is to the semantic version number. Otherwise, it'll be appended to the version number with an
+         * intercalar '-'.
+         *
+         * @param version the semantic version number
+         * @param suffix  the suffix, if any.
+         */
+        function makeVersion(version: string, suffix?: string): string {
+            if (!suffix) { return version; }
```
eladb: Seems over-designed, I would simply fail if the suffix does not begin with "-" or "."

RomainMuller: I don't like this, would rather *always* prepend a `-`, but it seems controversial enough that I want to be open to either, while making it easier to go with what I would prefer :D

PR #554, opened 2019-06-25
fix(kernel): properly deserialize structs passed in byref

### LINKED ISSUE #553: kernel: does not properly deserializes structs passed in by ref
During deserialization when the kernel expects a struct by gets a reference from the native language, which still happens in languages like .NET, it fails.

eladb:
The deserializer mistakingly used `objectReference` instead of `isObjRef` when trying to identify this case.

Test was added to verify this fix.

Fixes #553

Co-authored-by: Romain Muller <rmuller@amazon.com>


rix0rrr:
I'm not positive this is the right fix. Don't merge yet please.

PR #557, opened 2019-06-25
feat(pacmak): support adding suffix to .NET package versions

RomainMuller:
Allows setting a `jsii.targets.dotnet.versionSuffix` key defines a
string to be appended at the end of the version number of the NuGet
package.

---


packages/jsii-dotnet-generator/src/Amazon.JSII.Generator/VersionSuffixExtensions.cs
```diff
@@ -0,0 +1,26 @@
...
+            return MakeDecoratedVersion(package.Version, package.Targets?.DotNet?.VersionSuffix);
+        }
+
+        private static string MakeDecoratedVersion(string version, string suffix)
+        {
+            if (suffix == null)
+            {
+                return version;
+            }
+            return $"{version}-{suffix}";
```
eladb: Can we make this more consistent with java? Require that suffix will always begin with a `-` at the source

PR #563, opened 2019-06-29
fix(pacmak): fix .NET generated .csproj files using incorrect version for Amazon.JSII.Runtime

shivlaks:
The generated .csproj file is currently using the same version for Package Version and the
version for Amazon.JSII.Runtime.


assyadh:
They were. It's just that in the expected csproj, assembly.jsiiVersion and assembly.version carry the same value, so it had no impact on calc/calc-lib etc.

eladb:
Oh I see.... feels like we need some way to unit test this and/or add to conformance tests so this doesn’t repeat when we implement the next language generator
In

PR #568, opened 2019-07-01
fix(.net): occasional incorrect param type cast

RomainMuller:
When structures (as literal JSON blocks) get passed through a "map"
argument, a `JObject` would be passed to the method despite the declared
type on the .NET code is `IDictionary`, resulting in a crash.

This code forcefully converts `JObject` to `IDictionary<string, objetc>`
in such code paths.

Fixes awslabs/aws-cdk#3093

---


packages/jsii-dotnet-runtime/src/Amazon.JSII.Runtime/Amazon.JSII.Runtime.sln
```diff
@@ -0,0 +1,17 @@
+
```
eladb: Did we intend to add this file?

packages/jsii-dotnet-runtime/src/Amazon.JSII.Runtime/CallbackExtensions.cs
```diff
@@ -136,6 +137,17 @@ private static object FromKernel(object obj, IReferenceMap referenceMap)
                 {
                     return referenceMap.GetOrCreateNativeReference(new ByRefValue(prop.Value.Value<String>()));
                 }
+
+                var dict = ((JObject)obj).ToObject<Dictionary<string, object>>();
```
eladb: add a comment explaining what you are doing here and why

packages/jsii-dotnet-runtime-test/test/Amazon.JSII.Runtime.IntegrationTests/ComplianceTests.cs
```diff
@@ -915,6 +915,22 @@ public void CorrectlyReturnsFromVoidCallback()
             Assert.True(voidCallback.MethodWasCalled);
         }
 
+        [Fact(DisplayName = Prefix + nameof(CallbacksCorrectlyDeserializeArguments))]
```
eladb: We need those tests in all other languages...

packages/jsii-dotnet-runtime/src/Amazon.JSII.Runtime/Amazon.JSII.Runtime.sln
RomainMuller: Duh, no!

packages/jsii-python-runtime/tests/test_compliance.py
```diff
@@ -901,3 +902,11 @@ def consume_partially_initialized_this(self, obj, dt, en):
 def test_variadicMethodCanBeInvoked():
     variadic = VariadicMethod(1)
     assert variadic.as_array(3, 4, 5, 6) == [1, 3, 4, 5, 6]
+
+@pytest.mark.skip # I don't know how to make the .render() call work - it crashes saying it requires 2 named arguments.
```
eladb: @garnaat ?

RomainMuller:
I'm provisionally waiting to see if @garnaat can teach me how to do Python right before merging this. If we don't get around to it fast enough, I'll be fixing the Python test in a separate PR 🤞🏻

RomainMuller:
Talked with @garnaat  offline and managed to get the test to work. Whee.

PR #569, opened 2019-07-01
fix(jsii): correctly inherit initializer stability

rix0rrr:
Fix the inheriting of stability tags for the initializer. It used to
inherit from the package, it now correctly inherits from the class.

Update jsii-reflect with a "load all modules" method (to save time
in a complex dependency tree) and jsii-tree with options to show
stabilities.

### DIFF
--- a/packages/jsii-reflect/bin/jsii-tree.ts
+++ b/packages/jsii-reflect/bin/jsii-tree.ts
@@ -6,9 +6,10 @@ import { TypeSystem, TypeSystemTree } from '../lib';
 
 async function main() {
   const options = yargs
-    .usage('$0 <JSII-FILE | MODULE-DIR...>', 'Prints an ASCII tree representation of a jsii type system.', args => args
+    .usage('$0 [JSII-FILE | MODULE-DIR...]', 'Prints an ASCII tree representation of a jsii type system.', args => args
       .positional('JSII-FILE', { type: 'string', desc: 'path to a .jsii file to load, all dependency .jsii files must be explicitly supplied' })
       .positional('MODULE-DIR', { type: 'string', desc: 'path to an jsii npm module directory, all jsii dependencies will be loaded transitively' }))
+    .option('closure', { type: 'string', alias: 'c', desc: 'Load dependencies of package without assuming its a JSII package itself' })
     .option('all', { type: 'boolean', alias: 'a', desc: 'show all details', default: false })
     .option('colors', { type: 'boolean', desc: 'enable/disable ANSI colors in output', default: true })
     .option('dependencies', { type: 'boolean', alias: 'd', desc: 'show assembly dependencies', default: false })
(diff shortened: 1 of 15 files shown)


packages/jsii-reflect/bin/jsii-tree.ts
```diff
@@ -6,22 +6,29 @@ import { TypeSystem, TypeSystemTree } from '../lib';
...
       .positional('MODULE-DIR', { type: 'string', desc: 'path to an jsii npm module directory, all jsii dependencies will be loaded transitively' }))
+    .option('closure', { type: 'string', alias: 'c', desc: 'Load dependencies of package without assuming its a JSII package itself' })
     .option('all', { type: 'boolean', alias: 'a', desc: 'show all details', default: false })
     .option('colors', { type: 'boolean', desc: 'enable/disable ANSI colors in output', default: true })
     .option('dependencies', { type: 'boolean', alias: 'd', desc: 'show assembly dependencies', default: false })
     .option('inheritance', { type: 'boolean', alias: 'i', desc: 'show base classes and implemented interfaces', default: false })
     .option('members', { type: 'boolean', alias: 'm', desc: 'show type members', default: false })
     .option('signatures', { type: 'boolean', alias: 's', desc: 'show method and property signatures', default: false })
     .option('types', { type: 'boolean', alias: 't', desc: 'show types', default: false })
+    .option('validate', { type: 'boolean', alias: 'V', desc: 'Validate assemblies while loading', default: true })
```
eladb: Validate what?

packages/jsii-reflect/bin/jsii-tree.ts
```diff
@@ -6,22 +6,29 @@ import { TypeSystem, TypeSystemTree } from '../lib';
 
 async function main() {
   const options = yargs
-    .usage('$0 <JSII-FILE | MODULE-DIR...>', 'Prints an ASCII tree representation of a jsii type system.', args => args
+    .usage('$0 [JSII-FILE | MODULE-DIR...]', 'Prints an ASCII tree representation of a jsii type system.', args => args
       .positional('JSII-FILE', { type: 'string', desc: 'path to a .jsii file to load, all dependency .jsii files must be explicitly supplied' })
       .positional('MODULE-DIR', { type: 'string', desc: 'path to an jsii npm module directory, all jsii dependencies will be loaded transitively' }))
+    .option('closure', { type: 'string', alias: 'c', desc: 'Load dependencies of package without assuming its a JSII package itself' })
```
eladb: Not sure I understand this feature? What is the value passed in to this switch?

Feels like you could just auto-discover this when MODULE-DIR is not a jsii module

packages/jsii-reflect/lib/type-system.ts
```diff
@@ -26,6 +27,31 @@ export class TypeSystem {
 
   private readonly _assemblyLookup: { [name: string]: Assembly } = { };
 
+  /**
+   * Load all JSII dependencies of the given NPM package directory.
+   *
+   * The NPM package itself does *not* have to be a jsii package, and does
+   * NOT have to declare a JSII dependency on any of the packages.
+   */
+  public async loadAllNodeModules(packageRoot: string, options: { validate?: boolean } = {}) {
```
eladb: Would make more sense (to me) to look at `package.json` and just load all modules from it (through `require.resolve`). Technically, there could be modules in node_modules that reside on parent directories.

packages/jsii-reflect/bin/jsii-tree.ts
rix0rrr: Well `load()` (and `loadModule()`) expects to be able to return an Assembly, and we can't return an Assembly for a non-JSII package.

packages/jsii-reflect/lib/type-system.ts
rix0rrr: Good point.

PR #586, opened 2019-07-05
feat(python): idiomatic capitalization for structs

### LINKED ISSUE #537: python: another problem with variadic arguments
Sorry for the lengthy writeup here.  I wanted to provide enough context to enable a discussion about this issue because I'm not 100% sure of the right solution.

There was a previous issue (https://github.com/awslabs/jsii/issues/483) with variadic arguments in Python.  This issue involved methods with scalar variadic parameters such as https://github.com/awslabs/aws-cdk/blob/v0.34.0/packages/@aws-cdk/aws-ecs/lib/ec2/ec2-service.ts#L168:

```ts
 /**
   * Try to place tasks spread across instance attributes.
   *
   * You can use one of the built-in attributes found on `BuiltInAttributes`
   * or supply your own custom instance attributes. If more than one attribute
   * is supplied, spreading is done in order.
   *
   * @default attributes instanceId
   * @deprecated Use addPlacementStrategies() instead.
   */
  public placeSpreadAcross(...fields: string[]) {
    if (fields.length === 0) {
      this.addPlacementStrategies(PlacementStrategy.spreadAcrossInstances());
    } else {
      this.addPlacementStrategies(PlacementStrategy.spreadAcross(...fields));
    }
  }
```

This was mapped to a Python signature like this:

```python
 @jsii.member(jsii_name="placeSpreadAcross")
    def place_spread_across(self, *fields: str) -> None:
        return jsii.invoke(self, "placeSpreadAcross", [fields])
```

The problem was that the tuple containing ``fields`` was then passed directly to the Python runtime and it didn't know how to deal with it.  The fix was to change the Python code to this:

```python
 @jsii.member(jsii_name="placeSpreadAcross")
    def place_spread_across(self, *fields: str) -> None:
        return jsii.invoke(self, "placeSpreadAcross", [*fields])
```

This expanded the tuple and fixed the problem.  This was accomplished by changing the Python generator code as shown in https://github.com/awslabs/jsii/pull/513.  The change was very simple, just check if the parameter was variadic when generating the call to JSII and, if so, expand the tuple.

```ts
                // If the last arg is variadic, expand the tuple
	        const paramNames: string[] = [];
	        for (const param of this.parameters) {
	            paramNames.push((param.variadic ? '*' : '') + toPythonParameterName(param.name));
```

This change, however, introduced another issue.  This issue shows up, among other places, in  https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-ecs.ContainerDefinition.html#addportmappingsportmappings-void which looks like this:

```ts
 /**
   * Add one or more port mappings to this container
   */
  public addPortMappings(...portMappings: PortMapping[]) {
    this.portMappings.push(...portMappings.map(pm => {
      if (this.taskDefinition.networkMode === NetworkMode.AwsVpc || this.taskDefinition.networkMode === NetworkMode.Host) {
        if (pm.containerPort !== pm.hostPort && pm.hostPort !== undefined) {
          throw new Error(`Host port ${pm.hostPort} does not match container port ${pm.containerPort}.`);
        }
      }

      if (this.taskDefinition.networkMode === NetworkMode.Bridge) {
        if (pm.hostPort === undefined) {
          pm = {
            ...pm,
            hostPort: 0
          };
        }
      }

      return pm;
    }));
  }
```

In this case the variadic argument is not a scalar value but an object conforming to this interface https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-ecs.PortMapping.html.  The generated Python code looks like this:

```python
    @jsii.member(jsii_name="addPortMappings")
    def add_port_mappings(self, *, container_port: jsii.Number, host_port: typing.Optional[jsii.Number]=None, protocol: typing.Optional["Protocol"]=None) -> None:
        """Add one or more port mappings to this container.

        Arguments:
            portMappings: -
            containerPort: Port inside the container.
            hostPort: Port on the host. In AwsVpc or Host networking mode, leave this out or set it to the same value as containerPort. In Bridge networking mode, leave this out or set it to non-reserved non-ephemeral port.
            protocol: Protocol. Default: Tcp

        Stability:
            experimental
        """
        port_mappings: PortMapping = {"containerPort": container_port}

        if host_port is not None:
            port_mappings["hostPort"] = host_port

        if protocol is not None:
            port_mappings["protocol"] = protocol

        return jsii.invoke(self, "addPortMappings", [*port_mappings])
```

In this case, the parameters are passed as keyword args (the bare asterisk as the first param forces that), the PortMapping structure is built and then wrapped in a list and unpacked.  This will result in a list of the keys present in the dict being passed to ``jsii.invoke``, e.g. ``['containerPort', 'hostPort']`` which results in the following error:

```
  File "app.py", line 40, in <module>
    protocol=ecs.Protocol.Tcp
  File "/Users/garnaat/projects/cdkdev/aws-cdk-examples/python/ecs/ecs-service-with-task-placement/.env/lib/python3.7/site-packages/aws_cdk/aws_ecs/__init__.py", line 1869, in add_port_mappings
    return jsii.invoke(self, "addPortMappings", [*port_mappings])
  File "/Users/garnaat/projects/cdkdev/aws-cdk-examples/python/ecs/ecs-service-with-task-placement/.env/lib/python3.7/site-packages/jsii/_kernel/__init__.py", line 104, in wrapped
    return _recursize_dereference(kernel, fn(kernel, *args, **kwargs))
  File "/Users/garnaat/projects/cdkdev/aws-cdk-examples/python/ecs/ecs-service-with-task-placement/.env/lib/python3.7/site-packages/jsii/_kernel/__init__.py", line 258, in invoke
    args=_make_reference_for_native(self, args),
  File "/Users/garnaat/projects/cdkdev/aws-cdk-examples/python/ecs/ecs-service-with-task-placement/.env/lib/python3.7/site-packages/jsii/_kernel/providers/process.py", line 346, in invoke
    return self._process.send(request, InvokeResponse)
  File "/Users/garnaat/projects/cdkdev/aws-cdk-examples/python/ecs/ecs-service-with-task-placement/.env/lib/python3.7/site-packages/jsii/_kernel/providers/process.py", line 316, in send
    raise JSIIError(resp.error) from JavaScriptError(resp.stack)
jsii.errors.JSIIError: Expected object reference, got "containerPort"
Subprocess exited with error 1
```

Simply removing the ``*`` in the call to ``jsii.invoke`` fixes the problem but this raises two questions.

1. How to discriminate between scalar and non-scalar arguments in the Python code generator.
1. Isn't this signature for ``add_port_mappings`` wrong?  You should be able to pass a list of ``PortMapping`` objects but that is not possible here.

### LINKED ISSUE #577: python: issues when implementing an interface
This example:

```python
from aws_cdk.core import Stack, Construct, IAspect, IConstruct
import jsii

@jsii.implements(IAspect)
class MonitoringAspect:
    def visit(self, node: IConstruct) -> None:
        print("visit %s" % node)

class MyStack(Stack):
    def __init__(self, scope: Construct, cid: str, **kwarg) -> None:
        super().__init__(scope, cid, **kwarg)

        self.node.apply_aspect(MonitoringAspect()) # <-- WARNING: Expected type IAspect got MonitoringStack

__all__ = ["MyStack"]
```

Has the following issues:

1. Warning that says that `apply_aspect` expects `IAspect` and got `MonitoringStack`.
2. When `visit` is called, `node` is an `ObjRef` object instead of an `IConstruct`.

This is the output:

```
visit ObjRef(ref='@aws-cdk/core.Stack@10001')
visit ObjRef(ref='@aws-cdk/aws-dynamodb.Table@10004')
visit ObjRef(ref='@aws-cdk/aws-dynamodb.CfnTable@10019')
visit ObjRef(ref='@aws-cdk/core.Resource@10020')
visit ObjRef(ref='@aws-cdk/aws-lambda.Function@10008')
visit ObjRef(ref='@aws-cdk/aws-iam.Role@10021')
visit ObjRef(ref='@aws-cdk/aws-iam.CfnRole@10022')
visit ObjRef(ref='@aws-cdk/aws-iam.Policy@10023')
visit ObjRef(ref='@aws-cdk/aws-iam.CfnPolicy@10024')
visit ObjRef(ref='@aws-cdk/aws-lambda.CfnFunction@10025')
visit ObjRef(ref='@aws-cdk/aws-s3-assets.Asset@10026')
visit ObjRef(ref='@aws-cdk/assets.Staging@10027')
visit ObjRef(ref='@aws-cdk/core.CfnParameter@10028')
visit ObjRef(ref='@aws-cdk/core.CfnParameter@10029')
visit ObjRef(ref='@aws-cdk/core.CfnParameter@10030')
visit ObjRef(ref='@aws-cdk/core.Resource@10031')
visit ObjRef(ref='@aws-cdk/aws-lambda.CfnPermission@10032')
visit ObjRef(ref='@aws-cdk/aws-lambda.CfnPermission@10033')
```

Looks like the objref is not deserialized in the callback.

### LINKED ISSUE #578: python: data type field names are not capitalized correctly
For example:

```python
@jsii.data_type(jsii_type="@aws-cdk/aws-apigateway.DomainNameOptions", jsii_struct_bases=[_DomainNameOptions])
class DomainNameOptions(_DomainNameOptions):
    certificate: aws_cdk.aws_certificatemanager.ICertificate
    domainName: str
    """
```

I would have expected `domain_name` instead of `domainName`.

Furthermore, there is no ctor, so I don't get type completion when I am creating an instance. Would be nice to add a ctor for all the fields.

rix0rrr:
This change addresses 4 issues:

- Structs now use idiomatic (snake_case) capitalization of fields
  (instead of JSII-inherited camelCase).
- IDE support -- replace TypedDict usage with regular classes. This
  makes it so that we can't use dicts anymore, but mypy support in
  IDEs wasn't great and by using POPOs (Plain Old Python Objects)
  IDEs get their support back.
- Structs in a variadic argument use to be incorrectly lifted to
  keyword arguments, this no longer happens.
- Stop emitting "Stable" stabilities in docstrings, "Stable" is implied.

In order to make this change, I've had to make `jsii-pacmak` depend on
`jsii-reflect`. This is the proper layering of libraries anyway, since
`jsii-reflect` adds much-needed--and otherwise
duplicated--interpretation of the spec.

Complete refactoring of `jsii-pacmak` to build on `jsii-reflect` is too
much work right now, however, so I've added "escape hatching" where
generators can take advantage of the power of jsii-reflect if they want
to, but most of the code still works on the raw spec level.

Added a refactoring where we load the assembly once and reuse the same
instance for all generators, instead of loading the assembly for every
generator. Assembly-loading, especially with a lot of dependencies, takes
a non-negligible amount of time, so this has the side effect of making
the packaging step faster (shaves off 100 packages * 3 targets * a
couple of seconds).

Fixes #537 
Fixes #577 
Fixes #578 
Fixes #588


rix0rrr:
For reference, structs now look like this:

```python
@jsii.data_type(jsii_type="jsii-calc.TopLevelStruct", jsii_struct_bases=[], name_mapping={'required': 'required', 'second_level': 'secondLevel', 'optional': 'optional'})
class TopLevelStruct:
    def __init__(self, *, required: str, second_level: typing.Union[jsii.Number, "SecondLevelStruct"], optional: typing.Optional[str]=None):
        """
        Arguments:
            required: This is a required field.
            second_level: A union to really stress test our serialization.
            optional: You don't have to pass this.

        Stability:
            experimental
        """
        self._values = {
            'required': required,
            'second_level': second_level,
        }
        if optional is not None: self._values["optional"] = optional

    @property
    def required(self) -> str:
        """This is a required field.

        Stability:
            experimental
        """
        return self._values.get('required')

    @property
    def second_level(self) -> typing.Union[jsii.Number, "SecondLevelStruct"]:
        """A union to really stress test our serialization.

        Stability:
            experimental
        """
        return self._values.get('second_level')

    @property
    def optional(self) -> typing.Optional[str]:
        """You don't have to pass this.

        Stability:
            experimental
        """
        return self._values.get('optional')

    def _to_jsii(self) -> dict:
        return {
          'required': self.required,
          'secondLevel': self.second_level,
          'optional': self.optional,
        }

    def __eq__(self, rhs) -> bool:
        return isinstance(rhs, self.__class__) and rhs._values == self._values

    def __ne__(self, rhs) -> bool:
        return not (rhs == self)

    def __repr__(self) -> str:
        return 'TopLevelStruct(%s)' % ', '.join(k + '=' + repr(v) for k, v in self._values.items())
```

rix0rrr:
And a jsii function now looks like this:

```python
    @jsii.member(jsii_name="roundTrip")
    @classmethod
    def round_trip(cls, _positional: jsii.Number, *, required: str, second_level: typing.Union[jsii.Number, "SecondLevelStruct"], optional: typing.Optional[str]=None) -> "TopLevelStruct":
        """
        Arguments:
            _positional: -
            input: -
            required: This is a required field.
            second_level: A union to really stress test our serialization.
            optional: You don't have to pass this.

        Stability:
            experimental
        """
        input = TopLevelStruct(required=required, second_level=second_level, optional=optional)

        return jsii.sinvoke(cls, "roundTrip", jsii.structs_to_jsii([_positional, input]))


```

packages/jsii-python-runtime/src/jsii/_kernel/__init__.py
```diff
@@ -101,18 +101,29 @@ def _recursize_dereference(kernel, d):
 def _dereferenced(fn):
     @functools.wraps(fn)
     def wrapped(kernel, *args, **kwargs):
-        return _recursize_dereference(kernel, fn(kernel, *args, **kwargs))
+        output = fn(kernel, *args, **kwargs)
+        print(output)
```
garnaat: probably don't want to leave this in there

rix0rrr: Tisk :)

rix0rrr:
The last commit, switching to ReST-style docstrings (in order to help IDEs out, poor things!) is speculative.

It might be better with, or without. Any reviewers/mergers/editors of this PR should feel fully free to drop and/or amend it :).

garnaat:
There was a problem with ctors that take no parameters.  It was getting generated as:

```python
class foo():
    def __init__(self, *):
```

but you can't have a bare asterisk that isn't followed by any params.  I made this change:

```
diff --git a/packages/jsii-pacmak/lib/targets/python.ts b/packages/jsii-pacmak/lib/targets/python.ts
index a1a89786..ebd0131c 100644
--- a/packages/jsii-pacmak/lib/targets/python.ts
+++ b/packages/jsii-pacmak/lib/targets/python.ts
@@ -656,8 +656,13 @@ class Struct extends BasePythonClassType {
 
     private emitConstructor(code: CodeMaker, resolver: TypeResolver) {
         const members = this.allMembers;
+        let constructorArguments: string[] = [];
 
-        const constructorArguments = ['self', '*', ...members.map(m => m.constructorDecl(resolver))];
+        if (members.length > 0) {
+            constructorArguments = ['self', '*', ...members.map(m => m.constructorDecl(resolver))];
+        } else {
+            constructorArguments = ['self'];
+        }
 ```

and tested it and it worked.  There may be a better way to do it in TypeScript.  I don't think I can push the change here myself.

File: packages/jsii-pacmak/lib/reflect-hacks.ts [hunk 2]
```diff
@@ -0,0 +1,28 @@
+import * as reflect from 'jsii-reflect';
+import * as spec from 'jsii-spec';
+
+// All current functions are written against jsii-spec, but jsii-reflect
+// is more convenient.
+//
+// These functions exist to break the isolation barrier until we have time
+// to rewrite the code to properly use jsii-reflect.
+
```
eladb: Why not just expose spec as public in Jsii-reflect?

packages/jsii-pacmak/lib/targets/python.ts
```diff
@@ -630,103 +607,159 @@ class InterfaceProperty extends BaseProperty {
     protected readonly shouldEmitBody: boolean = false;
 }
 
-class TypedDict extends BasePythonClassType {
+class Struct extends BasePythonClassType {
+    protected directMembers = new Array<StructField>();
+
+    public addMember(member: PythonBase): void {
+        if (!(member instanceof StructField)) {
+            throw new Error('Must add TypedDictProperties to ValueType');
```
eladb: I am assuming this is `StructField`, no?

packages/jsii-pacmak/lib/targets/python.ts
```diff
@@ -630,103 +607,159 @@ class InterfaceProperty extends BaseProperty {
...
     public emit(code: CodeMaker, resolver: TypeResolver) {
         resolver = this.fqn ? resolver.bind(this.fqn) : resolver;
 
-        // MyPy doesn't let us mark some keys as optional, and some keys as mandatory,
-        // we can either mark either the entire class as mandatory or the entire class
-        // as optional. However, we can make two classes, one with all mandatory keys
-        // and one with all optional keys in order to emulate this. So we'll go ahead
-        // and implement this "split" class logic.
+        // Not actually rendering these as class params,
+        // we splat the entire
```
eladb: "the entire..." what?

packages/jsii-pacmak/lib/targets/python.ts
```diff
@@ -630,103 +607,159 @@ class InterfaceProperty extends BaseProperty {
...
+
+        const constructorArguments = kwargs.length > 0  ? ['self', '*', ...kwargs] : ['self'];
+
+        code.openBlock(`def __init__(${constructorArguments.join(', ')})`);
+        this.emitConstructorDocstring(code);
+
+        // Required properties, those will always be put into the dict
+        code.line('self._values = {');
+        for (const member of members.filter(m => !m.optional)) {
+            code.line(`    '${member.pythonName}': ${member.pythonName},`);
```
eladb: Do we need to validate that we actually got values in all required fields or are we leaving that to jsii?

packages/jsii-pacmak/test/expected.jsii-calc-base/python/src/scope/jsii_calc_base/__init__.py
```diff
@@ -23,18 +23,44 @@ def __init__(self) -> None:
...
 
 class _BaseProxy(Base):
     pass
 
-@jsii.data_type(jsii_type="@scope/jsii-calc-base.BaseProps", jsii_struct_bases=[scope.jsii_calc_base_of_base.VeryBaseProps])
-class BaseProps(scope.jsii_calc_base_of_base.VeryBaseProps, jsii.compat.TypedDict):
-    bar: str
+@jsii.data_type(jsii_type="@scope/jsii-calc-base.BaseProps", jsii_struct_bases=[scope.jsii_calc_base_of_base.VeryBaseProps], name_mapping={'foo': 'foo', 'bar': 'bar'})
+class BaseProps:
+    def __init__(self, *, foo: scope.jsii_calc_base_of_base.Very, bar: str):
```
eladb: Why do we need the "*"?

packages/jsii-pacmak/test/expected.jsii-calc/python/src/jsii_calc/__init__.py
```diff
@@ -1033,114 +1054,183 @@ def prop(self, value: str):
...
             jsii.create(DerivedClassHasNoProperties.Derived, self, [])
 
 
 
-@jsii.data_type_optionals(jsii_struct_bases=[scope.jsii_calc_lib.MyFirstStruct])
-class _DerivedStruct(scope.jsii_calc_lib.MyFirstStruct, jsii.compat.TypedDict, total=False):
-    anotherOptional: typing.Mapping[str,scope.jsii_calc_lib.Value]
-    """This is optional.
+@jsii.data_type(jsii_type="jsii-calc.DerivedStruct", jsii_struct_bases=[scope.jsii_calc_lib.MyFirstStruct], name_mapping={'anumber': 'anumber', 'astring': 'astring', 'first_optional': 'firstOptional', 'another_required': 'anotherRequired', 'bool': 'bool', 'non_primitive': 'nonPrimitive', 'another_optional': 'anotherOptional', 'optional_any': 'optionalAny', 'optional_array': 'optionalArray'})
+class DerivedStruct:
```
eladb: Don't we want this to inherit from `MyFirstStruct`? Did we test struct inheritence?

packages/jsii-python-runtime/src/jsii/_kernel/providers/process.py
```diff
@@ -141,7 +141,7 @@ def jdefault(obj):
         return {"$jsii.date": obj.isoformat()}
     elif isinstance(obj, datetime.datetime):
         raise TypeError("Naive datetimes are not supported, please add a timzone.")
-    raise TypeError
+    raise TypeError("Don't know how to convert object to JSON: %r" % obj)
```
eladb: YAS!

eladb:
@garnaat please review

packages/jsii-pacmak/lib/targets/python.ts
rix0rrr: Oh yes

packages/jsii-pacmak/lib/reflect-hacks.ts
rix0rrr: I went down that path first, but I think it's the wrong way round. Don't feel that spec should be exposed.

packages/jsii-pacmak/lib/targets/python.ts
rix0rrr: Fair point. It's easy enough to add, might as well do it.

packages/jsii-pacmak/test/expected.jsii-calc-base/python/src/scope/jsii_calc_base/__init__.py
rix0rrr: It makes it so you can't pass any of the following values without using the keyword. Which is required, otherwise adding optional parameters later on might change the order which is breaking.

packages/jsii-pacmak/test/expected.jsii-calc/python/src/jsii_calc/__init__.py
rix0rrr: Not even sure what that means.

packages/jsii-pacmak/lib/targets/python.ts
rix0rrr: Ah no, we don't have to, Python will complain. The constructor looks like this:

```ts
def __init__(self, *, required, optional=None):
  ...
```

So you MUST pass a value for required.

packages/jsii-pacmak/test/expected.jsii-calc/python/src/jsii_calc/__init__.py
rix0rrr: Okay it will be used for nominal typing, probably.

packages/jsii-pacmak/test/expected.jsii-calc-base/python/src/scope/jsii_calc_base/__init__.py
eladb: Oh wow. nice!

garnaat:
LGTM

PR #591, opened 2019-07-08
fix(kernel): validate presence of required struct properties

rix0rrr:
Detect when required struct properties are not supplied, so that
we can produce useful error messages to untyped jsii languages (in
particular, Python).

Also check for an array in place of a variadic argument, and throw
a hopefully helpful message to hint at a common misuse.

Fix the test to be actually variadic, which it wasn't :o.

### DIFF
--- a/packages/jsii-kernel/lib/serialization.ts
+++ b/packages/jsii-kernel/lib/serialization.ts
@@ -170,7 +170,7 @@ export const SERIALIZERS: {[k: string]: Serializer} = {
       return value;
     },
     deserialize(value, optionalValue) {
-      // /!\ Top-level "null" will turn to underfined, but any null nested in the value is valid JSON, so it'll stay!
+      // /!\ Top-level "null" will turn to undefined, but any null nested in the value is valid JSON, so it'll stay!
       if (nullAndOk(value, optionalValue)) { return undefined; }
       return value;
     },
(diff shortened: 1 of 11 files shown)


packages/jsii-kernel/lib/serialization.ts
```diff
@@ -296,16 +296,29 @@ export const SERIALIZERS: {[k: string]: Serializer} = {
         throw new Error(`Expected object reference, got ${JSON.stringify(value)}`);
       }
 
-      // Similarly to other end, we might be getting a reference type where we're
-      // expecting a value type. Accept this for now.
+      const namedType = host.lookupType((optionalValue.type as spec.NamedTypeReference).fqn);
+      const props = propertiesOf(namedType, host.lookupType);
+
+      if (Array.isArray(value)) {
```
eladb: Wouldn't it make more sense to verify that the type is the expected type instead of what it is *not*. What if it's a map for example or a primitive?

rix0rrr: A map is actually correct, that's what a struct is (or should be),  and we do already check that a primitive is not passed.

We just missed the Array because of the fact that `typeof array === 'object'` in JavaScript.

By doing the explicit check we also get a nicer error message.

PR #592, opened 2019-07-09
feat(jsii-diff): also check stability transitions

rix0rrr:
Once an API element is marked @stable, it is not allowed to
retract that stability guarantee anymore by changing it back to
@experimental. This is now verified by `jsii-diff`.

This was always intended to be checked, but was missed as an oversight
in the initial implementation.

### DIFF
--- a/packages/jsii-diff/lib/stability.ts
+++ b/packages/jsii-diff/lib/stability.ts
@@ -1,0 +1,45 @@
+import reflect = require('jsii-reflect');
+import spec = require('jsii-spec');
+import { ApiElement, ComparisonContext } from './types';
+
+export function compareStabilities(original: reflect.Documentable & ApiElement, updated: reflect.Documentable, context: ComparisonContext) {
+  // Nothing to do in these cases
+  if (original.docs.stability === undefined || original.docs.stability === updated.docs.stability) { return; }
+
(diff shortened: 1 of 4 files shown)


packages/jsii-diff/lib/stability.ts
```diff
@@ -0,0 +1,45 @@
+import reflect = require('jsii-reflect');
+import spec = require('jsii-spec');
+import { ApiElement, ComparisonContext } from './types';
+
+export function compareStabilities(original: reflect.Documentable & ApiElement, updated: reflect.Documentable, context: ComparisonContext) {
+  // Nothing to do in these cases
+  if (original.docs.stability === undefined || original.docs.stability === updated.docs.stability) { return; }
+
+  // Not allowed to disavow stability
```
eladb: Is that the case where stability is inherited? What happens if you change the stability of the oRent module/type, is it inherited by the Jsii compiler! Can we evened even have an undefined stability for an api?

Also, wasn’t this what we wanted to do for the CFN layer?

rix0rrr: Inheritance has already been calculated at this level. I thought for the CFN layer we introduce a new stability level.

PR #593, opened 2019-07-09
feat(jsii): configure outDir and rootDir for tsc

hoegertn:
This PR adds the possibility to configure the outDir and the rootDir for `tsc`

### DIFF
--- a/packages/jsii/lib/compiler.ts
+++ b/packages/jsii/lib/compiler.ts
@@ -98,10 +98,11 @@ export class Compiler implements Emitter {
         }
 
         const tsconf = this.typescriptConfig!;
+        const pi = this.options.projectInfo;
 
         const prog = ts.createProgram({
             rootNames: this.rootFiles.concat(_pathOfLibraries(this.compilerHost)),
-            options: {...COMPILER_OPTIONS, outDir: this.options.projectInfo.tscOutDir, rootDir: this.options.projectInfo.tscRootDir},
+            options: {...COMPILER_OPTIONS, outDir: pi.tsc && pi.tsc.outDir, rootDir: pi.tsc && pi.tsc.rootDir},
             // Make the references absolute for the compiler
             projectReferences: tsconf.references && tsconf.references.map(ref => ({ path: path.resolve(ref.path) })),
             host: this.compilerHost
@@ -115,14 +116,15 @@ export class Compiler implements Emitter {
      */
     private async _startWatch(): Promise<never> {
         return new Promise<never>(async () => {
-            const projectRoot = this.options.projectInfo.projectRoot;
+            const pi = this.options.projectInfo;
+            const projectRoot = pi.projectRoot;
             const host = ts.createWatchCompilerHost(
                 this.configPath,
                 {
                     ...COMPILER_OPTIONS,
                     noEmitOnError: false,
-                    outDir: this.options.projectInfo.tscOutDir,
-                    rootDir: this.options.projectInfo.tscRootDir,
+                    outDir: pi.tsc && pi.tsc.outDir,
+                    rootDir: pi.tsc && pi.tsc.rootDir,
                 },
                 { ...ts.sys, getCurrentDirectory() { return projectRoot; } }
             );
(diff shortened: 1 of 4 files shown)


packages/jsii/lib/project-info.ts
```diff
@@ -138,6 +140,8 @@ export async function loadProjectInfo(projectRoot: string, { fixPeerDependencies
 
         excludeTypescript: (pkg.jsii && pkg.jsii.excludeTypescript) || [],
         projectReferences: pkg.jsii && pkg.jsii.projectReferences,
+        tscOutDir: pkg.jsii && pkg.jsii.tscoutdir,
```
rix0rrr: Why not call it `tscOutDir` in the jsii config as well?

(thinking aloud) does this make sense in `package.json` or should they be command-line arguments? I guess we're mirroring the `tsconfig.json` here, right?

hoegertn: I named it according to the `outdir` variable but I am happy to rename it. I think this is something I want to configure for my project. I don't think there is much difference in configuring it inside the `jsii` block or in the scripts section ...

RomainMuller: I think I'd like to have options that'll land in `tsconfig.json` modeled in a separate namespace altogether:

```js
/// package.json
{
  // ...
  "jsii": {
    // ...
    "tsc": {
      "outDir": "dist",
      "rootDir": "lib",
    },
    // ...
  },
  // ...
}
```

Then, I'd use the exact same name that is in `tsconfig`'s `compilerOptions` block.

hoegertn: Sounds great. I will update this.

packages/jsii/lib/compiler.ts
```diff
@@ -98,10 +98,11 @@ export class Compiler implements Emitter {
         }
 
         const tsconf = this.typescriptConfig!;
+        const pi = this.options.projectInfo;
 
         const prog = ts.createProgram({
             rootNames: this.rootFiles.concat(_pathOfLibraries(this.compilerHost)),
-            options: COMPILER_OPTIONS,
+            options: {...COMPILER_OPTIONS, outDir: pi.tsc && pi.tsc.outDir, rootDir: pi.tsc && pi.tsc.rootDir},
```
RomainMuller: could be (since we decided to use the same name...):

```ts
{ ...pi.tsc, ...COMPILER_OPTIONS }
```

The order is important here - we don't want `pi.tsc` to be able to override the required settings we have in `COMPILER_OPTIONS`.

packages/jsii/lib/compiler.ts
```diff
@@ -115,10 +116,16 @@ export class Compiler implements Emitter {
...
+            const pi = this.options.projectInfo;
+            const projectRoot = pi.projectRoot;
             const host = ts.createWatchCompilerHost(
                 this.configPath,
-                { ...COMPILER_OPTIONS, noEmitOnError: false },
+                {
+                    ...COMPILER_OPTIONS,
+                    noEmitOnError: false,
+                    outDir: pi.tsc && pi.tsc.outDir,
+                    rootDir: pi.tsc && pi.tsc.rootDir,
```
RomainMuller: Same as above, here I'd just `...pi.tsc` __before__ `...COMPILER_OPTIONS`.

packages/jsii/lib/compiler.ts
```diff
@@ -188,11 +197,13 @@ export class Compiler implements Emitter {
                 lib: COMPILER_OPTIONS.lib && COMPILER_OPTIONS.lib.map(name => name.slice(4, name.length - 5)),
                 // Those int-enums, we need to output the names instead
                 module: COMPILER_OPTIONS.module && ts.ModuleKind[COMPILER_OPTIONS.module],
+                outDir: pi.tsc && pi.tsc.outDir,
```
RomainMuller: Same here :)

packages/jsii/lib/compiler.ts
hoegertn: I think for now this would work, but if someone adds rootDir or outDir to COMPILER_OPTIONS manual overrides would no longer be possible. But I am fine with the change.

PR #596, opened 2019-07-12
feat: add support for "external" stability

rix0rrr:
Add support for a new class of stability called "imported".

This stability class is treated like "stable" for documentation purposes,
but always lead to warnings (not errors) for stability comparison purposes.


packages/jsii/test/test.docs.ts
```diff
@@ -269,6 +269,28 @@ export = {
...
+        }
+      }
+    `);
+
+    const classType = assembly.types!['testpkg.Foo'] as spec.ClassType;
+    const method = classType.methods!.find(m => m.name === 'floop');
+
+    test.deepEqual(classType.docs!.stability, spec.Stability.Imported);
+    test.deepEqual(method!.docs!.stability, spec.Stability.Imported);
+    test.done();
```
eladb: Don’t we need a test which verifies that imported can break with only warnings?

rix0rrr: Yep: https://github.com/aws/jsii/pull/596/files#diff-bb49180729f436b791f8064a35fd60d4R24

rix0rrr:
We decided on `@stability external`

eladb:
Discussed face to face and decided on `@stability external`

PR #671, opened 2019-08-01
chore(ci): remove check against source tree from CodeBuild buildspec

shivlaks:
Modification to the check introduced in #668 

CodePipeline downloads the source as a zip from the source provider rather than doing a Git clone.
This means the .git folder won't be retained and git commands won't work in CodeBuild. 

---


shivlaks:
> I think @RomainMuller also added this to the aws-cdk repo...

good callout, will check and do the same once I verify that this fixes it

PR #680, opened 2019-08-06
feat(dotnet): handling optional and variadic parameters

### LINKED ISSUE #153: Support variadic parameters for .NET
This might be more involved that just adding the "params" keyword. Consider the following signature:

    public void DoSomething(params object[] myParam)
    {
    }

    // ...

    object[] foo = // ...
    // Because an object[] is also an object, the compiler treats foo as the first element
    // of the params array, rather than the params array itself.
    DoSomething(foo);

But in this case, there's no ambiguity:

    public void DoSomething(params string[] myParam)
    {
    }

    // ...

    string[] foo = // ...

    // The compiler treats foo as the params array, rather than the first element of the params array.
    DoSomething(foo);

This might not matter, since generated code does not include any invocations of the generated methods. Just something to keep in mind.

### LINKED ISSUE #210: dotnet: support optional arguments in methods and constructors
jsii methods/constructors can have optional arguments. The .NET generator should respect those and pass `null` in calls that don't supply them.

-- from @mpiroc:

`jsii-dotnet-generator` should supply default values for arguments with optional types

`jsii-dotnet-generator` ensures that `null` is a valid value for types marked `optional` in JSII. For value types, this means making them nullable, i.e. `int?`. For reference types, `null` is already a valid value.

When a type like this is used in a parameter list, `jsii-dotnet-generator` should create a default value of `null` for the parameter. Currently, the user has to explicitly pass `null`.

assyadh:
Handling optional and variadic parameters in .NET

Emits the =null or params[] keywords when required in constructors or methods.

Ran pack.sh in the CDK with this change, and the S3 construct now looks better:

**Optionals:**

`public Bucket(Amazon.CDK.Construct scope, string id, Amazon.CDK.AWS.S3.IBucketProps props = null): base(new DeputyProps(new object[]{scope, id, props}))
        {
        }
`

Making the C# call look like:

` var bucket = new Bucket(this, "bucketName");`

Rather than

` var bucket = new Bucket(this, "bucketName", null);`

**Variadic:**

Tested with null values, empty array, one value array, multiple values array.

```
// Array with no value in constructor params
var variadicClassNoParams = new VariadicMethod();

// Array with null value in constructor params
var variadicClassNullParams = new VariadicMethod(null);

// Array with one value in constructor params
var variadicClassOneParam = new VariadicMethod(1);

// Array with multiple values in constructor params
var variadicClassMultipleParams = new VariadicMethod(1, 2, 3, 4);
```

Fixes #153
Fixes #210
---


assyadh:
> Should `variadic` arguments be defaulted to `[]`?

In C#, a parameter array can not have a default value. It is by definition optional because it is passed as an empty array if not specified.

C# spec: http://download.microsoft.com/download/3/8/8/388e7205-bc10-4226-b2a8-75351c669b09/csharp%20language%20specification.doc

Section 7.4.1: 

> In particular, note that an empty array is created when there are zero arguments given for the parameter array.

This *means* no changes to the runtime should be required, as the parameter will be passed as an empty array.

packages/jsii-pacmak/lib/targets/dotnet/dotnetgenerator.ts
```diff
@@ -443,18 +437,36 @@ export class DotNetGenerator extends Generator {
...
+            for (const p of parameters) {
+                let optionalPrimitive = '';
+                let optionalKeyword = '';
+                let type = this.typeresolver.toDotNetType(p.type);
+                if (p.optional) {
+                    if (this.isOptionalPrimitive(p)) {
+                        optionalPrimitive = '?';
+                        optionalKeyword = ' = null';
+                    } else {
+                        optionalKeyword = ' = null';
```
costleya: The embedded if and else both set optionalKeyword to the same. Move this set to the parent if.

packages/jsii-pacmak/lib/targets/dotnet/dotnetgenerator.ts
```diff
@@ -443,18 +437,36 @@ export class DotNetGenerator extends Generator {
...
+                    }
+                }
+                if (p.variadic) {
```
costleya: is this an else if?

packages/jsii-pacmak/lib/targets/dotnet/dotnetgenerator.ts
assyadh: Oops, this is a refactoring leftover. Removing

packages/jsii-pacmak/lib/targets/dotnet/dotnetgenerator.ts
assyadh: It looks like JSII is already doing the "mutual exclusion", but no harm in enforcing it with an if/else

RomainMuller: Yeah actually **TypeScript** won't allow you to use *both* the variadic syntax and the optional syntax at the same time... So if it's variadic, it cannot be optional, and vice-versa. But there's definitely no harm in checking again.

assyadh:
> Have we made sure this works with the current runtime? Have the compliance tests been updated to reflect this change?

Tested with the CDKHello app (the S3 bucket instruction used to require a useless new BucketProps(), not anymore), calling 'cdk-synth' works fine.

Tested with the ported EC2/ECS sample stack, which used to require useless new Ec2TaskDefinitionProps() and useless new RepositoryImageProps(), calling 'cdk-synth' works fine.


Looks like the variadic might need some runtime work. Looking into this and holding off the PR so far.

assyadh:
This is ready to be reviewed/merged. tested with the CDK for Lambda and variadic layers.

Also added a full suite of compliance tests that run with the runtime with all sorts of tests around optionals and variadic params (nulls, empty, one value array, multiple values array etc..).

RomainMuller:
Looking super good. Love the small enhancements to readability that you brought in there & the comments... Sweet sweet stuff!

PR #692, opened 2019-08-09
fix(java): improve property override detection

RomainMuller:
The previous code only checked whether a methodd name was a get/setter
by checking if it's name started by `get` or `set` and was at least 4
characters long. This is insufficient, as a method called `settings`
would be misidentified.

Changed to expect getters to have zero parameters, and setters to have
exactly one parameter and to match a corresponding getter.

---

### DIFF
--- a/packages/jsii-java-runtime/project/src/main/java/software/amazon/jsii/Util.java
+++ b/packages/jsii-java-runtime/project/src/main/java/software/amazon/jsii/Util.java
@@ -43,8 +43,8 @@ final class Util {
 
     /**
      * Checks if the method name looks like a java property getter (getXxx).
-     * @param methodName The name of the method.
-     * @return True if the name looks like getXxx.
+     * @param method The reflected method that may be a get/setter
+     * @return true if the method is a get/setter.
      */
     static boolean isJavaPropertyMethod(final Method method) {
         final String methodName = method.getName();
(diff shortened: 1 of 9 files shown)


packages/jsii-java-runtime/project/src/main/java/software/amazon/jsii/Util.java
```diff
@@ -45,18 +46,46 @@ static String readString(final InputStream is) {
      * @param methodName The name of the method.
```
skinny85: This JavaDoc is a little out of date.

packages/jsii-java-runtime/project/src/main/java/software/amazon/jsii/Util.java
```diff
@@ -45,18 +46,46 @@ static String readString(final InputStream is) {
...
+        return false;
+    }
+
+    private static boolean isMatchingGetterPresent(final String getterName, final Class<?> returnType, final Class<?> declaring) {
+        try {
+            final Method getter = declaring.getDeclaredMethod(getterName);
+            return getter.getReturnType().equals(returnType);
+        } catch (final NoSuchMethodException nsme) {
+            return declaring.equals(Object.class)
+                                                  ? false
```
skinny85: This indent is weird...? Should be just indented one below the previous line.

packages/jsii-java-runtime/project/src/main/java/software/amazon/jsii/Util.java
```diff
@@ -45,18 +46,46 @@ static String readString(final InputStream is) {
...
+                                                  : isMatchingGetterPresent(getterName, returnType, declaring.getSuperclass());
+        }
     }
 
     /**
      * Convert a java property method name (getXxx/setXxx) to a javascript property name (xxx).
      * @param getterSetterMethod The java method name
```
skinny85: JavaDoc here as well.

packages/jsii-java-runtime/project/src/main/java/software/amazon/jsii/Util.java
RomainMuller: Uh - yeah!

RomainMuller:
I don't know what caused the `package-lock.json` changes but they keep coming back if I try to undo them. I am actually suspecting the no-diff enforcement of the Travis build is somehow not working here...

PR #696, opened 2019-08-12
chore: clean up repository root, shrinkwrap CLIs

RomainMuller:
Move all files that don't need to be at the repository root under the
dedicated sub-directory for those. This means all `.sh` files were moved
under `scripts`. Some of those were also renamed, and the common uses
have been aliased using `package.json` scripts (`npm run bootstrap`,
`npm run build`, `npm run package`, ...).

Additionally, ran `npm shrinkwrap` on the major CLIs (`jsii`,
`jsii-diff`, `jsii-pacmak`) as a way to improve overall reliability.

---


RomainMuller:
@eladb - the shrink-wrap is packaged with the library, whereas `package-lock.json` is *always* excluded from the `npm` package.

PR #703, opened 2019-08-13
fix(dotnet): add missing GetInterfaceType in the .NET runtime

assyadh:
Add support for return type being array of interfaces.

Tested with the customer code. 

https://docs.aws.amazon.com/cdk/api/latest/dotnet/api/Amazon.CDK.AWS.EC2.Vpc.html#Amazon_CDK_AWS_EC2_Vpc_PrivateSubnets

This was not supported by the runtime, as it only allowed for classes and enums:

```
public virtual IISubnet[] PrivateSubnets { get; }
```

Fixes aws/aws-cdk#2362

---


packages/jsii-dotnet-runtime/src/Amazon.JSII.Runtime/Services/TypeCache.cs
```diff
@@ -69,6 +69,12 @@ public Type GetFrameworkType(TypeReference typeReference, bool isOptional)
                 {
                     return MakeNullableIfOptional(enumType);
                 }
+                
+                Type interfaceType = GetInterfaceType(typeReference.FullyQualifiedName);
```
RomainMuller: Can we please add a test around this particular scenario?

PR #705, opened 2019-08-13
feat(java): detect & rename members named after reserved words

bmaizels:
---

### DIFF
--- a/packages/jsii-pacmak/lib/targets/java.ts
+++ b/packages/jsii-pacmak/lib/targets/java.ts
@@ -1,4 +1,5 @@
 import clone = require('clone');
+import {toPascalCase} from "codemaker/lib/case-utils";
 import fs = require('fs-extra');
 import spec = require('jsii-spec');
 import path = require('path');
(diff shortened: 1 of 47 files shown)


packages/jsii-pacmak/lib/targets/java.ts
```diff
@@ -183,6 +184,50 @@ interface JavaProp {
...
+
+        if (JavaGenerator.RESERVED_KEYWORDS.includes(propertyName)) {
+            return `${propertyName}Value`;
+        } else {
+            return propertyName;
+        }
+    }
+
+    /**
+     * Turns a raw javascript method name (eg: 'import') into a safe Java method name 9eg: 'doImport').
```
RomainMuller: 9eg?

packages/jsii-pacmak/lib/targets/java.ts
```diff
@@ -183,6 +184,50 @@ interface JavaProp {
...
+     * Turns a raw javascript property name (eg: 'default') into a safe Java property name (eg: 'defaultValue').
+     * @param propertyName the raw JSII property Name
+     */
+    private static safeJavaPropertyName(propertyName: string) {
+        if (!propertyName) {
+            return propertyName;
+        }
+
+        if (JavaGenerator.RESERVED_KEYWORDS.includes(propertyName)) {
+            return `${propertyName}Value`;
```
RomainMuller: Might be worth checking whether this is the name of an existing property on the same type... If that is so, just bail out saying you're out of options there...

packages/jsii-pacmak/lib/targets/java.ts
```diff
@@ -183,6 +184,50 @@ interface JavaProp {
...
+     * Turns a raw javascript method name (eg: 'import') into a safe Java method name 9eg: 'doImport').
+     * @param methodName
+     */
+    private static safeJavaMethodName(methodName: string) {
+        if (!methodName) {
+            return methodName;
+        }
+
+        if (JavaGenerator.RESERVED_KEYWORDS.includes(methodName)) {
+            return `do${toPascalCase(methodName)}`;
```
RomainMuller: Same as above, would be nice to have a collision check.

packages/jsii-pacmak/lib/targets/java.ts
bmaizels: About to be fixed.

packages/jsii-pacmak/lib/targets/java.ts
bmaizels: Difficult to do, so in the interests of unblocking CDK builds moving ahead without this functionality for now.

packages/jsii-pacmak/lib/targets/java.ts
bmaizels: Difficult to do, so in the interests of unblocking CDK builds moving ahead without this functionality for now.

RomainMuller:
Looks good - just waiting for a confirmation build of the CDK to succeed before I approve this... To be safe...

PR #717, opened 2019-08-17
feat(dotnet): [JsiiOptional] attribute on properties that are optionals + Roslyn Analyzer

assyadh:
On properties that are optionals, we now add a [JsiiOptional] attribute.

Ex:

```

[JsiiProperty(name: "optional1", typeJson: "{\"primitive\":\"string\"}", isOptional: true, isOverride: true)]
[System.Obsolete()]
[JsiiOptional]
public string Optional1
{
        get;
        set;
}
```

Also, This introduces a new package which needs to be pushed to NuGet: Amazon.Jsii.Analyzers.

It includes a Roslyn Analyzer which makes sure that inside nested props object initializations, all required properties are passed by the developer, and are not passed as null (syntax check only).

Example in Rider with a required property not passed:

Screen Shot 2019-08-16 at 5 13 35 PM (2)

Whit a required property passed with a null value:

Screen Shot 2019-08-16 at 5 14 12 PM (2)

With the required property correctly passed:

Screen Shot 2019-08-16 at 5 14 20 PM (2)

The idea here is to check inline initialization of data types within jsii classes.

---


packages/jsii-dotnet-analyzers/LICENSE
```diff
@@ -0,0 +1,202 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
RomainMuller: `2018-2019`

packages/jsii-dotnet-analyzers/build.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+set -euo pipefail
+
+dotnet build --force -c Release ./src/Amazon.JSII.Analyzers
+
+cp -f ./bin/Release/NuGet/*.nupkg .
```
RomainMuller: Should this be copied to `dist` instead?

packages/jsii-dotnet-analyzers/package.json
```diff
@@ -0,0 +1,31 @@
...
+  },
+  "author": {
+    "name": "Amazon Web Services",
+    "url": "https://aws.amazon.com",
+    "email": "aws-jsii@amazon.com"
+  },
+  "license": "Apache-2.0",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/aws/jsii.git"
```
RomainMuller: drop in `directory` here as well, pointing to the sub-directory where the package is.

packages/jsii-dotnet-analyzers/src/Amazon.JSII.Analyzers.UnitTests/Helpers/DiagnosticVerifier.Helper.cs
```diff
@@ -0,0 +1,170 @@
...
+        private static readonly MetadataReference SystemCoreReference = MetadataReference.CreateFromFile(typeof(Enumerable).Assembly.Location);
+        private static readonly MetadataReference CSharpSymbolsReference = MetadataReference.CreateFromFile(typeof(CSharpCompilation).Assembly.Location);
+        private static readonly MetadataReference CodeAnalysisReference = MetadataReference.CreateFromFile(typeof(Compilation).Assembly.Location);
+
+        internal static string DefaultFilePathPrefix = "Test";
+        internal static string CSharpDefaultFileExt = "cs";
+        internal static string VisualBasicDefaultExt = "vb";
+        internal static string TestProjectName = "TestProject";
+
+        #region  Get Diagnostics
```
RomainMuller: Wish the runtime library was full of those, too... ❤️

packages/jsii-dotnet-analyzers/src/Amazon.JSII.Analyzers.UnitTests/JsiiOptionalAnalyzerTests.cs
```diff
@@ -0,0 +1,114 @@
...
+                        OptionalProperty2 = ""test""
+                        RequiredProperty2 = ""test""
+                    );
+                }
+            }";
+            var expected = new List<DiagnosticResult>()
+            {
+                new DiagnosticResult()
+                {
+                    Id = "CDK001",
```
RomainMuller: None of those should be `CDK<anything>`. Should be `JSII<anything>`

packages/jsii-dotnet-analyzers/src/Amazon.JSII.Analyzers/JsiiOptionalAnalyzer.cs
```diff
@@ -0,0 +1,141 @@
...
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace Amazon.JSII.Analyzers
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public class JsiiOptionalAnalyzer : DiagnosticAnalyzer
+    {
+        private const string DiagnosticId = "CDK001";
```
RomainMuller: `JSII001`

packages/jsii-dotnet-analyzers/src/Amazon.JSII.Analyzers/JsiiOptionalAnalyzer.cs
```diff
@@ -0,0 +1,141 @@
...
+        private const string Title = "A required property is missing or null";
+        private const string MessageFormat = "The property is required and can not be null";
+        private const string MessageFormatWithPropertyName = "The property {0} is required and can not be null";
+        private const string Description = "The property is required and can not be null";
+        private const string DescriptionWitPropertyName = "The property {0} is required and can not be null";
```
RomainMuller: For all of those, I'd have written as `cannot` instead of `can not`. Maybe just me?

packages/jsii-dotnet-analyzers/src/Amazon.JSII.Analyzers/tools/install.ps1
```diff
@@ -0,0 +1,58 @@
+param($installPath, $toolsPath, $package, $project)
```
RomainMuller: What's the use of this? How are people supposed to be using this?

packages/jsii-dotnet-analyzers/src/Directory.Build.props
```diff
@@ -0,0 +1,6 @@
+<Project>
+  <PropertyGroup>
+    <JsiiVersion>0.15.0</JsiiVersion>
```
RomainMuller: Any chance this is actually a generated file that shouldn't be checked in?

packages/jsii-dotnet-analyzers/src/NuGet.Metadata.props
```diff
@@ -0,0 +1,16 @@
+<Project>
```
RomainMuller: And here, too... Generated file?

packages/jsii-dotnet-runtime/src/Amazon.JSII.Runtime/Deputy/JsiiOptionalAttribute.cs
```diff
@@ -0,0 +1,9 @@
+using System;
+
+namespace Amazon.JSII.Runtime.Deputy
+{
+    [AttributeUsage(AttributeTargets.Property)]
+    public class JsiiOptionalAttribute : Attribute
```
RomainMuller: I'd love some documentation on this... To explain exactly what it means (this one is obvious - but it's a good habit to build :D)

packages/jsii-dotnet-analyzers/build.sh
assyadh: I based this file on the jsii-dotnet-runtime's one. It does not have a dist directory?

packages/jsii-dotnet-analyzers/package.json
assyadh: Done

packages/jsii-dotnet-analyzers/src/Amazon.JSII.Analyzers.UnitTests/JsiiOptionalAnalyzerTests.cs
assyadh: Done!

packages/jsii-dotnet-analyzers/src/Amazon.JSII.Analyzers/JsiiOptionalAnalyzer.cs
assyadh: Both are acceptable, but I agree that the first one is more common

packages/jsii-dotnet-analyzers/src/Amazon.JSII.Analyzers/tools/install.ps1
assyadh: The only (ugly way) to use an analyzer inside an IDE is to reference its dll directly.

So, in order to avoid that, Microsoft gave Roslyn Analyzers vended by NuGet packages the option to be installed as analyzers, which really just adds this dll reference (not visible in the csproj though).

This is done by adding an install and uninstall scripts in the nupkg. When the analyzer is pulled from NuGet, it gets installed using that script in tools/install.ps1

I have tested this with Rider and Visual Studio. 

See: https://docs.microsoft.com/fr-fr/nuget/guides/analyzers-conventions

packages/jsii-dotnet-analyzers/src/Directory.Build.props
assyadh: Yes, removing them

packages/jsii-dotnet-analyzers/src/NuGet.Metadata.props
assyadh: Yes, removing them

packages/jsii-dotnet-runtime/src/Amazon.JSII.Runtime/Deputy/JsiiOptionalAttribute.cs
assyadh: Added a summary :-)

packages/jsii-dotnet-analyzers/src/Amazon.JSII.Analyzers.UnitTests/Helpers/DiagnosticVerifier.Helper.cs
assyadh: Pretty useful indeed. It can be tricky as everything has to be declared in a single file so that the diagnostic works (no references), but it works fine

assyadh:
Made some changes on the analyzer rules following our meeting this morning. 

See the Unit tests comments for some background :-)

assyadh:
Travis build is blocked due to a transient issue:

Error downloading packages:
  geronimo-jms-1.1.1-19.amzn2.noarch: [Errno 256] No more mirrors to try.

assyadh:
> [SuppressMessage("Usage", "JSII001")] is probably not quite what you want. Should we make it "Amazon.JSII" as the category?

It looks correct to me:

https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.suppressmessageattribute.-ctor?view=netframework-4.8#System_Diagnostics_CodeAnalysis_SuppressMessageAttribute__ctor_System_String_System_String_

PR #739, opened 2019-08-26
feat(jsii): improve locality of warnings emitted for reserved words

RomainMuller:
Instead of tagging the entire declaration, only tag the name part of it
when emitting warnings about usage of reserved words from one of the
supported languages. This improves the actionability of the messages,
and makes for shorter log entries, too.

---

### DIFF
--- a/packages/jsii/lib/assembler.ts
+++ b/packages/jsii/lib/assembler.ts
@@ -1018,7 +1018,7 @@ export class Assembler implements Emitter {
   private _warnAboutReservedWords(symbol: ts.Symbol) {
     const reservingLanguages = isReservedName(symbol.name);
     if (reservingLanguages) {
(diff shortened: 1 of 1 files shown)


packages/jsii/lib/assembler.ts
```diff
@@ -1018,7 +1018,7 @@ export class Assembler implements Emitter {
   private _warnAboutReservedWords(symbol: ts.Symbol) {
     const reservingLanguages = isReservedName(symbol.name);
     if (reservingLanguages) {
-      this._diagnostic(symbol.valueDeclaration,
+      this._diagnostic(ts.getNameOfDeclaration(symbol.valueDeclaration) || symbol.valueDeclaration,
```
rix0rrr: Is this something we need to be calling in more places?

PR #747, opened 2019-08-27
fix(jsii): allow more flexible package versioning

skinny85:
Before this change, the versions of a single dependency between `dependency`
and `peerDependency` sections of a JSII package had to be identical.
This loosens this requirement: they now have to be only intersecting
(in the Semantic Versioning sense of the word).

Needed for https://github.com/aws/aws-cdk/issues/3711

---

### DIFF
--- a/packages/jsii/lib/project-info.ts
+++ b/packages/jsii/lib/project-info.ts
@@ -77,7 +77,8 @@ export async function loadProjectInfo(projectRoot: string, { fixPeerDependencies
         version = _resolveVersion(version as any, projectRoot).version;
         pkg.peerDependencies = pkg.peerDependencies || {};
         const peerVersion = _resolveVersion(pkg.peerDependencies[name], projectRoot).version;
-        if (peerVersion === version) {
+        if ((!peerVersion && !version) || (peerVersion && version &&
+                new semver.Range(peerVersion).intersects(new semver.Range(version as string)))) {
             return;
         }
         if (!fixPeerDependencies) {
--- a/packages/jsii/test/negatives.test.ts
+++ b/packages/jsii/test/negatives.test.ts
@@ -6,6 +6,9 @@ import { ProjectInfo } from '../lib/project-info';
 
 const SOURCE_DIR = path.join(__dirname, 'negatives');
 
+// invoking the compiler often doesn't complete in Jest's default 5 second timeout
+jest.setTimeout(60_000);
+
 for (const source of fs.readdirSync(SOURCE_DIR)) {
     if (!source.startsWith('neg.') || !source.endsWith('.ts') || source.endsWith('.d.ts')) { continue; }
     const filePath = path.join(SOURCE_DIR, source);
--- a/packages/jsii/test/project-info.test.ts
+++ b/packages/jsii/test/project-info.test.ts
@@ -134,6 +134,30 @@ describe('loadProjectInfo', () => {
             info.peerDependencies[TEST_DEP_ASSEMBLY.name] = '^42.1337.0';
         })
     );
+
+    test('loads with a different, but intersecting, peerDependency version (when not auto-fixing)', () => _withTestProject(async projectRoot => {
+            await loadProjectInfo(projectRoot, { fixPeerDependencies: false });
+
+            const info = require(path.join(projectRoot, 'package.json'));
+            expect(info.dependencies[TEST_DEP_ASSEMBLY.name]).toBe("1.2.4");
+            expect(info.peerDependencies[TEST_DEP_ASSEMBLY.name]).toBe("^1.2.4");
+        }, info => {
+            info.dependencies[TEST_DEP_ASSEMBLY.name] = '1.2.4';
+            info.peerDependencies[TEST_DEP_ASSEMBLY.name] = '^1.2.4';
+        })
+    );
+
+    test('loads with a different, but intersecting, dependency version (when not auto-fixing)', () => _withTestProject(async projectRoot => {
+            await loadProjectInfo(projectRoot, { fixPeerDependencies: false });
+
+            const info = require(path.join(projectRoot, 'package.json'));
+            expect(info.dependencies[TEST_DEP_ASSEMBLY.name]).toBe("^1.2.4");
+            expect(info.peerDependencies[TEST_DEP_ASSEMBLY.name]).toBe("1.2.4");
+        }, info => {
+            info.dependencies[TEST_DEP_ASSEMBLY.name] = '^1.2.4';
+            info.peerDependencies[TEST_DEP_ASSEMBLY.name] = '1.2.4';
+        })
+    );
 });
 
 const TEST_DEP_ASSEMBLY: spec.Assembly = {


eladb:
I am wondering: why aren't we simply enforcing that peer dependencies will use pinned versions? Wouldn't the same issue we found with the CDK apply to any jsii module? If this is "the right thing" to do, then we shouldn't offer it as a degree of freedom.

@RomainMuller what do you think?

RomainMuller:
Yeah, I think the "degree of freedom" that the `peerDependency` buys is actually not real.

Most non-Javascript languages don't allow for multiple versions of the same package in a given dependency closure (and some might resolve in arbitrary ways when the requirements mention different versions at different points).

For that reason, `jsii` needs to ensure that the `npm` behavior will not get multiple versions at different points... Specifying a `peerDependency` that is not equal to the `dependency` will actually end up causing `npm` to install two different versions of the library - one under the `node_modules` of the package declaring the `dependency`, and one in the consuming package that declares a different version requirement... Where we want to maximize chances for `npm` to de-duplicate.

So TL;DR I think we actually still need to have `peerDependency[X] === dependency[X]`.

eladb:
Wait! This contradicts the conclusion @skinny85 made about the issue we saw in the CDK. My understanding is that peer deps should be fixed and normal deps can use semantic versioning, as long as it is compatible with the fixed peer dependency.

rix0rrr:
> My understanding is that peer deps should be fixed and normal deps can use semantic versioning

Assuming by "can use semantic versioning" you mean "can specify a version range", I believe we can't use a version range for dependencies.

1) NPM will **install** `dependencies`
2) NPM will only **warn about** `peerDependencies`

(Ideally we would only specify `peerDepencencies`, and have the user pull in all packages at the desired version, according to the version constraints specified by `peerDependencies`. However, that is too much onus on the user, so we populate every package's `dependencies` with a version that will auto-install the other packages, so they only have to specify the top of the dependency tree.)

On installing: if we *don't* do exact versions, users will not be able to roll back from buggy patch releases (if the version string is `"^1.5.0"` and our release `1.5.1` introduces a bug, users will not be able to roll back to `1.5.0`). That is exactly what we're trying to avoid.

Does this cause problems?

* If 2 different package.jsons specify two different exact versions we will get 2 copies of the package (bad, but should never occur since all CDK packages of the same version will have a dependency on the exact version).
* If one package.json specifies an exact version (CDK library package.json) and the other package.json specifies a version range (user package.json), I *hope* we will only get 1 copy of the package, matching the exact version. However, it could totally be that NPM will be silly and install 2 copies: `the exact version asked for` and `the *latest* version that matches the range`. If it did this, that would be bad. We need to test this somehow.

Because of (2), we would only get into an error situation if `dependencies` and `peerDependencies` don't intersect. As far as I can tell, peer dependencies don't actually lead to any version selection. 

For example, I think (warning: needs to be tested):

```
dependency: ^1.0.0
peerDependency: ~1.5.0
```

Would still select version `1.6.0` if it was available, even though that doesn't match the `peerDependency` range, and subsequently would lead to an error.

But since we have (1) and it's guaranteed that NPM will pull in an exact version of the package in `dependencies`, there is no need to have a range in `peerDependencies` either. The range could be wider, but the package is always guaranteed to be used with the exact version anyway.

So feels like our versioning should look like this:

```
USER package.json:
dependencies: {
  @aws-cdk/aws-ec2: ^1.5.0
}

@aws-cdk/core package.json:
dependencies: {
  @aws-cdk/core: 1.5.0
}
peerDependencies: {
  @aws-cdk/core: 1.5.0
}
```

Then package 1.5.1 and 1.6.0 of `aws-ec2` will be installed in by the user as they become available (which will transitively pull in the same version of the core package automatically).

nija-at:
> For that reason, jsii needs to ensure that the npm behavior will not get multiple versions at different 
> points... Specifying a peerDependency that is not equal to the dependency will actually end up
> causing npm to install two different versions of the library - one under the node_modules of the 
> package declaring the dependency, and one in the consuming package that declares a different 
> version requirement... Where we want to maximize chances for npm to de-duplicate.

Why is this bad? I thought this was the behaviour our customers wanted i.e. the original request in https://github.com/aws/aws-cdk/issues/3711

If the `package.json` of the app specifies
```json
"name": "cdk-app",
"version": "0.1.2",
"dependencies": {
  "@aws-cdk/aws-lambda": "1.2.0",
  "@aws-cdk/aws-iam": "^1.5.0"
}
```
Assuming `aws-lambda` depends on `aws-iam`, the `node_modules` folders for the app will be structured as (ignoring folders named `node_modules` for simplicity)
```sh
cdk-app@0.1.2
├── @aws-cdk/aws-iam@1.6.0
└─┬ @aws-cdk/aws-lambda@1.2.0
  └── @aws-cdk/aws-iam@1.2.2
```
In words, the app will use the latest 1.x release of aws-iam and the 1.2.0 release of aws-lambda. However, aws-lambda, will transitively use the 1.2.x version of its dependencies. 

**If this is the behaviour we want,** then we need to pin (or `~`, so bug patches get applied) the dependencies and `^` our peerDependencies.

rix0rrr:
> In words, the app will use the latest 1.x release of aws-iam and the 1.2.0 release of aws-lambda. However, aws-lambda, will transitively use the 1.2.x version of its dependencies.

This is well and all AS LONG AS no types from any of the dependencies are used in the public interface of a module. Another way to state that: as long as the dependencies are only used in the *implementation* of the module, to ensure some kind of behavior.  The behavior will be locked in place and won't change because the dependency version won't change.

As soon as types start to leak through public APIs though, the privately-versioned dependency thing doesn't work so well anymore. For example let's say `aws-lambda` returns a `PolicyStatement` object from its public interface. The `PolicyStatement@1.2.2` created and returned by `aws-lambda@1.2.0` escapes the boundaries of its "version set". It may get passed into a function  of `aws-iam@1.6.0`, which expects it to be a `PolicyStatement@1.6.0` and have some property or method that it doesn't (yet) have, call it and fail.

We'd have `peerDependencies` to fix this (all types used in the public API of a package must be from packages locked in `peerDependencies`). Since the CDK is all about classes from different packages working together, this happens all the time. We barely have any private implementation dependencies, and basically every construct that would be privately created escapes out through the construct tree anyway. So all our dependencies are also `peerDependencies` at the same version, removing basically all degrees of freedom, leading to a situation in which we might as well version-lock everything together.

Not to mention that people in other package manager ecosystems (Maven, PyPI) don't even have the choice of having "package-private" dependencies, so everything needs to version together anyway.

skinny85:
So what's the conclusion here? How do we want to model dependencies in the CDK, and do we need this change in JSII to accomplish that?

nija-at:
> For example let's say aws-lambda returns a PolicyStatement object from its public interface. The 
> PolicyStatement@1.2.2 created and returned by aws-lambda@1.2.0 escapes the boundaries of its 
> "version set". It may get passed into a function of aws-iam@1.6.0, which expects it to be a 
> PolicyStatement@1.6.0 and have some property or method that it doesn't (yet) have, call it and fail.

Good point. Would it be a ts compilation error, in this case? I suppose so but I can't really say; I've not hit this kind of problem.
Nevertheless, I see how this can be confusing, or error prone in some cases.

What's your recommendation here @rix0rrr? I couldn't quite gather this from your [second-to-last comment](https://github.com/aws/jsii/pull/747#issuecomment-525722160). 
Are you suggesting we force customers to use the same semver for all CDK packages?

rix0rrr:
> So what's the conclusion here? How do we want to model dependencies in the CDK, and do we need this change in JSII to accomplish that?

My best guess is the following, for intra-CDK references:

* The `dependency` refers to a point release, no range operator.
* The `peerDependency` *could* have a `^`, but I'm fairly sure it wouldn't buy us anything, so why bother? I vote no range operator there either.

jensbodal:
I feel like there's a slight misunderstanding of semver and the purpose of peer dependencies here and in https://github.com/aws/aws-cdk/issues/3711. 

Minor version bumps should **NOT** intentionally introduce breaking changes.  Minor version bumps should only intend to contain new backwards-compatible features, and patch version bumps should only intend to contain bugfixes. Major version bumps should only exist when non-backwards compatible changes are released, i.e. breaking changes.  

Peer dependencies should be pinned to an exact version if the peer dependency is another package that will always be released in cadence with the package declaring it as a peer dependency (e.g. new `@aws-cdk/aws-ec2` and `@aws-cdk/core` versions always being released simultaneously), and the intention that end users are always expected to update all related dependencies when updating a version.  Peer dependencies should generally otherwise be loosely pinned to minor versions (`^`) unless there's a known breaking change requiring it to be pinned to a loose patch (`~`) or exact version.

> However, that is too much onus on the user, so we populate every package's dependencies with a version that will auto-install the other packages, so they only have to specify the top of the dependency tree.

This is how you get to this problem though. The published `@aws-cdk/aws-ec2` package.json should not list `@aws-cdk/core` as a dependency, it should only be listed as a peer dependency with a fixed version (assuming they are released in cadence with each other). If `aws-ec2` or `core` updates can happen independently of each other, then `aws-ec2` should list `core` as a peer dependency with a loose minor version.  A minor version update in `@aws-cdk/core` should never intentionally break the same minor version of `@aws/aws-ec2` and there should be tests in place to ensure this.

The end result is if you require only one version of a package to be installed between related dependent packages and that package is part of the same group of packages, then that package should be moved exclusively to peer dependencies. If the runtime code depends on a specific version range of a library and it's irrelevant if there are multiple versions of that package installed, then it should be listed as a dependency.

---

A good example of this can be seen by installing the `@angular/cli` and starting a new project and looking at the resulting package.json files for the `@angular` packages.



**@angular/core@8.2.4**

```
  "dependencies": {
    "tslib": "^1.9.0"
  },
  "peerDependencies": {
    "rxjs": "^6.4.0",
    "zone.js": "~0.9.1"
  },
```

**@angular/router@8.2.4**

```
  "dependencies": {
    "tslib": "^1.9.0"
  },
  "peerDependencies": {
    "@angular/core": "8.2.4",
    "@angular/common": "8.2.4",
    "@angular/platform-browser": "8.2.4",
    "rxjs": "^6.4.0"
  },
```

**@angular/compiler-cli@8.2.4**

```
  "dependencies": {
    "reflect-metadata": "^0.1.2",
    "minimist": "^1.2.0",
    "canonical-path": "1.0.0",
    "chokidar": "^2.1.1",
    "convert-source-map": "^1.5.1",
    "dependency-graph": "^0.7.2",
    "magic-string": "^0.25.0",
    "source-map": "^0.6.1",
    "tslib": "^1.9.0",
    "yargs": "13.1.0"
  },
  "peerDependencies": {
    "@angular/compiler": "8.2.4",
    "typescript": ">=3.4 <3.6"
  },
```

rix0rrr:
> The published `@aws-cdk/aws-ec2` package.json should not list `@aws-cdk/core` as a dependency, it should only be listed as a peer dependency with a fixed version (assuming they are released in cadence with each other)

You are right. We're doing this 100% to improve the customer experience installing dependencies. An application declares their dependencies on the CDK packages they consume, and the rest automatically gets pulled in. If we didn't do this, the user would have to *know* the packages' dependency trees and satisfy them *for* the tool, by listing all transitively required packages at the top level. In practice, this would easily come down to 10+ packages included for the 1 or 2 a user would be using: not a good experience.

We've noticed that NPM does the required lifting of exact same dependencies, and because we list the dependency both in regular and in `peer` dependencies, NPM will at least give an error if it so happens that versions aren't deduplicated (and leading to an invalid closure). This is the best experience we can provide given the tools at the moment (since NPM dropped auto-installing peer dependencies as a feature).

rix0rrr:
Closing this for now. We don't seem to be getting any traction on it.

jensbodal:
This is being pedantic but npm won’t give in an error, it will give a warning. 

Forcing unnecessary dependencies on a user is not a good costumer experience.

If semantic versioning is being followed properly, there should be no issue with patch releases. Minor release should technically also not be an issue as they should only include backwards compatible changes. If a breaking change occurs, bump the version. I wouldn’t say this is easily enforced, but with proper build tools and unit tests it’s possible to heavily mitigate issues.

Regardless, pinning dependencies to a patch version is not a good customer experience if it is highly likely the consumer will have a different, compatible version installed but be forced to install a separate copy.

> We've noticed that NPM does the required lifting of exact same dependencies

While this is true, it is not limited to exact same dependencies. If you list `^1.2.3` as a dependency, and I have `1.9.9` explicitly I installed, I will just get one version of the package since `1.9.9` is a match for `1.2.3`.

skinny85:
> Closing this for now. We don't seem to be getting any traction on it.

@rix0rrr doesn't this add exactly what you complained you can't do in the commit message of https://github.com/aws/aws-cdk/commit/1d1b8bc22ae9f68f70cb00f1a69d02b5658a3832 ?

Issue #489, opened 2019-05-03
java: optimized setter for map entries (in builders)


fulghum:
From UX Feedback – customers would like a simpler interface for adding individual map entries to function prop objects.  Instead of having to construct the map object and then pass it in, customers are asking for a method like: myPropsObject.addConfiguration("key", "value").

eladb:
This is for builders I am assuming, correct?

fulghum:
Correct, only for the builders

PR #759, opened 2019-08-30
feat(java): enhanced handling of map and list collection types in bui…

bmaizels:
…lders

See issue for more context/background: https://github.com/aws/jsii/issues/489

Structs that have properties that generate into 'list' or 'map' types in Java will now have different features/characteristics in the generated Java builder for that struct:

1) When a list or map is passed into the builder setter, a copy of that collection will be stored in the builder instead of the builder taking a reference to it. This means that the builder will retain the original values at the time they were passed to the builder even if the collection is then modified before the struct is built.

2) An additional method will be added to the builder to allow single elements or entries to be added to the collection. For List it will be addTo[property-name](T value) and for map it will be putIn[property-name](String key, T value).

---

### DIFF
--- a/packages/jsii-java-runtime-test/project/src/test/java/software/amazon/jsii/testing/ComplianceTest.java
+++ b/packages/jsii-java-runtime-test/project/src/test/java/software/amazon/jsii/testing/ComplianceTest.java
@@ -52,6 +52,7 @@ import software.amazon.jsii.tests.calculator.ReferenceEnumFromScopedPackage;
 import software.amazon.jsii.tests.calculator.ReturnsPrivateImplementationOfInterface;
 import software.amazon.jsii.tests.calculator.StableStruct;
 import software.amazon.jsii.tests.calculator.Statics;
+import software.amazon.jsii.tests.calculator.StructWithCollections;
 import software.amazon.jsii.tests.calculator.StructWithJavaReservedWords;
 import software.amazon.jsii.tests.calculator.Sum;
 import software.amazon.jsii.tests.calculator.SyncVirtualMethods;
@@ -70,10 +71,15 @@ import software.amazon.jsii.tests.calculator.ConstructorPassesThisOut;
 import java.io.IOException;
 import java.time.Instant;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.contains;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.hasEntry;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotEquals;
@@ -1266,6 +1272,28 @@ public class ComplianceTest {
     public void canLoadEnumValues() {
         assertNotNull(EnumDispenser.randomStringLikeEnum());
         assertNotNull(EnumDispenser.randomIntegerLikeEnum());
+    }
+
+    @Test
+    public void listInStructCorrectlyBuilds() {
+        StructWithCollections structWithCollections = StructWithCollections.builder()
+            .array(Arrays.asList("one", "two"))
+            .addToArray("three")
+            .build();
+
+        assertThat(structWithCollections.getArray(), contains("one", "two", "three"));
+    }
+
+    @Test
+    public void mapInStructCorrectlyBuilds() {
+        StructWithCollections structWithCollections = StructWithCollections.builder()
+            .map(Collections.singletonMap("key1", "value1"))
+            .putInMap("key2", "value2")
+            .build();
+
+        assertThat(structWithCollections.getMap(), hasEntry("key1", "value1"));
+        assertThat(structWithCollections.getMap(), hasEntry("key2", "value2"));
+        assertThat(structWithCollections.getMap().size(), equalTo(2));
     }
 
     static class PartiallyInitializedThisConsumerImpl extends PartiallyInitializedThisConsumer {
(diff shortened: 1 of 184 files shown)


packages/jsii-java-runtime-test/project/src/test/java/software/amazon/jsii/testing/ComplianceTest.java
```diff
@@ -1268,6 +1274,27 @@ public void canLoadEnumValues() {
         assertNotNull(EnumDispenser.randomIntegerLikeEnum());
     }
 
+    public void listInStructCorrectlyBuilds() {
+        StructWithCollections structWithCollections = StructWithCollections.builder()
+            .array(Arrays.asList("one", "two"))
+            .addToArray("three")
```
eladb: What happens if a user has a property called `addToArray` in the same struct?

rix0rrr: Needs additional validation checks in jsii.

bmaizels:
> Please fix PR title and add information in the description that will provide background and rationale for this change.

The git issue is linked in the description (not very clearly, I'll make it a bit more obvious), and I tried to make the PR title as descriptive as possible. I'm not sure what else I can do there.

packages/jsii-java-runtime-test/project/src/test/java/software/amazon/jsii/testing/ComplianceTest.java
bmaizels: To answer Elad's question literally, they will get a compile error in the generated Java code. I'll look into adding a validation check in jsii as recommended.

eladb: What will the validation be? Not allow people to name a property `addToXxx`?

eladb:
@bmaizels the title is trimmed and doesn't reflect the actual user impact ("enhanced handling" is about the implementation and not about the user experience). The PR description should not just be a link to an issue but also provide pointers into what and where the change was done to support the new feature.

bmaizels:
> 
> 
> @bmaizels the title is trimmed and doesn't reflect the actual user impact ("enhanced handling" is about the implementation and not about the user experience). The PR description should not just be a link to an issue but also provide pointers into what and where the change was done to support the new feature.

Done.

packages/jsii-java-runtime-test/project/src/test/java/software/amazon/jsii/testing/ComplianceTest.java
bmaizels: ... or putInXXX. We considered mangling the java implementation if a conflict was detected, but having both methods side by side on the builder could be extremely confusing. Ideally we'd detect in JSII if the collision was actually going to happen by doing an exact property name match but it's going to be cheaper to do a global 'any property that starts with addTo or putIn' check.

bmaizels: Okay I've added in the JSII validation code. I also found and fixed an error with the pattern matching logic for error messages used in all the neg tests.

bmaizels:
> 
> 
> Looks good provided the additional checks on struct fields are added in jsii.

Added. Please take another look.

packages/jsii-java-runtime-test/project/src/test/java/software/amazon/jsii/testing/ComplianceTest.java
eladb: Feels messy that we are leaking the builder pattern naming conventions to all Jsii code. I can totally imagine someone wanting to define a struct that has a Boolean property “addToFoo”... feels wrong. Why did we get rid of the withXxx prefix. It will provide a more robust way to handle things like this and also forward looking as it will allow adding additional verbs in the future.

bmaizels: > Why did we get rid of the withXxx prefix

We were trying to emulate the style patterns currently used in the AWS Java v2 SDK and the popular libraries such as 'Immutables' that are strongly influencing the style of the larger Java community.

Alternatives we considered:
1) Allow conflicts in the model and then mangle the auto-generated method to not conflict on the java side. Eg: addToFoo(...) becomes addToFooArray(...) or something. The thing it falls back to could still theoretically conflict but we felt this was better than using a completely non-idiomatic form such as an underscore. We rejected this solution primarily on the grounds that seeing both methods side by side to each other in the builder may be very confusing.

2) Tune the conflict detection in JSII to only error in the specific case that an array is shadowing another property with the same name prefixed with addTo.

Having discussed this with the team we favor option 2 if the status-quo is unacceptable.

rix0rrr: > 1. Allow conflicts in the model and then mangle the auto-generated method to not conflict on the java side.

That unfortunately won't work because of library evolution.

Version 1 of the struct `{ foo: string[] }`, you'll generate an `addToFoo()` method in Java and people will use it.

Verstion 2 of the struct `{ foo: string[]; addToFoo: boolean; }`, the meaning of `addToFoo()` will have changed and the user's code will no longer compile.

It will have to be solution 2.

bmaizels: Yeah that's a good point. We'd have to mangle the 'new' thing to avoid backwards-compatibility issues but it wouldn't ever be clear what the new thing was, as it could either be the base attribute or the addTo attribute. Either way I think we're all in agreement that wasn't going to be the path forward.

packages/jsii/test/negatives/neg.property-name.4.ts
```diff
@@ -0,0 +1,5 @@
+///!MATCH_ERROR: Properties cannot have names like addToXxx() - those conflict with Java collection builder convenience methods
```
rix0rrr: Nice!

rix0rrr:
Unfortunately branch needs a conflict resolving merge from live :(

RomainMuller:
We decided against doing this due to the constraints it imposes on other languages in terms of what names must then be reserved. I'll close this without merging.

PR #762, opened 2019-09-03
feat(java): Indicate if method param is required

### LINKED ISSUE #365: java: when generating builders indicate for each property if it's required in the docs

dagnir:
This commit enhances javadoc generation for Java classes so that required
method/constructor parameters are indicated in their description in the javadoc.

This makes it easy for customers to see which paramaters must be provided when
interacting with libraries from Java.

Fixes #365

---

### DIFF
--- a/packages/jsii-pacmak/lib/targets/java.ts
+++ b/packages/jsii-pacmak/lib/targets/java.ts
@@ -896,11 +896,8 @@ class JavaGenerator extends Generator {
             this.code.line();
             this.code.line('/**');
             this.code.line(` * Sets the value of ${prop.propName}`);
-            if (prop.docs && prop.docs.summary) {
-                this.code.line(` * @param ${prop.fieldName} ${prop.docs.summary}`);
-            } else {
-                this.code.line(` * @param ${prop.fieldName} the value to be set`);
-            }
+            const summary = (prop.docs && prop.docs.summary) || "the value to be set";
+            this.code.line(` * ${paramJavadoc(prop.fieldName, prop.nullable, summary)}`);
             this.code.line(` * @return {@code this}`);
             if (prop.docs && prop.docs.deprecated) {
                 this.code.line(` * @deprecated ${prop.docs.deprecated}`);
@@ -1194,9 +1191,8 @@ class JavaGenerator extends Generator {
             const method = doc as spec.Method;
             if (method.parameters) {
                 for (const param of method.parameters) {
-                    if (param.docs && param.docs.summary) {
-                        tagLines.push(`@param ${param.name} ${param.docs.summary}`);
-                    }
+                    const summary = (param.docs && param.docs.summary) || undefined;
+                    tagLines.push(paramJavadoc(param.name, param.optional, summary));
                 }
             }
         }
@@ -1542,4 +1538,19 @@ function isNullable(optionalValue: spec.OptionalValue | undefined): boolean {
     return optionalValue.optional
         || (spec.isPrimitiveTypeReference(optionalValue.type)
             && optionalValue.type.primitive === spec.PrimitiveType.Any);
+}
+
+function paramJavadoc(name: string, optional?: boolean, summary?: string): string {
+    const parts = ['@param', name];
+    if (summary) { parts.push(endWithPeriod(summary)); }
+    if (!optional) { parts.push('This parameter is required.'); }
+
+    return parts.join(' ');
+}
+
+function endWithPeriod(s: string): string {
+    if (!s.endsWith('.')) {
+        return s + '.';
+    }
+    return s;
 }
--- a/packages/jsii-pacmak/test/expected.jsii-calc/java/src/main/java/software/amazon/jsii/tests/calculator/DeprecatedClass.java
+++ b/packages/jsii-pacmak/test/expected.jsii-calc/java/src/main/java/software/amazon/jsii/tests/calculator/DeprecatedClass.java
@@ -19,6 +19,8 @@ public class DeprecatedClass extends software.amazon.jsii.JsiiObject {
 
     /**
      * @deprecated this constructor is "just" okay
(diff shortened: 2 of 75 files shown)


packages/jsii-pacmak/test/expected.jsii-calc/java/src/main/java/software/amazon/jsii/tests/calculator/UsesInterfaceWithProperties.java
```diff
@@ -34,6 +36,8 @@ public UsesInterfaceWithProperties(final software.amazon.jsii.tests.calculator.I
 
     /**
      * EXPERIMENTAL
+     * 
+     * @param ext the value to be set. This parameter is required.
```
bmaizels: As discussed for method arguments we don't want to say 'the value to be set'.

bmaizels:
Looks like you removed 'the value to be set' from builder setters. I think we only wanted to remove them from method parameters and constructor args?

packages/jsii-pacmak/lib/targets/java.ts
```diff
@@ -1504,3 +1500,25 @@ function isNullable(optionalValue: spec.OptionalValue | undefined): boolean {
...
+        documentation = summary;
+    } else {
+        documentation = '';
+    }
+
+    if (documentation.length > 0) {
+        if (!documentation.endsWith('.')) {
+            documentation += '.';
+        }
+        documentation += " ";
```
rix0rrr: This space gets added even if the "required" suffix isn't added later on.

Feels like this method could be simplified to something like:


```ts
function paramJavadoc(name: string, optional?: boolean, summary?: string): string {
  const parts = ['@param', name];
  if (summary) { parts.push(endWithPeriod(summary)); }
  if (!optional) { parts.push('This parameter is required.'); }

  return parts.join(' ');
}
```

packages/jsii-pacmak/test/expected.jsii-calc/java/src/main/java/software/amazon/jsii/tests/calculator/UnionProperties.java
```diff
@@ -69,7 +69,7 @@ public Builder bar(software.amazon.jsii.tests.calculator.AllTypes bar) {
 
         /**
          * Sets the value of Foo
-         * @param foo the value to be set
+         * @param foo
```
rix0rrr: We are losing something here. Bit of a pity.

dagnir: This was unintended as @bmaizels. Will add back in!

eladb:
Please fix PR title and add rationale in description

packages/jsii-pacmak/lib/targets/java.ts
```diff
@@ -1504,3 +1500,18 @@ function isNullable(optionalValue: spec.OptionalValue | undefined): boolean {
...
+    const parts = ['@param', name];
+    if (summary) { parts.push(endWithPeriod(summary)); }
+    if (!optional) { parts.push('This parameter is required.'); }
+
+    return parts.join(' ');
+}
+
+function endWithPeriod(s: string): string {
+    if (!s.endsWith('.')) {
+        return s += '.';
```
rix0rrr: `return ... += `? Typo?

dagnir: Thanks!

dagnir:
@eladb Done

packages/jsii-pacmak/lib/targets/java.ts
```diff
@@ -1504,3 +1500,18 @@ function isNullable(optionalValue: spec.OptionalValue | undefined): boolean {
         || (spec.isPrimitiveTypeReference(optionalValue.type)
             && optionalValue.type.primitive === spec.PrimitiveType.Any);
 }
+
+function paramJavadoc(name: string, optional?: boolean, summary?: string): string {
+    const parts = ['@param', name];
+    if (summary) { parts.push(endWithPeriod(summary)); }
+    if (!optional) { parts.push('This parameter is required.'); }
```
eladb: Can we make this a bit more concise. Something like “(required)”

dagnir: I actually prefer "This parameter is required.". Yes, it's more verbose, but I think javadocs in libraries tend towards the verbose/descriptive side anyway, or at least do not try to be as short as possible. There is not much precedent for making javadocs more terse/concise that I can tell.

FWIW, others Java SDK team prefer ""This parameter is required." as well.

bmaizels:
Looks good. I've disabled auto-merging so we can get a core member of the CDK team to approve it too.

packages/jsii-pacmak/lib/targets/java.ts
eladb: Ok, cool.

packages/jsii-pacmak/test/expected.jsii-calc/java/src/main/java/software/amazon/jsii/tests/calculator/DeprecatedClass.java
```diff
@@ -19,6 +19,8 @@ protected DeprecatedClass(final software.amazon.jsii.JsiiObject.InitializationMo
 
     /**
      * @deprecated this constructor is "just" okay
+     * @param readonlyString This parameter is required.
+     * @param mutableNumber
```
eladb: shouldn't this say "the value to be set"?

dagnir: No, per previous comment from @bmaizels, we only want that description for params on builder setters, not constructors/class methods.

Issue #749, opened 2019-08-28
Invalid XML in documentation comments


RomainMuller:
### Affected Languages

- [ ] `TypeScript` or `Javascript`
- [ ] `Python`
- [ ] `Java`
- [X] .NET (`C#`, `F#`, ...)

### General Information
* **JSII Version:** `0.15.1`
* **Platform:**  `Darwin f01898256a91 18.7.0 Darwin Kernel Version 18.7.0: Tue Aug 20 16:57:14 PDT 2019; root:xnu-4903.271.2~2/RELEASE_X86_64 x86_64`

### What is the problem?
When creating documentation comments, XML-reserved characters are not correctly escaped, resulting in the literal copying of characters such as `<` and `>`, which make the documentation strings invalid for documentation website generators such as `docfx`.

Issue #714, opened 2019-08-16
.NET Generator csproj should target .NET Core and not .NET Standard


assyadh:
We don't want to support the .NET framework use cases.

.NET Generator csproj should target .NET Core and not .NET Standard

PR #772, opened 2019-09-11
fix(dotnet): fix deep type conversion across the process boundary, intelisense docs, set target to netcoreapp2.1

assyadh:
This PR fixes various issues due to how we convert collection elements, and how the type reference is passed when converting.

These two issues are similar and are due to nested Arrays of object not being casted properly:
Fixes https://github.com/aws/aws-cdk/issues/3244
Fixes https://github.com/aws/aws-cdk/issues/3672
Added a unit test to cover the case

This issue is due to Maps of Arrays of Anys not being casted properly:
Fixes https://github.com/aws/aws-cdk/issues/3813 
Added a unit test to cover the case

Also added support for xml documentation which is now added to the NuGet packages and should allow intellisense to provide in IDE help:
Fixes https://github.com/aws/jsii/issues/749
Fixes https://github.com/aws/aws-cdk/issues/1846

Migrated the target framework to netcoreapp2.1 instead of netstandard2.0:
Fixes https://github.com/aws/jsii/issues/714

---


packages/jsii-dotnet-runtime/src/Amazon.JSII.Runtime/Services/Converters/FrameworkToJsiiConverter.cs
```diff
@@ -285,6 +277,55 @@ protected override bool TryConvertString(object value, out object result)
...
+        /// </summary>
+        /// <param name="element">The element to convert in the collection</param>
+        /// <param name="referenceMap">The known references map</param>
+        /// <param name="elementType">The TypeReference of the element, as seen by Jsii</param>
+        /// <param name="convertedElement">out: the converted element</param>
+        /// <returns>True if the conversion was successful, false otherwise</returns>
+        private bool TryConvertCollectionElement(object element, IReferenceMap referenceMap, TypeReference elementType,
+            out object convertedElement)
+        {
+            bool converted;
```
costleya: why don't we just return directly instead of having temporary variables and a break?

assyadh: Done

assyadh:
This can be squashed and merge, double approved.

PR #795, opened 2019-09-18
feat: configure `engines` with `node >= 10.3.0`

RomainMuller:
As `node 8` is becoming EOL soon, we're upgrading the baseline `node`
version requirement from `>= 8.10.0` to `>= 10.3.0`. At this point, the
code generation is not affected, but this our first signal to consumers
that they need to upgrade their runtimes and/or codebases.

Additionally, normalized the order of fields in the `package.json` files
to make it nicer for people to work in the mono-repo.

Related to #794

---


RomainMuller:
It's possible the minimal version will need to be something higher (I think `10.3.0`) in order to ensure better coverage of ES2018.

PR #804, opened 2019-09-20
feat(java): offer short-hand list setter in builders

RomainMuller:
Leverage Java variadic methods to provide a list setter for builders
that converts the variadic parameter to a list (using `Arrays.asList`)
except if that is ambiguous (`List<Object>`).

---

### DIFF
--- a/packages/jsii-pacmak/lib/targets/java.ts
+++ b/packages/jsii-pacmak/lib/targets/java.ts
@@ -163,13 +163,13 @@ interface JavaProp {
   fieldName: string;
 
   // The java type for the property (eg: 'List<String>')
-  fieldJavaType: string;
+  fieldJavaType: JavaType;
 
   // The raw class type of the property that can be used for marshalling (eg: 'List.class')
   fieldJavaClass: string;
 
   // List of types that the property is assignable from. Used to overload setters.
-  javaTypes: string[];
+  javaTypes: JavaType[];
 
   // True if the property is optional.
   nullable: boolean;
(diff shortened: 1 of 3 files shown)


packages/jsii-pacmak/lib/targets/java.ts
```diff
@@ -869,30 +869,47 @@ class JavaGenerator extends Generator {
...
       }
       this.code.line(' */');
       this.emitStabilityAnnotations(prop.spec);
-      this.code.openBlock(`public ${builderName} ${prop.fieldName}(${type} ${prop.fieldName})`);
+      this.code.openBlock(`public ${builderName} ${prop.fieldName}(${type.syntax} ${prop.fieldName})`);
       this.code.line(`this.${prop.fieldName} = ${prop.fieldName};`);
       this.code.line('return this;');
       this.code.closeBlock();
+
+      if (type.listOf && type.listOf.syntax !== 'java.lang.Object') {
```
rix0rrr: This is just to catch types into which `java.util.List` could be assigned, right? I would make that a function to be super clear about that:

```
if (type.listOf && !listTypeAssignableTo(type.listOf)) { 
  ...
}
```

But is this even a good idea? Don't you think it would be confusing that you can call:

```java
builder.someStrings("a", "b", "c");
```

But will not be able to call:

```java
builder.someAnything(a, b, c); // Oops won't work! Why not? I've been trained that I can do that!
```

We do have a number of cases in the CDK where we use `any`...

PR #825, opened 2019-09-26
feat(kernel): annotate implemented interfaces on "ObjRef"s

### LINKED ISSUE #818: Allow "create" to specify additional interfaces
The kernel's `create` (and the private `objects.register`) call only requires the caller to specify a single `fqn` (that of the base type, which can be `Object` if nothing better is available). Unfortunately, this means that instances that implement one or more interface not inherited from a Javascript base class, the type information for interface members is lost when crossing the language boundary. This in turns causes all method arguments and return types to be inferred as `any`, further preventing type information from being available. This has been found to cause issues such as #807.

In order to address this, the kernel needs to allow the caller to also specify a list of implemented interfaces, so that it is able to access type information for members that are override in the foreign language. This is also useful when a client sub-classes a `jsii` type and adds new interface implementations to it, since that allows the type information for those to become available, too.

RomainMuller:
Annotating the interfaces implemented by an object instance with the
outbound Object Reference, and allowing the kernel's "create" call to
receive a list of interfaces allows the kernel to use more run-time type
information and operate safer.

Fixes #818

---

### DIFF
--- a/packages/jsii-java-runtime/project/src/main/java/software/amazon/jsii/JsiiEngine.java
+++ b/packages/jsii-java-runtime/project/src/main/java/software/amazon/jsii/JsiiEngine.java
@@ -198,6 +198,9 @@ public final class JsiiEngine implements JsiiCallbackHandler {
      * @return The Java class name.
      */
     private Class<?> resolveJavaClass(final String fqn) throws ClassNotFoundException {
+        if ("Object".equals(fqn)) {
+            return JsiiObject.class;
+        }
         String[] parts = fqn.split("\\.");
         if (parts.length < 2) {
             throw new JsiiException("Malformed FQN: " + fqn);
--- a/packages/jsii-kernel/lib/api.ts
+++ b/packages/jsii-kernel/lib/api.ts
@@ -1,9 +1,14 @@
 export const TOKEN_REF = '$jsii.byref';
+export const TOKEN_INTERFACES = '$jsii.interfaces';
 export const TOKEN_DATE = '$jsii.date';
 export const TOKEN_ENUM = '$jsii.enum';
 
 export interface ObjRef {
   [TOKEN_REF]: string;
+}
+
+export interface AnnotatedObjRef extends ObjRef {
+  [TOKEN_INTERFACES]?: string[];
 }
 
 export function isObjRef(value: any): value is ObjRef {
@@ -75,13 +80,29 @@ export interface LoadResponse {
 }
 
 export interface CreateRequest {
+  /**
+   * The FQN of the class of which an instance is requested (or "Object")
+   */
   fqn: string;
+  /**
+   * The FQNs of interfaces the instance implements, if any. Declaring
+   * interfaces that the class denoted by `fqn` implements is not necessary.
+   * This means that memebers of interfaces found in this property should
+   * declare members that are found in the `overrides` property.
+   */
+  interfaces?: string[];
+  /**
+   * Arguments to pass to the constructor of `fqn`. ("Object" accepts none)
+   */
   args?: any[];
+  /**
+   * Declarations of method overrides that should trigger callbacks
+   */
   overrides?: Override[];
 }
 
 /* eslint-disable @typescript-eslint/no-empty-interface */
-export interface CreateResponse extends ObjRef {}
+export interface CreateResponse extends AnnotatedObjRef {}
 /* eslint-enable @typescript-eslint/no-empty-interface */
 
 export interface DelRequest {
--- a/packages/jsii-kernel/lib/kernel.ts
+++ b/packages/jsii-kernel/lib/kernel.ts
@@ -17,7 +17,7 @@ export class Kernel {
   public traceEnabled = false;
 
   private assemblies: { [name: string]: Assembly } = { };
-  private readonly objects = new ObjectTable();
+  private readonly objects = new ObjectTable(this._typeInfoForFqn.bind(this));
   private cbs: { [cbid: string]: Callback } = { };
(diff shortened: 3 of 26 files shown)


packages/jsii-java-runtime/project/src/main/java/software/amazon/jsii/JsiiEngine.java
```diff
@@ -198,6 +198,9 @@ public Object getObject(final JsonNode objRefNode) {
      * @return The Java class name.
      */
     private Class<?> resolveJavaClass(final String fqn) throws ClassNotFoundException {
+        if ("Object".equals(fqn)) {
```
rix0rrr: This might be mentioned separately in the commit/merge message. Seems like a different fix.

packages/jsii-kernel/lib/api.ts
```diff
@@ -75,13 +80,29 @@ export interface LoadResponse {
 }
 
 export interface CreateRequest {
+  /**
+   * The FQN of the class of which an instance is requested (or "Object")
+   */
   fqn: string;
+  /**
+   * The FQNs of interfaces the instance implements, if any. Declaring
```
rix0rrr: > Declaring interfaces that the class denoted by `fqn` implements is not necessary.

Maybe describe these as "additional" interfaces then? In addition to the interfaces implemented by the class.

packages/jsii-kernel/lib/kernel.ts
```diff
@@ -756,83 +746,90 @@ export class Kernel {
...
+        }
       }
     }
 
     return undefined;
   }
 
-  private _typeInfoForProperty(fqn: string, property: string): spec.Property {
-    const typeInfo = this._tryTypeInfoForProperty(fqn, property);
+  private _typeInfoForProperty(property: string, fqn: string, interfaces?: string[]): spec.Property {
```
rix0rrr: In effect, shouldn't the last 2 arguments just be a list of FQNs, which can be either class or interface types?

packages/jsii-kernel/lib/objects.ts
```diff
@@ -79,26 +95,37 @@ export class ObjectTable {
...
     }
 
     const existingRef = objectReference(obj);
     if (existingRef) {
+      if (interfaces) {
+        const allIfaces = new Set(interfaces);
+        for (const iface of existingRef[api.TOKEN_INTERFACES] || []) {
+          allIfaces.add(iface);
+        }
+        existingRef[api.TOKEN_INTERFACES] = this.removeRedundant(Array.from(allIfaces), fqn);
```
rix0rrr: What if the same anonymous object gets returned under 2 different interfaces? This code path will not be hit, right? Does that interface get 'added' to the list of interfaces for this object?

rix0rrr: On the other hand, `removeRedundant` feels fairly expensive, I wouldn't want to do that every time an object crosses the jsii boundary.

packages/jsii-kernel/lib/api.ts
RomainMuller: Yeah that's how I have started to name them. I'll have to go on the comments & make sure it's homogeneously used.

packages/jsii-kernel/lib/kernel.ts
RomainMuller: There can only be ONE class there, and I feel like it's easier to represent that information if it's a separate argument. I probably should rename it though.

packages/jsii-kernel/lib/objects.ts
RomainMuller: This is still a WIP, but yeah, I have a local commit that also adds new interfaces (& removes redundant) when we return an object through a different "persona".

But I also worry about the intrinsic cost of doing this, so it might warrant a slightly deeper refactor to improve there...

RomainMuller:
@rix0rrr  - annotating from the active side is needed as well. This is still a draft PR as I'm working through the front-ends now... I wanted to have it out there at least so I don't lose anything if my hard disk crashes 😅

packages/jsii-kernel/lib/api.ts
```diff
@@ -75,13 +89,29 @@ export interface LoadResponse {
...
+   * declare members that are found in the `overrides` property.
+   */
+  interfaces?: string[];
+  /**
+   * Arguments to pass to the constructor of `fqn`. ("Object" accepts none)
+   */
   args?: any[];
+  /**
+   * Declarations of method overrides that should trigger callbacks
+   */
```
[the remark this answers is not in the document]
RomainMuller: Yeah - I also prefer with a blank line but I got lazy on this file 😆

packages/jsii-kernel/lib/objects.ts
```diff
@@ -132,9 +159,71 @@ export class ObjectTable {
...
+    if (ti.interfaces) {
+      for (const iface of ti.interfaces) {
+        if (this.interfaces.has(iface)) {
+          continue;
+        }
+        this.interfaces.add(iface);
+        this.addFromInterface(iface);
+      }
+    }
+
```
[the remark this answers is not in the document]
RomainMuller: This one was a merge artifact that I forgot to go fix...

RomainMuller:
@bmaizels - yup, the code was already mostly covered, and the tests that broke were un-broken by the updated runtime code. I'll still have a second pass over that to make sure the particular edge case this aims to solve is... indeed covered.

RomainMuller:
There was indeed a bug hiding in there (both in Java & .NET) due to insufficient test coverage of the feature. This is now fixed (including test)

PR #827, opened 2019-09-27
feat(sampiler): initial version of sample transpiler

rix0rrr:
This is the initial version of a tool that will parse the TypeScript
samples and emit equivalent sample code in Python.

For now, its use is limited:

- only supports Python
- doesn't support properly integrating/compiling with a multi-package
  build yet.

However, it will be an initial version we can demo and iterate on.

For a slightly outdated example of generated code, see: https://gist.github.com/rix0rrr/7c4a651f8eef1e94b006059209e805dd

---


packages/jsii-sampiler/bin/jsii-snippet
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env node
+require('./jsii-snippet.js');
```
eladb: Rename to jsii-samplier I guess

packages/jsii-sampiler/README.md
```diff
@@ -0,0 +1,147 @@
+# jsii-sampiler: a transpiler for code samples
+
+Utility to transcribe example code snippets from TypeScript to other
+jsii languages.
+
+Has knowledge about jsii language translation conventions to do the
+translations. Only supports a limited set of TypeScript language features.
+
```
eladb: Describe the usage of the tool

packages/jsii-sampiler/test/python/expressions.test.ts
```diff
@@ -0,0 +1,61 @@
...
+  'console.log(`Hello, ${x}, it works ${y}!`);',
+  ].join('\n'), `
+  x = "world"
+  y = "well"
+  print(f"Hello, {x}, it works {y}!")
+  `);
+});
+
+test('non-null expression', () => {
+  expectPython([
```
eladb: Maybe prepare for the next language: ```expect(input, { python: output })```

packages/jsii-sampiler/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "jsii-sampiler",
+  "version": "0.16.0",
+  "description": "Translate TypeScript code snippets to other languages",
+  "main": "lib/index.js",
+  "bin": {
```
eladb: Fix

packages/jsii-sampiler/test/python/expressions.test.ts
rix0rrr: Yep, totally planning to do that. It's the only way to cover all features across all languages.

PR #838, opened 2019-10-01
feat(jsii): enable strictPropertyInitialization checks

rix0rrr:
This will prevent forgotten initialization errors such as the root cause
of aws/aws-cdk#4286.

---


RomainMuller:
Strongly suggest checking whether this *currently* breaks the `aws-cdk` build. But I guess I can also get to it when I make the release/upgrade.

rix0rrr:
Oh it does, but I'm fixing it :)

PR #849, opened 2019-10-03
feat(pacmak): build all java targets at once

rix0rrr:
Improve build speed by building all jsii targets at the same time,
and using a special aggregate POM build for Java.

- Improves speed by not unnecessarily re-reading jsii assemblies
  for each individual build.
- Improves speed by using an aggregate POM to build all Java
  targets (can reuse JVM instances and compilation metadata by Maven).
- Improves wall-clock time by building all languages in parallel.

On my machine, reduces the time to pack the entire CDK repository
from 25min user/15min wall clock to 15min user/5 min wall clock.

Fixes # 

---


rix0rrr:
(Obviously this will not compile with the current jsii version)

packages/jsii-pacmak/bin/jsii-pacmak.ts
```diff
@@ -78,192 +87,153 @@ import { VERSION_DESC } from '../lib/version';
...
+  }
 
-    // outdir is either by package.json/jsii.outdir (relative to package root) or via command line (relative to cwd)
-    const outDir = argv.outdir !== undefined ? path.resolve(process.cwd(), argv.outdir) : path.resolve(packageDir, pkg.jsii.outdir);
-    const targets = argv.targets || [...Object.keys(pkg.jsii.targets), 'js']; // "js" is an implicit target.
+  await timers.recordAsync('npm pack', () => {
+    logging.info('Packaging NPM bundles');
+    return awaitAll(modulesToPackage
+      .map(m => () => m.npmPack()));
+  });
```
RomainMuller: Would we not extract the logging bit & make this a little less syntactically heavy?

packages/jsii-pacmak/bin/jsii-pacmak.ts
```diff
@@ -78,192 +87,153 @@ import { VERSION_DESC } from '../lib/version';
...
+  modules: JsiiModule[];
 }
 
-async function updateNpmIgnore(packageDir: string, excludeOutdir: string | undefined) {
-  const npmIgnorePath = path.join(packageDir, '.npmignore');
-  let lines = new Array<string>();
-  let modified = false;
-  if (await fs.pathExists(npmIgnorePath)) {
-    lines = (await fs.readFile(npmIgnorePath)).toString().split('\n');
+function sliceTargets(modules: JsiiModule[], requestedTargets: string[] | undefined, force: boolean) {
```
RomainMuller: I think this should make sure all available modules support all requested targets (or only return targets supported by all). If we brought in additional modules for dependencies, and some dependencies don't support a target, then the build is doomed to fail anyway...

packages/jsii-pacmak/bin/jsii-pacmak.ts
```diff
@@ -78,192 +87,153 @@ import { VERSION_DESC } from '../lib/version';
...
 
-      lines.push(pattern);
-    }
+function describePackages(target: TargetSet) {
+  if (target.modules.length > 0 && target.modules.length < 5) {
+    return target.modules.map(m => m.name).join(', ');
   }
-}
+  return `${target.modules.length} modules`;
+}
```
RomainMuller: Cute

packages/jsii-pacmak/lib/builder.ts
```diff
@@ -0,0 +1,123 @@
...
+ * Interface for classes that can build language targets
+ *
+ * Building can happen one target at a time, or multiple targets at a time.
+ */
+export interface TargetBuilder {
+  buildModules(modules: JsiiModule[], options: BuildOptions): Promise<void>;
+}
+
+/**
+ * Return the output directory if all modules have the same directory
```
RomainMuller: That comment isn't an accurate description of what the method does.

packages/jsii-pacmak/lib/npm-modules.ts
```diff
@@ -0,0 +1,115 @@
...
+ * Find all modules that need to be packagerd
+ *
+ * If the input list is empty, include the current directory. The result
+ * is NOT topologically sorted.
+ */
+export async function findJsiiModules(directories: string[], recurse: boolean) {
+  const ret: JsiiModule[] = [];
+  const visited = new Set<string>();
+  for (const dir of directories.length > 0 ? directories : ['.']) {
+    await visitPackage(dir, true);
```
RomainMuller: Would be mighty sweet if we could avoid `await` in `for` loops.

packages/jsii-pacmak/lib/npm-modules.ts
```diff
@@ -0,0 +1,115 @@
...
+  }
+  return ret;
+
+  async function visitPackage(dir: string, isRoot: boolean) {
+    const realPath = await fs.realpath(dir);
+    if (visited.has(realPath)) { return; } // Already visited
+    visited.add(realPath);
+
+    const pkg = await fs.readJson(path.join(realPath, 'package.json'));
```
RomainMuller: This could be `require`'d, which could then not require an `await.

packages/jsii-pacmak/lib/npm-modules.ts
```diff
@@ -0,0 +1,115 @@
...
+}
+
+export async function updateAllNpmIgnores(packages: JsiiModule[]) {
+  for (const pkg of packages) {
+    // updates .npmignore to exclude the output directory and include the .jsii file
+    await updateNpmIgnore(pkg.moduleDirectory, pkg.outputDirectory);
+  }
+}
+
+async function updateNpmIgnore(packageDir: string, excludeOutdir: string | undefined) {
```
RomainMuller: Seriously, this function is very long for the value it produces. I would be inclined to just stopping doing that.

packages/jsii-pacmak/lib/packaging.ts
```diff
@@ -0,0 +1,88 @@
...
+   * Identifier of the targets to build
+   */
+  availableTargets: string[];
+
+  /**
+   * Output directory where to package everything
+  */
+  defaultOutputDirectory: string;
+}
+export class JsiiModule {
```
RomainMuller: Missing blank line

packages/jsii-pacmak/lib/packaging.ts
```diff
@@ -0,0 +1,88 @@
...
+    return this._tarball.object;
+  }
+
+  public async load() {
+    this._assembly = await SHARED_TS.loadModule(this.moduleDirectory);
+  }
+
+  public get assembly(): reflect.Assembly {
+    if (!this._assembly) {
+      throw new Error('Assembly not available yet, call load() first');
```
RomainMuller: Couldn't this call `load()` instead of barfing? This class has so much temporal coupling in it's features 😕

packages/jsii-pacmak/lib/targets/index.ts
```diff
@@ -0,0 +1,19 @@
...
+
+import Dotnet from './dotnet';
+import { JavaBuilder } from './java';
+import JavaScript from './js';
+import Python from './python';
+import Ruby from './ruby';
+
+export type TargetName = 'dotnet' | 'java' | 'js' | 'python' | 'ruby';
+
+export const ALL_BUILDERS: {[key in TargetName]: TargetBuilder} = {
```
RomainMuller: Spaces within the curly braces, please?

RomainMuller: How about `{ readonly [key in TargetName]: TargetBuilder }`?

packages/jsii-pacmak/lib/targets/java.ts
```diff
@@ -9,13 +9,132 @@ import { Generator } from '../generator';
...
+    logging.debug(`Generating ${this.targetName} code into ${srcDir.directory}`);
+    await target.generateCode(srcDir.directory, module.tarball);
+
+    return srcDir;
+  }
+
+  private async generateAggregatePom(sourceDirectories: Array<Scratch<void>>) {
+    const parentDir = this.findSharedParentDirectory(sourceDirectories.map(s => s.directory));
+
+    const aggregatePom = xmlbuilder.create({
```
RomainMuller: Would be interesting if this mode could also produce a Maven `BOM` file!

packages/jsii-pacmak/bin/jsii-pacmak.ts
```diff
@@ -1,19 +1,17 @@
...
 import logging = require('../lib/logging');
-import { Target } from '../lib/target';
 import { Timers } from '../lib/timer';
-import { resolveDependencyDirectory, shell } from '../lib/util';
 import { VERSION_DESC } from '../lib/version';
+import { findJsiiModules, updateAllNpmIgnores } from '../lib/npm-modules';
+import { JsiiModule } from '../lib/packaging';
+import { Thunk, ConcurrencyLimiter } from '../lib/util';
+import { ALL_BUILDERS, TargetName } from '../lib/targets';
+import pLimit from 'p-limit';
```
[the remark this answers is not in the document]
rix0rrr: 100 jobs in parallel? I don't think so.

packages/jsii-pacmak/bin/jsii-pacmak.ts
```diff
@@ -22,9 +20,15 @@ import { VERSION_DESC } from '../lib/version';
       type: 'array',
       desc: 'target languages for which to generate bindings',
       defaultDescription: 'all targets defined in `package.json` will be generated',
-      choices: Object.keys(targetConstructors),
+      choices: Object.keys(ALL_BUILDERS),
```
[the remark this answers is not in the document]
rix0rrr: I understand, but I don't think it adds value right now. So I scrapped it under the notion of removing complexity we didn't need.

packages/jsii-pacmak/bin/jsii-pacmak.ts
```diff
@@ -54,6 +58,11 @@ import { VERSION_DESC } from '../lib/version';
       desc: 'force generation into a target-named subdirectory, even in single-target mode',
       default: true,
     })
+    .option('force-target', {
+      type: 'boolean',
+      desc: 'force generation of the given targets, even if the source package.json doesnt declare it',
```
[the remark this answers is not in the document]
rix0rrr: Hah. Well the Ruby runtime depended on it (wanted to generate ruby code for jsii-calc and friends, even though it wasn't declared and is otherwise untested). My first instinct was to get rid of the ruby runtime, but since I already overstepped my bounds a bit by removing the Sphinx target (ooohhh) I didn't want to make the scope of this PR even bigger and risk incurring more discussion.

I also did not want to enshrine the current Ruby generation into jsii-pacmak's `expected` tests, so this seemed like the smallest delta to keep the repository building.

packages/jsii-pacmak/bin/jsii-pacmak.ts
```diff
@@ -22,9 +20,15 @@ import { VERSION_DESC } from '../lib/version';
...
       desc: 'target languages for which to generate bindings',
       defaultDescription: 'all targets defined in `package.json` will be generated',
-      choices: Object.keys(targetConstructors),
+      choices: Object.keys(ALL_BUILDERS),
       required: false
     })
+    .option('concurrency', {
+      alias: 'C',
+      type: 'boolean',
+      desc: 'whether to work in parallel when possible',
```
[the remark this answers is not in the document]
rix0rrr: Your CPU might not like it.

Just, control to the user.

packages/jsii-pacmak/bin/jsii-pacmak.ts
```diff
@@ -78,192 +87,153 @@ import { VERSION_DESC } from '../lib/version';
       default: true
     })
     .version(VERSION_DESC)
+    .strict()
```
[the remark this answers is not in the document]
rix0rrr: Fucked myself over without this, too. Don't understand why it isn't the default.

packages/jsii-pacmak/bin/jsii-pacmak.ts
```diff
@@ -78,192 +87,153 @@ import { VERSION_DESC } from '../lib/version';
...
-        const depDir = resolveDependencyDirectory(packageDir, dep);
-        /* eslint-disable no-await-in-loop */
-        await buildPackage(depDir, /* isRoot */ false, forceSubdirectory);
-        /* eslint-enable no-await-in-loop */
-      }
-    }
+  } else if (argv.npmignore) {
+    // if outdir is coming from package.json, verify it is excluded by .npmignore. if it is explicitly
+    // defined via --out, don't perform this verification.
+    await updateAllNpmIgnores(modulesToPackage);
```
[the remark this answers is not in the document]
rix0rrr: Not sure what you're asking for. This behavior was already in there, I'm just keeping it (but moving it to a more logical place).

packages/jsii-pacmak/bin/jsii-pacmak.ts
```diff
@@ -78,192 +87,153 @@ import { VERSION_DESC } from '../lib/version';
...
+    return awaitAll(modulesToPackage
+      .map(m => () => m.load()));
+  });
 
-    if (argv.npmignore) {
-      // if outdir is coming from package.json, verify it is excluded by .npmignore. if it is explicitly
-      // defined via --out, don't perform this verification.
-      const npmIgnoreExclude = argv.outdir ? undefined : outDir;
+  try {
+    const requestedTargets = argv.targets && argv.targets.map(t => `${t}`)
```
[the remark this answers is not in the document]
rix0rrr: `Array<number | string>`. `@types/yargs` is a bit out of sync with how `yargs` actually behaves. Didn't feel like spending too much time on this one.

packages/jsii-pacmak/bin/jsii-pacmak.ts
```diff
@@ -78,192 +87,153 @@ import { VERSION_DESC } from '../lib/version';
...
-
-    const tmpdir = await fs.mkdtemp(path.join(os.tmpdir(), 'npm-pack'));
-    try {
-      const tarball = await timers.recordAsync('npm pack', () => {
-        return npmPack(packageDir, tmpdir);
-      });
-
-      const ts = new reflect.TypeSystem();
-      const assembly = await ts.loadModule(packageDir);
+    const perLanguageDirectory = targetSets.length > 1 || argv['force-subdirectory'];
```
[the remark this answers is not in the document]
rix0rrr: Maybe. Some runtime tests depend on that not being the case, and I don't want to dive into the labyrinth of bash build scripts for all the runtimes.

packages/jsii-pacmak/bin/jsii-pacmak.ts
rix0rrr: Could be, although to do that properly we would have to record the dependencies (not all packages are necessarily dependencies of each other).

So unrelated packages A and B could look like:

```
A: java+python
B: python
```

And I might want to build all targets. Though unlikely, I did not see a reason to prohibit that kind of build. Since the goal of this is speed, independent builds would require analyzing A twice, which would take away some of the purpose.

packages/jsii-pacmak/lib/packaging.ts
rix0rrr: No because `load()` is async and this getter isn't.

packages/jsii-pacmak/lib/targets/index.ts
rix0rrr: Noooooooooooo :)

packages/jsii-pacmak/lib/targets/java.ts
rix0rrr: What's that?

rix0rrr:
I removed the parallelism restriction. Can you re-review?

packages/jsii-pacmak/bin/jsii-pacmak.ts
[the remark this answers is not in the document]
RomainMuller: Oh I see. I wouldn't have opposed dropping Ruby (having side concerns about he lost effort, but that's another discussion).

Regardless - maybe just mention in the description that `--force-target` might cause `jsii-pacmak` to exit in error.

PR #880, opened 2019-10-14
feat: update Dockerfile to .NET SDK 3.1, improve NuGet metadata

RomainMuller:
---


packages/jsii-dotnet-analyzers/NuGet.Metadata.props.t.js
```diff
@@ -2,17 +2,25 @@ const package = require('./package.json');
...
-    <GeneratePackageOnBuild>True</GeneratePackageOnBuild>
-    <IncludeSymbols>True</IncludeSymbols>
-    <SymbolPackageFormat>snupkg</SymbolPackageFormat>
-    <IncludeSource>True</IncludeSource>
-    <PackageOutputPath>..\\..\\bin\\$(Configuration)\\NuGet\\</PackageOutputPath>
     <PackageVersion>$(JsiiVersion)</PackageVersion>
     <Description>${package.description}</Description>
-    <ProjectUrl>${package.homepage}</ProjectUrl>
+    <PackageLicenseExpression>${package.license}</PackageLicenseExpression>
     <LicenseUrl>https://spdx.org/licenses/${package.license}.html</LicenseUrl>
```
costleya: LicenseUrl is deprecated. It can be removed now that we are using PackageLicenseExpression.

Same with the other files in this PR.

packages/jsii-pacmak/lib/targets/dotnet/filegenerator.ts
```diff
@@ -60,6 +60,9 @@ export class FileGenerator {
     propertyGroup.ele('PackageId', packageId);
     propertyGroup.ele('Description', this.getDescription());
     propertyGroup.ele('ProjectUrl', assembly.homepage);
+    propertyGroup.ele('RepositoryUrl', assembly.repository.url);
+    propertyGroup.ele('RepositoryType', assembly.repository.type);
+    propertyGroup.ele('PackageLicenseExpression', assembly.license);
     propertyGroup.ele('LicenseUrl', `https://spdx.org/licenses/${assembly.license}.html`);
```
costleya: Make sure you remove it here too.

superchain/Dockerfile
```diff
@@ -1,29 +1,42 @@
 FROM amazonlinux:2
 
+# Install deltarpm as it can speed up the upgrade processes
+RUN yum -y install deltarpm
+
 # Install .NET Core, mono & PowerShell
 ENV DOTNET_CLI_TELEMETRY_OPTOUT=true                                                                                    \
-    DOTNET_SKIP_FIRST_TIME_EXPERIENCE=true
```
costleya: Why are we removing this? Is this because people are doing repeated builds on the same container, so the long-term optimization cost is worth it?

superchain/Dockerfile
```diff
@@ -1,29 +1,42 @@
...
+    DOTNET_RUNNING_IN_CONTAINER=true                                                                                    \
+    DOTNET_USE_POLLING_FILE_WATCHER=true                                                                                \
```
costleya: Same here - is this a development container? It it is only meant to build in a pipeline, all the files are already there - you don't need to ping for source file updates.

packages/jsii-dotnet-analyzers/NuGet.Metadata.props.t.js
```diff
@@ -2,17 +2,25 @@ const package = require('./package.json');
...
+
+    <ProjectUrl>${package.homepage}</ProjectUrl>
+    <RepositoryUrl>${package.repository.url}</RepositoryUrl>
```
costleya: <Optional>

Since we are now adding this information, do we want to enable source link?
https://github.com/dotnet/sourcelink/blob/master/README.md

superchain/Dockerfile
RomainMuller: I have actually looked at how Microsoft built *their* `dotnet` images, and the way they go past the "fist time experience" is by invoking `dotnet help` (which I have added).

superchain/Dockerfile
RomainMuller: Like for `FIRST_EXPERIENCE`, Microsoft does it in their container and notes that the `POLLING_FILE_WATCHER` is the only option that is supported in Docker. I'm not certain which context file watchers are being used, so I figured I'd just use the same environment that they do :)

packages/jsii-dotnet-analyzers/NuGet.Metadata.props.t.js
RomainMuller: The problem is that `SourceLink` seems to require that the `git` repository is present (or at least certain files like `.git/HEAD` and `.git/config`, which isn't present on *CodePipeline* executions.

costleya: This is fine for now, just wanted to raise the possibility.

superchain/Dockerfile
costleya: Yeah, I noticed where you probably got these from, and checked out the dockerfile. The difference is that that image is designed for development as well, why this may just be used for automation.

The performance impact is probably negligible - but just wanted to give some weight to what these options did, rather than simply adding/removing them.

superchain/Dockerfile
costleya: Yeah, I noticed where you probably got these from, and checked out the dockerfile. The difference is that that image is designed for development as well, why this may just be used for automation.

The performance impact is probably negligible - but just wanted to give some weight to what these options did, rather than simply adding/removing them.

packages/jsii-dotnet-analyzers/NuGet.Metadata.props.t.js
costleya: Agreed, just wanting to raise the possibility.

docs/configuration.md
```diff
@@ -246,7 +246,7 @@ the standard [`nuget push`][nuget-push] command.
 
 [NuGet]: https://www.nuget.org
 [nuget-push]: https://docs.microsoft.com/fr-fr/nuget/nuget-org/publish-a-package
-[.NET documentation]: https://docs.microsoft.com/en-us/dotnet/core/tools/csproj#packagelicenseurl
+[.NET documentation]: https://docs.microsoft.com/en-us/dotnet/core/tools/csproj#packageiconurl
```
costleya: This doesn't seem right. I think the #link was extraneous to begin with. Remove it?

[.NET documentation]: https://docs.microsoft.com/en-us/dotnet/core/tools/csproj

RomainMuller: Nah the link is in fact used a couple of lines above to point people at where Microsoft tells you what icons they suggest you use. It's not absolutely necessary, but it's actually kinda useful (as at least this can be expected to remain current on what Microsoft recommends)

RomainMuller:
Regarding the logo update - I reckon you're talking about the `jsii-calc` packages? In this case, those are currently intentionally left this way - there's some tricky things to do so we can support embedded icons for generated modules... And that deserves it's own separate work item.

PR #891, opened 2019-10-16
chore: migrate mono-repository to use "yarn workspaces"

RomainMuller:
This appears to improve the bootsrapping performance and makes the lock
files much easier to manager (`yarn` maintains a single `yarn.lock` file
at the root of the mono-repository instead of having one for each
package like `npm` does).

The `jsii-calc-bundled` package was moved out to an `@fixtures` subdirectory
as it is otherwise considered a "workspace package" and it's hoisting to the
root directory cannot be prevented. Since it is a bundled dependency of 
`jsii-calc`, hoisting needs to be disabled, otherwise `npm pack` (and
`yarn pack`) simply do not include it in the tarball.

As required, this adjusted the `jsii/superchain` Docker image to also
contain `yarn`, and changes how the `node` binary is installed to using
the `nodesource` repository (so as to satisfy the `yarn` RPM
dependencies without hacking around too much).

The `yarn` "engines" check has been disabled, as we are currently
building out of a `node 8.x` engine, as a way to ensure we don't
inadvertently cause breakage until we "finally" update our baseline
requirement.

---


.npmrc
```diff
@@ -0,0 +1,2 @@
+# Helps with using npm package-locally, while in a yarn workflow (avoids coupling scripts to yarn)
+scripts-prepend-node-path=auto
```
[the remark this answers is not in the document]
RomainMuller: This instructs node to prepend the "real"/"correct" node's standard library path to the load path... Instead of it using the default/system one. This removes the risk of your running a node engine version with a library that's from a different version.

`yarn` does "shim out" `node` to ensure that the _same_ runtime is used throughout the process... Which makes builds more reliable & repeatable.

.travis.yml
```diff
@@ -5,6 +5,10 @@ branches:
     - master
```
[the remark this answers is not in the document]
RomainMuller: Still needed to be able to publish the Docker image. I have to migrate this to CodeBuild, but I have more pressing things to attend to at this point... :(

.yarnrc
```diff
@@ -0,0 +1,2 @@
+--install.ignore-engines  true
```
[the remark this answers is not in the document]
RomainMuller: Good point (not sure the file supports comments, but I'll check!)

packages/jsii-dotnet-analyzers/package.json
```diff
@@ -30,9 +30,9 @@
     "package": "package-dotnet"
   },
   "dependencies": {
-    "jsii-dotnet-runtime": "file:../jsii-dotnet-runtime"
+    "jsii-dotnet-runtime": "^0.19.0"
```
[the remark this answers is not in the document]
RomainMuller: I could specify `*`, but packages will get published with that, too. For "private" dependencies, I could use that however, that shouldn't be a problem.

package.json
```diff
@@ -2,23 +2,30 @@
...
   },
   "devDependencies": {
     "@types/node": "^8.10.55",
-    "@typescript-eslint/eslint-plugin": "^2.3.3",
-    "@typescript-eslint/parser": "^2.3.3",
+    "@typescript-eslint/eslint-plugin": "^2.4.0",
+    "@typescript-eslint/parser": "^2.4.0",
     "eslint": "^6.5.1",
-    "lerna": "^3.17.0",
+    "lerna": "^3.18.1",
```
rix0rrr: Why do we still need lerna?

packages/jsii-dotnet-runtime/build.sh
```diff
@@ -4,7 +4,7 @@ set -euo pipefail
 # embed jsii-runtime as a resource
 bundle_dir="src/Amazon.JSII.Runtime/jsii-runtime"
 mkdir -p ${bundle_dir}
-rsync -av node_modules/jsii-runtime/webpack/ ${bundle_dir}
+rsync -av $(node -p 'require("path").dirname(require.resolve("jsii-runtime/package.json"))')/webpack/ ${bundle_dir}
```
rix0rrr: Don't understand why this doesn't have to be `-pe` ?

packages/jsii-ruby-runtime/package.json
```diff
@@ -32,10 +32,10 @@
     "package": "package-ruby"
   },
   "devDependencies": {
-    "jsii-build-tools": "file:../jsii-build-tools",
-    "jsii-calc": "file:../jsii-calc",
-    "jsii-pacmak": "^0.19.0",
-    "jsii-runtime": "^0.19.0",
+    "jsii-build-tools": "*",
```
rix0rrr: All the `*`s are somewhat scary to me. I understand why it doesn't matter here, but there are cases in which it DOES matter, and how will we ensure that everyone editing these files knows the difference AND doesn't make mistakes?

package.json
RomainMuller: Lerna provides high level primitives that are convenient and allow us to have topologically-sorted script execution, among other things. It is not entirely necessary, but makes a few things easier.

packages/jsii-dotnet-runtime/build.sh
RomainMuller: `-p` behaves like `-e`, so in fact it is not necessary to add the `-e`.

packages/jsii-ruby-runtime/package.json
RomainMuller: Well I felt people were generally in favor of not incurring so many changes on those, but happy to revert those if that scares you...

RomainMuller:
@rix0rrr  - the version upgrades are related to the introduction of `yarn` here. And the script changes are required for the project to build or test correctly with the way yarn installs dependencies (lots of code was making unsafe assumptions about the physical location of libraries).

PR #894, opened 2019-10-17
chore: upgrade to @types/node 10.x

### LINKED ISSUE #794: Upgrade node requirement to `>= 10`
As node 8 is going to be EOL soon (by the end of this year), we need to raise our baseline node requirement to be `>= 10`. This will also enable increasing our `@types/node` dependency as well as set our `tsc` target to `ES2018`.

- [x] Raise engine requirement in `package.json` files
- [x] Release new versions & notify customers (then wait about a month)
- [x] Upgrade `@types/node` dependency to the `^10.0.0` tree
- [x] Upgrade the `tsconfig.json` targets to `ES2018` (to generate more compact/efficient JS)

RomainMuller:
Actually upgrade the `node` engine support to 10.x tree, as has
been declared in the `package.json` files for a month now.

Additionally, standardized the `compilerOptions` in the `tsconfig.json` files
that are *not* generated by `jsii`, in order to use `ES2018` standard library,
enable `strict` mode (this includes `strictPropertyInitialization`, which was not
enabled previously and required a few code changes in `jsii-pacmak` that were
made ad-minima).

The library clause of  `tsconfig.json` files that are generated by `jsii` has not
been changed yet. This means you need to use `node >= 10.x` in order to use
`jsii`, but you cannot use it to build code that depend on APIs that are not yet
available in `node 8`. This behavior will be updated in a second phase.

Fixes #794

---


RomainMuller:
@nija-at so I have ran a full build of `aws-cdk` with this version of `jsii` and then proceeded to run all tests using the result in a Docker container running the `node:8` (containing `NODE_VERSION=8.16.2`) image (asserting that no code was making incompatible calls *yet*). This uncovered no issue.

PR #895, opened 2019-10-18
feat(java): offer Builders for certain Java classes

### LINKED ISSUE #488: java: Improve use of Builder pattern
UX Research Feedback – Customers have commented that the use of Builder objects for CDK function prop objects does not feel idiomatic.  They find passing in a builder as the third param for construct construction inconsistent with other Java APIs and the builder pattern.  

Customers expected to see the Builder pattern either used entirely for constructing constructs, or not at all in the constructor.

RomainMuller:
Provide a nested `Builder` class with a static `create` factory for all
classes that adhere to the following constraints:
- The class is not `abstract`
- The class has a non-protected `initializer` (aka `constructor`)
- The constructor has at least one `struct` (aka `datatype`) parameter

Such builders will request the *positional* parameters to be passed
directly to the `Builder#create` method and can (at least currently) not
be subsequently modified. All parameters from the `struct` are then set
direcly on the `Builder` instance. If there are multiple `struct` parameters,
all but the first one are handled like other positional parameters.

Fixes #488

---


packages/jsii-pacmak/test/expected.jsii-calc/java/src/main/java/software/amazon/jsii/tests/calculator/Calculator.java
```diff
@@ -18,14 +18,71 @@ protected Calculator(final software.amazon.jsii.JsiiObject.InitializationMode in
...
+            }
+            this.props.maximumValue(maximumValue);
+            return this;
+        }
+
+        /**
+         * @returns a new instance of {@link Calculator}
+         */
+        public Calculator build() {
+            return new software.amazon.jsii.tests.calculator.Calculator((this.props != null ? this.props.build() : null));
```
bmaizels: This code is kind of ugly, I would prefer multi-lining it.

packages/jsii-pacmak/test/expected.jsii-calc/java/src/main/java/software/amazon/jsii/tests/calculator/Calculator.java
```diff
@@ -18,14 +18,71 @@ protected Calculator(final software.amazon.jsii.JsiiObject.InitializationMode in
...
+        /**
+         * Defines the value of props.maximumValue
+         *
+         * @param maximumValue 
+         *
+         * @returns {@code this}
+         */
+        public Builder maximumValue(final java.lang.Number maximumValue) {
+            if (this.props == null) {
+                this.props = software.amazon.jsii.tests.calculator.CalculatorProps.builder();
```
bmaizels: If there are lots of nested properties it will look bad to duplicate this code over and over, maybe think about creating a private method to do this lazy initialization.

packages/jsii-pacmak/test/expected.jsii-calc/java/src/main/java/software/amazon/jsii/tests/calculator/Calculator.java
```diff
@@ -266,4 +208,54 @@ public void setUnionProperty(final software.amazon.jsii.tests.calculator.Multipl
...
+        }
+
+        /**
+         * @returns a newly built instance of {@link software.amazon.jsii.tests.calculator.Calculator}.
+         */
+        @software.amazon.jsii.Stability(software.amazon.jsii.Stability.Level.Experimental)
+        public software.amazon.jsii.tests.calculator.Calculator build() {
+            return new software.amazon.jsii.tests.calculator.Calculator(
+                this.props != null ? this.props.build() : null
+            );
```
bmaizels: This is more readable but not what I actually meant about multi-line. My apologies for not being explicit and clear. I think it should be split into a local variable assignment followed by passing that local variable in as an argument. This is more of a nit though so I wouldn't block the build on it, it's your call.

PR #900, opened 2019-10-22
feat(pacmak): put translated README into module docstring

rix0rrr:
We put in all this effort to convert code samples from TypeScript
to Python, but they only were written in translated form to a README
file that was used to publish to PyPI. When the packages are installed
to a Python library directory, the README is gone (and so cannot be used
by Sphinx to insert into the docs).

Our own Sphinx generator was reading the README from the jsii assembly,
but that README still contains the TypeScript examples.

Insert the README content as a module-level docstrings so that the doc
generator can read the right translated README version and put it in the
docs. We keep the module docstring in MarkDown, not convert it to ReST,
since our ReST conversion is not very reliable yet.

Fixes # 

---


package.json
```diff
@@ -27,5 +27,8 @@
     "nohoist": [
       "**/jsii-calc-bundled"
     ]
+  },
+  "dependencies": {
+    "yarn": "^1.19.1"
```
RomainMuller: I would rather not...

rix0rrr: This was actually added by accident, so I'm happy to take it out. But I do wonder: why not, actually? Isn't it a convenient way to bootstrap yarn by using NPM? And then we can keep it nice & isolated.

PR #906, opened 2019-10-24
fix(java,dotnet): emit default implementations for optional properties

### LINKED ISSUE #543: Java - problems implementing class/interface with optional property
When implementing an `abstract class` or `interface` that defines optional properties, but not actually specifying the property (valid in TypeScript), the Java code generator will never emit the concrete implementation of the `get`ter, resulting in an invalid class.

RomainMuller:
For optional properties on interfaces, emit default implementations (for .NET,
this required upgrading the target to `netcoreapp3.0`) so that adding a new
optional property on an interface does not break existing code that implements
the interface.

In it's current form, the getters always return `null`, while the setters throw an
exception unless overridden.

Fixes #543

---


packages/jsii-pacmak/lib/targets/dotnet/filegenerator.ts
```diff
@@ -50,7 +50,7 @@ export class FileGenerator {
     const rootNode = xmlbuilder.create('Project', { encoding: 'UTF-8', headless: true });
     rootNode.att('Sdk', 'Microsoft.NET.Sdk');
     const propertyGroup = rootNode.ele('PropertyGroup');
-    propertyGroup.ele('TargetFramework', 'netcoreapp2.1');
+    propertyGroup.ele('TargetFrameworks', 'netcoreapp2.1;netcoreapp3.0');
```
rix0rrr: Uh. Wut? The tag name changes?

packages/jsii-pacmak/test/diff-test.sh
```diff
@@ -29,11 +29,6 @@ function assert-generator() {
         rsync -a ${original_expected}/ ${expected}/
     fi
 
-    if [[ -d ${original_expected}/java/target ]]; then
```
rix0rrr: This helps me a lot. Please don't remove it?

packages/jsii-pacmak/lib/targets/dotnet/filegenerator.ts
RomainMuller: Yeah - there's a tag for singular target, and one for multi-targeting. Thank you (not) msbuild.

packages/jsii-pacmak/test/diff-test.sh
RomainMuller: What problem does it solve? I had no problem after removing this... I'm curious what I'm missing!

It did however cause me **MUCH** pain, because my IDE just auto-touches stuff, and that made my iteration "edit code - close IDE - clean up files - test - open IDE - edit code - ...". This is painful.

packages/jsii-pacmak/lib/targets/dotnet/dotnetgenerator.ts
```diff
@@ -183,10 +183,27 @@ export class DotNetGenerator extends Generator {
...
 
+    if (prop.optional) {
+      // Conditional compilation - default interface implementation requires C#8.0 support, which imples netcoreapp3.0 or above!
+      this.code.line('#if NETCOREAPP3_0');
+      this.code.openBlock('get');
+      this.code.line('return null;');
+      this.code.closeBlock();
+      if (!prop.immutable) {
+        this.code.openBlock('set');
+        this.code.line(`throw new System.NotSupportedException("'set' for '${propName}' is not implemented");`);
```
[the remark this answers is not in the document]
RomainMuller: The issue here is that interfaces cannot declare properties, and if I want to "store" values here, I have to have the default implementation use some external storage (maybe a static data store class somewhere). This is one thing I had considered by it feels fairly ugly...

What do you think about the addition of an `@Optional` annotation to denote those methods that you need to "opt into" manually?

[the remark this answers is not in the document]
rix0rrr: Well I was going to suggest that, but turns out this behavior actually correctly mimics TypeScript semantics. 

It was just surprising to me, so I suggest writing down somewhere that the current behavior is not only due to lack of other options, but even correct.

packages/jsii-pacmak/test/diff-test.sh
rix0rrr: Yes, but for me NOT doing this check just leads to a more cryptic error downstream (you will see an error about the diff failing because one dir has target, .settings, .project).

I added a test explicitly to tell you WHY that was happening, because it was confusing to me.

If your IDE compiles the expectation dir, the test set will fail, regardless of whether this test is here or not.

PR #919, opened 2019-10-30
fix(kernel): correct deserialization of structs in union contexts

### LINKED ISSUE #822: Kernel silently ignores additional properties when deserializing structs
### Affected Languages

- [X] `Python`
- [X] `Java`
- [X] .NET (`C#`, `F#`, ...)

### General Information
* **JSII Version:** `0.17.0`

### What is the problem?
The kernel struct deserialization silently ignores any properties that the type does not declare. When deserializing type unions where the first candidate has no required property, this means serialization will ignore everything and succeed (resulting in an empty object), instead of attempting the next candidate type.

This is the root cause of aws/aws-cdk#3917

RomainMuller:
When deserializing structures in a Union context, the Kernel used to try
options in an arbitrary order (actually - the declaration order which
often is alphanumerically sorted). In addition, it would ignore any
additional property encountered silently.

In cases where several of the union's possibilities had overlapping
required properties, the first attempted option would succeed, even if a
subsequent option would have consumed more properties. A pathologic case
of this is when the first candidate is a type with only optional
properties; as such a type would *always* successfully deserialize
anything, possibly ignoring all properties.

In order to address this, the `structs` can now be passed into the
Kernel by wrapping the object in a decorator box:
```js
{
  "$jsii.struct": {
    "fqn": "fully.qualified.struct.TypeName",
    "data": {
      /* the actual data included in the struct instance */
    }
  }
}
```

This enables "native" languages to correctly communicate the intended
type to the Kernel, so it can make an appropriate deserialization
decision.

The encoding of structs from the Kernel to "native" languages has not
changed: those are still passed by-reference in order to maximize the
compatibility; and to avoid undeterministic behavior in case where union
of structs are returned.

Fixes #822
Fixes aws/aws-cdk#3917
Fixes aws/aws-cdk#2013

---

### DIFF
--- a/packages/jsii-calc/lib/compliance.ts
+++ b/packages/jsii-calc/lib/compliance.ts
@@ -1976,3 +1976,44 @@ class PrivateType extends Implementation implements IAnonymouslyImplementMe {
         return 'to implement';
     }
 }
+
+/**
+ * We can serialize and deserialize structs without silently ignoring optional fields.
+ */
+export interface StructA {
+    readonly requiredString: string;
+    readonly optionalString?: string;
+    readonly optionalNumber?: number;
+}
+/**
+ * This intentionally overlaps with StructA (where only requiredString is provided) to test htat
+ * the kernel properly disambiguates those.
+ */
+export interface StructB {
+    readonly requiredString: string;
+    readonly optionalBoolean?: boolean;
+    readonly optionalStructA?: StructA;
+}
+export class StructUnionConsumer {
+    public static isStructA(struct: StructA | StructB): struct is StructA {
+        const keys = new Set(Object.keys(struct));
+        switch (keys.size) {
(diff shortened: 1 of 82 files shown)


rix0rrr:
It does not seem that this PR includes a solution for Python. Should be just as affected by this issue, no?

rix0rrr:
The most trivial solution to me seems just taking the by-value object from the wire, validating it, but not having the validation change the value that is used. Do we transform/deep-deserialize in some way that makes this impossible/impractical?

A potentially different way of going about this is having the `jsii` compiler, for every union, determine the correct deserialization order which allows it to umabiguously determine the variant? (And error out if it is not possible to determine such an order?)

It would be harder though.

RomainMuller:
@rix0rrr  - there's a change & new compliance test in the Python runtime... Are you saying this is insufficient?

PR #926, opened 2019-10-31
feat(jsii-diff): make assembly validation optional

rix0rrr:
Assembly validation is taking the majority of the time of doing the
API compatibility check, and the assemblies are almost guaranteed to
be correct.

Make it possible to skip the check for performance purposes.

Fixes # 

---

### DIFF
--- a/packages/jsii-diff/bin/jsii-diff.ts
+++ b/packages/jsii-diff/bin/jsii-diff.ts
@@ -18,7 +18,7 @@ async function main(): Promise<number> {
     .option('experimental-errors', { alias: 'e', type: 'boolean', default: false, desc: 'Error on experimental API changes' })
     .option('ignore-file', { alias: 'i', type: 'string', desc: 'Ignore API changes with keys from file (file may be missing)' })
     .option('keys', { alias: 'k', type: 'boolean', default: false, desc: 'Show diagnostic suppression keys' })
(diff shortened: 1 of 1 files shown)


packages/jsii-diff/bin/jsii-diff.ts
```diff
@@ -87,24 +88,24 @@ const NPM_REGEX = /^npm:(\/\/)?/;
  *
  * Supports downloading from NPM as well as from file or directory.
  */
-async function loadAssembly(requested: string): Promise<LoadAssemblyResult> {
+async function loadAssembly(requested: string, validate: boolean): Promise<LoadAssemblyResult> {
```
RomainMuller: Would we accept the boolean as part of an options hash? Then you can just pass `args as any` in...

rix0rrr:
> Sounds like the default should be no validation. 

That is fine by me too. I just didn't want to change the default behavior too much.

> What’s the use case for the validation? What are we protecting against?

At the edges of our system we should always validate user input before trying to operate on it. An arbitrary assembly file is user input, and I'd rather see `Not a valid assembly file because ...` than `Cannot read property 'fqn' of 'undefined'`.

In practice, I think we're most likely to screw this up by mismatching jsii tool version against jsii assembly version.

PR #927, opened 2019-10-31
fix(dotnet/analyzer): remove dependency on Runtime

RomainMuller:
Remove the dependency from the Amazon.JSII.Analyzers package to the
Amazon.JSII.Runtime package - as that dependency was not used and 
turned out to be a source of problems.

Also - pins the version of the runtimes to the exact version (this
could later be extended to a range).

---


packages/jsii-dotnet-analyzers/src/Amazon.JSII.Analyzers/Amazon.JSII.Analyzers.csproj
```diff
@@ -10,9 +10,7 @@
     </PropertyGroup>
 
     <ItemGroup>
-      <PackageReference Include="Amazon.JSII.Runtime" Version="$(JsiiVersion)" />
-      <PackageReference Include="Microsoft.CodeAnalysis" Version="3.2.0" />
-      <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="3.2.0" />
+      <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="3.2.0" PrivateAssets="all" />
```
costleya: Not sure why this needs to be PrivateAssets="all".

packages/jsii-pacmak/test/expected.jsii-calc-base/dotnet/Amazon.JSII.Tests.CalculatorPackageId.BasePackageId/Amazon.JSII.Tests.CalculatorPackageId.BasePackageId.csproj
```diff
@@ -24,7 +24,8 @@
     <EmbeddedResource Include="scope-jsii-calc-base-0.20.0.tgz" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="Amazon.JSII.Runtime" Version="0.20.0" />
+    <PackageReference Include="Amazon.JSII.Runtime" Version="[0.20.0]" />
```
costleya: The main issue with this is that when customers upgrade 1 dependency, they will have to upgrade them all. This is probably undesirable.

Are you sure we have to pin an exact version?

rix0rrr:
I don't feel qualified to comment on this PR at all

packages/jsii-dotnet-analyzers/src/Amazon.JSII.Analyzers/Amazon.JSII.Analyzers.csproj
RomainMuller: This is part of the (Microsoft-authored) tutorial I have found... So I fixed it up to match... I reckon the reason is that the compiler actually comes with it's own and it's better to not pollute, but I could be wrong.

packages/jsii-pacmak/test/expected.jsii-calc-base/dotnet/Amazon.JSII.Tests.CalculatorPackageId.BasePackageId/Amazon.JSII.Tests.CalculatorPackageId.BasePackageId.csproj
RomainMuller: You're right - I loosened this up to match the current major.

PR #928, opened 2019-11-01
fix(pacmak): put package README into the right Python module

rix0rrr:
The (translated) Python README used to be written to the
`<module>._jsii` module, as opposed to the main module.

Put it in the right module where people will actually look for it.

Fixes # 

---

### DIFF
--- a/packages/jsii-pacmak/test/expected.jsii-calc-lib/python/src/scope/jsii_calc_lib/__init__.py
+++ b/packages/jsii-pacmak/test/expected.jsii-calc-lib/python/src/scope/jsii_calc_lib/__init__.py
@@ -1,3 +1,6 @@
+"""
+
+"""
 import abc
 import datetime
 import enum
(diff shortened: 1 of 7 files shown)


packages/jsii-pacmak/test/expected.jsii-calc-lib/python/src/scope/jsii_calc_lib/__init__.py
```diff
@@ -1,3 +1,6 @@
+"""
```
eladb: does it make sense to include an empty docstring or perhaps omit it if it's empty?

packages/jsii-pacmak/test/expected.jsii-calc/python/src/jsii_calc/__init__.py
```diff
@@ -1,3 +1,24 @@
...
+
+## Sphinx
+
+This file will be incorporated into the sphinx documentation.
+
+If this file starts with an "H1" line (in our case `# jsii Calculator`), this
+heading will be used as the Sphinx topic name. Otherwise, the name of the module
+(`jsii-calc`) will be used instead.
+
+## Code Samples
```
eladb: Wow!

PR #932, opened 2019-11-05
fix(kernel): revert behavior change around `any` serialization

RomainMuller:
The change of behavior in `any` serialization introduced in #825 risks
breaking some emerging use-cases of the AWS CDK that are likely to be
prevalent in the installed user base. As a consequence, surgically
reverting this particular change in order to restore the behavior that
users might be dependent on.

This somehow uncovered an outstanding bug in the Python runtime that
was not triggering with the "reverted" behavior (it broke unit tests).
The runtime was fixed in order to address this bug.

---


RomainMuller:
Labelled with `do-not-merge` waiting for me to create a CDK build using this new version of *jsii* to double-check it achieves the desired result.

PR #953, opened 2019-11-07
fix(dotnet): fix callback issues

RomainMuller:
Corrects an issue with handling callbacks in various situations:
- When a callback for a method with optional parameters was invoked with
  some of the optional parameters not specified, an argument count
  mismatch error would occur. This is because dotnet reflection requires
  all arguments to be provided (with `null` if needed).
- When a callback for a variadic method was invoked, a type mismatch or
  a parameter count mismatch error would occur. This is because the
  variadic parameter is an array of it's value type, and has to be
  passed this way out to the reflection call.
- Objects would be created in Javascript without interface annotations,
  causing type information to be lost and incorrect callback argument
  serialization could happen in the Kernel.
- When a class implements members from an interface (and only from one),
  the overrides would not be correctly identified, as the Attributes on
  interface members are *not* visible on the implementors (whereas they
  are visible when they were on an extended class member!).
- Invalid type coertion could happen within the callback handling logic
  in certain edge cases. The complete type information is available, so
  switched to using the full conversion gear form the jsii runtime to
  achieve correct results.

All of those issues are fairly inter-linked and it was difficult to fix
them individually & issue a test that would not break due to another of
the issues... Hence the fixes are grouped in a single commit.

---


RomainMuller:
I reckon the two added tests actually cover all the fixed up behavior; however if you disagree & have an idea of use-cases that aren't tested, please let me know what those are. I've been heads down into this for long enough that I may be blind-sighted at this point 🤨

packages/jsii-dotnet-runtime-test/test/Amazon.JSII.Runtime.IntegrationTests/ComplianceTests.cs
```diff
@@ -509,10 +510,8 @@ class InterfaceWithProperties : DeputyBase, IInterfaceWithProperties
         {
             string _x;
 
-            [JsiiProperty(name: "readOnlyString", typeJson: "{\"primitive\":\"string\"}", isOverride: true)]
```
rix0rrr: Ah yeah. How did we miss this in the first place?

packages/jsii-dotnet-runtime-test/test/Amazon.JSII.Runtime.IntegrationTests/ComplianceTests.cs
```diff
@@ -1197,13 +1239,11 @@ public SubclassNativeFriendlyRandom()
                 _nextNumber = 100;
             }
 
-            [JsiiMethod(name: "hello", returnsJson: "{\"type\":{\"primitive\":\"string\"}}", isOverride: true)]
```
eladb: I am amazed that this was how these tests were written... Did we really expect users to annotate all their methods like this??

RomainMuller: 🤷🏻‍♂️ but to be honest figuring out how to get annotations from interfaces was... a challenge. But I do agree. In terms of ergonomics it was... _terrible_.

packages/jsii-dotnet-runtime/src/Amazon.JSII.Runtime/CallbackExtensions.cs
```diff
@@ -65,37 +67,95 @@ static CallbackResult InvokeCallbackCore(this Callback callback, IReferenceMap r
...
 
-            JsiiMethodAttribute attribute = methodInfo.GetCustomAttribute<JsiiMethodAttribute>();
-            var returnValue = methodInfo.Invoke(deputy, request.Arguments.Select(arg => FromKernel(arg, referenceMap)).ToArray());
-            return attribute?.Returns != null ? new CallbackResult(attribute?.Returns, returnValue) : null;
+            var attribute = methodInfo.GetAttribute<JsiiMethodAttribute>();
+            var parameters = methodInfo.GetParameters();
+            
+            var converter = ServiceContainer.ServiceProvider.GetRequiredService<IJsiiToFrameworkConverter>();
+            
+            var rehydratedArgs = Enumerable.Range(0, request.Arguments.Length)
```
rix0rrr: All this work should be in its own function, no? I have a hard time grokking it and I trust you that it's correct, but I'd like to see it factored out into more readable code units.

Seems like the same hydration/dehydration should happen in multiple places, so why isn't the code shared?

packages/jsii-dotnet-runtime/src/Amazon.JSII.Runtime/CallbackExtensions.cs
```diff
@@ -65,37 +67,95 @@ static CallbackResult InvokeCallbackCore(this Callback callback, IReferenceMap r
...
+                    
+                    if (!requiredType.IsInstanceOfType(value) && value is IConvertible)
+                    {
+                        value = Convert.ChangeType(value, requiredType);
+                    }
+
+                    return value;
+                }).ToArray();
+
+            var invokeParameters = Enumerable.Range(0, parameters.Length)
```
rix0rrr: Same here.

packages/jsii-dotnet-runtime/src/Amazon.JSII.Runtime/ReflectionUtils.cs
```diff
@@ -75,5 +76,71 @@ public static JsiiClassAttribute GetClassAttribute(Type type)
...
+        /// <summary>
+        /// Looks for a Property attribute, starting from the current class hierarchy, and searching on implemented
+        /// interfaces if none was found.
+        /// </summary>
+        /// <param name="property">The property for which an attribute is needed</param>
+        /// <param name="inherit">Whether inherited declarations should be searched for (if false, interfaces will NOT be searched)</param>
+        /// <typeparam name="T">The attribute type that is desired</typeparam>
+        /// <returns>The attribute if one was found, or null</returns>
+        [return: MaybeNull]
+        internal static T GetAttribute<T>(this PropertyInfo property, bool inherit = true) where T : Attribute
```
rix0rrr: These 2 methods look way too similar to me. Can this not be factored out with some generics?

PR #957, opened 2019-11-08
feat(kotlin): add Kotlin/JVM target to pacmak

yanex:
Add initial support for the Kotlin programming language in the `pacmak` generator.

Currently, the Kotlin generator can produce Java-like stubs for Kotlin/JVM.
It reuses the existing Java runtime library.

Current plans:

- [ ] Make stubs a bit more Kotlin-friendly;
    - [ ] Generate `data`-like constructors with named/optional parameters;
- [ ] Add Kotlin/JavaScript generation mode.

---


packages/jsii-calc-base-of-base/package.json
```diff
@@ -42,6 +42,13 @@
           "artifactId": "calculator-base-of-base"
         }
       },
+      "kotlin": {
```
RomainMuller: Would be nice to add the basic information about this to [`docs/configuration.md`](https://github.com/aws/jsii/blob/master/docs/configuration.md).

packages/jsii-pacmak/lib/generator.ts
```diff
@@ -48,18 +48,54 @@ export interface IGenerator {
   save(outdir: string, tarball: string): Promise<any>;
 }
 
+export abstract class GeneratorBase implements IGenerator {
+  abstract generate(fingerprint: boolean): void
+  abstract async load(packageDir: string, assembly: reflect.Assembly): Promise<void>
+  abstract async upToDate(outDir: string): Promise<boolean>
+  abstract async save(outdir: string, tarball: string): Promise<any>
```
RomainMuller: Make sure to have visibility modifiers consistently (I reckon this are `public`).

packages/jsii-pacmak/lib/generator.ts
```diff
@@ -48,18 +48,54 @@ export interface IGenerator {
...
-export abstract class Generator implements IGenerator {
+export abstract class Generator extends GeneratorBase {
   private readonly excludeTypes = new Array<string>();
   protected readonly code = new CodeMaker();
   private _assembly?: spec.Assembly;
   protected _reflectAssembly?: reflect.Assembly;
   private fingerprint?: string;
 
   public constructor(private readonly options: GeneratorOptions = {}) {
+    super();
```
RomainMuller: Might be interesting to actually relocate the `options` property to the `GeneratorBase` class and make it `protected`.

packages/jsii-pacmak/lib/generator.ts
```diff
@@ -110,19 +142,8 @@ export abstract class Generator implements IGenerator {
...
-      name = parts[0];
-    } else if (parts.length === 2 && parts[0].startsWith('@')) {
-      name = parts[1];
-    } else {
-      throw new Error('Malformed assembly name. Expecting either <name> or @<scope>/<name>');
-    }
-
-    return `${name}@${this.assembly.version}.jsii.tgz`
+  protected getAssemblyFileName(): string {
+    return this.getAssemblyArchiveFileName(this.reflectAssembly)
```
RomainMuller: Missing `;` -- I somehow thought I had an `eslint` rule configured to catch this... If that's not the case I have to re-verify that... 😅

packages/jsii-pacmak/lib/targets/jvm/jvmabi.ts
```diff
@@ -0,0 +1,64 @@
+import spec = require('jsii-spec');
```
RomainMuller: That's really interesting! I like what you're doing here!! <3

packages/jsii-pacmak/lib/targets/kotlin.ts
```diff
@@ -0,0 +1,66 @@
...
+import { KotlinGenerator } from './kotlin/kotlingenerator';
+import { getJvmPackageInfos, findJavaRuntimeLocalRepository } from './jvm/mavenutils';
+
+export class Kotlin extends Target {
+  public static toPackageInfos(assm: spec.Assembly): { [language: string]: PackageInfo } {
+    const kotlinTarget = assm.targets && assm.targets.kotlin;
+    const groupId = kotlinTarget && kotlinTarget.maven && kotlinTarget.maven.groupId;
+    const artifactId = kotlinTarget && kotlinTarget.maven && kotlinTarget.maven.artifactId;
+    const version = assm.version;
+    return getJvmPackageInfos('kotlin', groupId, artifactId, version);
```
RomainMuller: If you're able to assume that the `kotlin` target is configured at this stage (aka it was checked before you got there), you can `assm.targets!.kotlin` instead of using the `x && x.y` madness.

packages/jsii-pacmak/lib/targets/kotlin.ts
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  protected readonly generator = new KotlinGenerator();
+
+  public async build(sourceDir: string, outDir: string): Promise<void> {
+    const url = `file://${outDir}`;
+    const gradleArguments = new Array<string>();
+    for (const arg of Object.keys(this.arguments)) {
+      if (!arg.startsWith('gradle-')) { continue; }
+      gradleArguments.push(`--${arg.slice(4)}`);
```
RomainMuller: I think the `.slice(4)` needs to be `.slice(7)` -- I think that was copied from the Java generator where the stripped prefix is `mvn-`, and here you're stripping `gradle-` instead.

packages/jsii-pacmak/lib/targets/kotlin.ts
```diff
@@ -0,0 +1,66 @@
...
+    const customRepos = await this.getCustomRepos(outDir);
+
+    const args = [];
+    let index = 1;
+    for (const repo of customRepos) {
+      args.push(`-PcustomRepo${index}=file://${repo}`);
+      index += 1;
+    }
+
+    return Promise.resolve(args);
```
RomainMuller: The method is `async` so you can simply return `args` here. Right now this gets wrapped twice (once explicitly, once implicitly due to  `async`).

RomainMuller: This can be re-written to:

```ts
return (await this.getCustomRepos(outDir))
  .map((repo, index) => `-PcustomRepo${index}=file://${repo}`);
```

packages/jsii-pacmak/lib/targets/kotlin.ts
```diff
@@ -0,0 +1,66 @@
...
+    const repos = await this.findLocalDepsOutput(this.packageDir);
+
+    repos.push(outDir);
+    
+    const localJavaRuntime = await findJavaRuntimeLocalRepository();
+    if (localJavaRuntime) {
+      repos.push(localJavaRuntime);
+    }
+
+    return Promise.resolve(repos);
```
RomainMuller: Same as before - no need to wrap in `Promise.resolve` here.

packages/jsii-pacmak/lib/targets/kotlin/declarationnode.ts
```diff
@@ -0,0 +1,49 @@
...
+    return type.name;
+  }
+
+  private _type?: Type;
+  private _children: { [name: string]: DeclarationNode } = {};
+
+  private constructor(public readonly name: string) {}
+
+  public get type(): Type | undefined {
+    return this._type;
```
RomainMuller: Feels odd for me to have this being late-bound (but willing to give the benefits of doubt here)

packages/jsii-pacmak/lib/targets/kotlin/kdocbuilder.ts
```diff
@@ -0,0 +1,46 @@
+import { CodeMaker } from 'codemaker'
+
+export class KDocBuilder {
+  private readonly chunks: string[] = [];
+
+  public push(
+    chunk: string | undefined,
+    prefix: string | undefined = undefined
```
RomainMuller: Could make the last argument optional (this way `undefined` is the default value).

packages/jsii-pacmak/lib/targets/kotlin/kdocbuilder.ts
```diff
@@ -0,0 +1,46 @@
...
+  ) {
+    if (!chunk || chunk.length === 0) {
```
RomainMuller: `!chunk` actually already covers `chunk === ''` which is another way to say `chunk.length === 0`. The empty string is false-y in Javascript!

packages/jsii-pacmak/lib/targets/kotlin/kdocbuilder.ts
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  private getLines(): string[] {
+    const lines = [];
+    for (const chunk of this.chunks) {
+      if (lines.length > 0) {
+        lines.push('');
+      }
+
+      lines.push(...chunk.split('\n'));
```
RomainMuller: I wonder if the `split` could (should?) be done upfront. Not that this is bad, just a suggestion.

packages/jsii-pacmak/lib/targets/kotlin/kotlingenerator.ts
```diff
@@ -0,0 +1,254 @@
...
+      throw new Error('Assembly is not set');
+    }
+
+    return assembly;
+  }
+  
+  public async load(packageDir: string, assembly: Assembly): Promise<void> {
+    this.packageDir = packageDir;
+    this._assembly = assembly;
+    return Promise.resolve();
```
RomainMuller: You don't need that `return` clause at all here. I'll stop signaling the unnecessary `Promise`-wrapping of results of `async` functions from here one (I mean you probably got the point by now!)

packages/jsii-pacmak/lib/targets/kotlin/declarationnode.ts
yanex: As we traverse the declarations, it seems there is no guarantee that `foo` will appear earlier than `foo.bar`. So I firstly register a declaration name alone and bind the type when it's found.

packages/jsii-pacmak/lib/targets/kotlin/kotlingenerator.ts
yanex: Hm-m, eslint says that "Async method 'load' has no 'await' expression.eslint(@typescript-eslint/require-await)" if I remove the line.
I think it makes sense to remove the `async` things altogether both from `KotlinGenerator` and `Generator`, as they don't benefit from the coroutine functionality.

packages/jsii-pacmak/test/expected.jsii-calc-base-of-base/kotlin/src/main/kotlin/software/amazon/jsii/tests/calculator/baseofbase/VeryBaseProps.kt
```diff
@@ -0,0 +1,46 @@
+package software.amazon.jsii.tests.calculator.baseofbase
+
+@javax.annotation.Generated("jsii-pacmak")
+interface VeryBaseProps : software.amazon.jsii.JsiiSerializable {
+    val foo: software.amazon.jsii.tests.calculator.baseofbase.Very
+
+    class Builder {
```
kiiadi: be cool if we could do this in a more Kotlin DSLy kinda way, I'm thinking you could add something like this to the interface:

```kt
companion object {
    operator fun invoke(block: Builder.() -> Unit): VeryBaseProps {
        return Builder().apply(block).build()
    }
}
```

So then you can use it like so:

```kt
val props = VeryBaseProps {
    foo = Very.create()
}
```

packages/jsii-pacmak/test/expected.jsii-calc/kotlin/src/main/kotlin/software/amazon/jsii/tests/calculator/Add.kt
```diff
@@ -0,0 +1,40 @@
+package software.amazon.jsii.tests.calculator
+
+/**
+ * The "+" binary operation.
+ */
+@javax.annotation.Generated("jsii-pacmak")
+@software.amazon.jsii.Jsii(module = software.amazon.jsii.tests.calculator.`$Module`::class, fqn = "jsii-calc.Add")
+@software.amazon.jsii.Stability(software.amazon.jsii.Stability.Level.Experimental)
+open class Add : software.amazon.jsii.tests.calculator.BinaryOperation {
```
kiiadi: Does JSII have a concept of immutable types? e.g. objects that cannot be extended?

docs/configuration.md
```diff
@@ -211,6 +211,30 @@ The resulting artifact is a **Maven** package that can be deployed to
...
+
+Kotlin/JVM target configuration is [similar to Java](#configuring-java).
+* `maven` – `groupId`, `artifactId` and optional `versionSuffix`, as in Java.
+* `package` – the containing Kotlin qualified package name.
+
+Example:
+```js
+{
+  "jsii": {
+    "kotlin-jvm": {
```
kiiadi: Would you consider a multi-platform binding? Kotlin JS might give us more ability to interact with Typescript/JS directly.

packages/jsii-pacmak/test/expected.jsii-calc/kotlin/src/main/kotlin/software/amazon/jsii/tests/calculator/Add.kt
RomainMuller: TypeScript doesn't have such a thing, so no, we don't have that. We may come up with clever tricks to make ti possible in the future though (this is an idea that we have already touched on several times in the past - for there are really interesting aspects to it).

docs/configuration.md
RomainMuller: Our plan is to have a Kotlin-JVM and a Kotlin-JS target (separate). Nothing carved in stone though... But we definitely intend on supporting both flavors in one way or another.

packages/jsii-pacmak/test/expected.jsii-calc-base-of-base/kotlin/src/main/kotlin/software/amazon/jsii/tests/calculator/baseofbase/VeryBaseProps.kt
GrahamLea: Considering Kotlin's combo of named arguments and default parameter values, I feel there should be no need for a Builder pattern in Kotlin. In fact, I'd see it as a main advantage of a Kotlin impl to abandon the Builder pattern. A constructor for a props object with all the required parameters up front and the ones if default values following should function identically to a Builder at the call site, right?

GrahamLea:
I think it would make sense to make all Props classes `data` classes, especially due to the `copy()` function it would make available. I can see a use for making a Props instance that has a lot of common properties, then using `copy()` to make more specific instances for use in actual resource definitions.
(Sorry, I don't know this code-generation code enough well enough to see whether this has already been done or not, or against what code this comment belongs.)

packages/jsii-pacmak/test/expected.jsii-calc-base-of-base/kotlin/src/main/kotlin/software/amazon/jsii/tests/calculator/baseofbase/VeryBaseProps.kt
kiiadi: The main issues I see with this approach is backwards compatibility, when we're generating code - how do we know what was previously required vs not? It's currently legal for service teams to drop "required" on proprieties so the property order may change. There's nothing to force a consumer to explicitly name parameters at the call site. If the order of a bunch of `String` properties changes then consumers are going to be in for a bad time.

By going the "DSL Builder" route we force explicit names to properties.

RomainMuller:
Looks like this PR was closed in error as the `master` branch was renamed to `main`. I'll try to re-open this myself...

RomainMuller:
This had been closed in error as I renamed the `master` branch over to `main`. Sorry about that!

packages/jsii-pacmak/lib/targets/kotlin/kotlingenerator.ts
RomainMuller: Yup you're correct, this pesky rule... Would be kinda nice if it could shut up about overrides/implementations... but well well no biggie.

Stvad:
Is it still alive though 👀?

mrgrain:
Unfortunately this is too diverged from the current main branch now.

PR #968, opened 2019-11-11
fix(kernel): correctly de-serialize mappings as JSON

RomainMuller:
When mapping data (e.g. a Python `dict`) was passed through a JSON value,
it would not be deserialized correctly and the `$jsii.map` markers would
remain in the JS-visible map.

Fixes the cause of aws/aws-cdk#4957

---

### DIFF
--- a/packages/jsii-kernel/lib/serialization.ts
+++ b/packages/jsii-kernel/lib/serialization.ts
@@ -171,6 +171,11 @@ export const SERIALIZERS: {[k: string]: Serializer} = {
     deserialize(value, optionalValue, host) {
       // /!\ Top-level "null" will turn to undefined, but any null nested in the value is valid JSON, so it'll stay!
       if (nullAndOk(value, optionalValue)) { return undefined; }
+
+      // A mapping object can arrive though here. This would be the case if anything that is valid into a Map<string, ?>
+      // is passed into a JSON transfer point. Indeed, those are also valid JSON! For example, Python "dicts" will be
+      // serialized (by the Python runtime) as a $jsii.map (the mapping object). We need to de-serialize that as a
+      // Map<string, JSON> in order to obtain the correct output behavior here!
       if (isWireMap(value)) {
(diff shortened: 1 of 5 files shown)


packages/jsii-kernel/lib/serialization.ts
```diff
@@ -168,10 +168,33 @@ export const SERIALIZERS: {[k: string]: Serializer} = {
       // Just whatever. Dates will automatically serialize themselves to strings.
       return value;
     },
-    deserialize(value, optionalValue) {
+    deserialize(value, optionalValue, host) {
       // /!\ Top-level "null" will turn to undefined, but any null nested in the value is valid JSON, so it'll stay!
       if (nullAndOk(value, optionalValue)) { return undefined; }
-      return value;
+      if (isWireMap(value)) {
```
eladb: can you add a comment explaining why would a value be a "wire map" and what it is?

packages/jsii-python-runtime/tests/test_compliance.py
```diff
@@ -193,8 +193,8 @@ def test_primitiveTypes():
     assert types.date_property == datetime.fromtimestamp(123 / 1000.0, tz=timezone.utc)
 
     # json
-    types.json_property = {"Foo": 123}
-    assert types.json_property.get("Foo") == 123
+    types.json_property = { "Foo": { "bar": 123  } }
```
eladb: Would it make sense to do that for java and .net as well?

eladb: Did you verify that this indeed fails before the fix is applied?

(Sorry being a bit risk averse)

RomainMuller: Oh Java is going to be such fun to write!

RomainMuller: Nevermind... Java actually _prohibits_ this particular use-case through static typing!

RomainMuller:
@eladb done!!

PR #969, opened 2019-11-11
fix(python): pin dependency version to the exact match

RomainMuller:
This will make it easier for downstream consumers to revert to an older version of closures.
This is a temporary fix until we make further changes to forward the *exact* requirement that
was expressed on the source package.

Barring this, frighting one's way out of problems such as aws/aws-cdk#4957 is extremely
challenging.

---

### DIFF
--- a/packages/jsii-pacmak/lib/targets/python.ts
+++ b/packages/jsii-pacmak/lib/targets/python.ts
@@ -10,7 +10,7 @@ import { Generator, GeneratorOptions } from '../generator';
 import { warn } from '../logging';
 import { md2rst } from '../markdown';
 import { Target } from '../target';
-import { shell } from '../util';
+import { shell, nextMajorVersion } from '../util';
 
 /* eslint-disable @typescript-eslint/no-var-requires */
 const spdxLicenseList = require('spdx-license-list');
@@ -1140,18 +1140,7 @@ class Package {
     const expectedDeps = this.metadata.dependencies || {};
     for (const depName of Object.keys(expectedDeps)) {
       const depInfo = expectedDeps[depName];
-      // We need to figure out what our version range is.
-      // Basically, if it starts with Zero we want to restrict things to
-      // ~=X.Y.Z. If it does not start with zero, then we want to do ~=X.Y,>=X.Y.Z.
-      const versionParts = depInfo.version.split('.');
-      let versionSpecifier: string;
-      if (versionParts[0] === '0') {
-        versionSpecifier = `~=${versionParts.slice(0, 3).join('.')}`;
-      } else {
-        versionSpecifier = `~=${versionParts.slice(0, 2).join('.')},>=${versionParts.slice(0, 3).join('.')}`;
-      }
-
-      dependencies.push(`${depInfo.targets!.python!.distName}${versionSpecifier}`);
+      dependencies.push(`${depInfo.targets!.python!.distName}==${depInfo.version}`);
     }
 
     code.openFile('README.md');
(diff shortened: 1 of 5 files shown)


packages/jsii-pacmak/lib/targets/python.ts
```diff
@@ -1140,26 +1140,16 @@ class Package {
...
-      const versionParts = depInfo.version.split('.');
-      let versionSpecifier: string;
-      if (versionParts[0] === '0') {
-        versionSpecifier = `~=${versionParts.slice(0, 3).join('.')}`;
-      } else {
-        versionSpecifier = `~=${versionParts.slice(0, 2).join('.')},>=${versionParts.slice(0, 3).join('.')}`;
-      }
-
-      dependencies.push(`${depInfo.targets!.python!.distName}${versionSpecifier}`);
+      dependencies.push(`${depInfo.targets!.python!.distName}==${depInfo.version}`);
```
eladb: Does this have the same meaning like in all other languages?

joekiller:
@eladb per my comment here https://github.com/aws/jsii/pull/677#issuecomment-552633828 it should be ~=version instead of ==. ~= is fine with X.Y.Z but the current ~=X.Y... is incorrect

joekiller:
It really comes down to npm is like, people are dumb and 2.7 == 2.7.0 but python is like, well 2.7 isn't technically 2.7.0 because they don't enforcement major.minor.bug.

joekiller:
== is fine too :) I'm done for the day

RomainMuller:
Ideally, I want the dependency to match what is modeled in the source TypeScript package. This requires a few slightly deeper changes... I'll leave this for now, will get back to it tomorrow or so to re-consider on a cold head.

joekiller:
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-1.html#matching-behavior might help you reach a conclusion. 

Being that the dependency in typescript is an arbitrary number I'd suggest Python's equivalent of ===

https://www.python.org/dev/peps/pep-0440/#arbitrary-equality

joekiller:
arbitrary in the context that the typescript's module's version number is just as it is specified. Ie it is 1.15.0 not 1.15.0.1 not 1.15 it's just "1.15.0" and

RomainMuller:
The thing is here we're not talking about the version of typescript itself, but that of the declared dependencies of the node module we are wrapping into a nice Python shell. Those dependencies are expressed using semver ranges, and right now the generator ignores the exact specified range & generates it's own.

Currently, CDK dependencies are modeled with an exact match clause, and python actually breaches that with the current generated code... That's not great :(

joekiller:
I was noting that the semver techniques for typescript are the same as npm.

The cdk dependencies are modeled after the package.json file correct? Ie https://github.com/aws/aws-cdk/blob/f0a62fd5324cc368752bc75f0bdae620a13e826d/packages/%40aws-cdk/aws-codepipeline-actions/package.json#L78

Is there somewhere else doing some dependencies calculations?

It appears that for all python dependencies everything is an exact version on the package.json side. Ie an arbitrary version as it has not range and is precisely the desired version. To match the expectation of exactly matching in python an exact match operator is the triple equals, `===`. That should give you the same result every time. In package.json "node-pkg": "1.2.3" will always deliver 1.2.3.  In python "node-pkg===1.2.3" always delivers "1.2.3".


If I'm missing something totally obvious please forgive me. I have fought many a dependency battle in nearly any language that has a dependency tree.

joekiller:
I swear I'll stop commenting after this... pythonically ~=X.Y.Z is what you want to keep it on the bugfix version.

RomainMuller:
@joekiller  - I reckon we're in violent agreement at this stage 😂

RomainMuller:
Closing in favor of some upcoming work from @MrArnoldPalmer

PR #973, opened 2019-11-12
fix(python): correctly handle nested structs-as-dict

RomainMuller:
Added code to up-cast from `dict` to the relevant struct type in the generated
python code whenever the value type is some unambiguous struct. Whenever
a UNION of types is involved, the user must construct the struct type themselves,
as the runtime would then have no way to choose which type should be used on
it's own without additional information.

---

### DIFF
--- a/packages/jsii-pacmak/test/expected.jsii-calc/java/src/main/java/software/amazon/jsii/tests/calculator/RootStruct.java
+++ b/packages/jsii-pacmak/test/expected.jsii-calc/java/src/main/java/software/amazon/jsii/tests/calculator/RootStruct.java
@@ -1,0 +1,156 @@
+package software.amazon.jsii.tests.calculator;
+
+/**
+ * This is here to check that we can pass a nested struct into a kwargs by specifying it as an in-line dictionary.
+ * 
+ * This is cheating with the declared types, but Python people don't play by
+ * the rules much apparently.
+ * 
+ * EXPERIMENTAL
+ */
+@javax.annotation.Generated(value = "jsii-pacmak")
+@software.amazon.jsii.Jsii(module = software.amazon.jsii.tests.calculator.$Module.class, fqn = "jsii-calc.RootStruct")
+@software.amazon.jsii.Jsii.Proxy(RootStruct.Jsii$Proxy.class)
+@software.amazon.jsii.Stability(software.amazon.jsii.Stability.Level.Experimental)
+public interface RootStruct extends software.amazon.jsii.JsiiSerializable {
+
+    /**
+     * May not be empty.
+     * 
+     * EXPERIMENTAL
+     */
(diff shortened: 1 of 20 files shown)


RomainMuller:
There [is](https://github.com/aws/jsii/pull/973/files#diff-1c7d3c796506f8c6d9b6660813bc28f3) a [test](https://github.com/aws/jsii/pull/973/files#diff-1c7d3c796506f8c6d9b6660813bc28f3)

packages/jsii-python-runtime/tests/test_compliance.py
```diff
@@ -1000,6 +1002,15 @@ def test_consumer_calls_method_privateclass():
 def test_consumer_calls_method_typed_as_class():
     assert ConsumerCanRingBell().when_typed_as_class(PythonConcreteBellRinger())
 
+def test_can_pass_nested_struct_as_dict():
+    # Those shouldn't raise:
+    RootStructValidator.validate(string_prop= 'Pickle Rick!!!')
+    RootStructValidator.validate(
```
eladb: Please add a test that validates the None case (where there is no value passed into the nested struct)

RomainMuller: `None` is the default value of the second argument, so this is tested in the first case here... But I can add an explicit one, too, yeah.

packages/jsii-pacmak/test/expected.jsii-calc/java/src/main/java/software/amazon/jsii/tests/calculator/RootStruct.java
```diff
@@ -0,0 +1,156 @@
+package software.amazon.jsii.tests.calculator;
+
+/**
+ * This is here to check that we can pass a nested struct into a kwargs by specifying it as an in-line dictionary.
+ * 
+ * This is cheating with the declared types, but Python people don't play by
+ * the rules much apparently.
```
joekiller: I know y'all are opinionated but having pages and pages of this comment could be considered offensive by those using the target language as they are obviously playing by the rules.

eladb: You are absolutely right. This is not appropriate. We’ll remove this comment.

RomainMuller: Apologies - this was meant in a playful tone but didn't come across as intended & is indeed inappropriate.

Edited version at #975

joekiller: Yeah maybe "Python doesn't play by the same rules." Appreciate y'alls work.

PR #980, opened 2019-11-12
fix(python): correctly handle interface declarations on returned objects

### LINKED ISSUE #976: Python: return a subclass of a known class implementing an interface
### Affected Languages

- [x] `Python`
- [ ] `Java`
- [ ] .NET (`C#`, `F#`, ...)

### General Information
* **JSII Version:** 0.20.4

### What is the problem?

```ts
export interface IReturn {
  readonly foo: number;
}

export class Base {
}

export class SomeType {
  static returnReturn(): IReturn {
    class Derived extends Base implements IReturn {
      foo: 333
    }
  }
}
```

When calling `SomeTime.return_return()` from Python, the returned object will be `Base` instead of `IReturn`.

See aws/aws-cdk#4966

RomainMuller:
When the returned type had a known base type, the interfaces tagged on the object
reference tag. This addresses this problem.

Fixes #976 

---


eladb:
@RomainMuller I tool the liberty of implementing this test in .NET and Java, just in case, and also noticed that a related test is missing from python, so I added it (all passed).

dxunix:
using `cdk 1.16.3 (build 8c7f593)`

 'Resource' object has no attribute ... error still exists. 
`vpc = ec2.Vpc.from_lookup(self, "myvpc", vpc_id=VPC_ID)`

jvolkman:
@dxunix I had to update all of the Python dependencies in my virtual env before the issue was resolved.

PR #983, opened 2019-11-13
fix(dotnet): allow down-casting to parent interface type

### LINKED ISSUE #982: .NET won't cast to parent interface type
### Affected Languages

- [ ] `Python`
- [ ] `Java`
- [X] .NET (`C#`, `F#`, ...)

### General Information
* **JSII Version:** `0.20.x`
* **Platform:**     agnostic

### What is the problem?
The automated conversion of types in the .NET runtime checks too strictly and will
refuse to cast to any type that is not EXACTLY declared on the object reference.
In certain cases however, we are casting to a parent interface of the actual declared
interface and this is 100% legal.

## Related
Causes aws/aws-cdk#4989

RomainMuller:
Cast controls were too strict and prevented the framework from
successfully down-casting an object reference to a parent interface of
it's declared type; causing a cast error.

The new code looks for class compatibility using the standard .NET
primitives and successfully performs the requested cast.

Fixes #982

---


packages/jsii-dotnet-runtime-test/test/Amazon.JSII.Runtime.IntegrationTests/Amazon.JSII.Runtime.IntegrationTests.csproj
```diff
@@ -19,4 +19,8 @@
         <DotNetCliToolReference Include="dotnet-xunit" Version="2.3.1" />
     </ItemGroup>
 
+    <ItemGroup>
+      <ProjectReference Include="..\..\..\jsii-dotnet-runtime\src\Amazon.JSII.Runtime\Amazon.JSII.Runtime.csproj" />
+    </ItemGroup>
```
eladb: where is this coming from and how come it wasn't required so far?

RomainMuller: So. I've added this because it makes working with the integration tests in an IDE much nicer... It isn't _required_ per se, but ensures that the *local* source is used instead of what's in the symbols package (makes debugging more efficient, etc...)

I routinely add this when I work in the kernel, and since this is not a published artifact, I figured I'd leave it in this time... Because it'd help everyone in this situation.

RomainMuller:
Added the mighty Python test (I swear that was always the plan, but I **knew** this was not a problem there, and I had to turn to other problems for a moment :D)

PR #985, opened 2019-11-14
feat(rosetta): add skeleton of Java snippet translator

skinny85:
Includes a testing harness for the translators that executes
by automatically detecting files in a directory,
and using those files as the expected translation values.

---


packages/jsii-rosetta/lib/languages/java.ts
```diff
@@ -0,0 +1,629 @@
...
+import { DefaultVisitor } from './default';
+import { AstRenderer } from '../renderer';
+import { NO_SYNTAX, OTree } from '../o-tree';
+import { builtInTypeName, mapElementType, typeWithoutUndefinedUnion } from '../typescript/types';
+import { isReadOnly, matchAst, nodeOfType, quoteStringLiteral, visibility } from '../typescript/ast-utils';
+import { ImportStatement } from '../typescript/imports';
+import { jsiiTargetParam } from '../jsii/packages';
+
+interface JavaContext {
+  /** @default false */
```
rix0rrr: Some docs on what each property represents maybe?

packages/jsii-rosetta/lib/languages/java.ts
```diff
@@ -0,0 +1,629 @@
...
+    }
+    return new OTree(
+      [],
+      importStatement.imports.elements.map(importEl => `import ${namespace}.${importEl.sourceName};`),
+      { canBreakLine: true, separator: '\n' },
+    );
+  }
+
+  /*
+   * Because structs are represented in Java as interfaces with a JsiiProxy implementation,
```
rix0rrr: But what if a user wants to write a struct themselves in Java, to be used in Java? Then there will be no `JsiiProxy`. They would probably write a class to hold their "props". See the C# generator.

This translator is about code the user would write, not mirroring what jsii does for them.

packages/jsii-rosetta/lib/languages/java.ts
```diff
@@ -0,0 +1,629 @@
...
+    );
+  }
+
+  public printStatement(args: ts.NodeArray<ts.Expression>, renderer: JavaRenderer) {
+    return new OTree(
+      [
+        'System.out.println(',
+        args.length === 1
+          ? renderer.convert(args[0])
+          : new OTree([], ['""', ...renderer.convertAll(args)], { separator: ' + ' }),
```
rix0rrr: See Chime comment. Even though this may be semantically correct, I don't think this looks very nice to a user reading code samples. Can we simplify? If we have to sacrifice correctness in edge cases, that's okay, this doesn't have to be executed anyway.

packages/jsii-rosetta/lib/languages/java.ts
```diff
@@ -0,0 +1,629 @@
...
+      return fallback;
+    }
+
+    return this.renderType(type, renderer, fallback);
+  }
+
+  private renderType(type: ts.Type, renderer: JavaRenderer, fallback: string): string {
+    const nonUnionType = typeWithoutUndefinedUnion(type);
+    if (!nonUnionType) {
+      // renderer.report(typeNode, 'Type unions in examples are not supported');
```
rix0rrr: Why commented out?

packages/jsii-rosetta/lib/languages/java.ts
```diff
@@ -0,0 +1,629 @@
...
+    renderer: JavaRenderer,
+    methodOrConstructorName: ts.Node | undefined,
+    returnType: string | undefined): OTree {
+
+    const overloads = new Array<OTree>();
+    for (let i = 0 ; i < node.parameters.length; i++) {
+      const param = node.parameters[i];
+      if (!!param.questionToken || !!param.initializer) {
+        // The parameter is either optional, or has a default -
+        // render an overload that delegates to a version with one more parameter.
```
rix0rrr: Nice catch!

packages/jsii-rosetta/lib/languages/java.ts
skinny85: Fixed.

packages/jsii-rosetta/lib/languages/java.ts
skinny85: Changed.

packages/jsii-rosetta/lib/languages/java.ts
skinny85: Done.

packages/jsii-rosetta/lib/languages/java.ts
skinny85: Can you elaborate on what the translation should be? Let's say we have

```ts
interface MyStruct {
  readonly a: string;
  b: number;
}
```

in TypeScript. Should I generate

a) an interface with getters (and setters for non-readonly fields)?
b) a class with a constructor taking all fields as arguments, and just assigning them in that constructor?
c) a class that has a nested static Builder class that is used to create it, and the constructor of the class itself is private, similar to what JSII is doing?
d) a  simple class with just the fields, and Lombok annotations (`@Builder`, `@Data`, etc.)?
e) something else...?

rix0rrr: Yeah, good question. I'd like the syntactically most lightweight thing that roughly expresses the intent. I have a feeling the builder would add a lot of noise to make it work nicely at the consumer site, so I think I'd prefer either b or d (or b as a POJO).

I guess between those it depends on how reasonable it is for people to have Lombok installed. If not, I think POJOs are a reasonable compromise? Or honestly, publicly accessible fields without getters or setters (is what I did in C#). In fact, that is what structs are, just a bag of data. `readonly` doesn't really come into it, because this is only for structs and all fields are readonly by definition.

I can't really tell you the solution here, only what I value: readability and simplicity. I don't want all the syntax of a "super up-to-snuff" class definition get in the way of the rest of the example. People know how to translate from plain public attributes or POJOs to their preferred immutable/builder/deserialize/Lombok flavors that apply.

I think non-final public variables are the simplest solution. Only thing that gives me slight pause there is that people will get stuck up parroting how "that's an anti-pattern" to us and missing the forest for the trees.

Wrap those suckers up in getters and setters that don't do anything other than manipulating those same members (and crucially, never will, either!) and suddenly nobody bats an eye. Go figure.

I need your opinion here. Do what seems right to you.

skinny85: OK. I've settled on private fields + getters and fluent setters, so creating instances of the class looks something like this:

```java
new MyStruct()
        .a("a")
        .b(42);
```

, which seems the closest we can get in Java to JavaScript's object literal syntax without generating a separate Builder class, which would be very verbose, like you said.

Let me know if this works!

PR #987, opened 2019-11-14
fix(java): remove Jackson confusion with certain patterns

RomainMuller:
Upon deserializing (`ObjectMapper#treeToValue`), Jackson will first
validate the `valueType` for whether ti can be deserialized with the
"standard" bean deserializer. This will bail out if there are multiple
ambiguous setters for a given property (they are ambiguous if they have
parameters of non-trivial types, such as `List`, some class, ...).

In order to remove the problem, this changes the deserialization path so
that instead of asking for deserialization into the needed instance type
directly, `JsiiObject` will be requested instead when the declared type
is a sub-class of that. Since such types are passed by-reference, the
custom de-serializer modifier will correctly determine the "right" class
to use (the declared type, or a subtype of it), and return this... But
Jackson will not get the chance to be confused or unhappy about that
type's structure.

This was the root cause of aws/aws-cdk#4080

---


packages/jsii-calc/lib/compliance.ts
```diff
@@ -2266,3 +2266,18 @@ export class Demonstrate982 {
 
     public constructor() { }
 }
+
+/**
+ * This tries to confuse Jackson by having overloaded property setters.
+ *
+ * @see https://github.com/aws/aws-cdk/issues/4080
+ */
+export class ConfusingToJackson {
```
rix0rrr: Are we okay with serialization of structs that contain unions, and do we have a test for that?

RomainMuller: They're _also_ passed by reference; but let me add a test to this effect, still.

rix0rrr:
When is this revert going to be reverted?

PR #988, opened 2019-11-14
feat: upgrade to typescript 3.7

### LINKED ISSUE #996: noImplicitAny is temporarily disabled
To unblock #995 without requiring to upgrade TypeScript, the option `noImplicitAny` has been disabled from the `tsconfig.json` generated by the jsii compiler.

Once we upgrade to the 3.7.x version-line, we should be able to restore this validation, although might be tricky since it's a breaking change... Might need to wait until the next major version bump.

RomainMuller:
Enjoy the benefits of optional chaining, a null-coalescing operator, and other
fun new features and enhancements TypeScript brought in 3.7.  For more
information about the TypeScript features, refer to the [TypeScript release notes].

[TypeScript release notes]: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html

Closes #996

---


package.json
```diff
@@ -14,8 +14,7 @@
     "@typescript-eslint/eslint-plugin": "^2.7.0",
     "@typescript-eslint/parser": "^2.7.0",
     "eslint": "^6.6.0",
-    "lerna": "^3.18.4",
-    "typescript": "~3.6.4"
+    "lerna": "^3.18.4"
```
eladb: wouldn't it make sense to remove all dependencies from the root?

packages/jsii-diff/lib/type-analysis.ts
```diff
@@ -211,7 +211,7 @@ class TypePairs {
     if (!a.fqn || !b.fqn) { return false; } // Only for user-defined types
 
     const image = this.pairs.get(a.fqn);
-    return !!image && image.has(b.fqn);
+    return !!image?.has(b.fqn);
```
eladb: why do we need the `!!` at the beginning here?

package.json
RomainMuller: Probably would. I'll do this in a separate PR, though.

packages/jsii-diff/lib/type-analysis.ts
RomainMuller: Because I need to coerce this to a boolean context due to surrounding typing.

PR #997, opened 2019-11-18
fix(kernel): cannot pass decorated structs to kernel as "any"

RomainMuller:
When a struct is serialized with a $jsii.struct decoration from a jsii host language and passed into a value typed `any`, the kernel ignored the decoration. This resulted in a malformed object passed to the javascript code.

This fix undecorates the object and continues to deserialize it as `any`.

Fixes aws/aws-cdk#5066

---


packages/jsii-kernel/lib/serialization.ts
```diff
@@ -520,6 +520,15 @@ export const SERIALIZERS: {[k: string]: Serializer} = {
...
         return host.objects.findObject(value).instance;
       }
+    
+      // if the value has a struct token, it was serialized by a typed jsii
+      // struct, but since the deserialization target is ANY, all we can do is
+      // strip the data from $jsii.struct and continue to deserialize as ANY.
+      if (isWireStruct(value)) {
+        const { fqn, data } = value[TOKEN_STRUCT];
+        host.debug(`ANY is a struct of type ${fqn}`);
+        return SERIALIZERS[SerializationClass.Any].deserialize(data, _type, host);
```
RomainMuller: Could have passed to the `SerializationClass.Struct` deserialize instead... That'd kinda feel like a better way?

eladb: The struct deserializer assumes `type` is an FQN and crashes when called with `any`. Let me see what it entails to update it to special case `any`.

eladb: This is getting too hairy to fix in the struct deserializer.

RomainMuller: Care to add a note (TODO or whatever?)

eladb: Your advise to pass in the FQN worked, so I'll push another commit to fix.

RomainMuller: Oh sweet!

eladb: Done

PR #1008, opened 2019-11-19
chore: ignore jsii-calc and @scope/* in package.sh

RomainMuller:
Those packages contain no artifacts that should be published, and considering them may cause leftover artifacts from tests to be part of the release artifacts, which breaks integration tests.

---


RomainMuller:
Local build confirmed the right artifacts were present (and only those).

PR #1074, opened 2019-11-27
chore: configure import/no-extraneous-dependencies eslint rule

### LINKED ISSUE #1031: Unable to identify missing local dependencies during build
## :bug: Bug Report

Since yarn currently hoists all local dependencies into the root node_modules, modules can accidentally use a dependency that they don't declare in their `package.json`. This causes latent production issues such as #1030

We should figure out a way to disable local-dependency hoisting or not use yarn because this is a huge gap in our test coverage.

RomainMuller:
It allows to ensure that modules are not taking dependencies on things
they have not declared as part of their (direct) dependencies. Added
the necessary configuration so that all existing packages are duely
linted, and consequently addressed all violations in `jsii-rosetta`.

This hopefully also fixes the fact that the `eslint` plugin in VSCode
was not always able to lint files in-place (depending on how/where
you'd open the editor). It achieves this result by setting a better
form of the TypeScript parser for `eslint`, centralizing the "common"
parts of the TypeScript configurations in `tsconfig-base.json` and
updating all packages to extend from that, as well as adding a
"solution" `tsconfig.json` at the root of the repository that encodes
references to all individual (non-fixture) packages, per the
[TypeScript recommendation][1].

[1]: https://www.typescriptlang.org/docs/handbook/project-references.html#overall-structure

Fixes #1031

---


eslint-config.yaml
```diff
@@ -139,6 +141,12 @@ rules:
     - always
     - null: ignore
 
+  'import/no-extraneous-dependencies':
+    - error
+    - devDependencies: ['**/test/**'] # Only allow importing devDependencies from tests
```
eladb: nice!

packages/codemaker/tsconfig.json
```diff
@@ -1,19 +1,4 @@
...
-    "inlineSources": true,                 /* Emit the source alongside the sourcemaps within a single file. */
-    "noUnusedLocals": true,                /* Report errors on unused locals. */
-    "noUnusedParameters": true,            /* Report errors on unused parameters. */
-    "noImplicitReturns": true,             /* Report error when not all code paths in function return a value. */
-    "noFallthroughCasesInSwitch": true,    /* Report errors for fallthrough cases in switch statement. */
-    "resolveJsonModule": true,             /* Include modules imported with .json extension. */
-    "composite": true,                     /* Ensure TypeScript can determine where to find the outputs of the referenced project to compile project. */
-    "incremental": true,                   /* Enable incremental compilation by reading/writing information from prior compilations to a file on disk. */
-  }
+  "extends": "../../tsconfig-base",
```
eladb: nice!

eladb: Just to make sure, are we excluding tsconfig.json files in .npmignore?

packages/jsii-pacmak/lib/targets/dotnet.ts
```diff
@@ -124,19 +124,19 @@ export class DotnetBuilder implements TargetBuilder {
 
     // If dotnet-jsonmodel is checked-out and we can find a local repository, add it to the list.
     try {
-      /* eslint-disable @typescript-eslint/no-var-requires */
+      /* eslint-disable @typescript-eslint/no-var-requires,import/no-extraneous-dependencies */
```
eladb: is this intentional? why not model it as a dep? what happens at runtime? probably worth a comment explaining why this is ok

packages/jsii-rosetta/bin/jsii-rosetta.ts
```diff
@@ -10,7 +10,7 @@ import { readTablet as readTablet } from '../lib/commands/read';
 import { translateMarkdown } from '../lib/commands/convert';
 import { File, printDiagnostics, isErrorDiagnostic } from '../lib/util';
 
-async function main() {
```
eladb: what's up here? why is this not async anymore?

packages/codemaker/tsconfig.json
RomainMuller: Good call! There were a handful of places where it wasn't ignored (although those are private packages, so it wouldn't matter well much).

packages/jsii-pacmak/lib/targets/dotnet.ts
RomainMuller: The comment is literally on line 125... This is a "shadow-dep" so this can work with a local-build when working within the mono-repo. If we modeled that dependency, we'd be modeling a dependency on a _private_ package.

On the worst case, this would be a `devDependency`, which would still require explicit opt-in!

packages/jsii-rosetta/bin/jsii-rosetta.ts
RomainMuller: It does not return a `Promise` so it may not be async. It returns a plain `void`.

Issue #1073, opened 2019-11-27
Ability to silence JSII warnings selectively


RomainMuller:
### Affected Languages

- [X] `TypeScript` or `Javascript`
- [ ] `Python`
- [ ] `Java`
- [ ] .NET (`C#`, `F#`, ...)

### Proposed Solution
The `jsii` compilation emits warnings when language keywords are used for identifiers in TypeScript code, as this may result in sub-par code being generated in other languages. In certain cases, those warnings were introduced *after* the APIs were declared stable, so they cannot be addressed anymore without incurring an undesirable major version change.

The proposal is to allow user code to silence the warning on select offenses, noting that they have intentionally decided to leave it this way.

PR #1076, opened 2019-11-27
feat(jsii): switch to disable reserved words warnings

eladb:
By default, jsii issues a warning when it encounters a symbol that uses a word that is reserved in one of the supported languages. Practically this creates a tremendous amount of noise for large existing projects.

#1073 proposes to add support for a warning exclusion file, which is probably a better long term solution, but in the meantime, this switch allows opting-out of reserved words warnings in order to save the planet.

Tested by adding `--disable-reserved-words-warnings` to `jsii-calc` (which uses many reserved words already).

---


nija-at:
When would such a warning actually in an issue?

packages/jsii/lib/assembler.ts
```diff
@@ -22,13 +22,22 @@ const sortJson = require('sort-json');
 
 const LOG = log4js.getLogger('jsii/assembler');
 
+export interface AssemblerOptions {
+  /**
+   * Do not emit warnings for reserved words.
+   * @default false (warnings are emitted)
+   */
+  readonly reservedWordsWarningsDisabled?: boolean;
```
nija-at: Very wordy. Could shrink it down to `reservedWordWarns` which defaults to `true` or something like that.

RomainMuller: My (not so strong, but still) opinion on this subject is that optional `boolean` should default to `false`, not `true`. It's bugs waiting to happen otherwise as `null` and `undefined` are both falsey.

packages/jsii-calc/package.json
```diff
@@ -25,7 +25,7 @@
   "main": "lib/index.js",
   "types": "lib/index.d.ts",
   "scripts": {
-    "build": "jsii && jsii-rosetta --compile",
+    "build": "jsii --disable-reserved-words-warnings && jsii-rosetta --compile",
```
RomainMuller: This would be more idiomatic if it was `--no-reserved-words-warnings`. I'm also afraid that this is not going to scale super well in the future.

Maybe it'd be smoother if we had a `--silence-warnings=<comma-separated-list-of-types>`, e.g: `--silence-warnings=reserved-word`.

packages/jsii/lib/assembler.ts
eladb: I don't think optional booleans must default to `false` (especially given the double negative potential). In this specific case, however, since this is somewhat of an "escape hatch" I decided to be a bit less ergonomic (hence the long name and the false default).

packages/jsii-calc/package.json
eladb: I don't want to over engineer this. I will let you do that as part of #1073. This is a temporary quick fix.

eladb:
@nija-at wrote:
> When would such a warning actually create an issue, and how would we know?

Since we have thousands of these warnings right now in the CDK (and jsii), it teaches developers to ignore warnings. We should always operate in "treat warnings as errors" mode.

packages/jsii-calc/package.json
RomainMuller: As soon as it's part of the CLI signature, it's forever... So I don't care so much if you get the "over engineered" argument and then internally map it to the boolean signal, but I don't want to bring in new CLI arguments that we'll throw away later.

eladb: we can always deprecate (and also reminding you that jsii is pre 1.0)

eladb: But ok, I'll change this `--silence-warnings` if you feel strongly about it

RomainMuller: Please :)

nija-at:
> > When would such a warning actually create an issue, and how would we know?
>
> Since we have thousands of these warnings right now in the CDK (and jsii), it teaches developers to ignore warnings. We should always operate in "treat warnings as errors" mode.

Clarifying the intention behind my question - If there are hundreds and there are no problems, is this a legit warning? When would such a warning actually bite us or the customer?

packages/jsii/lib/assembler.ts
nija-at: I believe you're going to apply this across the CDK, no? Which means this "escape hatch" is going to be used everywhere.

My opinion on this is not strong enough to prevent approval of this PR, but pointing out that this wordy flag is going to be everywhere.

eladb: "everywhere" in the CDK v1.0. When we line up v2.0 we should re-enable and fix all the warnings.

RomainMuller:
@nija-at  - the warning is here to prevent adding more identifies that will be awkward to non-TS users. We can usually work around those issues in code generation, but this will often be at the expense of ergonomics... For example, if you call something `namespace`, you'll force C# users to consistently type `@namespace` instead...

nija-at:
Thanks @RomainMuller - makes sense.

This sounds like bad customer experience we should prevent for new APIs and API parameters. We should add the feature to exclude existing 'warnings' and change their behaviour to be an error (that fails the build). I would even say that this is time critical, to raise our bar on customer experience.

Silencing these warnings for all CDK(v1) is only going to delay this work and allow more of these to-be-avoided identifiers to slip through, isn't it?

packages/jsii/lib/warnings.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Indicates which warnings are currently enabled. By default all warnings are
+ * enabled, and can be silenced through the --silence-warning option.
+ */
+export const enabledWarnings = {
```
RomainMuller:
```suggestion
export const enabledWarnings: { [name]: boolean } = {
```

packages/jsii/bin/jsii.ts
```diff
@@ -38,11 +46,20 @@ import { VERSION } from '../lib/version';
 
   const projectInfo = await loadProjectInfo(projectRoot, { fixPeerDependencies: argv['fix-peer-dependencies'] });
 
+  // disable all silenced warnings
+  for (const key of argv['silence-warnings']) {
+    if (!(key in enabledWarnings)) {
+      throw new Error(`Unknown warning type ${key}. Must be one of: ${warningTypes}`);
+    }
+
+    (enabledWarnings as any)[key] = false;
```
RomainMuller: This `as any` shouldn't be necessary (see suggestion later in the review).

packages/jsii/bin/jsii.ts
```diff
@@ -38,11 +46,20 @@ import { VERSION } from '../lib/version';
 
   const projectInfo = await loadProjectInfo(projectRoot, { fixPeerDependencies: argv['fix-peer-dependencies'] });
 
+  // disable all silenced warnings
+  for (const key of argv['silence-warnings']) {
+    if (!(key in enabledWarnings)) {
+      throw new Error(`Unknown warning type ${key}. Must be one of: ${warningTypes}`);
```
RomainMuller: 👍🏻

packages/jsii/bin/jsii.ts
eladb: Fixed

PR #1078, opened 2019-11-28
Handle Python Dependency Versions Correctly

### LINKED ISSUE #676: Python dependency versions not handled correctly

MrArnoldPalmer:
Change logic of python dependency version ranges. Parse semver number either as specific version or range and output to setup.py.

Fix #676

---

### DIFF
--- a/packages/jsii-pacmak/lib/targets/python.ts
+++ b/packages/jsii-pacmak/lib/targets/python.ts
@@ -1,5 +1,5 @@
-import path = require('path');
-
+import * as path from 'path';
+import * as semver from 'semver';
 import { CodeMaker, toSnakeCase } from 'codemaker';
 import * as escapeStringRegexp from 'escape-string-regexp';
 import * as reflect from 'jsii-reflect';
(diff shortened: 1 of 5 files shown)


MrArnoldPalmer:
Some questions for reviewers:
1. Is throwing an error from the `write` method the best thing to do when an unknown version string is found?
2. Is diff testing enough coverage here or should we separate this into some free floating funciton and unit test separately.
3. jsii-calc currently declares "^0.28.0" for base dependencies but the .jsii output is 0.28.0. Should we preserve semver range shorthand in jsii output for code generation to leverage?

If we want to keep limiting the version numbers in jsii to not be ranges we can remove one of the branches here.

packages/jsii-pacmak/test/expected.jsii-calc-base/python/setup.py
```diff
@@ -32,7 +32,7 @@
     "install_requires": [
         "jsii~=0.20.8",
         "publication>=0.0.3",
-        "scope.jsii-calc-base-of-base~=0.20.8"
+        "scope.jsii-calc-base-of-base==0.20.8"
```
eladb: The original dependency is a caret dependency (`^`), does `==` means the same thing in Python?

https://github.com/aws/jsii/blob/8fc81ca1b5623ca5a8d51a0aa62a125b0681ef94/packages/jsii-calc-base/package.json#L32

packages/jsii-pacmak/lib/targets/python.ts
```diff
@@ -1161,17 +1161,21 @@ class Package {
...
+
+      // Convert parsed version info to expected format
+      if (parsedVersion) {
+        versionSpecifier = `==${version}`;
+      } else if (parsedVersionRange) {
+        versionSpecifier = parsedVersionRange.replace(' ', ', ');
       } else {
-        versionSpecifier = `~=${versionParts.slice(0, 2).join('.')},>=${versionParts.slice(0, 3).join('.')}`;
+        throw new Error(`Unexpected version format '${version}' for dependency: 'depName'`);
       }
```
RomainMuller: This feels like it should have tests. Also - similar work should be done on Java and .NET dependencies.

packages/jsii-pacmak/test/expected.jsii-calc-base/python/setup.py
RomainMuller: Yeah that looks suspicious; but I reckon because the `jsii` compiler forwards the "resolved" (not "declared") version. That needs to be fixed as well (possibly in a separate PR, but then we probably shouldn't merge this without that).

MrArnoldPalmer:
So seems like solution here is:

- [ ] Change output in .jsii to be the declared version and not the resolved one
- [ ] Check that the resolved version overlaps with the declared version range
- [ ] Output correct version range formats to target language dependency manifests
  - [ ] Python
  - [ ] Dotnet
  - [ ] Java
- [ ] Test coverage

Will reopen when this is ready for review

PR #1128, opened 2019-12-16
fix(java,dotnet): abstract properties have concrete implementations

### LINKED ISSUE #240: Abstract properties are generated in abstract classes.
The .NET generator will generate jsii properties marked abstract as concrete properties with implementations. This does not break anything - superclasses will override these properties with the same body. However, it is unnecessary.

Example:
```json
"jsii-calc.AbstractClassBase": {
      "abstract": true,
      "assembly": "jsii-calc",
      "fqn": "jsii-calc.AbstractClassBase",
      "initializer": {
        "initializer": true
      },
      "kind": "class",
      "name": "AbstractClassBase",
      "properties": [
        {
          "abstract": true,
          "immutable": true,
          "name": "abstractProperty",
          "type": {
            "primitive": "string"
          }
        }
      ]
    },
```

Generates:
```csharp
[JsiiClass(typeof(AbstractClassBase), "jsii-calc.AbstractClassBase", "[]")]
public abstract class AbstractClassBase : DeputyBase
{
    protected AbstractClassBase(): base(new DeputyProps(new object[]{}))
    {
    }

    protected AbstractClassBase(ByRefValue reference): base(reference)
    {
    }

    protected AbstractClassBase(DeputyProps props): base(props)
    {
    }

    [JsiiProperty("abstractProperty", "{\"primitive\":\"string\"}")]
    public virtual string AbstractProperty
    {
       get => GetInstanceProperty<string>();
    }
 }
```

### LINKED ISSUE #1011: Java/DotNet: abstract members are not marked as such on their .NET and Java representations.
## :bug: Bug Report

In this code:

https://github.com/rix0rrr/jsii-test-drive/blob/master/java/extends-abstract-class/src/main/java/com/myorg/MyCode.java﻿

The subclass of `Code` is not complete. It is missing an implementation of `public abstract readonly isInline: boolean`.

This class should have been needed to be marked abstract, but it it isn't.

### Impact to CDK
* Abstract members have an implementation in the generated **.NET** and **Java** code, allowing	classes extending those to omit implementation of some (or all) of the abstract members. While the resulting code compiles, it may fail at run-time due to the missing implementations.
* _Workaround_: when extending abstract classes, consult the [TypeScript API reference] to identify all abstract members of the extended type, and make sure to implement all of those. The members will be made abstract in a subsequent release of the CDK for Java and .NET, <u>**which will break invalid implementations**</u>.

[TypeScript API Reference]: https://docs.aws.amazon.com/cdk/api/latest/typescript/api/index.html

RomainMuller:
The generated code for abstract properties in Java and C# included fully
concrete implementations, instead of an abstract declaration. This made
it possible to subclass those types without actually implementing those
members, resulting in invalid code.

This changes the code generation to actually emit the `abstract` keyword
and not generate a full concrete implementation.

Fixes #240
Fixes #1011

---


RomainMuller:
This is currently marked as *draft* because I still have to double-check whether existing test coverage is sufficient to ensure this "works as-is" or not.

RomainMuller:
Added tests (and fixed the issues they uncovered in the .NET runtime...)

packages/jsii-pacmak/test/expected.jsii-calc/dotnet/Amazon.JSII.Tests.CalculatorPackageId/Amazon/JSII/Tests/CalculatorNamespace/AbstractSuite.cs
```diff
@@ -0,0 +1,60 @@
...
+        }
+
+        /// <summary></summary>
+        /// <param name="str"></param>
+        /// <remarks>
+        /// stability: Experimental
+        /// </remarks>
+        [JsiiMethod(name: "someMethod", returnsJson: "{\"type\":{\"primitive\":\"string\"}}", parametersJson: "[{\"name\":\"str\",\"type\":{\"primitive\":\"string\"}}]")]
+        protected abstract string SomeMethod(string str);
+
```
rix0rrr: Extra newline here.

PR #1133, opened 2019-12-17
docs(spec): initial pass on a jsii specification

RomainMuller:
## Reviewing Guide

This Pull Request introduces a collection of documents under `docs/specification/` that are effectively an RFC for the *jsii* technology.

### Foundational Elements

Since the specified behavior is due to evolve over time, the essential elements to nail down are the following:

1. Design Tenets (in `1-introduction.md`)
1. Updating the specification (in `1-introduction.md`)
1. Standard Compliance Suite (in `4-standard-compliance-suite.md`)
   + This is the core of the compliance & coverage plan
1. New language intake (in `5-new-language-intake.md`)

### Current Specification

Once the foundational elements are agreed upon (at least by the core team), the rest of the specification will be open for a review period (like any other RFC), before it is finally merged.

#### Why several documents and not just one?

There are presently several distinct documents, but this is merely a way to make the PR easier to assess in fragments, and the documents may be merged into a single `specification.md` document before merging.

In particular, GitHub's PR review UI has the "viewed" toggle that allows one to stop reviewing and come back to it later without having lost state of where they left off. This Feld particularly appealing, especially as the initial version of this document is large.

#### What are the block quotes for?

The documents attempt to specify the _current_ behavior of the *jsii* components and uses block-quotes introduced by a specific emoji:

- :construction: introduces remarks about behaviors that are not yet conform to the specification, where the current implementation is not homogenous, etc...
- :question: introduces remarks about open questions, or elements of the specification that may need re-designing.

---

### DIFF
--- a/docs/specifications/1-introduction.md
+++ b/docs/specifications/1-introduction.md
@@ -1,0 +1,205 @@
+# Introduction
+
+This document provides a high level overview of *jsii*, starting with it's
+design tenets. It introduces the concepts and components that compose *jsii*.
+
+
+## Updating the Specification
+
+### Introduction
+
+The *jsii* specification follows the guiding principles of an RFC. It is a
+living document that describes the current understanding of how the various
+[components](#components) of *jsii* are operating, as well as the approaches
+used to ensure consistent behavior across the various supported languages.
+
+The document is mastered with the *jsii* codebase, making it easy to determine
+what specification was in place at the time of a give *jsii* release (by ways of
+referring to the `vX.Y.Z` git tag). A single version of the specification is
+considered **active** at any given time: the version of the specification that
+is represented on the `HEAD` commit of the `master` branch of the [`aws/jsii`]
+repository. The **active** specification must be the base version for any update
+proposal.
+
+[`aws/jsii`]: https://github.com/aws/jsii
+
+The process to update the specification is intended to be as lightweight as
+possible, while ensuring sufficient conversation takes place before implementing
+significant (and breaking) changes. Since the process to update the
+specification is part of the specification itself, it is amenable to be changed
+following the process described in the currently **active** specification.
+
+### Process
+
+While the general process for updating the specification is to create a GitHub
+pull request against the [`aws/jsii`] repository, the exact requirements for
+what should be included in the pull request vary depending on the type of update
+that is proposed:
+
+- [:warning: Changing Behavior](#new-behavior) describes the process to be
+  followed when introducing changes to the behavior of any component of *jsii*:
+  new features, breaking changes to existing features, ...
+- [:mag: Addressing Gaps](#addressing-gaps) is the process used for adding
+  specification around existing but un-specified behavior.
+- [:thumbsup: Trivial Changes](#trivial) explains how to propose changes that
+  improve the specification without changing it's meaning.
+
+#### <a id="new-behavior"/> :warning: Changing Behavior
+
+If the change is **not backwards compatible** (it is a breaking change to an
+existing feature, or it is a new feature that requires all runtime libraries
+implement support immediately), a new RFC should be created in the
+[`awslabs/aws-cdk-rfcs`] repository, following the [RFC Process]. This ensures
+enough time is spent considering alternatives to breaking changes, and to create
+consensus that the change is desirable before time is spent implementing it.
+
+[`awslabs/aws-cdk-rfcs`]: https://github.com/awslabs/aws-cdk-rfcs
+[RFC Process]: https://github.com/aws/aws-cdk-rfcs#what-the-process-is
+
+> While going through the RFC process upfront is **strongly recommended**,
+> contributors may choose not to file an RFC for a behavior change. In this case
+> however, any core maintainer may decide that an RFC is required and block the
+> contribution until the RFC process has been followed.
+>
+> It is worth noting that a draft pull request with proposed modifications to
+> the specification (and possibly a proof-of-concept implementation), can b
+
+When the RFC is **ready**, a GitHub pull request is created that must contain:
+
+- Relevant additions or modifications to the specification documents
+- Relevant additions or modifications to the compliance suite
+- Implementation of the new behavior, including new or updated tests in all the
+  language bindings
+
+The pull request's body must reference the RFC if there has been one, and
+otherwise must include all discussion necessary to explain the reasoning behind
+the proposal (including alternatives considered, risks, ...).
+
+#### <a id="addressing-gaps"/> :mag: Addressing Gaps
+
+Proposal that increase the specification's coverage (desribing behavior that
+already exists) are handled as GitHub pull request that must contain the
+following elements:
+
+- Relevant additions to the specification documents
+- New compliance test(s) that enshrine the described behavior
+- Implementation of the new compliance test(s) for all *Generally Available*
+  language bindings
+
+The pull request body should provide pointers to any and all elements that can
+be used to verify that the behavior that is described is indeed what is
+currently implemented.
+
+#### <a id="trivial"/> :thumbsup: Trivial Changes
+
+Proposal of trivial changes, such as correcting typos in the document, or
+re-phrasing elements of the specification without altering the meaning
+(typically to improve clarity) are handled in a simple GitHub pull request.
+
+
+## Design Tenets (unless you know better ones)
+
+* *jsii* APIs strive to feel idiomatic in all supported languages.
+* *jsii* applications behave identically regardless of the language they are
+  written in. It favors correctness over performance.
+* *jsii* **does not** attempt to support all TypeScript idioms (many features of
+  TypeScript cannot be expressed in some target languages).
+  * Unsupported idioms will cause a compile-time error to be emitted.
+  * When prohibiting an idiom, *jsii* strives to provide an error message that
+    gives the user insight into why the pattern cannot be supported.
+* *jsii* does not force API design opinions on the developer:
+  * Reserved names are limited to a minimum.
+  * TypeScript API design patterns that are known to result in poor developer
+    experience when represented in other languages will cause warnings to be
+    issued, but the developer is ultimately entitled to decide whether they want
+    to take or leave the advice.
+* *jsii* produces artifacts compatible with idiomatic tools whenever possible:
+  * Generated libraries can be easily published to the "standard" package
+    repository for the language.
+  * Standard tools can be used to work with the generated libraries, and do not
+    require any special configuration.
+
+
+## Annotations
+
+Annotations are present in the *jsii* specification to provide additional
+information to the reader that is non-normative. Those take the form of
+block-quotes that use the following chart:
+
+- > :construction: Is used to annotate parts of the specification that are known
+  > to be partially or incorrectly implemented in the current releases. Those
+  > are known issues in the current implementation that will be addressed in the
+  > future.
+
+- > :question: Is used to annotate open questions. They are typically in parts
+  > of the specification that is likely to change in future releases, and that
+  > may be good candidates for introducing RFCs.
+
+- > :warning: Is used to draw the reader's attention on specific points. They
+  > are used primarily to help the reader identify areas of the specification
+  > that, when incorrectly implemented, may result in hard-to-troubleshoot bugs;
+  > or to identify behavior that is intentionally undefined.
+
+- > :information_source: Is used to provide additional context which may not be
+  > obvious to the reader. They typically contain trivia that can help the
+  > reader understand the motivation for certain behaviors that are not always
+  > intuitive.
+
+## Concepts
+
+*jsii* allows developers to author code once in **TypeScript**, while allowing
+use in a variety of other programming languages (including **C#**, **Java** and
+**Python**).
+
+### Assemblies
+
+The *jsii Assembly* document contains a specific representation of the API
+exported by the **TypeScript** module. Similar to a header file in the **C++**
+world, it contains only information about the signatures of APIs (type names
+with method and property signatures, associated documentation elements, ...) and
+no implementation.
+
+The *npm* package produced as a result of compiling the **TypeScript** source
+remains the source of truth with respects to implementation of the API.
+
+### Host & Kernel
+
+The [*jsii* runtime architecture] defines two processes:
+
+1. The *host* process runs the users' code native environment (a Java virtual
+   machine, the .NET Runtime, ...).
+2. The *kernel* process hosts the **JavaScript** code from the standard *npm*
+   package.
+
+The *host* process is responsible for starting the *kernel* process as needed. A
+designated *host runtime library* provides helper functions that will perform
+the necessary initialization when needed, so the *host* app does not need to
+include any special boilerplate code.
+
+The two processes exchange messages over a designated communication channel (for
+example, using pipes), using a *kernel API* that is standardized in the *jsii
+specification*.
+
+[*jsii* runtime architecture]: ../runtime-architecture.md
+
+
+## Components
+
+Several tools are involved in making this possible:
+
+* [`jsii`] is a modified **TypeScript** compiler. In addition to generating
+  **JavaScript** code from the source, it produces a *jsii Assembly* document.
+* [`jsii-pacmak`] generates language bindings from a package compiled using
+  `jsii`. It generates code in *host* languages that expose the API declared in
+  the *jsii Assembly* document.
+* *Host runtime libraries* centralize features used by code generated by
+  [`jsii-pacmak`], such as primitives to interact with the *kernel* process, so
+  that this code does not need to be duplicated in every generated module.
+* [`@jsii/kernel`] (and [`@jsii/runtime`]) provide the functionality exposed by
+  the *kernel* process, and allow the *host* code to seamlessly interact with
+  the **JavaScript** implementation.
+
+[`jsii`]: ../../../packages/jsii
+[`jsii-pacmak`]: ../../../packages/jsii-pacmak
+[`@jsii/kernel`]: ../../../packages/@jsii/kernel
+[`@jsii/runtime`]: ../../../packages/@jsii/runtime
--- a/docs/specifications/2-type-system.md
+++ b/docs/specifications/2-type-system.md
@@ -1,0 +1,316 @@
+# The *jsii* type system
+
+## Preamble
+The base language for authoring *jsii* libraries for re-use from other languages
+is **TypeScript**, which compiles to **JavaScript**. Consequently, the base type
+system that *jsii* sources from is that of **TypeScript**.
+
+When used from another language than **TypeScript** or **JavaScript**, *jsii*
+libraries are running the **JavaScript** code in a child *node* process, and
+data is exchanged using **JSON**-based protocol.
+
+This document describes how **TypeScript** types map into the *jsii* type
+system.
+
+The API represented by the *jsii* assembly only covers declarations that are
+exported from the main file in the **TypeScript** project (as specified in the
+`package.json` file by the `types` attribute). Restrictions described in this
+document only apply to such declarations, the rest of the module can leverage
+any **TypeScript** feature.
+
+## Basic Types
+
+### Introduction
+In order to build useful programs, the simplest units of data need to be
+modeled: booleans, numbers, strings, etc... Those basic building blocks are the
+foundations on which APIs stand. *jsii* supports much of the same types that
+**TypeScript** and **JavaScript** support, although with notable differences.
+
+### Boolean
+The *jsii* type system mirrors **TypeScript**'s `boolean`, which is the simplest
+primitive data types, with only two supported values: `true` and `false`.
+
+### Number
+The *jsii* type system mirrors **TypeScript**'s `number`. All numbers are
+floating point values.
+
+### String
+The *jsii* type system mirrors **TypeScript**'s `string`. Strings are used to
+represent textual data.
+
+### List
+**TypeScript** arrays (`Array<T>`, `T[]`, `ReadonlyArray<T>` and `readonly T[]`)
+are represented as *lists* in the *jsii* type model. Lists are shared between
+the *node* process and the host process by-value, meaning a copy of the array is
+produced each time it is passed through the process boundary.
+
+> :information_source: Items in the list may be passed by-reference (according
+> to their type's specification), in which case mutating operations performed on
+> those may be visible across the process boundary.
+
+### Enum
+As in many languages, `enum` can be used to represent a group of related
+constants. Whle **TypeScript** `enum` entries are associated with a value that
+is either a `string` or a `number`, the *jsii* type system does not allow for
+those to be down-casted to their value type (e.g: a `string`-valued `enum` entry
+cannot be directly passed into a `string` parameter).
+
+> :information_source: Unlike in certain languages such as **Java**, `enum`
+> types cannot declare new properties or methods.
+
+### Any and Unknown
+**TypeScript** defines two opaque types: `any` and `unknown` that can be used to
+represent a value of arbitary type. The difference between them is that while
+`any` is assignable to any other type, `unknown` requires a type assertion or
+explicit cast to be performed before it can be assigned.
+
+> :information_source: It is important to note that, contrary to the other types
+> in the **TypeScript** type system, `any` and `unknown` types are inherently
+> `null`-able.
+
+### Void
+As in most languages, the `void` type is used to denote a method does not return
+anything.
+
+### Null and Undefined
+**JavaScript** differentiates `undefined` and `null` values. While `undefined`
+denotes that *no value* has been set, `null` denotes an intentional signal of
+there being *no data*. Most other programming languages (particularly statically
+typed languages) however lack this distinction, and the *jsii* type model
+consequently considers `null` and `undefined` are semantically equivalent.
+
+> :information_source: Unlike certain other programming languages, such as
+> **Java**, **TypeScript** does not allow `null` (or `undefined`) values unless
+> the type signature expressedly supports that (with the exception of `any` and
+> `unknown`, which are implicitly `null`-able, as was discussed earlier).
+
+### Object
+**TypeScript**'s `object` type denotes anything that is not a *primitive* type,
+meaning anything other than a `number`, `srting`, `boolean`, `biging`, `symbol`,
+`null` or `undefined`.
+
+In the *jsii* type model, `object` indicates a block of structured data that can
+be shared by-value across the process boundary. As a consequence, they may not
+include any method.
+
+> :construction: This type is called `Json` in the current implementation.
+
+> :question: The by-value nature of `object` is problematic because
+> **TypeScript** makes no guarantee with respects to the absence of methods on
+> `object`, and properties may be dynamic.
+
+### Promises
+*jsii* supports asynchronous methods, and the **TypeScript** `Promise<T>` type
+has to be used as the result of `async` methods. Promises can only be used as
+the result type of methods, not as the type of a property or parameter.
+
+### Unsupported **TypeScript** basic types
+Due to how such types cannot be represented in many other programming languages,
+the *jsii* type model does not support the following **TypeScript** entities:
+- Tuples, a group of arbitrarily-typed values, often used as the result type for
+  multi-valued functions.
+- The `never` type, which is used as the return type of functions that will not
+  yield control back to their invoker (infinite loops, `process.exit()`, ...).
+- `bigint` and `symbol` don't have equivalents in many other programming
+  languages and are generally of limited value in API design.
+
+
+## Complex Types
+The goal of *jsii* is to enable cross-language re-use of class libraries.
+**TypeScript** enables representing classic object-oriented concepts such as
+*classes*, *interfaces* and such. The *jsii* type system supports some
+additional nuances on top of those, to better represent **TypeScript** and
+**JavaScript** idioms in a way that enables generating convenient APIs in other
+languages.
+
+### Classes
+Exported **TypeScript** classes are represented in the *jsii* type system, with
+the following restrictions from plain **TypeScript**:
+- Methods overloads are not supported.
+- Overridden methods or properties must retain the exact same type signature as
+  the one declared in a parent type. The **jsii** type system strictly enforces
+  the [Liskov substitution principle].
+
+[Liskov substitution principle]: https://en.wikipedia.org/wiki/Liskov_substitution_principle
+
+### Interfaces & Structs
+Exported **TypeScript** interfaces are interpreted as one of two entities in the
+*jsii* type system:
+- If the `interface` name is prefixed with an `I` (e.g: `ISomething`), it is
+  interpreted as a *behavioral interface*.
+- Otherwise (e.g: `Something`), it is interpreted as a *struct*.
+
+#### Behavioral Interfaces
+*Behavioral interfaces* are the usual object-oriented interface: they can extend
+other *behavioral interfaces*, and can be extended by *classes*. They may
+however not extend *structs*.
+
+#### Structs
+*Structs* are used to model the **JavaScript** idiom of receiving options as an
+object literal passed as the last parameter of a function. They are a formal
+description of a bag of properties, and are not meant to be implemented by other
+types. Since those types are used as inputs, they can be handled as pure-data,
+immutable objects, and the following restrictions apply:
+- A *struct* cannot declare any *method*: they must be kept behavior-free.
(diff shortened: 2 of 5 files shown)


docs/specifications/1-tenets.md
```diff
@@ -0,0 +1,21 @@
...
+* *jsii* applications behave identically regardless of the language they are
+  written in. It favors correctness over performance.
+* *jsii* **does not** attempt to support all TypeScript idioms (many features of
+  TypeScript cannot be expressed in some target languages).
+  * When prohibiting an idiom, *jsii* strives to provide an error message that
+    gives the user insight into why the pattern cannot be supported.
+* *jsii* guides (instead of forcing) library authors towards better APIs:
+  * Reserved names are limited to a minimum.
+  * Warn about patterns or names that are known to be awkward in other
+    programming languages, instead of blocking them.
```
eladb: I don't think that should be a jsii tenent - jsii should allow users to define any API they desire as long as it uses the supported features.

docs/specifications/4-new-language-intake.md
```diff
@@ -0,0 +1,7 @@
+---
+version: 1.0.0
+---
+# New Language Intake
```
eladb: why is this called "intake"?

docs/specifications/1-tenets.md
```diff
@@ -0,0 +1,21 @@
+---
```
eladb: I would recommend writing everything in a single page (at least initially) to keep it simple.

docs/specifications/1-tenets.md
RomainMuller: The tenet I want to propose is something along the lines of "don't _force_ opinions on how APIs should be designed", which implies minimizing the things *jsii* prohibits.

docs/specifications/4-new-language-intake.md
RomainMuller: Why not?

docs/specifications/1-tenets.md
RomainMuller: I personally have a preference for smaller files in higher quantities. But not opposed to being pragmatic. The organization of stuff will *definitely* change as the contents get fleshed out... right now I'm mostly trying to manage my own cognitive load / mitigate blank page syndrome / force myself to minimize "things you have to know for this to make sense" (aka - I would like all sections to make sense on their own as much as possible).

RomainMuller:
This is now ready for a first round of reviews. The current version is focused essentially on *what we have* at this point.

Note: the existing compliance suite needs to be retro-fitted into the document; this will happen once the format has been agreed upon to reduce the risk of expensive re-work.

RomainMuller:
Realized I forgot to add a section about the serialization behavior across the process boundary. I'll be adding this in the coming few days.

docs/specifications/2-type-system.md
```diff
@@ -0,0 +1,307 @@
...
+are serialized as JSON documents. They can be passed by value or by reference,
+depending on the type of the value as well as the declared type of the transfer
+point (method return type, property type, argument type, ...).
+
+The table below describes the serialization behavior applied for each possible
+*declared* type (rows) for a value of a given dynamic type (columns). The :x:
+sign expresses cases that are illegal and should cause immediate failure. The
+term *primitive* encompasses `boolean`, `string`, and `number`.
+
+&nbsp;      | `undefined` | `Date`      | *primitive* | `Array`     | *instance*  | `object`
```
RomainMuller: This is a little confusing - maybe break it down to one table per *jsii* type?

RomainMuller:
@MrArnoldPalmer mentioned also a need to add an intro block to lay out the grounds of what is jsii, what are the components, etc...

RomainMuller:
Need to get rid of all `Kernel` references, and simply refer to the JS runtime component.

docs/specifications/1-introduction.md
```diff
@@ -0,0 +1,172 @@
...
+currently implemented.
+
+#### <a id="trivial"/> :thumbsup: Trivial Changes
+
+Proposal of trivial changes, such as correcting typos in the document, or
+re-phrasing elements of the specification without altering the meaning
+(typically to improve clarity) are handled in a simple GitHub pull request.
+
+
+## Design Tenets (unless you know better ones)
```
eladb: LGTM 👍

docs/specifications/1-introduction.md
```diff
@@ -0,0 +1,172 @@
...
+[components](#components) of *jsii* are operating, as well as the approaches
+used to ensure consistent behavior across the various supported languages.
+
+The document is mastered with the *jsii* codebase, making it easy to determine
+what specification was in place at the time of a give *jsii* release (by ways of
+referring to the `vX.Y.Z` git tag). A single version of the specification is
+considered **active** at any given time: the version of the specification that
+is represented on the `HEAD` commit of the `master` branch of the [`aws/jsii`]
+repository. The **active** specification must be the base version for any update
+proposal.
```
eladb: This implies (and I like it) that the spec describes the *current* state of the project.

(we can define some documentation convention to annotate a paragraph/section and indicate what would be the desired behavior/spec that will be implemented in the future, maybe "not implemented yet").

docs/specifications/4-standard-compliance-suite.md
```diff
@@ -0,0 +1,166 @@
...
+the test without having to search for any additional information. Prefer a long,
+unambiguous description to a terse one that could be subject to interpretation.
+
+Attribute      | Value
+---------------|----------------------------------------------------------------
+Introduced     | 1.2.3
+Related Issues | #123, aws/aws-cdk#456
+
+##### Reference Implementation
+```ts
```
rix0rrr: This is not entirely clear to me. Looks like this consists of 2 parts:

* the parts that have to be present in TypeScript-land for the test to run (loaded into the kernel)
*  an assertion on the behavior of that code as it is being exercised, which needs to be rewritten into every client language.

Did I get that right?

I imagine there will be a lot of duplication between the kernel code of various tests. Is there going to be a single set of classes that cover everything, or are new classes going to be written for every individual test?

docs/specifications/1-introduction.md
```diff
@@ -0,0 +1,172 @@
...
+
+- [:warning: Changing Behavior](#new-behavior) describes the process to be
+  followed when introducing changes to the behavior of any component of *jsii*:
+  new features, breaking changes to existing features, ...
+- [:mag: Addressing Gaps](#addressing-gaps) is the process used for adding
+  specification around existing but un-specified behavior.
+- [:thumbsup: Trivial Changes](#trivial) explains how to propose changes that
+  improve the specification without changing it's meaning.
+
+#### <a id="new-behavior"/> :warning: Changing Behavior
```
eladb: Why isn't this called "Breaking Change"?

docs/specifications/1-introduction.md
```diff
@@ -0,0 +1,172 @@
...
+
+
+## Design Tenets (unless you know better ones)
+
+* *jsii* APIs strive to feel idiomatic in all supported languages.
+* *jsii* applications behave identically regardless of the language they are
+  written in. It favors correctness over performance.
+* *jsii* **does not** attempt to support all TypeScript idioms (many features of
+  TypeScript cannot be expressed in some target languages).
+  * When prohibiting an idiom, *jsii* strives to provide an error message that
```
nija-at: Design tenet should include 'unsupported idioms are prohibited' (i.e., will not silently ignore)

docs/specifications/1-introduction.md
```diff
@@ -0,0 +1,172 @@
...
+
+If the change is **not backwards compatible** (it is a breaking change to an
+existing feature, or it is a new feature that requires all runtime libraries
+implement support immediately), a new RFC should be created in the
+[`awslabs/aws-cdk-rfcs`] repository, following the [RFC Process]. This ensures
+enough time is spent considering alternatives to breaking changes, and to create
+consensus that the change is desirable before time is spent implementing it.
+
+[`awslabs/aws-cdk-rfcs`]: https://github.com/awslabs/aws-cdk-rfcs
+[RFC Process]: https://github.com/aws/aws-cdk-rfcs#what-the-process-is
```
eladb: Feels like a PR with a diff on the spec can go a long way for describing such changes. Maybe an RFC doc in that case would be accompanied with a PR to the jsii spec instead of having to describe everything twice?

docs/specifications/4-standard-compliance-suite.md
```diff
@@ -0,0 +1,166 @@
...
+##### Reference Kernel Messaging
+```
+< { "hello": "@jsii/runtime@1.2.3" }
+> { "load": { "name": "test-case-001", "version": "1.2.3", "tarball": "/tmp/jsii-kernel-test/lib.tgz" } }
+< { "assembly": "test-case-001", "types": 3 }
+```
+````
+
+## Compliance
+In order to be able to assert compliance of language binding libraries to the
```
rix0rrr: I like the idea of a standardized report, but why not make it some flavor of XUnit reporting XML? That has been semi-standardized and can be generated by nearly every unit testing tool easily enough.

Obviates the need for another XUnit -> JSON conversion tool.

docs/specifications/1-introduction.md
```diff
@@ -0,0 +1,172 @@
...
+* *jsii* APIs strive to feel idiomatic in all supported languages.
+* *jsii* applications behave identically regardless of the language they are
+  written in. It favors correctness over performance.
+* *jsii* **does not** attempt to support all TypeScript idioms (many features of
+  TypeScript cannot be expressed in some target languages).
+  * When prohibiting an idiom, *jsii* strives to provide an error message that
+    gives the user insight into why the pattern cannot be supported.
+* *jsii* does not force API design opinions on the developer:
+  * Reserved names are limited to a minimum.
+  * TypeScript API design patterns that will not transpose well in other
```
nija-at: What is the meaning of 'transpose well'? Tenets should make it clear when there will be an error and when there will be a warning.

docs/specifications/4-standard-compliance-suite.md
```diff
@@ -0,0 +1,166 @@
...
+> the exact Object IDs issued for created instances?
+
+#### Template
+
+Below is the template markdown to copy-paste when introducing a new test case in
+the compliance suite. New tests should always be added at the very end of the
+category they belong to, right after the last test in said category.
+
+````md
+### Test Category
```
rix0rrr: Categories might need symbolic names, will help later on in the report.

docs/specifications/1-introduction.md
RomainMuller: Yeah I should add a note about the block quotes I have been using to denotes that, and enshrine it as "the way to go" (I find it works nicely in the document flow)

docs/specifications/4-standard-compliance-suite.md
```diff
@@ -0,0 +1,166 @@
+# Standard Compliance Suite
+
+## Format
+In order to assert whether a new runtime implementation is correct, a formal
+compliance test suite is defined, that all new language runtimes must fully
```
eladb:
```suggestion
compliance test suite is defined, that all language runtimes must fully
```

docs/specifications/1-introduction.md
```diff
@@ -0,0 +1,172 @@
...
+    languages may cause a warning to be issued, but the developer is ultimately
+    entitled to decide whether they want to take or leave the advice.
+* *jsii* uses idiomatic tools whenever possible:
```
nija-at: 'uses' is ambiguous in this context. Is this for working on jsii, output of jsii or both?

docs/specifications/4-standard-compliance-suite.md
```diff
@@ -0,0 +1,166 @@
...
+  changing the tests case's semantic value) should not be included in the list.
+
+After the attributes table, a **TypeScript** block of code describes the
+canonical form of the test. It includes any type declaration that is used by the
+test (so the code example is self-contained). Assertions performed by the test
+should be written in the form of [`jest`] expectations.
+
+[`jest`]: https://jestjs.io/docs/en/getting-started
+
+Finally, another code block details the sequence of messages that should be
```
rix0rrr: Is this checked or does it mostly serve as a hint to implementors/debuggers?

docs/specifications/4-standard-compliance-suite.md
RomainMuller: That's correct - basically there needs to be:
- Type entities defined in TypeScript
- Expectation/assertions that need writing in each language.

The may end up being a lot of duplication, but I would rather keep the tests self-contained as much as possible (open to change if that turns out very inconvenient). The idea would be to code-generate the combination of all things in per-test namespaces, automatically.

docs/specifications/1-introduction.md
```diff
@@ -0,0 +1,172 @@
...
+*jsii* allows developers to author code once in **TypeScript**, while allowing
+use in a variety of other programming languages (including **C#**, **Java** and
+**Python**).
+
+### Assemblies
+
+The *jsii Assembly* document contains a specific representation of the API
+exported by the **TypeScript** module. Similar to a header file in the **C++**
+world, it contains only information about the signatures of APIs (type names
+with method and property signatures, ...) and no implementation.
```
nija-at: (jsii noob here) does it also not contain documentation for the methods and types?

docs/specifications/4-standard-compliance-suite.md
```diff
@@ -0,0 +1,166 @@
...
+  ```
+* Comments to improve readability of the trace:
+  ```
+  # Comment continues until the end of the line
+  ```
+* Blank lines can be added to logically group trace elements
+
+> :question: is there a need to support some form of a capture mechanism to
+> provision for non-deterministric results, or non-normative elements such as
+> the exact Object IDs issued for created instances?
```
eladb: object IDs are supposed to be deterministic as long as we reset state before every test (same as for kernel/runtime tests).

docs/specifications/1-introduction.md
RomainMuller: Because a new feature isn't necessarily a breaking change, but most often also should be handled through RFC (be it only to make sure the feature does not paint us into a corner).

docs/specifications/4-standard-compliance-suite.md
```diff
@@ -0,0 +1,166 @@
...
+compliance test suite is defined, that all new language runtimes must fully
+implement before they can be deemed eligible for General Availability.
+
+This document describes these tests, as well as a general approach for ensuring
+conformance requirements are met in a systematic manner.
+
+### Categories
+Test cases in the standard compliance suite are grouped by categories, which
+help implementors direct their effort in the early stages of the implementation
+of new language bindings. Each category is declared in an `H3` title (a line
```
rix0rrr: I understand your desire to be precise, but all the talk about how the document should be marked up feels like a waste of space to me. Write some examples and people will naturally start to copy/paste.

Are you planning to automatically process this MarkDown file using tools? If so, then that is something worthwhile to call out in a separate section, and you can specify the markup requirements there. Otherwise, I wouldn't bother.

docs/specifications/1-introduction.md
```diff
@@ -0,0 +1,172 @@
...
+
+The *npm* package produced as a result of compiling the **TypeScript** source
+remains the source of truth with respects to implementation of the API.
+
+### Host & Kernel
+
+The [*jsii* runtime architecture] defines two processes:
+
+1. The *host* process runs the users' code natural environment (a Java virtual
```
nija-at: s/natural/native/

docs/specifications/4-standard-compliance-suite.md
eladb: +1, I also think test case names should be symbolic because they are often used as method names

docs/specifications/4-standard-compliance-suite.md
rix0rrr: Is it going to including kernel initialization, or only the communication after the module has been loaded?

docs/specifications/1-introduction.md
```diff
@@ -0,0 +1,172 @@
...
+   machine, the .NET Runtime, ...).
+2. The *kernel* process hosts the **JavaScript** code from the standard *npm*
+   package.
+
+The *host* process is responsible for starting the *kernel* process as needed,
+and a designated *host runtime library* is provided to make this transparent to
+the user.
+
+The two processes exchange messages over a designated communication channel (for
+exmaple, using pipes), using a *kernel API* that is standardized in the *jsii
```
nija-at: typo: example

docs/specifications/4-standard-compliance-suite.md
```diff
@@ -0,0 +1,166 @@
...
+Related Issues | #123, aws/aws-cdk#456
+
+##### Reference Implementation
+```ts
+// Declare any type used by the test
+export class Foo { /* ... */ }
+
+// Instructions the test will execute, using `jest`-like expectation syntax
+expect(new Foo().bar().baz)
+  .toBeUndefined();
```
eladb: What are we doing with the typescript test? Is it executed against the kernel somehow? Do we have plans to transpile it using @rix0rrr's magic stone?

docs/specifications/1-introduction.md
RomainMuller: If the RFC happens before-hand, then the specification change can be mostly a copy of text from the RFC. But I reckon there is some value in openly stating that a PR on the jsii spec is a valid documentation element for the RFC...

docs/specifications/1-introduction.md
```diff
@@ -0,0 +1,172 @@
...
+The [*jsii* runtime architecture] defines two processes:
+
+1. The *host* process runs the users' code natural environment (a Java virtual
+   machine, the .NET Runtime, ...).
+2. The *kernel* process hosts the **JavaScript** code from the standard *npm*
+   package.
+
+The *host* process is responsible for starting the *kernel* process as needed,
+and a designated *host runtime library* is provided to make this transparent to
+the user.
```
nija-at: Not sure I follow what is made transparent to the user.

docs/specifications/4-standard-compliance-suite.md
RomainMuller: I mean if there is a standard format that satisfies the requirements I'm all for it. XML is meh to me, but if there's an XML standard, that wins. I'll investigate.

docs/specifications/1-introduction.md
```diff
@@ -0,0 +1,172 @@
...
+## Components
+
+Several tools are involved in making this possible:
+
+* [`jsii`] is a modified **TypeScript** compiler. In addition to generating
+  **JavaScript** code from the source, it produces a *jsii Assembly* document.
+* [`jsii-pacmak`] generates language bindings from a package compiled using
+  `jsii`. It generates code in *host* languages that expose the API declared in
+  the *jsii Assembly* document.
+* *Host runtime libraries* provide basic features that the code generated by
```
nija-at: I'm a bit lost on this host runtime library and why it's needed; all the others are fairly clear to me. Can you be clearer?

docs/specifications/4-standard-compliance-suite.md
RomainMuller: Could symbolic be a snake_cased form of the textual title?

docs/specifications/4-standard-compliance-suite.md
fulghum: I was also really curious to see how this works with each binding language – what's the experience to implement the tests in those languages?

docs/specifications/4-standard-compliance-suite.md
eladb: I am not sure I understand what categories are? Who defines them? Why do we need them?

docs/specifications/4-standard-compliance-suite.md
RomainMuller: The sequence of messages is normative, it is necessary to ensure consistent behavior across languages, at least from a side-effects perspective.

The initialization *might* be truncated off, for sure (it seems very inconvenient indeed).

docs/specifications/4-standard-compliance-suite.md
RomainMuller: It becomes fun when you test asynchronous workflows, though.

docs/specifications/4-standard-compliance-suite.md
```diff
@@ -0,0 +1,166 @@
...
+
+## Format
+In order to assert whether a new runtime implementation is correct, a formal
+compliance test suite is defined, that all new language runtimes must fully
+implement before they can be deemed eligible for General Availability.
+
+This document describes these tests, as well as a general approach for ensuring
+conformance requirements are met in a systematic manner.
+
+### Categories
```
eladb: I am not sure I understand what categories are? Who defines them? Why do we need them?

docs/specifications/4-standard-compliance-suite.md
rix0rrr: I imagine it will be thinks like "values", "callbacks", "async", "interfaces", "structs", etc.

RomainMuller: I am definitely planning on lifting data from the Markdown document by ways of a tool.

docs/specifications/4-standard-compliance-suite.md
RomainMuller: We probably could do something like this.

docs/specifications/5-new-language-intake.md
```diff
@@ -0,0 +1,136 @@
+# New Language Intake
+This document outlines the process to be followed when introducing a new *jsii*
```
eladb: add estimated total duration (4-6 months)

docs/specifications/4-standard-compliance-suite.md
RomainMuller: I actually mean to define a set of standard categories that tests need to be in. Refrained from trying to create that list upfront though, so I also didn't mention that would exist... But it was my idea (until it is proven bad by experience?)

docs/specifications/5-new-language-intake.md
RomainMuller: Good point!

docs/specifications/4-standard-compliance-suite.md
```diff
@@ -0,0 +1,166 @@
+# Standard Compliance Suite
+
+## Format
+In order to assert whether a new runtime implementation is correct, a formal
+compliance test suite is defined, that all language runtimes must fully
+implement before they can be deemed eligible for General Availability.
+
+This document describes these tests, as well as a general approach for ensuring
+conformance requirements are met in a systematic manner.
```
nija-at: I'm wondering if we should manage all this via markdown at all. 

Can we instead turn this into a JSON spec or something similar where there is a list of categories under which there is a list of test cases with path/name of the test file that implements it in that language? This can be the source of truth for compliance tests, i.e., if the test case is not in the spec file it's not part of the compliance test suite, it the test for a specific langunage is not in this spec file, that's a coverage gap.
We can then render it into a markdown table if you still like.

This may be over-engineering, I can't judge since I don't have the full picture; but my primary concern is that we (programmers and software engineers, in general) are terrible at maintaining documentation and keeping it consistent.

docs/specifications/4-standard-compliance-suite.md
```diff
@@ -0,0 +1,166 @@
...
+  included the test. This implies previous versions of `jsii` may not have
+  conformed to the test.
+- **Related Issues** lists any and all issues related to the test case. This
+  includes the issue that was created to track the introduction of the test, and
+  other issues such as bugs that prompted the introduction of a new compliance
+  test.
+- **Revised** lists all versions of `jsii` after the initial introduction of the
+  test case, for which the release included a revision of the test case. Trivial
+  revisions (such as correcting typos, or re-phrasing the description without
+  changing the tests case's semantic value) should not be included in the list.
```
nija-at: This all sounds cool, but I don't see why we need to hold all of this information. What are the requirements for this?
Why is `git blame` not a sufficient approximation? Sure, it's not precise but it's not clear why we need this level of precision.

Same concern as above. I'm afraid we'll struggle to keep all of this consistent.

docs/specifications/4-standard-compliance-suite.md
```diff
@@ -0,0 +1,166 @@
...
+After the attributes table, a **TypeScript** block of code describes the
+canonical form of the test. It includes any type declaration that is used by the
+test (so the code example is self-contained). Assertions performed by the test
+should be written in the form of [`jest`] expectations.
+
+[`jest`]: https://jestjs.io/docs/en/getting-started
+
+Finally, another code block details the sequence of messages that should be
+exchanged between the *host* and `node` processes during the execution of the
+test case, such that implementations can assert conherent behavior. The dialogue
```
nija-at: Typo: coherent

docs/specifications/4-standard-compliance-suite.md
```diff
@@ -0,0 +1,166 @@
+# Standard Compliance Suite
```
nija-at: I would love for us to start by defining the requirements of a the compliance suite.

What is it? What is it going to be used for, by who and what information must it carry?

docs/specifications/4-standard-compliance-suite.md
```diff
@@ -0,0 +1,166 @@
...
+> { "load": { "name": "test-case-001", "version": "1.2.3", "tarball": "/tmp/jsii-kernel-test/lib.tgz" } }
+< { "assembly": "test-case-001", "types": 3 }
+```
+````
+
+## Compliance
+In order to be able to assert compliance of language binding libraries to the
+standard test suite, implementations are responsible for provide a test harness
+that can produce a standardized test report in the form of a JSON document that
+follows the following schema:
```
nija-at: I'm a bit lost in here. Where would this test harness be checked in? What are other requirements of this harness; what is its entry point and exit points? It feels like we've not fully defined the boundary here and the interactions.

Based on my reading so far, it seems that there is a compliance test runner somewhere that interacts with the harness which further interacts with test categories and cases. It's not clear to me what these interactions and components are. Are these going to be clarified elsewhere? 
On the other hand, is this a bit heavy handed for the first round of compliance tests? Anything simpler we can do here?

docs/specifications/4-standard-compliance-suite.md
RomainMuller: I agree. Dropped all that.

docs/specifications/4-standard-compliance-suite.md
RomainMuller: I want to think of the spec as the source of truth - which is a golden master against which all implementations are assessed. Tools will be built and integrated as part of the continuous integration workflow to make sure consistency is there.

I don't want to have to reference all language implementation's tests in the spec, because this could not possibly include implementations made outside of this repository....

docs/specifications/4-standard-compliance-suite.md
RomainMuller: My perspective is each implementation defines it's own test suite & harness. The contract is it must produce the correct output document (as specified here); and a CLI will be able to consume that output document + the test suite markdown document; and and provides an assessment of coverage (%Compliance, or otherwise) -- this is then made part of the CI/CD pipeline.

RomainMuller:
Cleaned up from other changes - I think it is time to start the "Final Comments Period" for this RFC.

docs/specifications/4-standard-compliance-suite.md
```diff
@@ -0,0 +1,500 @@
...
+##### Kernel Trace
+```
+```
+</details>
+
+#### Partially initialized object consumption
+When a constructor passes `this` out from **JavaScript** to the *host* app, the
+reference must be correctly identified and passed across.
+
+> :construction: The .NET Runtime does not currently honor object identity,
```
rix0rrr: That is interesting. Means that jsii needs a notion of what it means to be "the same object". I believe because of anonymous objects being returned under 2 different interfaces, we must also emit 2 proxies in general (Java etc) and so it just might be that comparing objects by identity using `==` is explicitly disallowed by the JSII spec. 

It might be worthwhile adding a `JSII.isSameObject(x, y)` static to the runtimes then to fill that gap, for example.

docs/specifications/4-standard-compliance-suite.md
```diff
@@ -0,0 +1,500 @@
...
+
+  public consumePartiallyInitializedThis(obj: ConstructorPassesThisOut) {
+    this.obj = obj;
+  }
+}
+const consumer = new MyConsumer();
+const object = new ConstructorPassesThisOut(consumer);
+
+// THEN
+expect(consumer.obj).toBe(object);
```
rix0rrr: So rather than saying "C# fails this test", we don't have this test, I suppose?

Or the test should be about observable side effects, not identity. Changes made to the object that is passed out are visible via the other reference as well, and so we know they "must be" the same object, or at least behave as the same object, regardless of the strict object identity.

docs/specifications/4-standard-compliance-suite.md
```diff
@@ -0,0 +1,500 @@
...
+> {"complete":{"api":"complete","cbid":"jsii::callback::20000"}}
+< {"ok":{"$jsii.byref":"test.ConstructorPassesThisOut@10001"}}
+```
+</details>
+
+### Interfaces
+Tests in this section ensure correct behavior of *behavioral interfaces*.
+
+#### Host app can implement an interface "from scratch"
+It is possible for a "pure" host interface implementation to be passed across
```
rix0rrr: I'm only now noticing that you are planning to write descriptive notes and language-specific implementation notes with every test. So is the plan to collect a giant swath of tests, with fragments of prose surrounding them, and the aggregation of all of those is the spec? 

I would have expected to see a long-form document that talks me through the behavior of jsii in general, a document per language with implementation notes, and references out to a set of tests.

docs/specifications/1-introduction.md
```diff
@@ -0,0 +1,205 @@
...
+The *npm* package produced as a result of compiling the **TypeScript** source
+remains the source of truth with respects to implementation of the API.
+
+### Host & Kernel
+
+The [*jsii* runtime architecture] defines two processes:
+
+1. The *host* process runs the users' code native environment (a Java virtual
+   machine, the .NET Runtime, ...).
+2. The *kernel* process hosts the **JavaScript** code from the standard *npm*
```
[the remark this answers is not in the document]
RomainMuller: Absolutely! Since this PR auto-merged (silly I forgot to "do not merge" it), I'll introduce this (and other changes) in a subsequent PR.

docs/specifications/2-type-system.md
```diff
@@ -0,0 +1,316 @@
...
+additional nuances on top of those, to better represent **TypeScript** and
+**JavaScript** idioms in a way that enables generating convenient APIs in other
+languages.
+
+### Classes
+Exported **TypeScript** classes are represented in the *jsii* type system, with
+the following restrictions from plain **TypeScript**:
+- Methods overloads are not supported.
+- Overridden methods or properties must retain the exact same type signature as
+  the one declared in a parent type. The **jsii** type system strictly enforces
```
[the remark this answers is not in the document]
RomainMuller: Well it is "strictly enforced" as in *jsii* will not let you compile code that does not honor it... We have checks in place for that...

Language support actually becomes problematic here in the case of C# where you *cannot* change the signature of an overridden call (you'd define a new signature, with an entirely new resolution group of its own).

docs/specifications/4-standard-compliance-suite.md
RomainMuller: That's a great idea and I'm stealing it.

docs/specifications/4-standard-compliance-suite.md
RomainMuller: This is a case where they *could* (and in fact, *should*) be the same object. I've intentionally left this test here (in this state) because I think this particular predicate (same reference with same static type must be the same object instance on the host) is valuable, and is definitely feasible.

I want to go with it this way for a while and possibly revisit if we find out the benefits is not worth the cost (I *guess* the cost isn't very high now... but I could guess wrong!)

docs/specifications/4-standard-compliance-suite.md
RomainMuller: The organization is obviously subject to change as we iterate and discover what is pragmatic or not.

The language-specific notes at this point are mostly tokens of known existing non-compliance of runtimes that we however already offer as "Generally Available". They help a reader not get confused in a way akin to:

>The specification says X but I can observe X', what the eff?

I am *extremely* favorable to building up a document that details the normative behavior in a clear & concise way, and then replace most of the prose attached to the tests with references to specific points in that document (or/and vice-versa). I believe the test suite will help formalize this document better.

RomainMuller:
#1259 contains addendums to this PR pushed after it was inadvertently auto-merged.

Issue #1137, opened 2019-12-18
Peer dependencies


eladb:
## :rocket: Feature Request

jsii should allow specifying a dependency as only a peer dependency without requiring that it will also be defined in the `dependencies` section. Peer dependencies are legitimate and in fact seem to be needed to allow implementing 3rd party CDK modules that can actually be used with multiple versions of the CDK.

PR #1141, opened 2019-12-19
fix(pacmak): generated dependencies are not consistent with source npm module

### LINKED ISSUE #676: Python dependency versions not handled correctly

RomainMuller:
The dependency statements generated in the Java, .NET and Python
packages generated by `jsii-pacmak` were not reflecting the dependency
statement modeled on the source package. In certain pathological cases,
such as Python, the dependency declaration was often more permissive
than what the source package allowed, resulting in surprising behavior
as well as difficult to troubleshoot problems.

This updates several elements involved in this problem:
1. The `jsii` compiler no longer carries version information for
  transitive dependencies of the assembly in the `dependencyClosure`
  property, and instead only carries the `targets` configuration for
  those.
2. The version statements in the `dependencies` section of the `.jsii`
  file no longer contains the "compile-time resolved" version number,
  but the actual SemVer expression declared in the package's source.
  Also, the `targets` object was dropped (only the SemVer range
  statement remains), as this incurred needless duplication of the
  target configuration (in `dependencies` and `dependencyClosure`).
3. `jsii-pacmak` renders the declared SemVer range in the relevant form
  for the target being generated.

Fixes #676
Related to #1137

---


packages/@jsii/spec/lib/assembly.ts
```diff
@@ -151,6 +145,19 @@ export interface Assembly extends Documentable {
...
+ * Shareable configuration of a jsii Assembly.
+ */
+export interface AssemblyConfiguration {
+  /**
+   * A map of target name to configuration, which is used when generating
+   * packages for various languages.
+   *
+   * @default none
+   */
+  targets?: AssemblyTargets;
```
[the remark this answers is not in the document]
rix0rrr: And I wonder if embedding the "target" information of dependencies isn't going to lead to breakage at some point in the future. What if it changes? Should we not have a hash/check on it? If I change the namespace of a Java package, for example.

packages/jsii-pacmak/lib/targets/java.ts
```diff
@@ -786,25 +786,74 @@ class JavaGenerator extends Generator {
...
 
+    /**
+     * Formats a SemVer version range into the equivalent Maven dependency notation.
+     *
+     * @see https://cwiki.apache.org/confluence/display/MAVENOLD/Dependency+Mediation+and+Conflict+Resolution
+     *
+     * @param version the semantic version number
+     * @param suffix a suffix to be appended to versions
+     */
+    function semverToMavenRange(version: string, suffix?: string): string {
```
rix0rrr: Worth a couple of unit tests? Same for the other vdep wranglers.

packages/jsii/lib/assembler.ts
```diff
@@ -1624,35 +1557,20 @@ function* intersect<T>(xs: Set<T>, ys: Set<T>) {
...
-  return Array.prototype.concat.call([], ...xs);
-}
-
 function noEmptyDict<T>(xs: {[key: string]: T}): {[key: string]: T} | undefined {
   if (Object.keys(xs).length === 0) { return undefined; }
   return xs;
 }
 
+function toDependencyClosure(assemblies: readonly spec.Assembly[]): { [name: string]: spec.AssemblyConfiguration } {
+  const result: { [name: string]: spec.AssemblyTargets } = {};
```
rix0rrr: I see you got rid of the version conflict checkage. Are you sure it can't happen?

rix0rrr: BTW, personally I find `Record<string, spec.AssemblyTargets>` easier on the eyes. This comment goes for all instances of `{ [key: string]: T}`.

packages/jsii-pacmak/lib/targets/python.ts
```diff
@@ -1182,19 +1183,15 @@ class Package {
...
-      if (versionParts[0] === '0') {
-        versionSpecifier = `~=${versionParts.slice(0, 3).join('.')}`;
-      } else {
-        versionSpecifier = `~=${versionParts.slice(0, 2).join('.')},>=${versionParts.slice(0, 3).join('.')}`;
-      }
+    for (const [depName, version] of Object.entries(this.metadata.dependencies ?? {})) {
+      const depInfo = this.metadata.dependencyClosure![depName];
+
+      const range = new semver.Range(version);
+      const versionSpecifier = range.set.map(
```
rix0rrr: Would be nice to put this in a function just like for the other languages, so it can be unit tested.

packages/jsii-pacmak/lib/targets/python.ts
```diff
@@ -1182,19 +1183,15 @@ class Package {
...
+        set => set.map(
+          comp => `${comp.operator || '=='}${comp.semver.raw}`
```
rix0rrr: What if `comp.operator` is equal to `"="`? Shouldn't it turn to `"=="` for pip?

What if `comp.operator` is equal to `"~"`? Shouldn't it turn to `"~="`?

packages/@jsii/spec/lib/assembly.ts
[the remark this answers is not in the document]
RomainMuller: That's be a breaking change and should be accompanied with a major version bump. However we currently have no validation for this.

packages/jsii-pacmak/lib/targets/python.ts
RomainMuller: Per the [type definitions](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/semver/index.d.ts#L275) `comp.operator` **cannot** be `~`. It however can be `=` in which case we'd actually render a single `=` which is no bueno. Good catch!

packages/jsii/lib/assembler.ts
RomainMuller: That's a very good point. I somehow though it was no longer necessary but that's not true. I'll update with a modified version where the "no conflict" check happens upfront (when loading the `ProjectInfo` object) and reduces the dependency statement to the intersection of all requirements found in the closure (which is the "most correct" way to go here, I feel, and will detect non-overlapping requirements at the same time)

RomainMuller: I have mixed feelings here. I tend to agree but `Record<T>`  does not have a `readonly` form, whereas I can `{ readonly [name: string]: T }`, and I like the fact the syntax remains consistent. Decided not to change this time around (it felt like unnecessary churn), but probably will gradually adopt in the future in places where `readonly` is not useful.

packages/jsii-pacmak/test/targets/version-utils.test.ts
```diff
@@ -0,0 +1,77 @@
+import { toMavenVersionRange, toNuGetVersionRange, toPythonVersionRange } from '../../lib/targets/version-utils';
+
+const examples: Record<string, { maven: string, nuget: string, python: string }> = {
```
rix0rrr: Sweet, I like <3

Issue #1142, opened 2019-12-19
Stop relying on synchronous reads from STDIN


RomainMuller:
### Affected Languages

- [ ] `TypeScript` or `Javascript`
- [X] `Python`
- [X] `Java`
- [X] .NET (`C#`, `F#`, ...)

### What is the problem?
`node` opens it's STDIN file descriptor in `O_NONBLOCK` mode, making attempts to synchronously read unsafe. Starting with node `13.2`, there has been a significant increase in the incidence of hitting EAGAIN in this process. Synchronous mitigations are tacky at best, and the proper way to fix is to move away from synchronous IO.

See aws/aws-cdk#5187

eladb:
I don't think it will be possible to move away from synchronous reads since you need to invoke host callbacks during synchronous javascript function calls, so you need to block the javascript execution.

eladb:
Can you also provide a reference to the node.js documentation that says that stdin is opened with O_NONBLOCK. From https://github.com/joyent/libuv/commit/1988f5e is seems like it's actually opened without O_NONBLOCK

RomainMuller:
> Can you also provide a reference to the node.js documentation that says that stdin is opened with O_NONBLOCK? From [joyent/libuv@1988f5e](https://github.com/joyent/libuv/commit/1988f5e) is seems like it's actually opened without O_NONBLOCK

The code you are pointing at is the code that initializes the STDIN/OUT for child processes, which was introduced around nodejs/node-v0.x-archive#3994, where people very specifically call out that, unlike `node`, most other programs don't have a non-blocking STDIN and hit problems when they do.

RomainMuller:
That being said, I think you're correct about it not being possible to make this process synchronous (or it'd require the application's code performs NO asynchronous operation ever).

PR #1143, opened 2019-12-19
fix(runtime): runtime crashes with EAGAIN trying to read from STDIN

RomainMuller:
When using node >= 13.2, attempts to synchronously read from STDIN often
result in `EAGAIN` being raised. This is due to the STDIN file
descriptor being opened with `O_NONBLOCK`.

This introduces a temporary workaround that catches the EAGAIN exception
and re-attempts reading from STDIN "immediately". While this is not the
ideal solution, it can be done right away and appears to have minimal
impact on the overall performance of applications.

A longer term fix is to move away from synchronous IO operations, but
this is a significantly mroe involved fix and migth be a while before
this can be delivered. The long term solution is tracked in #1142.

Related to aws/aws-cdk#5187

---

### DIFF
--- a/packages/@jsii/runtime/lib/sync-stdio.ts
+++ b/packages/@jsii/runtime/lib/sync-stdio.ts
@@ -22,21 +22,33 @@ export class SyncStdio {
       return this.inputQueue.shift();
     }
 
-    const buff = Buffer.alloc(INPUT_BUFFER_SIZE);
-    const read = fs.readSync(STDIN_FD, buff, 0, buff.length, null);
+    try {
+      const buff = Buffer.alloc(INPUT_BUFFER_SIZE);
+      const read = fs.readSync(STDIN_FD, buff, 0, buff.length, null);
 
-    if (read === 0) {
-      return undefined;
-    }
+      if (read === 0) {
+        return undefined;
+      }
 
-    const str = buff.slice(0, read).toString();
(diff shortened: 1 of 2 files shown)


rix0rrr:
Do you have a reference to the Node issue tracker for this issue somewhere? Do they consider it a feature?

rix0rrr:
BTW we can probably switch `stdin` back to blocking.

rix0rrr:
Can we add the tiniest of blocking sleep? Probably doesn't exist eh?

RomainMuller:
`node` (or `libuv`?) has  had `STDIN` open in `O_NONBLOCK` since some point in `0.x` (virtually forever), so the code has *always* been susceptible to `EAGAIN`. Generally speaking, the behavior around this is *different* in Windows versus \*NIXes (that makes things even funnier, doesn't it?). This [StackOverflow response](https://stackoverflow.com/a/16048083/223339) has a pretty good summary of the situation.

As a matter of fact, the `process.stdin` object is a `net.Socket` in the failing executions, and that class is fully asynchronous in the `node` standard library.

I also could not find a way to access `fcntl` from js-land, which would be required to flip the `O_NONBLOCK` flag around for safe synchronous operations.

Regarding introducing of a `sleep` instruction... Well as you have it... there isn't a synchronous `sleep` possibly in `node`, short of shelling out to `sleep`.

PR #1150, opened 2019-12-23
fix(jsii): `--watch` causes immediate failure

### LINKED ISSUE #1149: jsii --watch is broken
### Affected Languages

- [ ] `Python`
- [ ] `Java`
- [ ] .NET (`C#`, `F#`, ...)

### General Information
* **JSII Version:** 0.20.11 (build d4fd44f), typescript 3.7.3
* **Platform:** Mac

### What is the problem?

`jsii --watch` always exits with the following error:

```
Error: Error: Unexpectedly returned from createWatchProgram
    at Compiler._startWatch (/Users/benisrae/code/cdk/aws-cdk/node_modules/jsii/lib/compiler.js:111:31)
    at Compiler.emit (/Users/benisrae/code/cdk/aws-cdk/node_modules/jsii/lib/compiler.js:60:25
```

RomainMuller:
A change in the TypeScript compiler API around how the compiler host
reacts when invoked in `--watch` mode caused a failure in `jsii` due to
the method changing behaviors from not returning at all when watching to
instead returning an object that can be used to inspect/influence the
background watch process.

Changed the behavior of the `Compiler` interface so it now acts as an
`EventEmitter` when operating in `--watch` mode, and introduced a test
that confirms that operating with `--watch` results in the correct
behavior of watching for file changes and dynamically re-compiling.

Fixes #1149

---


packages/jsii/bin/jsii.ts
```diff
@@ -57,12 +57,17 @@ const warningTypes = Object.keys(enabledWarnings);
...
-    watch: argv.watch,
     projectReferences: argv['project-references'],
     failOnWarnings: argv['fail-on-warnings'],
   });
 
-  return { projectRoot, emitResult: await compiler.emit() };
+  compiler.on('statusChanged', diagnostic => utils.logDiagnostic(diagnostic, projectRoot));
+
+  const result = argv.watch
+    // When watching, return a promise that never resolves... #magic
```
eladb: why isn't this hidden inside `compiler.watch()`. That's how I'd expect it to behave without wrapping it with #magic

packages/jsii/lib/compiler.ts
```diff
@@ -72,28 +72,70 @@ export class Compiler implements Emitter {
...
-      return this._startWatch();
-    }
     return this._buildOnce();
 
   }
 
+  public async watch(): Promise<Watch> {
+    await this.buildTypeScriptConfig();
+    await this.writeTypeScriptConfig();
+    this.rootFiles = this.determineSources([]);
```
eladb: Can we reuse these three lines with `emit()`? `this.prepareForBuild()` or something

packages/jsii/lib/compiler.ts
```diff
@@ -72,28 +72,70 @@ export class Compiler implements Emitter {
...
+   *
+   * @param event    the 'result' event
+   * @param listener the listener to be called with the EmitResult.
+   */
+  public once(event: 'result', listener: (result: EmitResult) => any): this;
+  /**
+   * Adds a once-only listener for the 'statusChanged' event that is called with a Diagnostic
+   * whenever the status of the current watch process changes.
+   *
+   * @param event    the 'statusChanged' event
```
eladb: not sure I understand why was this needed as part of this change? Also feels like it might be risky...

packages/jsii/lib/compiler.ts
```diff
@@ -329,6 +374,16 @@ export class Compiler implements Emitter {
   }
 }
 
+/**
+ * A handle to a watch-mode operation.
+ */
+export interface Watch {
```
eladb: Is this really needed?

packages/jsii/lib/project-info.ts
```diff
@@ -177,7 +177,7 @@ async function _loadDependencies(
     const pkg = _tryResolveAssembly(name, localPackage, searchPath);
     LOG.debug(`Resolved dependency ${name} to ${pkg}`);
     // eslint-disable-next-line no-await-in-loop
-    const assm = await loadAndValidateAssembly(pkg);
+    const assm = await loadAndValidateAssembly(pkg).catch(e => { throw new Error(`Error loading ${pkg}: ${e}`); });
```
eladb: Use `try/catch` please

packages/jsii/bin/jsii.ts
RomainMuller: That makes `watch`... impossible to test...

packages/jsii/lib/compiler.ts
RomainMuller: This is 100% motivated by the desire/need to be able to unit-test `watch` mode to ensure it actually does watch.

packages/jsii/lib/compiler.ts
RomainMuller: It's virtually impossible to unit test without that. Or there is no way to make the test finish.

packages/jsii/lib/compiler.ts
```diff
@@ -126,7 +168,10 @@ export class Compiler implements Emitter {
         ...COMPILER_OPTIONS,
         noEmitOnError: false,
       },
-      { ...ts.sys, getCurrentDirectory() { return projectRoot; } }
+      { ...ts.sys, getCurrentDirectory() { return projectRoot; } },
+      ts.createSemanticDiagnosticsBuilderProgram,
+      undefined,
+      diag => this.eventEmitter.emit('statusChanged', diag)
```
eladb: Why does this need to go through an event emitter?

packages/jsii/lib/compiler.ts
```diff
@@ -72,28 +72,79 @@ export class Compiler implements Emitter {
...
+    return this._buildOnce();
+
+  }
+
+  /**
+   * Watches for file changes and dynamically re-compiles whenever needed.
+   */
+  public async watch(): Promise<Watch> {
+    await this._prepareForBuild();
+    return this._startWatch();
```
eladb: Why do we need `_startWatch` as a separate method?

packages/jsii/lib/compiler.ts
```diff
@@ -72,28 +72,79 @@ export class Compiler implements Emitter {
   }
 
   /**
-     * Compiles the configured program.
-     *
-     * @param files can be specified to override the standard source code location logic. Useful for example when testing "negatives".
-     */
+   * Adds a listener for the 'result' event that is called with EmitResult.
```
eladb: Document the meaning of the `result` event. It is not super intuitive by the name that this event is emitted every the code is compiled

packages/jsii/lib/compiler.ts
```diff
@@ -72,28 +72,79 @@ export class Compiler implements Emitter {
...
+    return this;
+  }
+
+  /**
+   * Adds a once-only listener for the 'result' event that is called with EmitResult.
+   *
+   * @param event    the 'result' event
+   * @param listener the listener to be called with the EmitResult.
+   */
+  public once(event: 'result', listener: (result: EmitResult) => any): this;
```
eladb: Why do you need the `once` method here? I don't see a reason for it.

packages/jsii/lib/compiler.ts
```diff
@@ -72,28 +72,79 @@ export class Compiler implements Emitter {
...
+  /**
+   * Adds a once-only listener for the 'statusChanged' event that is called with a Diagnostic
+   * whenever the status of the current watch process changes.
+   *
+   * @param event    the 'statusChanged' event
+   * @param listener the listener to be called with the Diagnostic.
+   */
+  public once(event: 'statusChanged', listener: (message: Diagnostic) => any): this;
```
eladb: Why does this change introduce this event? How is it related to what is being fixed?

packages/jsii/lib/compiler.ts
RomainMuller: Good point. It's probably a remnant of multiple successive refactors that gradually thinned the `watch` method until nothing was left in it :)

packages/jsii/lib/compiler.ts
RomainMuller: Yeah I actually have a better name for it now... I'll call it `compilationComplete` I think

packages/jsii/lib/compiler.ts
RomainMuller: It's convenient in the unit test, because I can do something different for the __first__ compilation result. But it's not the only way to achieve this for sure!

packages/jsii/lib/compiler.ts
RomainMuller: That event is needed just to make sure the (console) output behavior of the CLI does not change. It's a little unfortunate -- I'll be looking at whether I can make this less invasive/obnoxious.

packages/jsii/lib/compiler.ts
RomainMuller: This was to preserve the console output behavior (while avoiding the standard behavior to clear out the console when testing, which is SUPER inconvenient), but I believe I have a better way...

packages/jsii/bin/jsii.ts
```diff
@@ -57,12 +57,15 @@ const warningTypes = Object.keys(enabledWarnings);
...
   const compiler = new Compiler({
     projectInfo,
-    watch: argv.watch,
     projectReferences: argv['project-references'],
     failOnWarnings: argv['fail-on-warnings'],
   });
 
-  return { projectRoot, emitResult: await compiler.emit() };
+  const result = argv.watch
+    // When watching, return a promise that never resolves... #magic
```
eladb:
```suggestion
```

packages/jsii/bin/jsii.ts
```diff
@@ -57,12 +57,15 @@ const warningTypes = Object.keys(enabledWarnings);
...
+    ? await compiler.watch().then(watch => watch.block())
```
eladb: I'd still rather this looked like `await compiler.watch()` and the non-blocking version used by tests is `await compiler.watch({ nonblock: true });`

packages/jsii/lib/compiler-watch.ts
```diff
@@ -0,0 +1,155 @@
...
+
+    this.watch = createWatchProgram(host);
+    this.watchComplete = new Promise((resolve, reject) => {
+      this.onWatchEnd = resolve;
+      this.onWatchFailed = reject;
+    });
+  }
+
+  /**
+   * Adds a listener for the 'result' event that is called with EmitResult.
```
eladb: fix documentation

packages/jsii/lib/compiler-watch.ts
```diff
@@ -0,0 +1,155 @@
...
+   * @param timeoutMillis the maximum amount of time to wait for the watch to complete. When
+   *                      provided, the returned promise will reject if it is not resolved by the
+   *                      end of this amount of time.
+   *
+   * @returns a `Promise` that only resolves once the `.close()` method was invoked.
+   */
+  public async block(timeoutMillis?: number): Promise<EmitResult> {
+    const promises = [this.watchComplete];
+
+    if (timeoutMillis != null) {
```
eladb: why is this needed?

packages/jsii/lib/compiler-watch.ts
```diff
@@ -0,0 +1,155 @@
...
+    return Promise.race(promises);
+  }
+
+  /**
+   * Terminates the watch process.
+   */
+  public close(): void {
+    // Calling `this.watch.close()` will "synchronize" the watch, and emit once last time
+    this.eventEmitter.once(CompilerWatch.EVENT_COMPILATION_COMPLETE, this.onWatchEnd);
+    this.watch.close();
```
eladb: wouldn't it be easier to just expose the raw `watch` object for unit tests to directly access the typescript API instead of duplicating this logic in this layer? Since this is an internal API, I don't see a reason why it can't simply be a leaky API for unit tests, which is a common practice.

packages/jsii/lib/compiler.ts
```diff
@@ -72,28 +69,57 @@ export class Compiler implements Emitter {
...
+  /**
+   * Watches for file changes and dynamically re-compiles whenever needed.
+   */
+  public async watch(
+    onStatusChanged?: ts.WatchStatusReporter,
+    onDiagnostic = logDiagnostic,
+  ): Promise<CompilerWatch> {
+    await this._prepareForBuild();
+
+    return new CompilerWatch(
```
eladb: Instead of wrapping the typescript API, why not just expose the TS watch object as @internal?

packages/jsii/bin/jsii.ts
RomainMuller: Did something of the likes... I get a few more options :)

packages/jsii/lib/compiler-watch.ts
RomainMuller: That method no longer exists.

packages/jsii/lib/compiler-watch.ts
RomainMuller: This was added safety in case the test never succeeds (because watch is broken?). This is no longer needed.

packages/jsii/lib/compiler-watch.ts
RomainMuller: It's a common practice I dislike, because it encourages poor separation of concerns in my opinion. However, this is not a hill I'm going to die on, so I changed as you're asking.

packages/jsii/lib/compiler.ts
RomainMuller: I like containing the TypeScript compiler API as much as I can because it's not been very stable in the past.

PR #1169, opened 2020-01-03
fix(rosetta): crashes on behavioral interfaces

rix0rrr:
Behavioral interfaces (`ISomething`) weren't translated at all, and used
to crash the Java translator. Add translations for them.

Make the `jsii-rosetta` command-line tool more robust against errors in
the translator: catch and report the error, instead of crashing.

Add support for shorthand property assignments in object literals
(`someFunction({ foo });`) for Java as well, which wasn't covered
by any test yet.

---

### DIFF
--- a/packages/jsii-rosetta/lib/commands/extract.ts
+++ b/packages/jsii-rosetta/lib/commands/extract.ts
@@ -96,12 +96,25 @@ async function translateAll(snippets: IterableIterator<TypeScriptSnippet>, inclu
 export function singleThreadedTranslateAll(snippets: IterableIterator<TypeScriptSnippet>, includeCompilerDiagnostics: boolean): TranslateAllResult {
   const translatedSnippets = new Array<TranslatedSnippet>();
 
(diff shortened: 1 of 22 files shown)


packages/jsii-rosetta/lib/commands/extract.ts
```diff
@@ -96,12 +96,25 @@ async function translateAll(snippets: IterableIterator<TypeScriptSnippet>, inclu
...
   const translator = new Translator(includeCompilerDiagnostics);
   for (const block of snippets) {
-    translatedSnippets.push(translator.translate(block));
+    try {
+      translatedSnippets.push(translator.translate(block));
+    } catch(e) {
+      failures.push({
+        category: ts.DiagnosticCategory.Error,
+        code: 999,
+        messageText: `Error translating snippet: ${e}\n${block.completeSource}`,
```
eladb: Prefix with “Jsii-Rosetta” or something like this.

PR #1184, opened 2020-01-08
fix(rosetta): properly escape C# comments as XML

rix0rrr:
The C# compiler will completely drop comment blocks that aren't
well-formed XML; the `` tag we were outputting wasn't, and
so namespace headings weren't being generated.

At the same time, pay more attention to escaping of arbitrary
HTML and the text inside attributes to minimize the chances
of things going wrong in the future.

---


rix0rrr:
That is a fair question. I will endeavor to find out.

Issue #1182, opened 2020-01-07
Improve error messages in kernel-found type mismatches


RomainMuller:
### Affected Languages

- [ ] `TypeScript` or `Javascript`
- [X] `Python`
- [ ] `Java`
- [ ] .NET (`C#`, `F#`, ...)

### General Information
* **JSII Version:** `0.21.0`
* **Platform:** any

* [X] I may be able to implement this feature request

* [ ] This feature might incur a breaking change

### Description
When using non-statically typed languages (such as Python), users may inadvertently pass an incorrectly typed value though an argument. In such cases, the *jsii* runtime will raise a type checking error, which can look like the following:

```
Error: Expected Scalar, got {"$jsii.byref":"@aws-cdk/core.Reference@10024"}
```

While this describes the problem, it does not do so in a way that users not familiar with the internals of jsii can understand.

### Proposed Solution
The proposed solution is to produce a more semantically relevant message that does not expose internals of jsii, such as:

```
Error: Expected a 'string' value, but received an instance of '@aws-cdk/core.Reference'
```

If possible, the error should also include contextual information that provides better hits as to where the invalid value was provided, for example:

```
Error:
    Illegal value provided for the 'file_system_id' property of '@aws-cdk/aws-efs.CfnMountTargetProps':
        Expected a 'string' value, but received an instance of '@aws-cdk/core.Reference'
```

---

Related to aws/aws-cdk#5500

PR #1187, opened 2020-01-09
fix: poor error message on scalar deserialization type mismatch

RomainMuller:
When deserializing a scalar (string, boolean, number, ...), and receiving a value that is not
a scalar, the error message would not mention what kind of scalar value was required, making
the message less useful than it could be.

Used the description of the type reference in the error message so that the specific scalar
type is direction mentioned in the error message there. The description of the incorrect
value that was received remains unchanged (typically the JSON representation of the value),
and may be improved in a subsequent change.

Improves the situation identified in #1182, however does not fix the whole problem.

---

### DIFF
--- a/packages/@jsii/kernel/package.json
+++ b/packages/@jsii/kernel/package.json
@@ -53,7 +53,6 @@
   "jest": {
     "collectCoverage": true,
     "collectCoverageFrom": [
-      "**/bin/**/*.js",
       "**/lib/**/*.js"
     ],
     "coverageReporters": [
--- a/yarn.lock
+++ b/yarn.lock
@@ -1328,11 +1328,6 @@
   dependencies:
     "@types/node" "*"
 
-"@types/xmldom@^0.1.29":
-  version "0.1.29"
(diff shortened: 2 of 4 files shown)


packages/@jsii/kernel/package.json
```diff
@@ -53,7 +53,6 @@
   "jest": {
     "collectCoverage": true,
     "collectCoverageFrom": [
-      "**/bin/**/*.js",
```
RomainMuller: This is because there is no `bin` in there anymore :)

yarn.lock
```diff
@@ -1328,11 +1328,6 @@
   dependencies:
     "@types/node" "*"
 
-"@types/xmldom@^0.1.29":
-  version "0.1.29"
-  resolved "https://registry.yarnpkg.com/@types/xmldom/-/xmldom-0.1.29.tgz#c4428b0ca86d3b881475726fd94980b38a27c381"
-  integrity sha1-xEKLDKhtO4gUdXJv2UmAs4onw4E=
-
```
RomainMuller: I think this was previously introduced by error but is no longer depended on. It's unrelated to the change at hand, but it's small enough I don't mind bundling.

PR #1197, opened 2020-01-14
fix(java): invalid collections returned with non-class elements

### LINKED ISSUE #1196: Invalid collections returned when element type is not a class
### Affected Languages

- [ ] `Python`
- [X] `Java`
- [ ] .NET (`C#`, `F#`, ...)

### General Information
* **JSII Version:** `0.21.1 (build 9ff44cb), typescript 3.7.4`
* **Platform:**     *any*

### What is the problem?
When a return value is typed as a `List<SomeStruct>`, the Java runtime does not correctly perform the conversion to the correct Java type, and instead returns a `List<JsiiObject>` where the `JsiiObject` instances however contain all the information necessary to permit the conversion.

The runtime needs to be corrected so that it actually returns a `List<SomeStruct>`, or else attempting to use the contents of the list is likely to cause a `ClassCastException`.

It is probable the same issue would affect `Map<String, SomeStruct>` instances, so this has to be checked as well.

### Minimal Reproduction
A minimal reproduction using the AWS CDK can be achieved:

```java
package com.myorg;

import software.amazon.awscdk.core.*;
import software.amazon.awscdk.services.cloudwatch.Dimension;
import software.amazon.awscdk.services.cloudwatch.IMetric;
import software.amazon.awscdk.services.cloudwatch.MetricConfig;
import software.amazon.awscdk.services.lambda.Code;
import software.amazon.awscdk.services.lambda.Function;
import software.amazon.awscdk.services.lambda.Runtime;

import java.util.List;

public class CwMetricsApp {
    public static void main(final String[] args) {
        final App app = new App();
        final Stack stack = new Stack(app, "ReproStack");

        final Function funct = Function.Builder.create(stack, "Function")
                .code(Code.fromInline("export function handler() {}"))
                .handler("index.handler")
                .runtime(Runtime.NODEJS_10_X)
                .build();

        final IMetric metric = funct.metricInvocations();

        final MetricConfig config = metric.toMetricConfig();

        // This list is, in fact (and incorrectly), a List<JsiiObject>
        final List<Dimension> dimensions = config.getMetricStat().getDimensions();
        for (final Dimension dimension : dimensions) {
            System.out.println(String.format("Dimension<%s, %s>", dimension.getName(), dimension.getValue()));
        }

        app.synth();
    }
}
```

RomainMuller:
The Java runtime for jsii used to not be able to necessarily hold the
type information necessary to correctly process the contents of
collections, as the generic information is lost.

In order to address this, introduced a new `NativeType<T>` class that is
able to carry:
1. The actual declared type of the value (via generic type parameter)
2. The raw type of the value (reified)
3. The correct JavaType to pass to Jackson to obtain the correct result

It also preserves the specific behavior around de-serialization of
`JsiiObject` sub-classes as well as jsii interfaces, correctly
requesting de-serialization as the `JsiiObject` class, and in a second
phase `transform` this into the correct value type as needed.

Introduced a new test that validates the `List<T>` and `Map<String, T>`
values received from `node` are deserialized to instances with the
appropriate content type.

Fixes #1196

---

---

## Commit Message
fix(java): invalid collections returned with non-class elements (#1197)

The Java runtime for jsii used to not be able to necessarily hold the
type information necessary to correctly process the contents of
collections, as the generic information is lost.

In order to address this, introduced a new `NativeType<T>` class that is
able to carry:
1. The actual declared type of the value (via generic type parameter)
2. The raw type of the value (reified)
3. The correct JavaType to pass to Jackson to obtain the correct result

It also preserves the specific behavior around de-serialization of
`JsiiObject` sub-classes as well as jsii interfaces, correctly
requesting de-serialization as the `JsiiObject` class, and in a second
phase `transform` this into the correct value type as needed.

Introduced a new test that validates the `List<T>` and `Map<String, T>`
values received from `node` are deserialized to instances with the
appropriate content type.

Fixes #1196


RomainMuller:
Labelled with `do-not-merge` as I'm pondering whether the old code paths need to be preserved in order to ensure proper interoperability with code generated against older versions of the Java runtime.

rix0rrr:
To help me review, since this change is ungodly:

This was flagged as a regression (since jsii 0.17 if I'm reading the breadcrumb trail directly). Can you explain succinctly what the changes were that caused this breakage (starting from the wire protocol changes, if any) and what the approach for the fix is?

To the uncultured plebeian that I am, it would seem that if the wire protocol says:

```
[ { "jsii$ref": "my.MyObject@1" }]
```

It should be easy enough to construct a `List` with a `MyObject` in it?

It's probably trickier than that but I need it explained to me like I'm 5 :).

RomainMuller:
I significantly reduced the amount of regression test file changes by preserving the previous API (and having it delegate into the new one); and only switching generated code to the new API when the older one was unsafe.

To answer @rix0rrr's question, here's what happened:
* In #825, the Object ID generation was changed, specifically around interfaces.
  + The format of Object IDs generally speaking is `<type-fqn>@<some-number>`.
  + Prior to #825, the `<type-fqn>` part could be an interface if the value was exiting JavaScript through an interface-typed method or property.
  + After #825, the `<type-fqn>` part is *required* to be a class name, or `Object` (pure interface implementations, and by-reference structs). Necessary interface declarations are passed on a separate field of the `ObjRef`.
* This means that since then, first pass hydration in Java cannot yield an interface at all... Obtaining an interface view of an object requires on additional step (`JsiiObject.asInterfaceProxy(Class<?>)`).
* This additional step is done for single values, but was missed from `List`s and `Map`s; because the only signal that is passed to the deserializer was `List.class` or `Map.class`, which yields no information about the expected element type.
* Add to this that Java does not allow providing customized implicit conversions (the ways C# does), nor creating a dynamic proxy object with a base class other than Object (the way it can be done in Python), and you get the broken behavior.

Hopefully this helps clarify the problem...

Now - to the approach for solving this: I introduced a way to represent complex types to provide directions to the deserializer, in the form of the `NativeType` class. This is a Jackson (our JSON handling tool of choice) friendly way to express the expected type to be deserialized, including any specializations such as element type.

eladb:
I am not sure I fully understand if this is a breaking change. I don't think we can afford breaking changes neither of the manifest nor of the runtime protocol.

RomainMuller:
@eladb - I don't think this changes the runtime protocol. The only part that may be "broken" here is technically not working correctly in the current state of things. The only one thing which I think may still need "easing in" (so new runtime works with code generated by the previous generator) is with how the "old API" was modified to delegate into the "new API", which will throw if one attempts to use an api that returns a List or a Map (those could have returned collection with "illegal" items in them).

I think I can work out a way to make the old behavior preserved, so older generated code keeps behaving the same as it currently does (still broken when it was, but not failing more).

I can then attempt to run a number of scenarios using the current CDK libraries with the new Java runtime ahead of the class path to validate it is indeed backwards-compatible?

I'd add unit tests to assert persistence of the older behavior, but I'd be skeptical as to the coverage of such (new) tests being sufficient...

rix0rrr:
> * This means that since then, first pass hydration in Java cannot yield an interface at all...

This part I don't get.  It used to be:

```
{ "jsii$ref": "my.MyInterface@1" }
```

And now it's:

```
{ "jsii$ref": "Object@1", "jsii$interfaces": ["my.MyInterface"] }
```

We can restore the "old" behavior by picking the first interface from the associated list and instantiating the proxy class for that type, right?

I guess we're doing all this work so that if there are 2 interfaces in the wire list, we'll pick the one that the current  method has been declared to return? (Without looking at the wire data, looks like, just assuming the types will work out as declared?)

I am *slightly* concerned about all the introspection happening on every JSII call; every call constructs a set of new `NativeType` objects, which could be cached because they're always the same. But I guess this is not the most expensive work happening on jsii calls :)

RomainMuller:
I added a test of the *previous* behavior with respects to `List` and `Map`, in order to ensure the new code path does not cause older generated code to crash. I also fixed the new code path so that... it... does not crash on such code paths.

packages/@jsii/java-runtime/pom.xml.t.js
```diff
@@ -113,10 +113,10 @@ process.stdout.write(`<?xml version="1.0" encoding="UTF-8"?>
             <scope>test</scope>
         </dependency>
 
-        <!-- https://mvnrepository.com/artifact/org.mockito/mockito-core -->
+        <!-- https://mvnrepository.com/artifact/org.mockito/mockito-junit-jupiter -->
```
eladb: Why was this changed?

packages/@jsii/java-runtime/project/src/main/java/software/amazon/jsii/NativeType.java
```diff
@@ -0,0 +1,216 @@
...
+
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public abstract class NativeType<T> {
```
eladb: Missing documentation

packages/@jsii/java-runtime/project/src/main/java/software/amazon/jsii/NativeType.java
```diff
@@ -0,0 +1,216 @@
...
+    protected static final JavaType[] NO_TYPE_PARAMS = {};
```
eladb: Missing docs

PR #1205, opened 2020-01-16
feat(superchain): upgrade maven to 3.6.3

RomainMuller:
This involves installing maven from tarball instead of installing it from
the RPM repositories provided in the AmazonLinux 2 base image. This is
somewhat unfortunate, but is (currently) the only way to get a recent build
of maven without depending on third-party repositories which would need
vetting before they can be safely used.

---

### DIFF
--- a/superchain/Dockerfile
+++ b/superchain/Dockerfile
@@ -1,7 +1,7 @@
 FROM amazonlinux:2
 
-# Install deltarpm as it can speed up the upgrade processes
-RUN yum -y install deltarpm
+# Install deltarpm as it can speed up the upgrade processes, and tar as it's needed for installing Maven
+RUN yum -y install deltarpm tar
 
 # Install .NET Core, mono & PowerShell
 ENV DOTNET_CLI_TELEMETRY_OPTOUT=true                                                                                    \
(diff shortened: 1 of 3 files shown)


superchain/Dockerfile
```diff
@@ -35,19 +35,39 @@ RUN amazon-linux-extras install ruby2.4
...
+    M2=/usr/local/apache-maven/bin                                                                                      \
+    MAVEN_OPTS="-Xms256m -Xmx512m"
 
 # Install Docker
 RUN amazon-linux-extras install docker                                                                                  \
   && yum clean all && rm -rf /var/cache/yum
 VOLUME /var/lib/docker
 
 # Install shared dependencies
-RUN yum -y install awscli git gzip openssl rsync tar unzip which zip                                                    \
```
eladb: Why was “tar” removed? Sounds like a useful thing to have in the image.

superchain/Dockerfile
```diff
@@ -35,19 +35,39 @@ RUN amazon-linux-extras install ruby2.4
...
+# Install JDK8 (Corretto)
+RUN amazon-linux-extras enable corretto8                                                                                \
+  && yum -y install java-1.8.0-amazon-corretto-devel                                                                    \
   && yum clean all && rm -rf /var/cache/yum
+
+# Install Maven
+ENV M2_VERSION 3.6.3
+RUN curl -sL https://www.apache.org/dist/maven/maven-3/${M2_VERSION}/binaries/apache-maven-${M2_VERSION}-bin.tar.gz     \
+         -o /tmp/apache-maven.tar.gz                                                                                    \
+  && curl -sL https://www.apache.org/dist/maven/maven-3/3.6.3/binaries/apache-maven-3.6.3-bin.tar.gz.asc                \
```
rix0rrr: Some more `$M2_VERSION`s missing here.

superchain/Dockerfile
RomainMuller: `tar` wasn't removed, it was moved upfront because it's needed before we get to this point (to install `maven` from tarball, of course 😁)

superchain/Dockerfile
RomainMuller: Ohhh good catch!!!

RomainMuller:
@rix0rrr  - we mustn't, but the only RPM repository I could find with a non-antique version of `apache-maven` is described on https://gist.github.com/sebsto/19b99f1fa1f32cae5d00, and I don't know how trustworthy that source is... nor what it's maintenance profile is like...

PR #1211, opened 2020-01-21
chore: annotate Docker image with author info

RomainMuller:
This produces more comprehensive metadata on the image and
clearly identifies a maintainer.

---


.travis.yml
```diff
@@ -15,7 +15,7 @@ services:
 script:
   - set -eo pipefail
   # Building the Docker image
-  - docker build --pull --build-arg BUILD_TIMESTAMP=$(date -u +'%Y-%m-%dT%H:%M:%SZ') --build-arg COMMIT_ID="${TRAVIS_COMMIT}" -t "jsii/superchain:nightly" ./superchain
+  - docker build --pull --build-arg BUILD_TIMESTAMP=$(date -u +'%Y-%m-%dT%H:%M:%SZ') --build-arg COMMIT_ID="${TRAVIS_COMMIT}" --build-arg JSII_VERSION=$(node -p "require('./lerna.json').version") -t "jsii/superchain:nightly" ./superchain
```
eladb: this assumes that the image for the *current* version was already pushed to docker hub. Does that make sense in terms of workflow?

RomainMuller: For the `release` tag it definitely does... But I'm guessing you mean to talk about the `nightly` tag, which technically is for the upcoming version, not the current one...

If interpreted as "the current version of jsii when this image was released" then it is always valid. But how about instead, for the `nightly` tag we make this value be the version followed by some other sting, for example: `0.21.1-HEAD`?

Issue #1209, opened 2020-01-21
Build aws-cdk as an integration test for jsii


RomainMuller:
As a way to identify testing gap in `jsii`.

eladb:
Yey!

PR #1219, opened 2020-01-27
feat: CDK Build Integration Test

MrArnoldPalmer:
Adds @jsii/integ-test private module for defining new integration tests.
Adds a new integration test that downloads the latest CDK release source
code and builds it with the local version of jsii and jsii-pacmak.

This unit test requires a github access token defined in the environment
to get the latest release version and download the asset.

Fixes: #1209

---


MrArnoldPalmer:
@RomainMuller 

1. Where is the best place to actually add `yarn test:integ` in our CD pipeline. I would think in the jsii-master pipeline in delivlib but adding it to buildspec.yml means it will run on all PRs and other events?
2. Specified a 60min timeout which on my machine it finishes comfortably. That feel right to you?
3. Process management stuff is interesting, potentially we could do other integration testing in a similar way. I would like to build integration tests for the jsii-config tool with similar interactions via child_process.

packages/@jsii/integ-test/utils/index.ts
```diff
@@ -0,0 +1,118 @@
...
+ * downloadReleaseAsset
+ *
+ * Wrap http calls to download release asset in a promise. Github responds with
+ * a 302 sometimes which is required to be handled. Returns the buffer to be
+ * streamed to destination fs stream.
+ */
+export const downloadReleaseAsset = (url: string): Promise<any> => new Promise((resolve, reject) => {
+  const config = {
+    headers: {
+      'User-Agent': 'aws-cdk',
```
RomainMuller: Well technically.... `@jsii/integ-test`.

packages/@jsii/integ-test/utils/index.ts
```diff
@@ -0,0 +1,118 @@
...
+      Authorization: `token ${process.env.GITHUB_TOKEN}`,
+      Accept: 'application/json'
```
RomainMuller: Are you required to accept `application/json` despite you're downloading a tarball? Is `application/x-tar+gzip` (or a more generic `application/octet-stream`) not working?

packages/@jsii/integ-test/utils/index.ts
```diff
@@ -0,0 +1,118 @@
...
+    }
+  };
+
+  https.get(url, config, response => {
+    if (response.statusCode! < 200 && response.statusCode !== 302) {
```
RomainMuller: This `!` here... has a potential for weird surprises (because JavaScript!):

```
null < 200;      //=> true
undefined < 200; // => false
```

packages/@jsii/integ-test/utils/index.ts
```diff
@@ -0,0 +1,118 @@
...
+      reject(new Error(`Status Code: ${response.statusCode}`));
+    }
+    if (response.statusCode === 302 && response.headers.location) {
```
RomainMuller: You're falling through here, possibly extending unnecessary effort (you may have already `reject`ed at this point).

You can either `return reject(...);`, or use an `else if` here.

I would not check for the presence of `response.headers.location` here, but instead throw if it is absent (meaning you received an invalid `302`).

packages/@jsii/integ-test/utils/index.ts
```diff
@@ -0,0 +1,118 @@
...
+      return https.get(response.headers.location, config, response => {
+        return resolve(response);
+      });
+    }
+
+    return resolve(response);
```
RomainMuller: At this point, the `response` you're resolving with has the following guarantees:
* `response.statusCode >= 200`
* `response.statusCode !== 302 || !response.headers.location`

This means you would `resolve` a possible `404`, `500`, etc... I feel like you should reject if the response is not a `response.statusCode >= 200 && response.statusCode < 300`... For otherwise the result may be confusing.

packages/@jsii/integ-test/test/build-cdk.test.ts
```diff
@@ -0,0 +1,81 @@
...
+import * as path from 'path';
+import * as Octokit from '@octokit/rest';
+import { downloadReleaseAsset, minutes, ProcessManager, rmdirRecursive } from '../utils';
+import * as dotenv from 'dotenv';
+
+const { mkdtemp } = fs.promises;
+
+dotenv.config();
+const JSII_DIR = path.resolve(require.resolve('jsii'), '..', '..');
+const JSII_PACMAK_DIR = path.resolve(require.resolve('jsii-pacmak'), '..', '..');
```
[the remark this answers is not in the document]
MrArnoldPalmer: Yeah I was looking for a different strategy here. This is much better.

packages/@jsii/integ-test/test/build-cdk.test.ts
```diff
@@ -0,0 +1,81 @@
...
+    // install cdk dependencies
+    await processes.spawn('yarn', ['install'], {
+      cwd: srcDir
+    });
+
+    // link local jsii/jsii-pacmak builds
+    await processes.spawn('rm', ['-rf', './node_modules/jsii'], { cwd: srcDir });
+    await processes.spawn('rm', ['-rf', './node_modules/jsii-pacmak'], { cwd: srcDir });
+    await processes.spawn('ln', ['-s', JSII_DIR, './node_modules'], { cwd: srcDir });
+    await processes.spawn('ln', ['-s', JSII_PACMAK_DIR, './node_modules'], { cwd: srcDir });
```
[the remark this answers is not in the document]
MrArnoldPalmer: yeah this is the meat of the test and gave me a tough time.

Trying to use both `npm link` and `npm install` results in errors. `npm link` fails when trying to resolve private modules that are local dependencies managed by yarn workspaces, e.g. jsii-build-tools. `npm install` fails with `no such file or directory` while looking for files in `node_modules/.staging`.

As far as .bin goes. All of the files in the local .bin directory of the build are also symlinks. `builDir/node_modules/.bin/jsii` points to  `buildDir/node_modules/jsii/bin/jsii`.

packages/@jsii/integ-test/utils/index.ts
```diff
@@ -0,0 +1,118 @@
...
+
+  async killAll() {
+    const values = Object.values(this.processes);
+    values.forEach(procObj => procObj.proc.kill());
+    await Promise.all(values.map(proc => proc.promise));
+    this.processes = {};
+  }
+
+  private add(proc: cp.ChildProcess, promise: Promise<void>) {
+    const { pid } = proc;
```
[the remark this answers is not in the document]
MrArnoldPalmer: leftover from some previous thing I swear! 😜

packages/@jsii/integ-test/utils/index.ts
MrArnoldPalmer: Problem here is the first request will fail if it redirects when accept !== 'application/json'. Not specifying Accept works though so removing.

packages/@jsii/integ-test/test/build-cdk.test.ts
```diff
@@ -41,16 +38,11 @@ describe('Build CDK', () => {
...
-    // save to file and wait to finish
-    code.pipe(codeStream);
-    await new Promise(resolve => codeStream.on('close', () => {
-      resolve();
-    }));
+    await writeFileStream(code, tarFile);
 
     // unzip tar archive
-    await processes.spawn('tar', ['-xzvf', fileName], {
+    await processes.spawn('tar', ['-xzf', fileName], {
```
RomainMuller: I think there's a `tar` npm module of some sort we're using somewhere else. We should probably revisit later to stop shelling out here. But I don't reckon it's a blocker.

packages/@jsii/integ-test/utils/index.ts
```diff
@@ -1,84 +1,78 @@
...
-      await rmdirRecursive(currentPath);
-    } else {
-      await fs.unlink(currentPath);
-    }
-  }));
-
-  await fs.rmdir(dir);
-};
+export function minutes(num: number): number {
+  return num * 1000 * 60;
```
RomainMuller: I still think we should skip the runtime multiplication that can be done upfront (`60_000`). But whatevs.

packages/@jsii/integ-test/utils/index.ts
```diff
@@ -1,84 +1,78 @@
...
-    this.processes = {};
-  }
-
-  async killAll() {
+  } = {};
+
+  /**
+   * kill all still running processes
+   *
+   * @param [signal] - signal sent to terminate process
```
RomainMuller: I don't think the `[]` are needed here...

packages/@jsii/integ-test/test/build-cdk.test.ts
MrArnoldPalmer: yeah I noticed that earlier today, its an easy switch so gonna make the change.

packages/@jsii/integ-test/utils/index.ts
MrArnoldPalmer: my thinking here is if we continue to add long running integ-tests having this util may be handy?

packages/@jsii/integ-test/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "@jsii/integ-test",
+  "version": "0.21.2",
```
RomainMuller:
```suggestion
  "version": "0.22.0",
```

packages/@jsii/integ-test/package.json
```diff
@@ -0,0 +1,33 @@
...
+  "keywords": [],
+  "author": {
+    "name": "Amazon Web Services",
+    "url": "https://aws.amazon.com"
+  },
+  "license": "Apache-2.0",
+  "devDependencies": {
+    "@octokit/rest": "^16.36.0",
+    "dotenv": "^8.2.0",
+    "fs-extra": "^8.1.0",
```
RomainMuller: Odd this doesn't come with `@types/fs-extra`, probably also drue for `tar` & `dotenv`.

packages/@jsii/integ-test/test/build-cdk.test.ts
```diff
@@ -0,0 +1,96 @@
+import { mkdtemp, remove } from 'fs-extra';
+import * as path from 'path';
+import * as Octokit from '@octokit/rest';
+import { downloadReleaseAsset, minutes, ProcessManager, extractFileStream } from '../utils';
+import * as dotenv from 'dotenv';
```
RomainMuller: Make sure to order the imports:
- external, alphanumerically sorted (`@octokit/rest`, `doting`, `fs-extra`, `path`)
- internal (`../utils`)

packages/@jsii/integ-test/test/build-cdk.test.ts
```diff
@@ -0,0 +1,96 @@
...
+      repo: 'aws-cdk'
+    });
+
+    // download and extract code
+    const code = await downloadReleaseAsset(`https://api.github.com/repos/aws/aws-cdk/tarball/${release.data.tag_name}`);
+    const srcDir = path.join(buildDir, `aws-aws-cdk-${release.data.target_commitish.substring(0, 7)}`);
+    await extractFileStream(code, buildDir);
+
+    // install cdk dependencies
+    await processes.spawn('yarn', ['install'], {
```
RomainMuller: For good measure, I'd add a `--frozen-lockfile` option, to be safe.

```suggestion
    await processes.spawn('yarn', ['install', '--frozen-lockfile'], {
```

packages/@jsii/integ-test/test/build-cdk.test.ts
```diff
@@ -0,0 +1,96 @@
...
+      'lerna',
+      'run',
+      '--scope',
+      'cdk-build-tools',
+      '--scope',
+      'pkglint',
+      '--scope',
+      'awslint',
+      'build',
+    ], { cwd: srcDir });
```
RomainMuller: Maybe add `--stream`?

Also (and that is true for the other spawns, too), I kinda like logically grouping stuff, e.g:

```ts
    await processes.spawn('npx', [
        'lerna',
        '--stream',
        '--scope', 'cdk-build-tools',
        '--scope', 'pkglint',
        '--scope', 'awslint',
        'run', 'build',
      ], { cwd: srcDir });
```

packages/@jsii/integ-test/test/build-cdk.test.ts
```diff
@@ -0,0 +1,96 @@
...
+      '--stream',
+      '--scope',
+      '@aws-cdk/*',
+      '--scope',
+      'aws-cdk',
+      'build',
+      '--',
+      '--jsii',
+      path.join(JSII_DIR, 'bin', 'jsii'),
+    ], { cwd: srcDir });
```
RomainMuller: Instead of using the `--jsii path/to/jsii` argument, why not use the `CDK_BUILD_JSII` environment variable? Is there a difference?

packages/@jsii/integ-test/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+  "compilerOptions": {
+    "composite": false,
```
RomainMuller: Why? I reckon this also disables `incremental`...

packages/@jsii/integ-test/utils/index.ts
RomainMuller: I mean, sure, but you're still doing the `1000 * 60` multiplication at run-time, when it's a fairly obvious value... to hard-code beforehand?

packages/@jsii/integ-test/tsconfig.json
MrArnoldPalmer: tsc throws `tsconfig.json(3,5): error TS6304: Composite projects may not disable declaration emit.` ?

RomainMuller: Oh yeah, right. Composite requires declarations... but why bother?

MrArnoldPalmer: yeah idk, I wasn't quite sure the ideal solve here. Outputting declarations is fine? Just doesn't feel like we should need to do.

rix0rrr:
> This unit test requires a github access token defined in the environment

Wait, why not clone the `release` branch of the public repo, or download the latest release's source? Neither of those should need a GitHub token, rite?

MrArnoldPalmer:
> Wait, why not clone the `release` branch of the public repo, or download the latest release's source? Neither of those should need a GitHub token, rite?

We use the token to hit the GH API and get the tag of the latest release. Then using that tag we download the source from the archive. Neither of these actually requires a token but it gets rate limited pretty quickly without one.

Doing a `git clone` and checkout of the release branch wouldn't require a token, but I'd rather us know the specific tag we are checking out and not check out the latest `release` branch. In my mind its more robust? 🧐 Asking for "latest published release" vs `git checkout release` feels more decoupled.

PR #1234, opened 2020-02-03
feat(java): annotate nullability

RomainMuller:
This commit annotates values (parameters, properties, and method return
values) with the JetBrains annotations associated to the use-case:
- `@org.jetbrains.annotations.NotNull`
- `@org.jetbrains.annotations.Nullable`

While there is no standard nullability annotation in the Java Language Specification,
several alternatives exist - including [JSR-305](https://jcp.org/en/jsr/detail?id=305)
and [FindBugs](https://mvnrepository.com/artifact/com.google.code.findbugs/annotations).
The decision to use JetBrains' is based on the following reasoning:
- JSR-305 has been dormant, and may incur licensing issues (see google/guava#2960)
- JetBrains' annotations are licensed as Apache-2.0 (same as jsii itself), while FindBugs is LGPL
- One of the values of annotating is to improve Kotlin interoperability, so JetBrains annotations make the most sense in this context.

This can help IDEs provide static analysis of the code for discovery
of possible bugs; and will make usage from Kotlin nicer as it enables
use of the correct platform types.

Implements aws/aws-cdk#6026

---

### DIFF
--- a/packages/jsii-pacmak/lib/targets/java.ts
+++ b/packages/jsii-pacmak/lib/targets/java.ts
@@ -21,6 +21,9 @@ const spdxLicenseList = require('spdx-license-list');
 
 const BUILDER_CLASS_NAME = 'Builder';
 
+const ANN_NOT_NULL = '@org.jetbrains.annotations.NotNull';
+const ANN_NULLABLE = '@org.jetbrains.annotations.Nullable';
+
 /**
  * Build Java packages all together, by generating an aggregate POM
  *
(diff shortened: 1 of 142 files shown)


packages/jsii-pacmak/lib/targets/java.ts
```diff
@@ -1580,6 +1583,16 @@ class JavaGenerator extends Generator {
...
   }
 
+  private toDecoratedJavaType(optionalValue: spec.OptionalValue): string {
+    const nakedType = this.toJavaType(optionalValue.type);
+    return `${optionalValue.optional ? ANN_NULLABLE : ANN_NOT_NULL} ${nakedType}`;
+  }
+
+  private toDecoratedJavaTypes(optionalValue: spec.OptionalValue): string[] {
+    return this.toJavaTypes(optionalValue.type).map(nakedType =>
+      `${optionalValue.optional ? ANN_NULLABLE : ANN_NOT_NULL} ${nakedType}`);
```
eladb: can you reuse `toDecoratedJavaType` or otherwise find a way to avoid the duplicated code?

eladb:
Minor comments, added _do-not-merge_

PR #1241, opened 2020-02-06
feat(csharp): upgrade to netcoreapp3.1

RomainMuller:
From `netcoreap3.0`, in order to reflect the current
recommended framework version for .NET 3.x.

---

## Commit Message
feat(csharp): upgrade to netcoreapp3.1 (#1241)

As `netcoreapp3.0` is going to be EOL pretty soon (early March 2020),
upgrading our target framework to `netcoreapp3.1` early in order to
catch any surprises. Since `netcoreapp3.1` is nothing but an LTS version
of `netcoreapp3.0` (plus some bug fixes), it should be transparent to users
who may only need to upgrade their .NET SDK.


rix0rrr:
What is the impact of this? Specifically, what are the benefits and what are the costs?

* Are people using .NET core 3.1 currently unable to use jsii because we're targeting 3.0?
* Do people using .NET core 3 now have to upgrade to use jsii after this change?
* What are the advantages of upping our requirements?
* Does .NET core 3.1 have new features we are planning to take advantage of?
* Is .NET core 3.0 end of life?

To me, it seems nice to our customers to have as broad compatibility as possible for our software, to not force them to upgrade unnecessarily. It's why we target Java 8 (instead of 13) and NodeJS 10 (instead of 13) and Python 3.6 (instead of 3.7).

RomainMuller:
@rix0rrr - `netcoreapp3.0`is going EOL on March 3, 2020. I'd rather upgrade early and be able to revert while we work out wrinkles than do it after EOL. `netcoreapp3.1` is the LTS release.

The upgrade was recommended by our .NET SDK engineers, who strongly advised we upgrade our target framework to 3.1 on the first occasion (it was not released yet when we moved to `netcoreapp3.0`).

Ref: https://dotnet.microsoft.com/platform/support/policy/dotnet-core

RomainMuller:
(Note that this code requires a new Superchain image to be live before it'll pass PR validations)

PR #1261, opened 2020-02-15
chore: Add Prettier for Code Formatting

MrArnoldPalmer:
Adds prettier to auto-format code. Also adds eslint-plugin-prettier to
embed formatting into the lint workflow. Running `yarn lint` will report
any formatting mis-match as a lint error, `yarn lint:fix` will fix any
formatting related issues.

We are pretty close to the prettier defaults, if we cared to not have any extra configuration we could delete the .prettierrc.yaml file and use doublequotes for strings as the default. I don't really care either way personally so continuing with singlequotes just means less of a diff. This also removes all of the style-only related lint rules from our eslint config just to trim it down.

This obviously adds a lot of noise to the git history and can affect the usefulness of git blame. Just something to consider.

---


rix0rrr:
Although I'm personally not sure we need this (I don't feel like this is solving a problem I personally ever had, though I can't speak for other people obviously) I'm provisionally okay with this, but I must have IDE support for this otherwise it's going to cause grief to no end when my build fails for trivial reasons after I'm done typing. Can you write a little something in the CONTRIBUTING guide on how to configure inline feedback in the IDE?

Style-wise, not too keen on all the newlines it's adding everywhere, but I guess that's the price we pay...

packages/@jsii/integ-test/test/build-cdk.test.ts
```diff
@@ -27,62 +36,98 @@ describe('Build CDK', () => {
...
+          'aws-cdk',
+          '--ignore',
+          'cdk-build-tools',
+          '--ignore',
+          'pkglint',
+          '--ignore',
+          'awslint',
+          'run',
+          'build'
+        ],
```
RomainMuller: I don't like this change :/ It hurts readability :/

packages/@jsii/kernel/lib/api.ts
```diff
@@ -220,44 +220,46 @@ export interface NamingRequest {
 }
 
 export interface NamingResponse {
-  readonly naming: { readonly [language: string]: { readonly [key: string]: any } | undefined };
+  readonly naming: {
+    readonly [language: string]: { readonly [key: string]: any } | undefined;
+  };
```
RomainMuller: I find it odd that it applies to `[lamguage: string]` but not to `[key: string]`... I feel like they should be treated the same.

packages/@jsii/integ-test/test/build-cdk.test.ts
rix0rrr: I agree. Too many newlines.

skinny85:
I don't like the fact it removes trailing commas (not using trailing commas makes diffs look bad) :/. Not sure if this is a matter of configuration, or if prettier doesn't support them.

MrArnoldPalmer:
@rix0rrr Yeah in JSII I don't think this is really solving a lot of problems that we currently have. I think it would have more of an impact in aws-cdk if we decided we like this toolchain and want to add it there. I did this now after talking with @nija-at about inconsistencies in newlines at EOF in aws-cdk and possible fixes.

I agree that linebreaks are not ideal in some places. Prettier allows ignoring a node in the AST with `// prettier-ignore`. Added this on spawn `[...args]` to demonstrate.

Added notes about IDE tooling, but the long term idea is to add a github action that will run formatting on PRs and re-commit automatically. Then users can write new code however they prefer and it just gets normalized as it moves to master.

@skinny85 I agree, I changed the prettier "trailingComma" option to "all" instead of default which is "es5".

RomainMuller:
This looks like it got severely merge-conflicted. Would you mind rebasing/re-prettiftying so we can merge?

MrArnoldPalmer:
@RomainMuller dismissed review because I upgraded prettier to 2.0 which has better support for complex types in TS so some of the outputs have changed. This was required for TS 3.3 + private fields support.

PR #1263, opened 2020-02-17
chore(java): get rid of JSR-305 references

RomainMuller:
Those have been replaced with JetBrains annotations, which does not have the
possible licensing issues that JSR-305 has, and is generally speaking still being
maintained (while JSR-305 has been dormant for years now).

While touching on the `pom.xml` files, upgraded a couple of dependencies
that had newer and better versions, added the necessary plug-ins to make
sure this is easy to maintain going forward (`versions-maven-plugin`), and
that an appropriate version of `mvn` is used (`maven-enforcer-plugin`).

Finally, improved meta-information included in the `pom.xml` files around
licenses and attribution, to be more homogenous & de-duplicated.

None of the changes beyond replacing `JSII-305` with the JetBrains
annotations is really material in terms of the contents of the packages,
as those are essentially meta-information changes only. They could be
extracted out if we feel like we should split changes up.

---


packages/@jsii/java-runtime-test/pom.xml.t.js
```diff
@@ -29,7 +29,7 @@ process.stdout.write(`<?xml version="1.0" encoding="UTF-8"?>
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
-            <version>4.12</version>
+            <version>4.13</version>
```
rix0rrr: Title nor body mention this change.

packages/@jsii/java-runtime-test/pom.xml.t.js
```diff
@@ -53,6 +53,36 @@ process.stdout.write(`<?xml version="1.0" encoding="UTF-8"?>
                     <target>1.8</target>
                 </configuration>
             </plugin>
+
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-enforcer-plugin</artifactId>
```
rix0rrr: Title nor body mention this change.

packages/@jsii/java-runtime/pom.xml.t.js
```diff
@@ -20,18 +20,25 @@ process.stdout.write(`<?xml version="1.0" encoding="UTF-8"?>
 
     <licenses>
         <license>
-            <name>Apache License 2.0</name>
-            <url>http://www.apache.org/licenses/LICENSE-2.0</url>
+            <name>${packageInfo.license}</name>
```
rix0rrr: Title nor body mention this change.

packages/@jsii/java-runtime/pom.xml.t.js
```diff
@@ -40,7 +47,9 @@ process.stdout.write(`<?xml version="1.0" encoding="UTF-8"?>
 
     <scm>
         <connection>scm:${packageInfo.repository.type}:${packageInfo.repository.url}</connection>
+        <developerConnection>scm:${packageInfo.repository.type}:${packageInfo.repository.url}</developerConnection>
```
rix0rrr: Title nor body mention this change.

PR #1269, opened 2020-02-18
chore(mergify): no-squash label

eladb:
If a PR is labeled with "no-squash" this will perform a merge without squashing.

---

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -25,7 +25,7 @@ pull_request_rules:
       - status-success=continuous-integration/travis-ci/pr
       - status-success=Semantic Pull Request
 
-  - name: Synchronize that PR to upstream and merge it
+  - name: Synchronize that PR to upstream and merge it (squash)
     actions:
       dismiss_reviews:
         approved: true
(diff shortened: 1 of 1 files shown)


.mergify.yml
```diff
@@ -41,6 +41,7 @@ pull_request_rules:
       - label!=blocked
       - label!=do-not-merge
       - label=pr-status/ready-to-merge
+      - label=!no-squash
```
rix0rrr: !=

.mergify.yml
```diff
@@ -50,6 +51,33 @@ pull_request_rules:
...
+        method: merge
+        strict_method: merge
+    conditions:
+      - author!=dependabot[bot]
+      - author!=dependabot-preview[bot]
+      - -title~=(WIP|wip)
+      - label!=blocked
+      - label!=do-not-merge
+      - label=pr-status/ready-to-merge
+      - label=no-squash
```
rix0rrr: Is it `no-squash` or `pr-status/no-squash`?

In the other repo we do a `label~=~` for substring matching.

eladb: Yes, `pr-status/no-squash`. I will update.

Also, mergify docs say `=` should be sufficient to match against one label from the array.  `~=` is for regular expressions.

PR #1284, opened 2020-02-22
chore: Fix Integ Test Module Build

MrArnoldPalmer:
Adds an .npmignore file to @jsii/integ-test to specify inclusion of
built .js files in packaged tarball. Moved dependencies needed during
test runtime from devDependencies to dependencies so they are present
when running from tarball.

---


packages/@jsii/integ-test/package.json
```diff
@@ -6,7 +6,8 @@
   "scripts": {
     "build": "tsc --build && npm run lint",
     "lint": "eslint . --ext .ts --ignore-path=.gitignore",
-    "test:integ": "jest"
+    "test:integ": "jest",
+    "package": "package-js"
```
eladb: Curious why we need to package this if it's a private package?

MrArnoldPalmer:
Yeah we want to be able run the integ-tests against the built artifacts in our pipeline and not necessarily against the source code. This way we can take the build step artifacts and use the tarball to run our tests without recloning or rebuilding.

RomainMuller:
@MrArnoldPalmer the issue is that the publisher will actually try to publish that (and fail, because it's private). We need to somehow inhibit this behavior before it becomes a problem :)

MrArnoldPalmer:
@eladb @RomainMuller I added a `package-private` script to jsii-build tools for use in private packages that we still want to access post build. This puts the build artifact in `dist/private/` directory so it wont be published by delivlib.

PR #1287, opened 2020-02-24
fix(dotnet): missing ? on nullable interface members

### LINKED ISSUE #1285: jsii 1.0.0 fails aws-cdk build
## :bug: Bug Report

When upgrading the CDK to use jsii 1.0.0, the build is failing.

See aws/aws-cdk#6407

RomainMuller:
When interfaces declare nullable members, the code generator erroneously omitted the `?` that denotes it, while generated implementations of the interface would have it, causing compilation failures.

Fixes #1285

---

### DIFF
--- a/packages/jsii-pacmak/lib/targets/java.ts
+++ b/packages/jsii-pacmak/lib/targets/java.ts
@@ -605,7 +605,7 @@ class JavaGenerator extends Generator {
 
   protected onInterfaceMethod(_ifc: spec.InterfaceType, method: spec.Method) {
     this.code.line();
-    const returnType = method.returns ? this.toJavaType(method.returns.type) : 'void';
+    const returnType = method.returns ? this.toDecoratedJavaType(method.returns) : 'void';
     this.addJavaDocs(method);
     this.emitStabilityAnnotations(method);
     this.code.line(`${returnType} ${method.name}(${this.renderMethodParameters(method)});`);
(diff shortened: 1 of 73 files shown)


RomainMuller:
I have locally code-generated for the AWS CDK with this patch enabled, and all modules compiled fine.

packages/jsii-pacmak/lib/targets/java.ts
```diff
@@ -616,8 +616,8 @@ class JavaGenerator extends Generator {
   }
 
   protected onInterfaceProperty(_ifc: spec.InterfaceType, prop: spec.Property) {
-    const getterType = this.toJavaType(prop.type);
-    const setterTypes = this.toJavaTypes(prop.type);
+    const getterType = this.toDecoratedJavaType(prop);
+    const setterTypes = this.toDecoratedJavaTypes(prop);
```
eladb: How do we know we didn't forget more of these?

eladb:
This is a potential BREAKING CHANGE in .NET and Java, no?

RomainMuller:
In Java, I don't look at it as breaking (the annotations do not have any intrinsic compilation or runtime behavior. They may trigger linter rules by their presence, but generally speaking this will be in highlight of possible bugs, and shouldn't prevent compiling the code, still).

In DotNet, this may be breaking if the user's project enabled the `Nullable` feature, and configured to handle warnings as errors. All the nullable reference types do is add meta-information that will be used to emit new __warnings__ if code violates apparent nullability rules. While this could break some user's code, I believe it's a narrow enough niche that it may not be considered breaking still.

PR #1297, opened 2020-02-27
feat(jsii): introduce submodules feature

### LINKED ISSUE #1286: Submodules
### Affected Languages

- [x] `Python`
- [x] `Java`
- [x] .NET (`C#`, `F#`, ...)

### Description

We are looking into releasing the entire AWS CDK as a single module (see https://github.com/aws/aws-cdk-rfcs/pull/122).

In order to be able to organize the namespace of that huge single module we need a way to group types into "submodules".

The current way these types are exported is:

```ts
import * as aws_s3 from './aws-s3';
export aws_s3;

import * as aws_sns from './aws-sns';
export aws_sns;
```

This offers a good experience in TypeScript:

```ts
import { aws_s3, aws_sns } from 'monocdk';

new aws_s3.Bucket(this, ...);
```

We need this to translate properly to all supported jsii languages.

### Proposed Solution

A few things to consider:

1. What are we doing with submodule README files? How do we model them inside the jsii manifest? Maybe we need an entry for the submodule itself.
2. Ideally we want to be able to control the name of the submodule per-language. For example, Java submodules should translate into java package names, and we wish to keep the java package names in tact as we bundle all of the code into a single JAR. This means that we can't just use the name of the TypeScript submodule but rather need some form of translation table. It could be a custom jsdocs  attribute such as `@java.package foo.bar.foo`.

RomainMuller:
## Commit Message
feat(jsii): introduce submodules feature (#1297)

Introduces a `jsii` *Submodule* feature. *Submodules* are introduced
by declaring a `namespace`:
```ts
// Using a namespaced export statement
export * as ns from './subfolder';

// Using a namespace declaration
export ns {
  // ...
}
```

*Submodules* can be nested in other *submodules*, but the
dependency graph between submodules (and the top-level module)
must not result in a cycle (jsii` does currently **not** check for this
particular issue, although it probably will in the future).

Although this is out-of-scope for this change, they could be
enhanced so a specific `README.md` is attached to them, and so
that dedicated code-generating configuration can be specified.

Fixes #1286 

## End Commit Message

---


packages/jsii/lib/assembler.ts
```diff
@@ -292,7 +305,10 @@ export class Assembler implements Emitter {
       this._diagnostic(node, ts.DiagnosticCategory.Error, `Could not find module for ${modulePath}`);
       return `unknown.${typeName}`;
     }
-    const fqn = `${pkg.name}.${typeName}`;
+    const submoduleNs = this._namespaceMap.get(type.symbol)?.name;
+    const fqn = submoduleNs == null
+      ? `${pkg.name}.${typeName}`
+      : `${pkg.name}.${submoduleNs}.${typeName}`;
```
eladb: nit style: I would reverse the logic:

```suggestion
    const fqn = submoduleNs
      ? `${pkg.name}.${submoduleNs}.${typeName}`
      : `${pkg.name}.${typeName}`;
```

Feels more readable to me: "if there is a namespace, then use it, otherwise, dont"

packages/jsii/lib/assembler.ts
```diff
@@ -30,6 +30,9 @@ export class Assembler implements Emitter {
   private _deferred = new Array<DeferredRecord>();
   private _types: { [fqn: string]: spec.Type } = {};
 
+  /** Map of Symbol to namespace export Symbol */
+  private readonly _namespaceMap = new Map<ts.Symbol, ts.Symbol>();
```
eladb: Maybe rename to something like `submoduleMap` so it has semantic connection to submodules

packages/jsii/lib/assembler.ts
```diff
@@ -292,7 +305,10 @@ export class Assembler implements Emitter {
       this._diagnostic(node, ts.DiagnosticCategory.Error, `Could not find module for ${modulePath}`);
       return `unknown.${typeName}`;
     }
-    const fqn = `${pkg.name}.${typeName}`;
+    const submoduleNs = this._namespaceMap.get(type.symbol)?.name;
```
eladb:
```suggestion
    const submodule = this._submoduleMap.get(type.symbol)?.name;
```

packages/jsii/lib/assembler.ts
```diff
@@ -311,13 +327,105 @@ export class Assembler implements Emitter {
     }
   }
 
+  private _registerNamespaces(symbol: ts.Symbol): void {
```
eladb:
```suggestion
  private _registerSubmodule(symbol: ts.Symbol): void {
```

packages/jsii/lib/assembler.ts
```diff
@@ -311,13 +327,105 @@ export class Assembler implements Emitter {
...
+    );
+    if (resolution.resolvedModule == null) {
+      // Unresolvable module... We'll let tsc report this for us.
+      return;
+    }
+    if (resolution.resolvedModule.isExternalLibraryImport) {
+      // External re-exports are "pure-javascript" sugar; they need not be
+      // represented in the jsii Assembly since the types in there will be
+      // resolved through dependencies.
+      return;
```
eladb: Is this something we want to report/warn?

packages/jsii/lib/assembler.ts
```diff
@@ -311,13 +327,105 @@ export class Assembler implements Emitter {
...
+    const sourceFile = this.program.getSourceFile(resolution.resolvedModule.resolvedFileName)!;
+    const sourceModule = this._typeChecker.getSymbolAtLocation(sourceFile);
+    // If there's no module, it's a syntax error, and tsc will have reported it for us.
+    if (sourceModule) {
+      this._addToNamespace(symbol, sourceModule);
+    }
+  }
+
+  private _addToNamespace(ns: ts.Symbol, moduleLike: ts.Symbol) {
+    for (const symbol of this._typeChecker.getExportsOfModule(moduleLike)) {
```
eladb: Add a comment explaining what this loop is doing

packages/jsii/lib/assembler.ts
```diff
@@ -311,13 +327,105 @@ export class Assembler implements Emitter {
...
+      if (this._namespaceMap.has(symbol)) {
+        const currNs = this._namespaceMap.get(symbol)!;
+        if (currNs.name !== ns.name) {
```
eladb: Add a comment explaining what this check is

packages/jsii/lib/assembler.ts
```diff
@@ -311,13 +327,105 @@ export class Assembler implements Emitter {
...
+        }
+        // Found two re-exports, which is odd, but they use the same namespace,
+        // so it's probably okay? That's likely a tsc error, which will have
+        // been reported for us already anyway.
+        continue;
+      }
+      this._namespaceMap.set(symbol, ns);
+
+      const decl = symbol.declarations?.[0];
+      if (decl != null
```
eladb: add a comment such as "if this is a supported jsii type"...

Also, do we want a warning if it's not?

packages/jsii/lib/assembler.ts
```diff
@@ -311,13 +327,105 @@ export class Assembler implements Emitter {
...
+  }
+
   /**
    * Register exported types in ``this.types``.
    *
    * @param node       a node found in a module
    * @param namePrefix the prefix for the types' namespaces
    */
   private async _visitNode(node: ts.Declaration, context: EmitContext): Promise<spec.Type[]> {
+    if (ts.isNamespaceExport(node)) {
```
eladb: comment

packages/jsii/lib/assembler.ts
RomainMuller: I don't think so. I feel it's a legitimate use-case, and the fact we need not represent those in *jsii*-land is actually just an implementation detail... Of course this is subject to change if we discover this is surprising in some way.

packages/jsii/lib/assembler.ts
RomainMuller: I don't think the warning is useful (exports that cannot be modeled by *jsii* are logged somewhere else as part of the next pass).

docs/specifications/2-type-system.md
```diff
@@ -308,6 +308,86 @@ document.
...
+### Overview
+
+Typescript allows grouping declarations together in *namespaces*, which are
+interpreted by *jsii* as *submodules*. *Submodules* names are the fully
+qualified name of the namespace from the package's root (if a package `foo`
+defines a namespace `ns1`, which itself contains `ns2`, the submodule for `ns2`
+will be named `foo.ns1.ns2`).
+
+*Submodules* may use differnt [code-generation configuration](#code-generation)
+than thier parent submodule or package.
```
eladb:
```suggestion
than their parent submodule or package.
```

packages/jsii/test/negatives/neg.submodules-cannot-have-colliding-names.ts
```diff
@@ -0,0 +1,7 @@
+///!MATCH_ERROR: Submodule "ns1" conflicts with "Ns1".
+
+export * as ns1 from './namespaced';
+
+export class Ns1 {
```
eladb: Would `my_module` and `myModule` collide as well?

packages/jsii/lib/assembler.ts
```diff
@@ -348,6 +492,31 @@ export class Assembler implements Emitter {
...
+        submodule.name,
+        Case.camel(submodule.name),
+        Case.pascal(submodule.name),
+        Case.snake(submodule.name),
+      ]));
+      const colliding = candidates.find(name => `${this.projectInfo.name}.${name}` === jsiiType!.fqn);
+      if (colliding != null) {
+        const submoduleDecl = submodule.valueDeclaration ?? submodule.declarations[0];
+        this._diagnostic(node, ts.DiagnosticCategory.Error,
+          `Submodule "${submodule.name}" conflicts with "${jsiiType.name}". Restricted names are: ${candidates.join(', ')}`,
```
eladb: I guess top level types should always be pascal case (interfaces, classes, enums), so I think by definition they can't conflict with namespace names...

packages/jsii-reflect/test/__snapshots__/jsii-tree.test.js.snap
```diff
@@ -74,6 +41,192 @@ exports[`jsii-tree --all 1`] = `
...
+ │   │   ├─┬ operationsMap property (experimental)
+ │   │   │ ├── immutable
+ │   │   │ └── type: Map<string => Array<@scope/jsii-calc-lib.Value>>
+ │   │   ├─┬ curr property (experimental)
+ │   │   │ └── type: @scope/jsii-calc-lib.Value
+ │   │   ├─┬ maxValue property (experimental)
+ │   │   │ └── type: Optional<number>
+ │   │   └─┬ unionProperty property (experimental)
+ │   │     └── type: Optional<jsii-calc.Add | jsii-calc.Multiply | jsii-calc.Power>
+ │   ├─┬ class MethodNamedProperty (experimental)
```
eladb: Missing namespaces?

packages/jsii/test/negatives/neg.submodules-cannot-have-colliding-names.ts
RomainMuller: Yes. I guess I could add more coverage on this... Initially I only guarded against `PascalCase` / `camelCase` collisions, but figured Java & Python would use `snake_case`.

packages/jsii/lib/assembler.ts
RomainMuller: They can because the conflict space is `PascalCase` (preferred for C#),  `camelCase` (preferred for JS), `snake_case` (preferred for Python & Java).

packages/jsii-reflect/test/__snapshots__/jsii-tree.test.js.snap
RomainMuller: Yah the tool doesn't show those... Hmm. I'll see what should be the best way here.

RomainMuller:
I sure can undo the name spacing of that much code. It felt *appropriate*, but I reckon this can be re-introduced separately (assuming the feature works correctly, the review burden for that would be lower).

RomainMuller:
This is now de-namespaced & submodules render in `jsii-reflect`.

packages/@jsii/kernel/test/kernel.test.ts
```diff
@@ -79,7 +79,7 @@ defineTest('deleteObject will remove the reference', (sandbox) => {
 });
 
 defineTest('in/out primitive types', (sandbox) => {
-  const alltypes = sandbox.create({ fqn: 'jsii-calc.AllTypes', args: [] });
+  const alltypes = sandbox.create({ fqn: 'jsii-calc.compliance.AllTypes', args: [] });
```
eladb: Didn't you remove the submodules from jsii-calc?

packages/jsii-pacmak/test/expected.jsii-calc-base-of-base/python/src/scope/jsii_calc_base_of_base/__init__.py
```diff
@@ -2,13 +2,13 @@
...
 import enum
+import publication
 import typing
 
 import jsii
 import jsii.compat
-import publication
 
-__jsii_assembly__ = jsii.JSIIAssembly.load("@scope/jsii-calc-base-of-base", "1.1.0", __name__, "jsii-calc-base-of-base@1.1.0.jsii.tgz")
+__jsii_assembly__ = jsii.JSIIAssembly.load("@scope/jsii-calc-base-of-base", "1.1.0", "scope.jsii_calc_base_of_base", "jsii-calc-base-of-base@1.1.0.jsii.tgz")
```
eladb: Will this break the world, or is `__name__` equivalent?

packages/jsii-pacmak/test/expected.jsii-calc-lib/python/src/scope/jsii_calc_lib/_jsii/__init__.py
```diff
@@ -2,13 +2,14 @@
...
 import enum
+import publication
 import typing
 
 import jsii
 import jsii.compat
-import publication
 
 import scope.jsii_calc_base
+
```
eladb: Is this change actually needed or can we revert? (try to reduce the blast radius of your change)

packages/jsii-pacmak/test/expected.jsii-calc/dotnet/Amazon.JSII.Tests.CalculatorPackageId/Amazon/JSII/Tests/CalculatorNamespace/Calculator.cs
```diff
@@ -28,7 +28,7 @@ namespace Amazon.JSII.Tests.CalculatorNamespace
     /// Console.WriteLine(calculator.Expression.Value);</code>
     /// </example>
     [JsiiClass(nativeType: typeof(Amazon.JSII.Tests.CalculatorNamespace.Calculator), fullyQualifiedName: "jsii-calc.Calculator", parametersJson: "[{\"docs\":{\"summary\":\"Initialization properties.\"},\"name\":\"props\",\"optional\":true,\"type\":{\"fqn\":\"jsii-calc.CalculatorProps\"}}]")]
-    public class Calculator : Amazon.JSII.Tests.CalculatorNamespace.composition.CompositeOperation
+    public class Calculator : Amazon.JSII.Tests.CalculatorNamespace.Composition.CompositeOperation
```
eladb: This technically indicates that we have a breaking change here. I am okay with it since we don't use it in the CDK, but I am wondering what's the right thing to do here..

packages/jsii-pacmak/test/expected.jsii-calc/dotnet/Amazon.JSII.Tests.CalculatorPackageId/Amazon/JSII/Tests/CalculatorNamespace/Submodule/Child/Structure.cs
```diff
@@ -0,0 +1,25 @@
+using Amazon.JSII.Runtime.Deputy;
+
+#pragma warning disable CS0672,CS0809,CS1591
+
+namespace Amazon.JSII.Tests.CalculatorNamespace.Submodule.Child
+{
+    #pragma warning disable CS8618
```
eladb: Why is this needed? Curious...

packages/jsii-pacmak/test/expected.jsii-calc/java/src/main/java/software/amazon/jsii/tests/calculator/derived_class_has_no_properties/Base.java
```diff
@@ -1,4 +1,4 @@
-package software.amazon.jsii.tests.calculator.DerivedClassHasNoProperties;
+package software.amazon.jsii.tests.calculator.derived_class_has_no_properties;
```
eladb: Why is this happening?

packages/jsii-pacmak/test/expected.jsii-calc/python/src/jsii_calc/__init__.py
```diff
@@ -31,17 +31,18 @@
 import builtins
```
eladb: I expected 0 changes in this file as proof that this is not going to break anyone... Can we avoid?

packages/jsii-pacmak/test/expected.jsii-calc/java/src/main/java/software/amazon/jsii/tests/calculator/derived_class_has_no_properties/Base.java
RomainMuller: Because `DerivedClassHasNoProperty` is a namespace... so it gets case-adjusted like it should.

packages/jsii-pacmak/test/expected.jsii-calc/dotnet/Amazon.JSII.Tests.CalculatorPackageId/Amazon/JSII/Tests/CalculatorNamespace/Submodule/Child/Structure.cs
RomainMuller: `CS8618` is a warning about non-nullable fields that have no initializer... This warning is immensely difficult to avoid in the generated code because of how we use dynamic properties... There's no initial assignment because the value is never stored in C# (it is read from and written to the JS process).

The disable is emitted systematically in all generated files so that stuff compiles "cleanly".

RomainMuller:
@elad

> in monocdk, I want to put all the "core" types under the root namespace and then submodules will reference those types

Do you mean `export * from './core';` and then a bunch of `export * as aws_svc from './aws-svc';`?

packages/jsii-pacmak/test/expected.jsii-calc/python/src/jsii_calc/__init__.py
RomainMuller: Could, but it breaks the logical grouping of imports and it's a little sucky. This is literally just a line moving a couple lines higher... I don't see how this is going to break anything or anyone...

packages/jsii-pacmak/test/expected.jsii-calc-base-of-base/python/src/scope/jsii_calc_base_of_base/__init__.py
RomainMuller: `__name__` is equivalent in the existing places, and is not (as needed -- which is why it changed) in submodules.

packages/jsii-pacmak/test/expected.jsii-calc/dotnet/Amazon.JSII.Tests.CalculatorPackageId/Amazon/JSII/Tests/CalculatorNamespace/Calculator.cs
RomainMuller: Not doing it means permanently adopting non-idiomatic names for namespaces in the various front-end languages.

packages/@jsii/kernel/test/kernel.test.ts
RomainMuller: This is a namespace that was _already_ there.

RomainMuller: This explains why tests won't pass :)

packages/jsii-pacmak/test/expected.jsii-calc/dotnet/Amazon.JSII.Tests.CalculatorPackageId/Amazon/JSII/Tests/CalculatorNamespace/Calculator.cs
eladb: I am okay with hoping for the best.

packages/jsii-pacmak/test/expected.jsii-calc/dotnet/Amazon.JSII.Tests.CalculatorPackageId/Amazon/JSII/Tests/CalculatorNamespace/Submodule/Child/Structure.cs
eladb: ok. thanks

packages/jsii-pacmak/test/expected.jsii-calc/java/src/main/java/software/amazon/jsii/tests/calculator/derived_class_has_no_properties/Base.java
eladb: got it. ok

packages/jsii-pacmak/test/expected.jsii-calc/python/src/jsii_calc/__init__.py
eladb: I am fine with leaving this as is, but in the future, please resist the urge to introduce unrelated changes in PRs. As much as I understand the desire to make small improvements like this, by definition it introduces risk to the project because it makes the PR harder to review and harder to reason about the implications of the actual change. If you want to reorganize the imports, you can easily just submit a PR that includes only this change and it will be easy to approve and merge...

eladb:
> @eladb
> 
> > in monocdk, I want to put all the "core" types under the root namespace and then submodules will reference those types
> 
> Do you mean `export * from './core';` and then a bunch of `export * as aws_svc from './aws-svc';`?

Exactly, I want the user experience to look like this:

```ts
import { Construct, App, aws_s3, aws_sns } from 'monocdk-experiment';
```

Makes sense?

RomainMuller:
Makes sense. I'm not certain it's the best ergonomics, but you don't know before you try, and it's still a fairly reasonable thing to do anyway.

eladb:
> Makes sense. I'm not certain it's the best ergonomics, but you don't know before you try, and it's still a fairly reasonable thing to do anyway.

Regardless if we are doing this in monocdk or not, it should be supported and ideally we should have a test case for it in jsii-calc.

packages/jsii-pacmak/lib/targets/python.ts
```diff
@@ -982,51 +1072,56 @@ class EnumMember implements PythonBase {
...
     this.package = opts.package;
-    this.members = [];
   }
 
   public addMember(member: PythonBase) {
     this.members.push(member);
   }
 
+  public dependsOnModules(resolver: TypeResolver) {
+    resolver = this.fqn ? resolver.bind(this.fqn, this.pythonName) : resolver;
```
eladb: this.boundResolver?

RomainMuller: That's not a thing on this particular type, due to peculiarities in the inheritance chain here.

PR #1301, opened 2020-02-28
chore: add contribution/core label to core team PRs

nija-at:
---


.mergify.yml
```diff
@@ -1,6 +1,13 @@
 # See https://doc.mergify.io
 
 pull_request_rules:
+  - name: label core
```
eladb: Can we reuse this label in discard stale reviews below, so we don't have to maintain this twice?

nija-at: Nice idea.

PR #1305, opened 2020-02-29
chore: Fix integ test module install strategy

MrArnoldPalmer:
Change the way we are installing local versions of jsii modules to
ensure that they are present where they are needed during cdk build. A
conflict of versions was causing packaging to stall out during dotnet
build.

---

### DIFF
--- a/packages/@jsii/integ-test/test/build-cdk.test.ts
+++ b/packages/@jsii/integ-test/test/build-cdk.test.ts
@@ -1,6 +1,6 @@
 import * as Octokit from '@octokit/rest';
 import * as dotenv from 'dotenv';
-import { mkdtemp, remove } from 'fs-extra';
+import { readdir, mkdtemp, remove } from 'fs-extra';
 import * as path from 'path';
 import { downloadReleaseAsset, minutes, ProcessManager, extractFileStream } from '../utils';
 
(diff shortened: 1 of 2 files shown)


MrArnoldPalmer:
I'm not quite sure why building specific modules in the order we were before and passing in the jsii executable locations was causing this. I think that there was a problem between the workspace global dependencies (jsii-rosetta) and the workspace local dependencies in cdk-build (jsii(?-pacmak)). This way we don't have to know so much about the modules in CDK and just install local versions into cdk-build and the global workspace then run `build` + `pack`.

packages/@jsii/integ-test/test/build-cdk.test.ts
```diff
@@ -44,45 +44,22 @@ describe('Build CDK', () => {
...
-    await processes.spawn('./pack.sh', [], {
-      cwd: srcDir,
-      env: {
-        PACMAK: path.join(JSII_PACMAK_DIR, 'bin', 'jsii-pacmak'),
-        ROSETTA: path.join(JSII_ROSETTA_DIR, 'bin', 'jsii-rosetta')
-      }
+    // package modules with pacmak
+    await processes.spawn('yarn', ['pack'], {
+      cwd: srcDir
     });
```
nija-at: Should we sanity check that the generated pack(s) contain some basic set of artifacts? To prevent this test from succeeding but not actually doing anything.

MrArnoldPalmer: good call out, I will add a step to verify the presence of the various language target directories in `srcDir/dist`

PR #1318, opened 2020-03-06
chore: align versions in build test

MrArnoldPalmer:
Update integ test to include new step for aligning versions. This was
breaking packaging in other languages because it was trying to build
against v0.0.0.

---

### DIFF
--- a/packages/@jsii/integ-test/test/build-cdk.test.ts
+++ b/packages/@jsii/integ-test/test/build-cdk.test.ts
@@ -52,6 +52,11 @@ describe('Build CDK', () => {
       cwd: srcDir
     });
 
+    // align versions for packaging
+    await processes.spawn('./scripts/align-version.sh', [], {
+      cwd: srcDir
+    });
+
     // build cdk modules
     await processes.spawn('npx', ['lerna', 'run', 'build'], {
       cwd: srcDir


packages/@jsii/integ-test/test/build-cdk.test.ts
```diff
@@ -52,6 +52,11 @@ describe('Build CDK', () => {
       cwd: srcDir
     });
 
+    // align versions for packaging
+    await processes.spawn('./scripts/align-version.sh', [], {
+      cwd: srcDir
+    });
+
```
nija-at: This whole integ test feels very flaky, in the way it's written.
Any change to the cdk build process is going to impact this test and needs to be updated. 

Why not run, `./install.sh`, `./build.sh` and `./pack.sh` instead, and somewhere in-between those commands, configure lerna/yarn to use the local jsii packages? 
One option could be to make these parameters of the `install.sh`.

MrArnoldPalmer: We can do that but that's essentially what's already happening here.

`npx lerna run build` is just running `build` instead of `build+test` which is run in build.sh.

We can change to calling install.sh and pack.sh, but `yarn install` and `yarn run pack` feel like they actually require less knowledge of the cdks build system than knowing where/what the specific scripts are.

None of install/build/package scripts call align-versions.sh currently, so we still have to run that from our end and this still would have broken.

We went the route of passing different jsii executables to the package script, but the build script has no real way of handling this because it doesn't differentiate between calling build on a jsii package, and on a non jsii package and so it breaks passing an invalid cli arg to a yargs program that's in strict mode (aka fails on unknown arguments).

PR #1330, opened 2020-03-11
fix(python): `self` as property name leads to assignment error

skorfmann:
When a property with the name `self` exists in a class, it leads to
a runtime error. This slugifies the conventional `self` parameter
as needed to avoid this problem without compromising on the
developer experience.

---


RomainMuller:
Thanks for this contribution!

This is currently missing a test that demonstrates the issue is fixed.

skorfmann:
> This is currently missing a test that demonstrates the issue is fixed.

Could you point me to a similar test? 

> I think right now this will cause a warning to be emitted, but it will not fix code generation so the name is correctly slugified.

Yes, I think so too. What would be the correct way to slugify this?

RomainMuller:
With respects to the slugification, there appears to be a duplicated reserved words list in `jsii-pacmak` that fails to include `self` (bonus points if you find a way to de-duplicate, but I'm cool if you don't want to bother with that, too):
https://github.com/aws/jsii/blob/1ff25571a08652c59b23d6a2938dd3f426cd2665/packages/jsii-pacmak/lib/targets/python.ts#L71-L76

---

With respects to the tests... This is going to involve:
1. Adding stuff in `jsii-calc` to reproduce the issue you've identified:
    https://github.com/aws/jsii/blob/1ff25571a08652c59b23d6a2938dd3f426cd2665/packages/jsii-calc/lib/compliance.ts#L896-L961
2. Adding a new test in the Python runtime's test suite that "runs into the bug" (which would pass with the fix in, so you know you've nailed it for good!)
   https://github.com/aws/jsii/blob/1ff25571a08652c59b23d6a2938dd3f426cd2665/packages/%40jsii/python-runtime/tests/test_compliance.py#L1168-L1170

In any case - we'll happily provide help if you don't know how to proceed!

RomainMuller:
Reading up a bit, it appears that `self` is *technically* not a keyword in Python, so we're abusing a little here if we add it to the `PYTHON_KEYWORDS` list in `jsii-pacmak` (it's actually "fine" to have a warning about it in `jsii`, still).

I reckon the real fix is going to possibly be with code-generation of instance methods in `jsii-pacmak` (the `targets/python.ts` file is where it's at).

I can let you have a look here and see if you find your way around what needs to change... And then I can have a look, too, if you get nowhere or can't afford to spend that kind of time on this :)

skorfmann:
Cool, thanks for providing me the context. I'll take a look at it.

RomainMuller:
I have improved the solution here, but realized the current Submodules feature of `jsii` generates invalid Python code for the test I would like to introduce (making it impossible to load).

This would need fixing before the compliance test can be added, however we *could* ship as-is since the original bug is likely addressed. I'd still kind of prefer being able to add the next test however (because I'm not sure the instance property/method name conflict results in code that'll behave as intended).

PR #1335, opened 2020-03-15
feat: add autobump script

MrArnoldPalmer:
Changes all version numbers in package.json to 0.0.0. to centralize current version to a single location in lerna.json. This removes noise from release PRs.

Adds scripts for aligning package versions to the current version. Changes content of `scripts/bump.sh` to only bump

**Note**:
Changing to using `standard-version` for bumping and changelog from lerna means that all changes are documented in the root `CHANGELOG.md` instead of within individual package folders.

It is possible to continue to use lerna to bump versions, however a couple of issues come up with that.

1. running `lerna publish` or `lerna version` requires you to be attached to a branch. Most CI environments check out an individual commit and are are in a detached state. This means we would have to re-attach somehow.

2. bumping with lerna replaces all of the `"version": "0.0.0"` fields in package's `package.json` files with the newly bumped version. So release PRs still have the version noise.

In summay, if we want to have changelogs in package dirs, we probably can't have "v0.0.0" in source code. If we want "v0.0.0" in source code, we should go this way and have a central changelog. Central version = central changelog.

[Example Release PR](https://github.com/MrArnoldPalmer/jsii/pull/4)

---


scripts/align-version.js
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env node
```
eladb: Should we extract this from CDK into a stupid little node module instead of copying and pasting?

scripts/align-version.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
eladb: same

scripts/bump.sh
```diff
@@ -1,29 +1,23 @@
...
-
-# rebuild everything
-yarn build
-
-# update test expectations
-yarn test:update
+# /bin/bash ./install.sh
 
+# Generate CHANGELOG and create a commit (see .versionrc.json)
+npx standard-version --release-as ${version}
```
eladb: I am totally fine with moving to standard-version here

scripts/align-version.js
MrArnoldPalmer: I'm cool with it. Cute name ideas?

eladb:
@MrArnoldPalmer shall we get this merged?

MrArnoldPalmer:
Yeah I'm fixing the tests here to remove version #s instead of aligning before tests. Will do this now.

packages/@scope/jsii-calc-base/test/assembly.jsii
```diff
@@ -150,6 +150,6 @@
       "name": "IBaseInterface"
     }
   },
-  "version": "1.1.0",
-  "fingerprint": "piaiqNxSj1XkycnXHPRG4697t6DpKGgkya6effAlXKI="
+  "version": "0.0.0",
```
eladb: Oh cool we no longer need to update tests in version bumps!

MrArnoldPalmer: yeah, this does add some complexity to how we build though.

Since tests rely on version numbers not changing, we have to build+test before we run align-version in our build pipeline. Then after running align-version, I'm wary of running package.sh without running build again because if for some reason we changed how align-version works or otherwise do something that could affect the build, then I'd like to know about that failure.

Because of this I actually changed buildspec.yml to do build+test+align-version+build+pack. This seems to be the safest thing for now? @eladb am I being irrational by worrying about this?

eladb: Yes that sounds like what I would do. Not ideal but will work I believe until we figure out some other way...

PR #1344, opened 2020-03-17
fix(jsii): out-of-source builds are broken

### LINKED ISSUE #1273: JSII does not work properly with out of source build
### Affected Languages
This bug affects the generation of jsii assemblies

- [x] `Python`
- [x] `Java`
- [x] .NET (`C#`, `F#`, ...)

### General Information
* **JSII Version:** 0.22.0 (build 14afdde), typescript 3.7.5
* **Platform:**     OSX 10.14.6

### What is the problem?

When using the jsii options to override tsc outDir, no types are included in the .jsii assembly. This appears to be because of how the main file for the assembly is located: https://github.com/aws/jsii/blob/591a5d97c1f1f55019ae88a39fc9a09ca6f8e67e/packages/jsii/lib/assembler.ts#L74 The use of the regex here to replace .d.ts with .ts assumes that the source file that created the typings file is in the same directory, which it won't be when tsc outDir is set. It seems jsii should either use the typings file as is (if that's possible), or have a separate configuration option for where the main *source* file is.

### Verbose Log

No useful logs.

RomainMuller:
When configuring tsc.outDir, the jsii compiler would not look for the
correct source file and captured no types from the module. This
commit fixes this problem by reversing the types attribute in order
to resolve back to the corresponding source file.

Fixes #1273

---

### DIFF
--- a/packages/@scope/jsii-calc-lib/package.json
+++ b/packages/@scope/jsii-calc-lib/package.json
@@ -23,8 +23,8 @@
   "engines": {
     "node": ">= 10.3.0"
   },
-  "main": "lib/index.js",
-  "types": "lib/index.d.ts",
+  "main": "build/index.js",
+  "types": "build/index.d.ts",
   "scripts": {
     "build": "jsii && jsii-rosetta",
     "test": "diff-test test/assembly.jsii .jsii",
@@ -34,7 +34,8 @@
     "@scope/jsii-calc-base": "^1.1.0"
   },
   "peerDependencies": {
-    "@scope/jsii-calc-base": "^1.1.0"
+    "@scope/jsii-calc-base": "^1.1.0",
+    "@scope/jsii-calc-base-of-base": "^1.1.0"
   },
   "devDependencies": {
     "@types/node": "^10.17.17",
@@ -62,6 +63,9 @@
         "distName": "scope.jsii-calc-lib",
         "module": "scope.jsii_calc_lib"
       }
+    },
+    "tsc": {
+      "outDir": "build"
     },
     "versionFormat": "short"
   }
--- a/packages/jsii-pacmak/test/expected.jsii-calc-lib/python/src/scope/jsii_calc_lib/_jsii/__init__.py
+++ b/packages/jsii-pacmak/test/expected.jsii-calc-lib/python/src/scope/jsii_calc_lib/_jsii/__init__.py
@@ -9,6 +9,7 @@ import jsii.compat
 import publication
 
 import scope.jsii_calc_base
+import scope.jsii_calc_base_of_base
 __all__ = []
 
 publication.publish()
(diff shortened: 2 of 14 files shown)


packages/@scope/jsii-calc-lib/package.json
```diff
@@ -34,7 +34,8 @@
     "@scope/jsii-calc-base": "^1.1.0"
   },
   "peerDependencies": {
-    "@scope/jsii-calc-base": "^1.1.0"
+    "@scope/jsii-calc-base": "^1.1.0",
+    "@scope/jsii-calc-base-of-base": "^1.1.0"
```
eladb: Why is this now required? Weird.

packages/jsii-pacmak/test/expected.jsii-calc-lib/python/src/scope/jsii_calc_lib/_jsii/__init__.py
```diff
@@ -9,6 +9,7 @@
 import publication
 
 import scope.jsii_calc_base
+import scope.jsii_calc_base_of_base
```
eladb: How is this change related to out of source builds..?

packages/jsii/lib/assembler.ts
```diff
@@ -38,7 +40,22 @@ export class Assembler implements Emitter {
   public constructor(
     public readonly projectInfo: ProjectInfo,
     public readonly program: ts.Program,
-    public readonly stdlib: string) { }
+    public readonly stdlib: string) {
+    const dts = projectInfo.types;
+    let mainFile = dts.replace(/\.d\.ts(x?)$/, '.ts$1');
+
+    const tscOutDir = program.getCompilerOptions().outDir;
```
eladb: Add a comment explaining what's going on here and why it's needed.

packages/jsii/lib/compiler.ts
```diff
@@ -223,8 +223,18 @@ export class Compiler implements Emitter {
...
-      exclude: ['node_modules'].concat(pi.excludeTypescript),
+      include: [
+        pi.tsc?.rootDir
+          ? path.join(pi.tsc.rootDir, '**', '*.ts')
+          : path.join('**', '*.ts')
+      ],
+      exclude: [
+        'node_modules',
+        ...pi.excludeTypescript,
+        ...pi.tsc?.outDir != null
```
eladb: For consistency:

```suggestion
        ...pi.tsc?.outDir
```

packages/@scope/jsii-calc-lib/package.json
RomainMuller: Always has been, jsii emits a warning for it, which has always been ignored. The package exposes a type from jsii-calc-base-of-base, so it needs to declare a peered on it.

packages/jsii-pacmak/test/expected.jsii-calc-lib/python/src/scope/jsii_calc_lib/_jsii/__init__.py
RomainMuller: This is due to the new peerDependency

packages/jsii/lib/compiler.ts
RomainMuller: I'd rather fix the other way around (empty sting is falsey)

PR #1354, opened 2020-03-19
fix(diff): check base types

rix0rrr:
Check that an updated type still extends all of its previous base types,
so that objects of the updated type can still be assigned to variables
or passed into parameters of the declared supertype.

This check would have prevented a deployment failure during a big
refactoring of the CDK.

---

### DIFF
--- a/packages/jsii-diff/lib/classes-ifaces.ts
+++ b/packages/jsii-diff/lib/classes-ifaces.ts
@@ -1,7 +1,7 @@
 import * as reflect from 'jsii-reflect';
 import * as log4js from 'log4js';
 import { compareStabilities } from './stability';
-import { Analysis, FailedAnalysis, isSuperType } from './type-analysis';
+import { Analysis, FailedAnalysis, isSuperType, isNominalSuperType } from './type-analysis';
 import { ComparisonContext } from './types';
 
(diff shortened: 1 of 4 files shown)


packages/jsii-diff/lib/classes-ifaces.ts
```diff
@@ -296,3 +314,49 @@ function isCompatibleArgumentType(original: reflect.TypeReference, updated: refl
...
+ * ```
+ * const variable: B = new T();
+ * ```
+ *
+ * This code needs to be valid in the updated assembly, so for each
+ * B an updated type B' needs to exist in the new assembly which is
+ * still a supertype of T'.
+ */
+function assignableToAllBaseTypes(original: reflect.ReferenceType, updated: reflect.ReferenceType): Analysis {
+  for (const B of baseTypes(original)) {
```
eladb:
```suggestion
  for (const base of baseTypes(original)) {
```

rix0rrr: Yeah. Well it was `B` on purpose to match the docstring just above it.

 But sure.

rix0rrr:
Cool. Now we just have to wait for @MrArnoldPalmer

PR #1368, opened 2020-03-24
feat(compliance): produce compliance test reports

RomainMuller:
## Commit Message
feat(compliance): produce compliance test reports (#1368)

Generates reports (xUnit or JSON) out of compliance test executions
so that we can assert conformance based on the spec.

## End Commit Message

---


RomainMuller:
@MrArnoldPalmer - a review would be appreciated :)

PR #1406, opened 2020-03-26
fix(jsii): single-valued enums are not processed correctly

RomainMuller:
## Commit Message

TypeScript represents enums similar to union types, and single-valued
enums are "simplified" to the sole member (the TypeChecker fiercely
refuses to give a handle to the actual `enum` type). This resulted in
`jsii` incorrectly tagging the type at usage sites.

This commit adds the necessary infrastructure to detect single-valued
enums and tweak the FQN generation to obtain the correct result.

A new test was added to validate this whole endeavor works correctly
even when the single-valued enum is within a submodule (which adds even
more complexity to the mix).

References: aws/aws-cdk#6712 aws/aws-cdk#6948

## End Commit Message

---


RomainMuller:
@Mergifyio refresh

PR #1461, opened 2020-04-02
chore(python): improve test harness slightly

RomainMuller:
Tuned the test script so that it does not depend on globally installed
pip packages (such as `wheel`), and uses what is in the `venv` more
consistently.

Additionally, upgraded versions of Python dependencies.

---


RomainMuller:
@mergify refresh

PR #1491, opened 2020-04-06
feat(jsii): exclude files in tsconfig.json

nija-at:
## Commit Message

This change lets package.json of a jsii package to specify additional
'exclude' paths. They are encoded into tsconfig.json.

Motivation
jsconfig is a common way to configure tools like eslint and jest. These
files have the 'js' extension but should not be compiled into the
package. This flag lets the jsii compiled packages to configure this.
## End Commit Message

---

### DIFF
--- a/packages/jsii/lib/project-info.ts
+++ b/packages/jsii/lib/project-info.ts
@@ -45,7 +45,9 @@ export interface ProjectInfo {
   readonly description?: string;
   readonly homepage?: string;
   readonly contributors?: readonly spec.Person[];
-  readonly excludeTypescript: string[];
+  /** @deprecated - use excludes */
+  readonly excludeTypescript?: string[];
+  readonly exclude: string[];
   readonly projectReferences?: boolean;
   readonly tsc?: TSCompilerOptions;
 }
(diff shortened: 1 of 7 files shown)


eladb:
Not sure what’s I am seeing here? Seems like we already have this feature under a different name, no?

nija-at:
Are you referring to the property `excludeTypescript`? It wasn't clear why there's the `typescript` suffix, so I've tried to unify it under one property called `exclude`.

eladb:
> Are you referring to the property `excludeTypescript`? It wasn't clear why there's the `typescript` suffix, so I've tried to unify it under one property called `exclude`.

I’ll let @RomainMuller chime in on this.

RomainMuller:
> > Are you referring to the property `excludeTypescript`? It wasn't clear why there's the `typescript` suffix, so I've tried to unify it under one property called `exclude`.
> 
> I’ll let @RomainMuller chime in on this.

Well that was introduced in #343, and since the PR was mostly about something else it flew in.

But I'm generally favorable of modeling all the TypeScript compilation tweaks as an object under the `jsii.tsc` key, where the key for each option is the same as what's in `tsconfig.json`.

I am also wondering if we took a wrong turn in generating the `tsconfig.json` systematically, and whether we should actually allow a hand-written file with the other `jsii` configuration in there -- but this is a conversation for later.

packages/jsii/lib/project-info.ts
```diff
@@ -142,6 +144,7 @@ export async function loadProjectInfo(projectRoot: string, { fixPeerDependencies
             && (pkg.contributors as any[]).map((contrib, index) => _toPerson(contrib, `contributors[${index}]`, 'contributor')),
 
     excludeTypescript: (pkg.jsii && pkg.jsii.excludeTypescript) || [],
+    exclude: (pkg.jsii && pkg.jsii.exclude) || [],
```
RomainMuller, line 147: Breaks surrounding style, and you might not want to fix them all up, so take it or leave it:

```suggestion
    exclude: pkg.jsii?.exclude ?? [],
```

Issue #1501, opened 2020-04-08
Python is broken in jsii 1.2.0 when jsii modules are installed as submodules


eladb:
### Affected Languages

- [x] `Python`

### What is the problem?

[cdk8s](https://github.com/awslabs/cdk8s) has an `import` command which uses jsii to generate python _code_ on-demand and puts it in the user project under `imports/<module>`. 

For example:

```
cdk8s import k8s
```

Will create a jsii python module under `$project_root/imports/k8s` which users then import via:

```py
from imports import k8s
```

When the cdk8s CLI is compiled with jsii 1.2.0, the generated python code will produce a module that cannot be imported this way and fails with the following error:

```
Traceback (most recent call last):
  File "__init__.py", line 14, in <module>
    __jsii_assembly__ = jsii.JSIIAssembly.load("k8s", "0.0.0", "k8s", "k8s@0.0.0.jsii.tgz")
  File "/Users/benisrae/.local/share/virtualenvs/test-sodC6YKc/lib/python3.7/site-packages/jsii/_runtime.py", line 39, in load
    f"{assembly.module}._jsii", assembly.filename
  File "/usr/local/opt/python/Frameworks/Python.framework/Versions/3.7/lib/python3.7/contextlib.py", line 112, in __enter__
    return next(self.gen)
  File "/usr/local/opt/python/Frameworks/Python.framework/Versions/3.7/lib/python3.7/importlib/resources.py", line 184, in path
    package = _get_package(package)
  File "/usr/local/opt/python/Frameworks/Python.framework/Versions/3.7/lib/python3.7/importlib/resources.py", line 47, in _get_package
    module = import_module(package)
  File "/usr/local/opt/python/Frameworks/Python.framework/Versions/3.7/lib/python3.7/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1006, in _gcd_import
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 953, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1006, in _gcd_import
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 965, in _find_and_load_unlocked
[...]

PR #1502, opened 2020-04-08
feat(submodules): allow python submodule name configuration

RomainMuller:
Using a `.jsiirc.json` file, the name generation for entities within a
submodule can be configured ot use a different name than the default.

---


packages/jsii-pacmak/test/expected.jsii-calc/python/src/jsii_calc/submodule/nested_submodule/__init__.py
```diff
@@ -8,15 +8,20 @@
...
+    Goodness as _Goodness_2df26737,
+)
+from .deeply_nested import (
+    INamespaced as _INamespaced_e2f386ad,
+)
 import scope.jsii_calc_base
 import scope.jsii_calc_base_of_base
 import scope.jsii_calc_lib
 
 __jsii_assembly__ = jsii.JSIIAssembly.load("jsii-calc", "0.0.0", "jsii_calc", "jsii-calc@0.0.0.jsii.tgz")
```
eladb, line 21: Are you planning to also fix #1501 as part of this commit (i.e. always import the root module so you can use `__name__` instead of an explicit name in the `.load()` call?

packages/jsii-pacmak/test/build-test.sh
```diff
@@ -17,12 +17,22 @@ clean_dists() {
...
+final_cleanup() {
+    rm -rf ${outdir}
+}
+trap final_cleanup EXIT
+
+# Prepare Python venv to avoid depending on system stuff
+venv="${outdir}/.env"
+python3 -m venv ${venv}
+. ${venv}/bin/activate
+pip install pip~=20.0.2 setuptools~=46.1.3 wheel~=0.34.2 twine~=3.1.1
```
eladb, line 30: Where do these version numbers come from?

packages/jsii-calc/lib/submodule/refers-to-parent/.jsiirc.json
```diff
@@ -0,0 +1,13 @@
+{
+  "targets": {
+    "java": {
+      "package": "software.amazon.jsii.tests.calculator.refers_to_parent"
```
eladb, line 4: Is the plan to also support java and .net in this PR or as a separate PR?

packages/@jsii/spec/lib/assembly.ts
```diff
@@ -156,6 +156,15 @@ export interface AssemblyConfiguration {
...
   targets?: AssemblyTargets;
+
+  /**
+   * The submodules included in this Assembly or submodule. Submodules can
+   * include nested submodules, as denoted by the `AssemblyConfiguration` value
+   * in this map.
+   *
+   * @default none
+   */
+  submodules?: { [name: string]: AssemblyConfiguration };
```
eladb, line 167: Requires updating the spec document, no?

packages/@jsii/python-runtime/tests/test_compliance.py
```diff
@@ -1168,3 +1168,8 @@ def test_collection_of_interfaces_map_of_structs():
 def test_collection_of_interfaces_map_of_interfaces():
     for elt in InterfaceCollections.map_of_interfaces().values():
         assert getattr(elt, "ring") is not None
+
+def test_load_submodules():
```
eladb, line 1172: Do we also want to verify that they can be loaded out of order?

packages/jsii/lib/assembler.ts
```diff
@@ -31,11 +32,14 @@ export class Assembler implements Emitter {
...
   private _deferred = new Array<DeferredRecord>();
-  private _types: { [fqn: string]: spec.Type } = {};
+  private _types: Record<string, spec.Type> = {};
+  private _submodules: Record<string, spec.AssemblyConfiguration> = {};
 
   /** Map of Symbol to namespace export Symbol */
-  private readonly _submoduleMap = new Map<ts.Symbol, ts.Symbol>();
-  private readonly _submodules = new Set<ts.Symbol>();
+  // eslint-disable-next-line @typescript-eslint/explicit-member-accessibility
+  readonly #submoduleMap = new Map<ts.Symbol, ts.Symbol>();
```
eladb, line 40: why not use `private`?

packages/jsii-pacmak/test/expected.jsii-calc/python/src/jsii_calc/submodule/nested_submodule/__init__.py
RomainMuller: I've actually fixed it separately because this would land too late.

packages/jsii-pacmak/test/build-test.sh
RomainMuller: THey're the "latest" from `pypi`. Just some housekeeping since those aren't managed by dependabot.

packages/jsii-calc/lib/submodule/refers-to-parent/.jsiirc.json
RomainMuller: I was thinking probably subsequent PR, not settled yet.

packages/@jsii/spec/lib/assembly.ts
RomainMuller: Hence this is a draft PR at this point :)

RomainMuller:
@eladb - can't you just make monocdk's root Java package be `software.amazon.awscdk.core`, and then customize the Java package name of submodules in there? What's missing?

eladb:
> @eladb - can't you just make monocdk's root Java package be `software.amazon.awscdk.core`, and then customize the Java package name of submodules in there? What's missing?

I guess you are right. @MrArnoldPalmer can you acknowledge that you saw this discussion?

RomainMuller:
I'm dropping this PR because it'll be superceded by a different (and I believe slightly simpler) way to model the data in the Assembly schema.

PR #1528, opened 2020-04-10
fix(python): imports between subpackages are broken

RomainMuller:
## Commit Message
fix(python): imports between subpackages are broken (#1528)

Using absolute imports from a sub-package to another of the same root
package causes cycles in the module loading machinery, resulting in
obscure errors (typically, the loader complaining that the root module
name is not defined).

This changes the name generation mechanism so that relative imports are
used to refer to sub-packages within a same root, as per the best
practices. This allows the module loaded to flatten the loop and
successfully load everything (granted, as always, there is no
intractible loop between the sub-packages).

## End Commit Message

---


packages/@jsii/python-runtime/tests/test_compliance.py
```diff
@@ -1168,3 +1168,9 @@ def test_collection_of_interfaces_map_of_structs():
 def test_collection_of_interfaces_map_of_interfaces():
     for elt in InterfaceCollections.map_of_interfaces().values():
         assert getattr(elt, "ring") is not None
+
+
+@pytest.mark.skip # Currently broken because submodule names aren't case-adjusted correctly :(
```
RomainMuller, line 1173: This is a pre-existing bug I'm merely shedding some light on. In order to fix this, I believe we'll need the kind of work in #1502 - specifically modeling the *submodules* information (including of dependencies) as part of the Assembly, so we can tell a PascalCased submodule name apart from a nested type's parent.

packages/jsii-pacmak/lib/targets/python.ts
```diff
@@ -66,27 +68,15 @@ export default class Python extends Target {
...
-const PYTHON_BUILTIN_TYPES = ['bool', 'str', 'None'];
-
-const PYTHON_KEYWORDS = [
-  'False', 'None', 'True', 'and', 'as', 'assert', 'async', 'await', 'break', 'class',
-  'continue', 'def', 'del', 'elif', 'else', 'except', 'finally', 'for', 'from',
-  'global', 'if', 'import', 'in', 'is', 'lambda', 'nonlocal', 'not', 'or', 'pass',
-  'raise', 'return', 'try', 'while', 'with', 'yield'
-];
+interface EmitContext extends NamingContext {
+  /** @deprecated The TypeResolver */
```
RomainMuller, line 72: I'm marking this deprecated because *in fine*, I intend on getting rid of the `TypeResolver` class as a whole.  It's grown overly complex and I believe its purposes can be achieved in more reliable (expressive?) ways with the information present in the assemblies.

But this is tricky and this PR is already large enough that I didn't want to go forward with it at this moment.

packages/jsii-pacmak/lib/targets/python/type-name.ts
```diff
@@ -0,0 +1,354 @@
...
+  readonly ignoreOptional?: boolean;
+}
+
+export function toTypeName(ref?: OptionalValue | TypeReference): TypeName {
+  if (ref == null) {
+    return Primitive.NONE;
+  }
+
+  const type = isOptionalValue(ref) ? ref.type : ref;
+  const optional = isOptionalValue(ref) && ref.optional;
```
eladb, line 65: doesn't `isOptionalValue` imply `ref.optional === true`?

packages/jsii-pacmak/lib/targets/python/type-name.ts
```diff
@@ -0,0 +1,354 @@
...
+      for (const item of items) {
+        result[packageName].add(item);
+      }
+    }
+  }
+  return result;
+}
+
+function isOptionalValue(type: OptionalValue | TypeReference): type is OptionalValue {
+  return (type as unknown as OptionalValue).type != null;
```
eladb, line 103: the fact that `type` has `type` indicates that it's an optional value? Sounds odd.

packages/jsii-pacmak/lib/targets/python/type-name.ts
```diff
@@ -0,0 +1,354 @@
...
+}
+
+class Dict implements TypeName {
+  // eslint-disable-next-line @typescript-eslint/explicit-member-accessibility
+  readonly #element: TypeName;
+
+  public constructor(element: TypeName) {
```
eladb, line 110:
```suggestion
  public constructor(private readonly element: TypeName) {
```

No?

packages/jsii-pacmak/lib/targets/python/type-name.ts
```diff
@@ -0,0 +1,354 @@
...
+    this.#element = element;
+  }
+
+  public pythonType(context: NamingContext) {
+    return `typing.Mapping[str,${this.#element.pythonType(context)}]`;
+  }
+
+  public requiredImports(context: NamingContext) {
+    return mergePythonImports(this.#element.requiredImports(context), { typing: new Set(['']) });
```
eladb, line 119: why is this needed? It's very unclear.

packages/jsii-pacmak/lib/targets/python/type-name.ts
```diff
@@ -0,0 +1,354 @@
...
+    return mergePythonImports(wrapped, { typing: new Set(['']) });
+  }
+}
+
+class Primitive implements TypeName {
+  private static readonly BOOL = new Primitive('bool', {});
+  private static readonly DATE = new Primitive('datetime.datetime', { 'datetime': new Set(['']) });
+  private static readonly JSII_NUMBER = new Primitive('jsii.Number', {}); // "jsii" is always already imported!
+  private static readonly STR = new Primitive('str', {});
+  private static readonly JSON = new Primitive('typing.Mapping[typing.Any, typing.Any]', { 'typing': new Set(['']) });
```
eladb, line 170: These private statics feel unnecessary... I would just inline them inside `of()`.

packages/jsii-pacmak/lib/targets/python/type-name.ts
```diff
@@ -0,0 +1,354 @@
...
+      case PrimitiveType.Json:
+        return Primitive.JSON;
+      case PrimitiveType.Any:
+      default:
+        return Primitive.ANY;
+    }
+  }
+
+  // eslint-disable-next-line @typescript-eslint/explicit-member-accessibility
+  readonly #pythonType: string;
```
eladb, line 194: What's the benefit of using the new javascript private annotations over typescript's `private`? I would expect typescript to compile `private` to `#`, but otherwise from a readability and coding standards perspective it seems like it's not adding much value.

packages/jsii-pacmak/lib/targets/python/type-name.ts
```diff
@@ -0,0 +1,354 @@
...
+  // eslint-disable-next-line @typescript-eslint/explicit-member-accessibility
+  readonly #requiredImports: PythonImports;
+
+  private constructor(pythonType: string, requiredImports: PythonImports) {
```
eladb, line 198:
```suggestion
  private constructor(private readonly pythonType: string, private readonly requiredImports: PythonImports) {
```

packages/jsii-pacmak/lib/targets/python/type-name.ts
```diff
@@ -0,0 +1,354 @@
...
+  }
+
+  public pythonType(context: NamingContext) {
+    return `typing.Union[${this.#options.map(o => o.pythonType(context)).join(', ')}]`;
+  }
+
+  public requiredImports(context: NamingContext) {
+    return mergePythonImports(
+      ...this.#options.map(o => o.requiredImports(context)),
+      { typing: new Set(['']) },
```
eladb, line 227: This seems to be always needed, any way to generalize and avoid the duplication everywhere?

packages/jsii-pacmak/lib/targets/python/type-name.ts
```diff
@@ -0,0 +1,354 @@
...
+  }
+
+  private resolve({ assembly, submodule, nestingScope, typeAnnotation = true }: NamingContext) {
+    const { assemblyName, pythonFqn } = toPythonFqn(this.#fqn, assembly);
+    if (assemblyName !== assembly.name) {
+      return {
+        pythonType: pythonFqn,
+        requiredImport: {
+          // TODO: Migth we need to import a subpackage instead?
+          sourcePackage: toPythonFqn(assemblyName, assembly).pythonFqn,
```
eladb, line 259: do we have a test that imports across from a subpackage?

packages/jsii-pacmak/lib/targets/python/type-name.ts
```diff
@@ -0,0 +1,354 @@
...
+ * @param toPkg   the package that needs to be relatively imported.
+ *
+ * @returns a relative import path.
+ *
+ * @example
+ *  relativeImportPath('A.B.C.D', 'A.B.E') === '...E';
+ *  relativeImportPath('A.B.C', 'A.B')     === '..';
+ *  relativeImportPath('A.B', 'A.B.C')     === '.C';
+ */
+function relativeImportPath(fromPkg: string, toPkg: string): string {
```
eladb, line 324: Would it make sense instead of these aliases, always import from the root? See example in a comment below.

packages/jsii-pacmak/test/expected.jsii-calc/python/src/jsii_calc/submodule/back_references/__init__.py
```diff
@@ -1,23 +1,18 @@
...
-
 from ..._jsii import *
 
+from .. import (MyClass as _MyClass_a2fdc0b6)
+
 
 @jsii.data_type(jsii_type="jsii-calc.submodule.back_references.MyClassReference", jsii_struct_bases=[], name_mapping={'reference': 'reference'})
 class MyClassReference():
-    def __init__(self, *, reference: jsii_calc.submodule.MyClass):
+    def __init__(self, *, reference: _MyClass_a2fdc0b6) -> None:
```
eladb, line 15: Is it possible, instead of these hashified aliases, to always import from the root? So this would be:

```py
from ... import submodule
```

And then the usage would be:

```py
submodule.MyClass
```

packages/jsii-pacmak/lib/targets/python/type-name.ts
RomainMuller: No - `isOptionalValue(x) -> x is OptionalValue`. I'd agree the naming is a little awkward here (but at the time I didn't think of anything better 🤷🏻‍♂️)

packages/jsii-pacmak/lib/targets/python/type-name.ts
RomainMuller: `OptionalValue` is an interface that has `{ type: TypeReference, optional?: boolean }`. The only "always there" field is `type`.

packages/jsii-pacmak/lib/targets/python/type-name.ts
RomainMuller: I want to gradually replace TS-only private fields with JS private #fields. Unfortunately, you cannot define those inline in the constructor signature.

packages/jsii-pacmak/lib/targets/python/type-name.ts
RomainMuller: Well because I need to import:
1. What needs to be imported for the element type
2. `typing` (so I can have `typing.Mapping` in scope)

packages/jsii-pacmak/lib/targets/python/type-name.ts
RomainMuller: They help reduce churn. Inlining in `of` would take more work that would hurt the overall performance (and I'd argue wouldn't improve readability).

packages/jsii-pacmak/lib/targets/python/type-name.ts
RomainMuller: TypeScript does *not* compile `private` to `#`. The semantic is different. Using `private foo`, I can always `as any` my way to access the field; where `#` actually makes access to the value impossible from outside of the defining type. It's better encapsulation and less risk of "inadvertently" overriding a private.

packages/jsii-pacmak/lib/targets/python/type-name.ts
RomainMuller: It's not always needed. It's *often* needed. I can introduce a constant for `IMPORT_TYPING`, but that would be as much as I can de-duplicate here...

packages/jsii-pacmak/test/expected.jsii-calc/python/src/jsii_calc/submodule/back_references/__init__.py
RomainMuller: That does not work when you import from the root itself... In this case, I need to always alias... And it makes the code simpler to always do the same, rather than ad yet another branch...

packages/jsii-pacmak/lib/targets/python/type-name.ts
RomainMuller: There's one critical case where that fails (see my answer to your comment below). I'd rather have only one way to manage those than have to distinguish "from root" versus "from non-root".

packages/jsii-pacmak/lib/targets/python/type-name.ts
RomainMuller: Oh I had forgotten this TODO. Definitely need to add one there!

packages/jsii-pacmak/test/expected.jsii-calc/python/src/jsii_calc/submodule/back_references/__init__.py
RomainMuller: Oh and it also breaks down if we import something named `submodule` from two different packages (for example, one local, one from a dependency).

eladb: Why do you need to alias? Can't you do `from . import Type` or something?

packages/jsii-pacmak/lib/targets/python/type-name.ts
eladb: Maybe `isOptionalValueType`?

packages/jsii-pacmak/lib/targets/python/type-name.ts
eladb: Rename `type` to `optionalType` or something like that so it will be easier to distinguish (and avoid bugs).

packages/jsii-pacmak/lib/targets/python/type-name.ts
eladb: Why do you want to replace TS-only private fields with JS private fields?

eladb: We don't *have* to use every shiny new thing :-)

packages/jsii-pacmak/lib/targets/python/type-name.ts
eladb: Seems like typing is always needed, why specify every time?

packages/jsii-pacmak/lib/targets/python/type-name.ts
eladb: Would definitely be more readably and less error prone. You can also change this to something like:

```ts
return importsWithTyping(this.#options.map(o => o.requiredImports(context))
```

Or even:

```ts
return importsWithTyping(this.#options);
```

packages/jsii-pacmak/test/expected.jsii-calc/python/src/jsii_calc/submodule/back_references/__init__.py
eladb: `submodule` can't be a type name (all lowercase), only a module name, no?

packages/jsii-pacmak/lib/targets/python/type-name.ts
RomainMuller: But I _want_ to!

packages/jsii-pacmak/lib/targets/python/type-name.ts
RomainMuller: Typing isn't needed for user-defined types (named references). It's often needed, not always.

packages/jsii-pacmak/test/expected.jsii-calc/python/src/jsii_calc/submodule/back_references/__init__.py
RomainMuller: That won't save me when I have the same submodule "leaf name" at different points...

```
from .foo import submodule
from .bar import submodule

submodule.FancyType # <-- Which "submodule" is this?
```

And while a submodule name cannot be a type name, I can have the same type name defined in several submodules:

```
from .foo import MyType
from .bar import MyType

MyType() # <-- Which is that?
```

When using relative imports, once has no choice but to use the `from .blah import thing`, and that appears to prevent the use fo the fully qualified type name further down (which would have been possible with an absolute import of the form `import foo.bar`). Unfortunately, I have to bring types into the present namespace, which means I have to deal with collisions myself, too.

Issue #83, opened 2018-07-12
Language Support: Golang 


eladb:
## Description

- [ ] Go as a target language (collaboration with JetBrains)

## Progress

- [x] Tracking Issue Created
- [x] RFC PR Created https://github.com/aws/aws-cdk-rfcs/pull/206
- [x] Core Team Member Assigned
- [x] Initial Approval / Final Comment Period
- [x] Alpha: https://github.com/aws/jsii/releases/tag/v1.15.0
- [x] Dev Preview: https://github.com/aws/jsii/projects/2
- [x] GA

RomainMuller:
Go is now GA.

PR #1551, opened 2020-04-15
feat(golang): introduce Golang code generation

RomainMuller:
Adds a new `GoLang` code generation handler. The handler defines Go specific
generators for JSII submodules, interfaces, structs, and enums. The directory
structure of the generated code mirrors public submodules and namespaces defined
in the JSII module, ie, there is one file per namespace. The generated type
model is in progress and subject to change.

Adds documentation to `docs/configuration.md` about configuring the Go target.
Go code generation can be enabled by adding the `go` key to the `jsii` section
of your package.json.

```json
{
  "jsii": {
    "targets": {
      "go": {}
    }
  }
}
```

The Go target is not currently supported for production use. This is the first
iteration of code generation and does not define a target runtime library.

Relates to #83

---


packages/jsii-pacmak/lib/targets/golang/interface.ts
```diff
@@ -0,0 +1,124 @@
...
+    );
+  }
+
+  private emitClassProperty(property: Property) {
+    const name = goNameFromJs(property.name);
+    const type = new TypeMapper(property.type).mapType();
+
+    this.code.line(`${name} ${type}`);
+  }
+
```
RomainMuller, line 109: Chances are you'll use those when creating proxies for interfaces, structs and abstract classes... In such cases, it might make things easier to have a Method / Property emitters so you can re-use.

PR #1557, opened 2020-04-16
fix(jsii): unable to use type from dependencies' submodules

RomainMuller:
## Commit Message

fix(jsii): unable to use type from dependencies' submodules (#1557)

When referring to a type defined in a submodule of a dependency (but
not using the "inline `namespace`" syntax style), the `jsii` compiler
failed to account for the namespace segment and generated an incorrect
`fqn`, leading to a type resolution error failing the compilation.

This adds a pass in the compiler to gather submodule layout from
dependencies and correctly register them in the compilation context, so
correct `fqn` are emitted.

New types have been introduced in `@scope/jsii-calc-lib` and `jsii-calc`
to verify that the compiler produces reasonable output. Those will be
leveraged in new compliance tests in a subsequent PR, as some
outstanding code-generation issues come in the way of these tests.

## End Commit Message

---


RomainMuller:
And of course the Python generated code won't load because it lacks necessary `import` statements, which cannot be retroactively added without knowing the list of submodules from dependencies...

PR #1563, opened 2020-04-17
feat(spec): model submodules in the Assembly schema

RomainMuller:
## Commit Message

feat(spec): model submodules in the Assembly schema (#1563)

Adds a `submodules` property on the `Assembly` structure, and carry it
forward in the `dependencyClosure`, so the information can later be used
to improve code generation for languages such as Python where the
submodule structure is modeled as first-class entity that needs to be
propertly dealt with. It can also help with properly adjusting the
submodule names so they look more "native" in target languages, without
facing problems when trying to generate type names in dependent packages.

The forwarding of dependent submodules is not exercized in the current
regression test suite because of a pair of other bugs (#1528, #1557)
that need to be addressed before the generated Python code can
successfully load. The last of those PRs to be merged will incldude the
necessary test coverage. This change is necessary for these PRs to
be able to fix their respective issues.

## End Commit Message

---


packages/jsii-calc/test/assembly.jsii
```diff
@@ -137,6 +137,17 @@
...
+  "submodules": {
+    "jsii-calc.DerivedClassHasNoProperties": {},
+    "jsii-calc.InterfaceInNamespaceIncludesClasses": {},
+    "jsii-calc.InterfaceInNamespaceOnlyInterface": {},
+    "jsii-calc.composition": {},
+    "jsii-calc.submodule": {},
+    "jsii-calc.submodule.back_references": {},
+    "jsii-calc.submodule.child": {},
+    "jsii-calc.submodule.nested_submodule": {},
+    "jsii-calc.submodule.nested_submodule.deeplyNested": {}
```
eladb, line 149: I think empty objects are a bad way to model something. Do we really need those here?

RomainMuller: Why do you want instead? Those would eventually contain the submodule-specific configuration block... I don't want to resort to type unions (`true | SubmoduleConfiguration`?), and if I put `undefined` as a value, then the keys will be missing from the serialized form...

packages/@jsii/spec/lib/assembly.ts
```diff
@@ -148,7 +148,19 @@ export interface Assembly extends AssemblyConfiguration, Documentable {
...
  * Shareable configuration of a jsii Assembly.
  */
-export interface AssemblyConfiguration {
+export interface AssemblyConfiguration extends Targetable {
+  /**
+   * Submodules declared in this assembly.
+   *
+   * @default none
+   */
+  submodules?: { [fqn: string]: SourceLocatable & Targetable };
```
eladb, line 157: Where is the `readme` contents going to be? Is it under `Targetable`?

RomainMuller: Could be added to `Targetable` (moved there from `Assembly`) yeah...

I would do this when I actually start processing READMEs in a subsequent PR. I don't want to bloat this one :)

PR #1566, opened 2020-04-17
chore(dotnet): move all packages to single solution

RomainMuller:
## Commit Message

chore(dotnet): move all packages to single solution (#1566)

The multiple solutions introduced for managing the DotNet runtime
packages result in unnecessary complexity in managing the mono-repo.
Consequently, they have been rolled into the `@jsii/dotnet-runtime`
package to make things easier going forward.

Since the projects are all on a single solution, it is also possible to
use a solution-wide declaration of dependency versions instead of
duplicating the values everywhere.

## End Commit Message

---


RomainMuller:
> I trust you

packages/@jsii/dotnet-analyzers/CHANGELOG.md
```diff
@@ -1,214 +0,0 @@
...
-
-
-
-## [0.20.0](https://github.com/aws/jsii/compare/v0.19.0...v0.20.0) (2019-10-30)
-
-
-### Bug Fixes
-
-* **java,dotnet:** emit default implementations for optional properties ([#906](https://github.com/aws/jsii/issues/906)) ([37ddfd5](https://github.com/aws/jsii/commit/37ddfd5fde1399274ca1541542c7268b75e026c2)), closes [#543](https://github.com/aws/jsii/issues/543)
-
```
nija-at, line 155: Should this get merged in with the changelog in dotnet-runtime ?

I'll leave this to you since I don't know how these changelogs are used.

packages/@jsii/dotnet-analyzers/src/NuGet.config
```diff
@@ -1,6 +0,0 @@
-<?xml version="1.0" encoding="utf-8"?>
```
nija-at, line 1: Why do these have a NuGet.config? Are these published to NuGet?

.dependabot/config.yml
```diff
@@ -7,26 +7,9 @@ update_configs:
     package_manager: javascript
     update_schedule: live
     version_requirement_updates: increase_versions
-    allowed_updates:
-      - match:
-          update_type: 'all'
```
RomainMuller, line 12: This is an intentional delete - I find it causing too much churn.

packages/@jsii/dotnet-jsonmodel/.npmignore
```diff
@@ -1,7 +0,0 @@
-# by default, exclude everything
-*
-
-# bundle only...
-!*.nupkg
-!package.json
-!lib/index.js
```
nija-at, line 7: Does this need to be carried over?

packages/@jsii/dotnet-analyzers/test.sh
```diff
@@ -1,6 +0,0 @@
-#!/bin/bash
-set -euo pipefail
-
-# Run unit tests
-echo "Running library unit tests"
-dotnet test -c Release ./src/Amazon.JSII.Analyzers.UnitTests
```
nija-at, line 6: Verify that these unit tests are still executed.

packages/@jsii/dotnet-jsonmodel/test.sh
```diff
@@ -1,5 +0,0 @@
-#!/bin/bash
-set -euo pipefail
-
-# Run unit tests
-dotnet test -c Release ./src/Amazon.JSII.JsonModel.UnitTests
```
nija-at, line 5: Verify that these unit tests are still executed.

packages/@jsii/dotnet-runtime/src/Directory.Build.targets
```diff
@@ -0,0 +1,16 @@
...
+    <PackageReference Update="Microsoft.Extensions.Logging" Version="3.1.3" />
+    <PackageReference Update="Microsoft.Extensions.Logging.Console" Version="3.1.3" />
+
+    <PackageReference Update="Microsoft.NET.Test.Sdk" Version="16.5.0" />
+    <PackageReference Update="NSubstitute" Version="4.2.1" />
+    <PackageReference Update="xunit" Version="2.4.1" />
+    <PackageReference Update="xunit.runner.visualstudio" Version="2.4.1" />
+    <DotNetCliToolReference Update="dotnet-xunit" Version="2.3.1" />
+
+    <PackageReference Update="Newtonsoft.Json" Version="12.0.3" />
```
nija-at, line 14: Should you drop the Version property here, just like in the other files?

packages/@jsii/dotnet-analyzers/CHANGELOG.md
RomainMuller: Good point. I'd rather get rid of per-package changelogs in the long run... so I'l just accept this as civilian causalties.

packages/@jsii/dotnet-analyzers/src/NuGet.config
RomainMuller: The NuGet.config file was used to configure a "local source" so local builds work before publishing. Since now in a single solution, this is no longer needed!

packages/@jsii/dotnet-jsonmodel/.npmignore
RomainMuller: The other package already has that, since they were mostly copies of one-another from the boilerplate standpoint.

packages/@jsii/dotnet-analyzers/test.sh
RomainMuller: They definitely are :) I've checked the logs and all.

packages/@jsii/dotnet-runtime/src/Directory.Build.targets
RomainMuller: This is the place that actually specifies the versions for all of them (which is why `Update` instead of `Include`). This is how dotnet has you set up solution-wide version configuration for your dependencies.

PR #1568, opened 2020-04-17
fix(jsii-pacmak): invalid dotnet version suffixes

MrArnoldPalmer:
Removes version suffix from version ranges in `.csproj` files generated
by jsii-pacmak.

Dotnet doesn't allow version suffixes that start with a number. Version
ranges by default don't include prereleases so the resolved version is
as intended without the suffix in this case.

---

### DIFF
--- a/packages/jsii-pacmak/lib/targets/version-utils.ts
+++ b/packages/jsii-pacmak/lib/targets/version-utils.ts
@@ -20,7 +20,7 @@ export function toMavenVersionRange(semverRange: string, suffix?: string): strin
  * @see https://docs.microsoft.com/en-us/nuget/concepts/package-versioning#version-ranges-and-wildcards
  */
 export function toNuGetVersionRange(semverRange: string): string {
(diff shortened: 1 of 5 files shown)


eladb:
Curious @MrArnoldPalmer what introduced this issue?

MrArnoldPalmer:
@eladb An [update to node-semver](https://github.com/npm/node-semver/blob/master/CHANGELOG.md#723) changed how prerelease suffixes are parsed/output.

We [bumped semver](https://github.com/aws/jsii/pull/1546) and changed some code to handle the new behavior where needed.

It wasn't apparent that the changed output broke dotnet because it doesn't cause issues until publishing. It's my understanding that it's a valid range and can be used in a .csproj for a binary project, it just can't be published to nuget because it breaks legacy versions of the toolchain. 🤷‍♂️

eladb:
Cool. Thanks for the details. Worth mentioning the root cause in the PR/commit description next time :-)

PR #1605, opened 2020-04-22
fix(java): compilation fails with "code too large"

RomainMuller:
## Commit Message
fix(java): compilation fails with "code too large" (#1605)

In cases where a module includes a large amount of types, compilation of
Java packages could fail with the "code too large" error. This is due to
the presence of a single method in the `$Module` class that contains a
reference to all types in the package, resulting in a large amount of
generated bytecode (more than Java's maximum of 64KB).

In order to remediate, moved the FQN to Class mapping to a resource file
that gets loaded when the `$Module` class is initialized. The individual
`Class` entries are then resolved at the last minute using
`Class.forName` to ensure the reoslution happens *after* the full
mapping has been processed.

## End Commit Message

---


packages/jsii-pacmak/test/expected.jsii-calc/java/src/main/java/software/amazon/jsii/tests/calculator/$Module.java
```diff
@@ -2,10 +2,47 @@
...
 import java.util.List;
+import java.util.Map;
+
 import software.amazon.jsii.JsiiModule;
 
 public final class $Module extends JsiiModule {
+    private static final Map<String, String> MODULE_TYPES = load();
+
+    private static Map<String, String> load() {
+        final Map<String, String> result = new HashMap<>();
```
eladb, line 24: Probably makes sense to implement this in `JsiiModule`, no?

PR #1633, opened 2020-05-06
fix(pacmak): label "External" stability as "Stable"

rix0rrr:
The distinction between "Stable" (checked stability) and "External"
(unchecked stability) is lost on users, and honestly we shouldn't bother
them with it.

Treat them the same when generating code.

---


packages/jsii-pacmak/test/expected.jsii-calc/dotnet/Amazon.JSII.Tests.CalculatorPackageId/Amazon/JSII/Tests/CalculatorNamespace/IStableInterfaceProxy.cs
```diff
@@ -4,19 +4,13 @@
 
 namespace Amazon.JSII.Tests.CalculatorNamespace
 {
-    /// <remarks>
-    /// <strong>Stability</strong>: Stable
-    /// </remarks>
```
eladb, line 9: Why is this now omitted? Sounds like a useful thing to put in the docs, no?

rix0rrr: It actually makes more sense to me to do the opposite: stable unless indicated otherwise. We already did that for Python and it makes sense.

It's also how every other API is documented. Nobody else just slathers on "stable" everywhere, they *might* note "deprected" or "experimental" as it applies.

packages/jsii/test/negatives/neg.submodules-cannot-share-symbols.ts
```diff
@@ -1,4 +1,4 @@
-///!MATCH_ERROR: Symbol is re-exported under two distinct submodules (ns1 and ns2)
+///!MATCH_ERROR: Symbol is re-exported under two distinct submodules
```
RomainMuller, line 1: Why this change?

rix0rrr: Because on my machine this was breaking the build. The error message I got said `(ns2 and ns1)` (order different way around), and I did not want to fight with apparent nondeterminism.

RomainMuller: Ohhhh yeah makes sense!

PR #1650, opened 2020-05-12
chore(repo): add stale github action

SomayaB:
- Sends a warning and then closes issues that haven't gotten a response in 2 years.
- Adds a closing-soon label to issues that have had the response-requested label for a week with no response.
- Sends a warning and then closes issues that have a closing-soon label and haven't had a response in 4 days.

---


.github/workflows/close-stale-issues.yml
```diff
@@ -0,0 +1,46 @@
+name: "Close Stale Issues"
+
+# Controls when the action will run.
+on:
+  schedule:
+  - cron: "0 * * * *"
```
RomainMuller, line 6: Isn't every hour a little excessive? Seems to me that once a day would be plenty sufficient?

NGL321: I agree with this! Otherwise LGTM

SomayaB: Sure, I copied the one from the CDK where I do think it's worth checking every hour because if someone responds I want labels to update pretty quickly so that they are not filtered out from the metrics email but we can make it once a day for jsii.

RomainMuller: I mean you could also time the daily run to happen an hour or so before the metrics are extracted? But that's up to you. I only mean to be conservative of free minutes of GitHub Actions executions :P

PR #1654, opened 2020-05-12
feat(java): add a Builder<T> interface implemented by all builders

### LINKED ISSUE #1652: Java APIs: Add a Builder interface implemented by all Builders
In Java, each Builder class should implement a generic Builder<T> interface.

### Use Case
We can use extension methods in Kotlin to make it less verbose to build props, like this:
```
        fun stackProps(): StackProps =
            StackProps.Builder().props {
                env(Environment.Builder().props {
                    account(accountToUse)
                    region(regionToUse)
                })
            }
```

which works by using this extension method:
```
        fun <B : Any, T> B.props(init: B.() -> Unit): T {
            init(this)
            val buildFunction = this::class.memberFunctions.single { it.name == "build" }
            return buildFunction.call(this) as T
        }
```

Unfortunately, the extension method has to use unsafe reflection and casting due to there being no relationship in the type system between Builders and the types they build.

A simple Builder interface would help make helper methods like these simpler and type-safe, like this:
```
        fun <B: Builder<T>, T> B.props(init: B.() -> Unit): T {
            init(this)
            return this.build()
        }
```

### Proposed Solution
The Java API should include this interface:

```
package software.amazon.awscdk.core;

@FunctionalInterface
public interface Builder<T> {
    T build();
}
```

and each Builder class should implement it:

```
public interface StackProps extends software.amazon.jsii.JsiiSerializable {
    ....
    public static final class Builder implements software.amazon.awscdk.core.Builder<StackProps> {
        ....
```

### Other

This should be backwards-compatible for almost all cases. In theory, it simply adds an interface for functions that are already present. In the wild, if some code relies on the fact that Builder classes don't implement an interface, that code may break.

* [x] :wave: I may be able to implement this feature request
* [ ] :warning: This feature might incur a breaking change

---

This is a :rocket: Feature Request

RomainMuller:
## Commit Message 
feat(java): add a Builder<T> interface implemented by all builders (#1654)

The common super-interface for all `Builder` classes creates a nice and
clean extension point that can be leveraged to improve the experience of
developers using Kotlin extensions, or doing fancy AOP things. For example,
this enables introduction of Kotlin extensions to improve/simplify the syntax
needed to initialize types that provide a `Builder` further from what the
Java implementation natively allows (see the related issue for more info).

The new interface does not change the current `Builder` layout, merely
adding the new super-interface to pre-declare an already-existing method.

Fixes #1652
## End Commit Message

---

### DIFF
--- a/packages/@jsii/java-runtime/project/src/main/java/software/amazon/jsii/Builder.java
+++ b/packages/@jsii/java-runtime/project/src/main/java/software/amazon/jsii/Builder.java
@@ -3,7 +3,7 @@ package software.amazon.jsii;
 /**
  * A superinterface common to instance builders.
  */
- @FunctionalInterface
+@FunctionalInterface
 public interface Builder<T> {
   /**
    * Builds the instance given the current builder configuration.
--- a/packages/@jsii/java-runtime/project/src/main/java/software/amazon/jsii/Builder.java
+++ b/packages/@jsii/java-runtime/project/src/main/java/software/amazon/jsii/Builder.java
@@ -1,0 +1,14 @@
+package software.amazon.jsii;
+
+/**
+ * A superinterface common to instance builders.
+ */
+ @FunctionalInterface
+public interface Builder<T> {
+  /**
+   * Builds the instance given the current builder configuration.
+   *
+   * @return the built instance.
+   */
+  T build();
+}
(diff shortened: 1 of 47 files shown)


packages/@jsii/java-runtime/project/src/main/java/software/amazon/jsii/Builder.java
```diff
@@ -0,0 +1,14 @@
+package software.amazon.jsii;
+
+/**
+ * A superinterface common to instance builders.
+ */
+ @FunctionalInterface
```
nija-at, line 6:
```suggestion
@FunctionalInterface
```

PR #1690, opened 2020-05-19
feat: allow per-submodule naming configuration

RomainMuller:
Introduces a new `.jsiirc.json` file leveraged by sumodules defined
using the `export * as namespace from './namespace';`  syntax. The
`.jsiirc.json` file is a sibling to what `./namespace` resolves to (since
this typically resolves to `./namespace/index.ts`, the `.jsiirc.json` file
must be at `./namespace/.jsiirc.json`).

The `targets` key can je set within this `.jsiirc.json` document and
include specific configuration for various targets:

```js
{
  "targets": {
    "dotnet": {
      // Only the "namespace" key is considered/supported at this point
      "namespace": "Some.DotNet.Namespace"
    },
    "java": {
      // Only the "package" key is considered/supported at this point
      "package": "some.java.package_name"
    },
    "python": {
      // Only the "module" key is considered/supported at this point
      // The configured module name for a submodule _MUST_ be nested
      // under the root module name for the package.
      "module": "some.module_name"
    }
  }
}
```

Related to #1286 

---


packages/jsii-pacmak/lib/targets/dotnet/dotnetgenerator.ts
```diff
@@ -463,16 +462,15 @@ export class DotNetGenerator extends Generator {
...
     while (ns != null && assm.types?.[`${assm.name}.${ns}`] != null) {
       const nesting = assm.types[`${assm.name}.${ns}`];
       ns = nesting.namespace;
     }
     if (ns != null) {
-      parts.push(...ns.split('.').map((n) => toPascalCase(n)));
+      return this.typeresolver.resolveNamespace(assm, assm.name, ns);
     }
-    return parts.join('.');
+    return assm.targets!.dotnet!.namespace;
```
MrArnoldPalmer, line 473: Can the 'non-nullability' of these fields not be represented in the assembly spec because of the presence of the dotnet target isn't guaranteed?

RomainMuller: Problem is the spec doesn't actually represent per-language configurations... so they're opaque stuff entirely.

PR #1692, opened 2020-05-19
feat(jsii): use incremental builders

RomainMuller:
Using the `IncrementalBuilder` to compiler code instead of a regular
`Builder` allows saving time when certain files are aloready up-to-date.

The incremental support is only enabled when `--project-references` is
passed to the compiler, as it relies on project references and the
composite model to work it's magic (as far as `tsc` is concerned, the
`incremental` option implies the `composite` option).

---

### DIFF
--- a/packages/jsii/lib/compiler.ts
+++ b/packages/jsii/lib/compiler.ts
@@ -63,10 +63,13 @@ export class Compiler implements Emitter {
   private readonly projectReferences: boolean;
 
   public constructor(private readonly options: CompilerOptions) {
-    this.compilerHost = ts.createIncrementalCompilerHost(BASE_COMPILER_OPTIONS, {
-      ...ts.sys,
-      getCurrentDirectory: () => this.options.projectInfo.projectRoot,
-    });
+    this.compilerHost = ts.createIncrementalCompilerHost(
+      BASE_COMPILER_OPTIONS,
+      {
+        ...ts.sys,
+        getCurrentDirectory: () => this.options.projectInfo.projectRoot,
+      },
+    );
 
     this.configPath = path.join(
       this.options.projectInfo.projectRoot,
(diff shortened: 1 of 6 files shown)


RomainMuller:
Tested against `aws-cdk`, this yielded a 30% improvement in rebuild speed (for the dependency closure up to and including `@aws-cdk/aws-iam`).

packages/jsii/lib/compiler.ts
```diff
@@ -255,29 +260,22 @@ export class Compiler implements Emitter {
   private async buildTypeScriptConfig() {
     let references: string[] | undefined;
     let composite: boolean | undefined;
+    let incremental: boolean | undefined;
     if (this.projectReferences) {
       references = await this.findProjectReferences();
-      composite = true;
+      composite = incremental = true;
```
MrArnoldPalmer, line 269:
```suggestion
      composite = true;
      incremental = true;
```
This feels like its gonna unnecessarily confuse people but maybe I'm just easily confused.

packages/jsii/lib/compiler.ts
```diff
@@ -489,18 +511,18 @@ export interface NonBlockingWatchOptions {
 function _pathOfLibraries(
   host: ts.CompilerHost | ts.WatchCompilerHost<any>,
 ): string[] {
-  if (!COMPILER_OPTIONS.lib || COMPILER_OPTIONS.lib.length === 0) {
+  if (!BASE_COMPILER_OPTIONS.lib || BASE_COMPILER_OPTIONS.lib.length === 0) {
```
MrArnoldPalmer, line 514: Does this conditional ever return `true`? `BASE_COMPILER_OPTIONS` is hardcoded by us no?

RomainMuller: Hmm yeah right... but this may not be hardcoded forever. Since this is just the rename I'd leave it as-is for now.

packages/jsii/lib/compiler.ts
MrArnoldPalmer: Already done.

Issue #1678, opened 2020-05-15
Expected behavior for * dependencies when compiling into Java


campionfellin:
### Affected Languages

- [ ] `TypeScript` or `Javascript`
- [ ] `Python`
- [x] `Java`
- [ ] .NET (`C#`, `F#`, ...)

### General Information
* **JSII Version:** 1.1
* **Platform:**  macOS Catalina

### The Question

What is the expected behavior when compiling `*` dependencies into Java? Here's part of an example `package.json`:

```
 jsii: {
      outdir: "dist",
      targets: {
        java: {
          package: "main",
          maven: {
            groupId: "main",
            artifactId: "main"
          }
        }
      }
    },    
    dependencies: {
      "constructs": "*",
      "cdk8s": "*",
    },
```

This leads to my dependencies in `pom.xml` looking like:

```
  <dependency>
      <groupId>org.cdk8s</groupId>
      <artifactId>cdk8s</artifactId>
      <version>[undefined]</version>
    </dependency>
    <dependency>
      <groupId>software.constructs</groupId>
      <artifactId>constructs</artifactId>
      <version>[undefined]</version>
    </dependency>
```

Changing versions from `*` to a numbered version like `2.0.1` or `0.21.0` seems to fix the issue.

Is this expected?
Any solutions besides explicitly adding the version?

RomainMuller:
How'd you model such a dependency in `pom.xml`?

The reason I as is because I see two ways of addressing this:
1. Generate a valid pom.xml from this configuration (probably also for other languages)
2. Prohibit `*` dependencies

I'd instinctively lean towards option 2, however I also don't like being more restrictive than needed (so if I can express `*` in other languages, I don't see why I'd refrain from doing it).

Generally speaking however, `jsii` packages are expected to adhere to semantic versioning, and having a `*` dependency would mean you're accepting any major version (including incompatible ones) - that doesn't sound like something you'd want to be doing?

campionfellin:
I think for (1), something like this could be acceptable for maven/Java at least:

```
    <dependency>
      <groupId>org.cdk8s</groupId>
      <artifactId>cdk8s</artifactId>
      <version>(0.0.0,]</version>
    </dependency>
```

(2) is also fine with me, if it can be added to the documentation somewhere!

campionfellin:
@RomainMuller either option is fine with me, but I'd prefer (1). I can work on this if you'd like

RomainMuller:
[...]

PR #1696, opened 2020-05-20
feat(jsii-pacmak): add maven and nuget support for * dependencies

campionfellin:
For https://github.com/aws/jsii/issues/1678 in case you want to go with route (1) of supporting `*` dependencies.

Signed-off-by: campionfellin <campionfellin@gmail.com>

---


packages/jsii-pacmak/lib/targets/version-utils.ts
```diff
@@ -65,6 +65,10 @@ function toBracketNotation(
     .map((set) => {
       if (set.length === 1) {
         const version = set[0].semver.raw;
+        if (!version) {
+          // Case where version is '*'
+          return `(0.0.0,]`;
```
RomainMuller, line 70: This probably should be *inclusive* of `0.0.0`.

```suggestion
          return `[0.0.0,]`;
```

I kinda wonder if the '*' shouldn't be detected upfront, and maybe "replaced" with `>= 0.0.0` instead?

packages/jsii-pacmak/test/targets/version-utils.test.ts
```diff
@@ -63,6 +63,18 @@ const examples: Record<
...
     python: '<=0.1.2',
   },
+
+  // Somewhat unusual ranges
+  '*': {
+    maven: '(0.0.0,]',
+    nuget: '(0.0.0,]',
+    python: '>=0.0.0',
+  },
+  '^1.2.*': {
```
RomainMuller, line 73: Is this a valid `semver` range notation? It appears confusing to me as two different intentions are expressed:
- `1.2.*` is typically used to mean minor-pinned (`>= 1.2.0, <= 1.3.0-0`)
- `^1.2.0` means major-pinned (`>= 1.2.0, < 2.0.0-0`)

packages/jsii-pacmak/lib/targets/version-utils.ts
campionfellin: I think it's better to detect it here within JSII (rather than in `packages/jsii-pacmak/lib/targets/java.ts`, etc), so it's just done once, rather than in both the maven-related code and nuget-related code.

I also don't think consumers should have to worry about this, I think style-wise `*` is more common than `>=0.0.0`.

packages/jsii-pacmak/test/targets/version-utils.test.ts
campionfellin: Great callout! Switched it to `1.2.*` :)

packages/jsii-pacmak/lib/targets/version-utils.ts
RomainMuller: Oh no you misunderstood me! I meant catch it at the beginning of *this function*, adding something along the lines of:

```
if (semverRange === '*') {
  semverRange = '>=0.0.0';
}
```

And then letting the rest of the code work it's magic unchanged.

packages/jsii-pacmak/test/targets/version-utils.test.ts
```diff
@@ -63,6 +63,18 @@ const examples: Record<
...
+  '*': {
+    maven: '[0.0.0,]',
+    nuget: '[0.0.0,]',
+    python: '>=0.0.0',
+  },
+  '1.2.*': {
+    maven: '[1.2.0,1.3.0)',
+    nuget: '[1.2.0,1.3.0)',
+    python: '>=1.2.0, <1.3.0',
+  },
```
RomainMuller, line 77: Can you also add one for `1.*`? (I reckon this would be a valid way to denote major-pinned, isn't it?)

packages/jsii-pacmak/lib/targets/version-utils.ts
campionfellin: Ahh! Will do that :)

campionfellin: The problem I'm running into here is that:

```
new Range('>=0.0.0');
```

leads to this:

```
     Range {
        options: { loose: false, includePrerelease: false },
        loose: false,
        includePrerelease: false,
        raw: '>=0.0.0',
        set: [ [ [Comparator] ] ],
        range: '' }
```

And `range.set[0].semver` is `Symbol(SemVer ANY)`

and thus `range.set[0].semver.raw` is undefined :/

RomainMuller: Well this sucks! I guess you've tried 🤷🏻‍♂️

campionfellin: Would you suggest a mix of checking in both places, or just where I had previously? I can also check `range.raw>=0.0.0` (instead of just `set[0].semver.raw` undefined

campionfellin:
@RomainMuller seems that builds are just generally failing now (unrelated to this change), but this should be good to go.

RomainMuller:
Yeah looks like MyPY again started checking things it used to ignore. Duh.

RomainMuller:
@MergifyIO refresh

PR #1702, opened 2020-05-25
fix(kernel): error raised during rename operation on win32

### LINKED ISSUE #992: Rename operation throwing exception
### Affected Languages

- [X] `Python`
- [ ] `Java`
- [ ] .NET (`C#`, `F#`, ...)

### General Information
* **JSII Version:**
* **Platform:**  Windows 10

### What is the problem?
Invocation of `cdk` cli fails due to a permission error while renaming a file.
As reported here https://stackoverflow.com/a/50297753 the cause likely to be the antivirus that blocks the operation. In my scenario I couldn't disable the A/V due to domain policy, so instead I've added a retry mechanism in the `doRename` function (`cdk_venv/python3.7/site-packages/jsii/_embedded/jsii/jsii-runtime.js:10408`):
(please keep in mind that I have **NO experience** at all with JS, i'm sure that there are much better way to achieve this  by using async/await for example, this is just to provide a minimal example)

```
function doRename (src, dest, overwrite) {
  let i = 0;
  
  for(i=0; i<10; i++){
    try {
      if (overwrite) {
        removeSync(dest)
        return rename(src, dest, overwrite)
      }
      if (fs.existsSync(dest)) throw new Error('dest already exists.')
      return rename(src, dest, overwrite)

    } 
    catch (err) {
      console.error(`Failed rename, try ${i}/10`);
      if(i==9) throw err

      var now = new Date().getTime();
      var millisecondsToWait = 250; /* i.e. 1 second */
      while ( new Date().getTime() < now + millisecondsToWait ) {}
    }
  }
}
```

### Verbose Log
```
jsii.errors.JavaScriptError:
  Error: EPERM: operation not permitted, rename 'C:\Users\F87D8~1.GEN\AppData\Local\Temp\jsii-kernel-install-staging-7hkALG\package' -> 'C:\Users\F87D8~1.GEN\AppData\Local\Temp\jsii-kernel-EHW0R3\node_modules\@aws-cdk\core'
      at Object.renameSync (fs.js:643:3)
      at rename (C:\Users\f.gentile\Desktop\test\cdk-t\.win_env\lib\site-packages\jsii\_embedded\jsii\jsii-runtime.js:10408:8)
      at doRename (C:\Users\f.gentile\Desktop\test\cdk-t\.win_env\lib\site-packages\jsii\_embedded\jsii\jsii-runtime.js:10403:10)
      at Object.moveSync (C:\Users\f.gentile\Desktop\test\cdk-t\.win_env\lib\site-packages\jsii\_embedded\jsii\jsii-runtime.js:10394:10)
      at Kernel.load (C:\Users\f.gentile\Desktop\test\cdk-t\.win_env\lib\site-packages\jsii\_embedded\jsii\jsii-runtime.js:7479:16)
      at KernelHost.processRequest (C:\Users\f.gentile\Desktop\test\cdk-t\.win_env\lib\site-packages\jsii\_embedded\jsii\jsii-runtime.js:7285:28)
      at KernelHost.run (C:\Users\f.gentile\Desktop\test\cdk-t\.win_env\lib\site-packages\jsii\_embedded\jsii\jsii-runtime.js:7225:14)
      at Immediate._onImmediate (C:\Users\f.gentile\Desktop\test\cdk-t\.win_env\lib\site-packages\jsii\_embedded\jsii\jsii-runtime.js:7228:37)
      at processImmediate (internal/timers.js:439:21)

The above exception was the direct cause of the following exception:

  File "app.py", line 3, in <module>
    from aws_cdk import core
  File "C:\Users\f.gentile\Desktop\test\cdk-t\.win_env\lib\site-packages\aws_cdk\core\__init__.py", line 413, in <module>
    __jsii_assembly__ = jsii.JSIIAssembly.load("@aws-cdk/core", "1.16.3", __name__, "core@1.16.3.jsii.tgz")
  File "C:\Users\f.gentile\Desktop\test\cdk-t\.win_env\lib\site-packages\jsii\_runtime.py", line 40, in load
    _kernel.load(assembly.name, assembly.version, os.fspath(assembly_path))
  File "C:\Users\f.gentile\Desktop\test\cdk-t\.win_env\lib\site-packages\jsii\_kernel\__init__.py", line 209, in load
    self.provider.load(LoadRequest(name=name, version=version, tarball=tarball))
  File "C:\Users\f.gentile\Desktop\test\cdk-t\.win_env\lib\site-packages\jsii\_kernel\providers\process.py", line 330, in load
    return self._process.send(request, LoadResponse)
  File "C:\Users\f.gentile\Desktop\test\cdk-t\.win_env\lib\site-packages\jsii\_kernel\providers\process.py", line 318, in send
    raise JSIIError(resp.error) from JavaScriptError(resp.stack)
jsii.errors.JSIIError: EPERM: operation not permitted, rename 'C:\Users\F87D8~1.GEN\AppData\Local\Temp\jsii-kernel-install-staging-7hkALG\package' -> 'C:\Users\F87D8~1.GEN\AppData\Local\Temp\jsii-kernel-EHW0R3\node_modules\@aws-cdk\core'
Subprocess exited with error 1
Error: Subprocess exited with error 1
    at ChildProcess.<anonymous> (C:\Users\f.gentile\AppData\Roaming\npm\node_modules\aws-cdk\lib\api\cxapp\exec.ts:115:23)
    at ChildProcess.emit (events.js:210:5)
    at ChildProcess.EventEmitter.emit (domain.js:476:20)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:272:12)
```
Full log https://gist.github.com/FabioGentile/29350501e3890305a94feff4b54ff1f8

Could it make sense to introduce something similar in the upstream version?

RomainMuller:
When loading a new library into the jsii kernel, the provided `tarball`
was extracted to a temporary directory, then moved into it's final
location. On Windows, this operation could fail on an `EACCESS` or
`EPERM` error (often due to malware scanners accessing the file for
inspection on file systems which do not support renaming files that are
being accessed).

This changes how the `load` API works so taht the `tarball` is extracted
directly in it's final install location, so that no rename operation is
needed.

Fixes #992 

---


RomainMuller:
Note: the change appears much bigger than it really is because it involves changing the indentation of a large chunk of code (no longer wrapped under a `try { ... } finally { ... }` block). You'll get a much better view on the actual change if you [hide whitespace changes](https://github.blog/2018-05-01-ignore-white-space-in-code-review/).

rix0rrr:
Guess I'm not necessarily opposed to this, but generally we do moves for a reason: to preserve atomicity.

Can you elaborate why this would not be an issue in this case? Why did we do the move in the first place?

rix0rrr:
Okay talked to Elad, it's not for atomicity reasons. Kay then.

RomainMuller:
> Guess I'm not necessarily opposed to this, but generally we do moves for a reason: to preserve atomicity.
> 
> Can you elaborate why this would not be an issue in this case? Why did we do the move in the first place?

So the move was quite probably due to the fact the contents of the tarball is nested under an undesirable `package` directory, while we need to install into `node_modules/<package-name>` instead. Either the `strip` feature was unavailable (or undocumented) at the time we first wrote this, or it didn't occur to us we could use it.

rix0rrr:
> assume

Just going to leave this right here 😛 

But yeah, let's try it, probably worth the risk.

PR #1706, opened 2020-06-02
fix(jsii-pacmak): computeIfAbsent throws ConcurrentModificationException

NetaNir:
---

Java generated modules use cache to store the result of `resolveClass` which is called when ever `jsiiStaticGet`/`jsiiGet` is called. The cache is implemented as a  `HashMap`, to fill the cache we use `computeIfAbsent`:

```Java
this.cache.computeIfAbsent(MODULE_TYPES.get(fqn), this::findClass);
```
> [code](https://github.com/aws/jsii/blob/master/packages/jsii-pacmak/lib/targets/java.ts#L2533)

When either of the get methods is called on a type which in itself includes a type, the first get (the root type get) will call get on the inner type. When `computeIfAbsent` is called from the root get, the compute method will trigger another call to `computeIfAbsent` to resolve the inner type, since all types share the same cache it means that the inner `computeIfAbsent` call will modify the cache while the root `computeIfAbsent` is not yet completed, resulting it a `ConcurrentModificationException`. 

This effects all Java >= 9 applications as a change to the behavior of `computeIfAbsent` was introduced in Java 9.  From the [Javadocs](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/HashMap.html#computeIfAbsent(K,java.util.function.Function)):

```
Throws:
ConcurrentModificationException - if it is detected that the mapping function modified this map
```
The use of  `computeIfAbsent `  was introduced in https://github.com/aws/jsii/pull/1605 and released in `v1.4.1`. 

This PR replaces `computeIfAbsent` with explicitly checking if the value exists in the cache and if not adding it (essentially old style `computeIfAbsent`).

### Test
Steps to verify fix:

1. Build and package a sample module (see repro) using a local version of `jsii` and `jsii-pacmak`.
2. Add the generated jar to a local `mvn` repo.
3. Added the package as a dependency to a Java11 application.
4. Verify the exception is no longer thrown. 

### Repro

A repro can be found in this [repo](https://github.com/NetaNir/jsii-nested-get-repro)


NetaNir:
copy @iliapolo

eladb:
🤩 wow

eladb:
@NetaNir What did we change that suddenly caused this? Can you please include a reference to the PR that introduced this break?

packages/jsii-pacmak/lib/targets/java.ts
```diff
@@ -2529,11 +2529,12 @@ class JavaGenerator extends Generator {
       'throw new ClassNotFoundException("Unknown JSII type: " + fqn);',
     );
     this.code.closeBlock();
-    this.code.line(
-      'return this.cache.computeIfAbsent(MODULE_TYPES.get(fqn), this::findClass);',
-    );
+    this.code.line('String className = MODULE_TYPES.get(fqn);');
+    this.code.openBlock('if (!this.cache.containsKey(className))');
+    this.code.line('this.cache.put(className, this.findClass(className));');
```
eladb, line 2534: Any chance this can cause an infinite recursion (which I assume is why `computeIfAbsent` throws this error)?

NetaNir:
> @NetaNir What did we change that suddenly caused this? Can you please include a reference to the PR that introduced this break?

@eladb updated description to include the [PR](https://github.com/aws/jsii/pull/1605) which introduced the use of `computeIfAbsent`

packages/jsii-pacmak/lib/targets/java.ts
NetaNir: I think that there is a circuit breaker [here](https://github.com/aws/jsii/blob/master/packages/@jsii/java-runtime/project/src/main/java/software/amazon/jsii/JsiiEngine.java#L224), but I'll @RomainMuller comment

eladb:
> > @NetaNir What did we change that suddenly caused this? Can you please include a reference to the PR that introduced this break?
> 
> 
> 
> @eladb updated description to include the [PR](https://github.com/aws/jsii/pull/1605) which introduced the use of `computeIfAbsent`
> 
> 

Thanks!

rix0rrr:
> When `computeIfAbsent` is called from the root get, the compute method will trigger another call to `computeIfAbsent` to resolve the inner type,

We kinda knew that this would have to be the issue, but do we know **HOW/WHY**?

How in the hell does `Class.forName(...)` ultimately lead back to `$Module.resolveClass()`?

-----------------

(I also originally expected `$Module.resolveClass()` to be in the stack trace twice, but of course that's not true. The ultimate complete stack trace would look like this:

```
(stack goes ⬇️ that way)
(...some stack frames...)
$Module.resolveClass
HashMap.computeIfAbsent      <----- 2️⃣ Error occurs here *AFTER* the frames below have unwound
$Module.findClass
(... what goes here...?)
$Module.resolveClass
HashMap.computeIfAbsent       <---- 1️⃣ This completes successfully first
```

packages/jsii-pacmak/lib/targets/java.ts
RomainMuller: I think this is going to be enough indeed. The problem was that further mutation to the map happened as part of `this.findClass` within the lambda expression passed to `computeIfAbsent`. This approach makes "atomic" mutations here (and this is safe because `jsii` apps can be expected to be single-threaded).

RomainMuller:
@rix0rrr to answer your `(... what goes here...)` section:
* `Class.forName` will possibly load up a newly discovered class (let's call it `Foo`)
  - Can be a base class/interface, some property's type, ...
* Class initialization of `Foo` might involve `jsii` calls
  - Getting values of `const` fields, for example
* The result of those calls may be something that `$Module.resolveClass` is used for (e.g: class/interface instance, struct, enum)
  - At this point, we'll get back to `computeIfAbsent`

RomainMuller:
I'm trying to come up with a test that blows up before the fix is introduced.

rix0rrr:
> * Getting values of `const` fields, for example

Got it. So that should be easy enough to write a test for then. So it's basically a matter of:

```ts
class SomeClass {
  public static const OTHER: OtherClass = new OtherClass();
}
```

?

rix0rrr:
Although of course that won't break if `OtherClass` is referenced in any other capacity...

RomainMuller:
> > * Getting values of `const` fields, for example
> 
> Got it. So that should be easy enough to write a test for then. So it's basically a matter of:

Yeah that's what I thought but it appears the patch to trigger this situation depends on a bunch of other factors I'm trying to identify (possibly the properties being typed "any" or something.

RomainMuller:
Oh I missed a critical element of the repro: it only happens with Java >= 9. It seems related to a change in class loading behavior between 8 and 9+.

iliapolo:
@RomainMuller Its not a change in class loading, its a change in the implementation of `computeIfAbsent`, which is Java >= 9 throws the `ConcurrentModificationException` when it detects a `dirty` mutation. 

https://stackoverflow.com/questions/54824656/since-java-9-hashmap-computeifabsent-throws-concurrentmodificationexception-on

Regarding the repro, did you look at @NetaNir's [repo](https://github.com/NetaNir/jsii-nested-get-repro/blob/master/lib/index.ts)?

It wasn't that straight-forward to reproduce, but I do think the repro can probably be simplified. 

Also, since the recursion stems from `Class.forName`, it can only happen when a class has static initializers that eventually call `jsiiGet` - right? Since `Class.forName` only loads the class, and doesn't instantiate if.

RomainMuller:
So the flow is:
- Resolving some class involves loading another
- Loading the other class involves resolving yet another

An easy way to trigger is to have a const property causing loading of an enum (it'll then try to resolve the enum when loading up it's values)

Added a test to this effect. Interestingly, it works in Java 8, fails in Java 9. I suspect Java 9 resolves classes more eagerly than 8.

iliapolo:
@RomainMuller 

>  Interestingly, it works in Java 8, fails in Java 9. I suspect Java 9 resolves classes more eagerly than 8.

https://stackoverflow.com/questions/54824656/since-java-9-hashmap-computeifabsent-throws-concurrentmodificationexception-on

RomainMuller:
> @RomainMuller 
> 
> 
> 
> >  Interestingly, it works in Java 8, fails in Java 9. I suspect Java 9 resolves classes more eagerly than 8.
> 
> 
> 
> https://stackoverflow.com/questions/54824656/since-java-9-hashmap-computeifabsent-throws-concurrentmodificationexception-on

Ohhhhh

RomainMuller:
@MergifyIO refresh

PR #1708, opened 2020-06-02
fix(jsii): build succeeds using Omit<T, K>

### LINKED ISSUE #1707: jsii build succeeds when using Omit<T, K>
### What is the problem?

jsii does not support the [`Omit` type](https://www.typescriptlang.org/docs/handbook/utility-types.html#omittk) but jsii build does not fail. It's likely that the jsii compiler is ignoring the `Omit` type.

Props class uses `Omit` utility type - [PR commit](https://github.com/aws/aws-cdk/pull/8281/commits/5771bcb56d93b823365751f7e25509df2093947f#diff-f869db0e3fd8a48f38c33737566ebf70)

Successful build (account: 260708760616) - https://us-east-1.console.aws.amazon.com/codesuite/codebuild/260708760616/projects/AutoBuildProject6AEA49D1-qxepHUsryhcu/build/AutoBuildProject6AEA49D1-qxepHUsryhcu%3A531f3f72-e613-425b-b692-c041b756d515?region=us-east-1

RomainMuller:
When using the `Omit<T, K>` helper type (a mapped type) on an exported
API, `jsii` should detect this and fail, instead of allowing the
application to compile.

An additional bug caused use of this type to result in a successful exit
code from `jsii`, while the `.jsii` file was never written. This was caused
by the mapped types symbols not having a `valueDeclaration`, triggering
an attempt to read `undefined.parent`, which threw an exception caught
by the `Compiler` class, but not signaled back to the caller.

Fixes #1707

---


packages/jsii/lib/assembler.ts
```diff
@@ -1587,7 +1628,7 @@ export class Assembler implements Emitter {
       for (const member of declaringType.getProperties()) {
         if (
           !(declaringType.symbol.getDeclarations() ?? []).find(
-            (decl) => decl === member.valueDeclaration.parent,
+            (decl) => decl === member.valueDeclaration?.parent,
```
RomainMuller, line 1631: The `?` here is due to `valueDeclaration` being nullable (despite the type declarations do not mention it).

packages/jsii/lib/compiler.ts
```diff
@@ -246,6 +246,7 @@ export class Compiler implements Emitter {
       diagnostics.push(...assmEmit.diagnostics);
     } catch (e) {
       LOG.error(`Error during type model analysis: ${e}\n${e.stack}`);
+      hasErrors = true;
```
RomainMuller, line 249: This previously caused exceptions thrown from the assembler to be "swallowed" (after logging to `STDERR`) without necessarily causing the process to exit with non-0 (`hasErrors` could be `false` when this `catch` block triggers!).

packages/jsii/lib/assembler.ts
```diff
@@ -892,6 +894,45 @@ export class Assembler implements Emitter {
...
 
+  private _validateHeritageClauses(clauses?: ts.NodeArray<ts.HeritageClause>) {
+    if (clauses == null || clauses.length === 0) {
+      // Nothing to do.
+      return;
+    }
+    for (const clause of clauses) {
+      for (const node of clause.types) {
+        const parentType = this._typeChecker.getTypeAtLocation(node);
+        // For some reason, we cannot trust parentType.isClassOrInterface()
```
nija-at, line 905: Expand a bit with an example of when it has lost your trust?

packages/jsii/lib/assembler.ts
```diff
@@ -892,6 +894,45 @@ export class Assembler implements Emitter {
...
+            !ts.isInterfaceDeclaration(decl) && // <-- local interfaces
+            !ts.isModuleDeclaration(decl), // <-- imported types
+        );
+        if (badDecl != null) {
+          this._diagnostic(
+            node,
+            ts.DiagnosticCategory.Error,
+            `Illegal "${clauseType(clause.token)}" value for an exported API: ${
+              ts.SyntaxKind[badDecl.kind]
+            }`,
```
nija-at, line 918: What does this error look like the `neg.omit.*.ts`? 

Make sure it's easy to understand for an external contributor to the CDK.

packages/jsii/lib/assembler.ts
RomainMuller: Basically, imported types respond `false` to all `is*` methods. :(

packages/jsii/lib/assembler.ts
RomainMuller:
```
Illegal "extends" value for an exported API: Omit<SomeInterface, 'somefield'>
```

I'd have wanted to make it more specific (mapped/utility types are not allowed here), however the type checker does not really give me a way to do this with confidence...

nija-at: This looks fairly good.

PR #1709, opened 2020-06-02
feat(pacmak/java): isolate maven repositories

RomainMuller:
When preparing a Java package using Maven, `jsii-pacmak` now configures
a dedicated `localRepository` for its own use. This prevents artifacts
cached from previous builds as well as local build artifacts to possibly
interfere with the compilation process.

This has the side-effect to force a re-download of all dependencies on
each build (except those that are generated as part of the same build),
which results in a performance degradation. It should be possible to
make the local repository location user-configurable so this impact can
be mitigated on a case-by-case basis.

---


packages/@jsii/java-runtime/.gitignore
```diff
@@ -1,3 +1,4 @@
+.m2
```
rix0rrr, line 1: Why oh why not in a tempdir somewhere?

RomainMuller: Because local iterations will be *much* faster for not re-downloading things every single time. You can nuke this directory when you need it...

PR #1717, opened 2020-06-08
fix(@jsii/runtime): "maximum call stack size exceeded" in SyncStdio.readLine

RomainMuller:
When using node `>= 12.17`, `EAGAIN` errors consistently occur when
trying to read from `stdin` when there is no available data. The retry
mechanism for this was to recursively call back `SyncStdio.readLine`,
which could evtnually lead to a "Maximum call stack size exceeded" error
to occur (possibly hidden from the consumer, and later causing a "Stream
closed" error).

This changes how the retry mechanism works so it operates in a `while`
loop instead of making a recursive call, completely avoiding to run into
the growing stack issue.

Fixes aws/aws-cdk#8288
Fixes aws/aws-cdk#5187
Fixes aws/aws-cdk#8397

---


packages/@jsii/runtime/lib/sync-stdio.ts
```diff
@@ -18,45 +18,41 @@ export class SyncStdio {
...
+        for (const ch of str) {
+          if (ch === '\n') {
+            this.inputQueue.push(this.currentLine);
+            this.currentLine = '';
+          } else {
+            this.currentLine += ch;
+          }
+        }
+      } catch (e) {
+        // HACK: node may set O_NONBLOCK on it's STDIN depending on what kind of input it is made
```
eladb, line 41: We control both sides of the pipe. In what conditions will node use non blocking? Is this something we can configure on the runtime side?

rix0rrr: According to the linked thread, something as simple as calling `console.log()` may switch `process.stdin` to non-blocking mode.

packages/@jsii/runtime/lib/sync-stdio.ts
```diff
@@ -18,45 +18,41 @@ export class SyncStdio {
...
+        // of (see https://github.com/nodejs/help/issues/2663). When STDIN has O_NONBLOCK, calls may
+        // result in EAGAIN. In such cases, the call should be retried until it succeeds. This kind
+        // of polling will result in horrible CPU thrashing, but there does not seem to be a way to
+        // force a O_SYNC access to STDIN in a reliable way within node.
+        // In order to make this stop we need to either stop depending on synchronous reads, or to
```
rix0rrr, line 46: > stop depending on synchronous reads

That would effectively mean callbacks would stop working, right?

RomainMuller: Why no. This means asynchronous user-code suddenly would throw everything off. Pure-async kernel would work okay if we could guarantee client code NEVER does anything asynchronous by itself.

PR #1736, opened 2020-06-15
fix(jsii): excessive "exclude" in "tsconfig.json"

RomainMuller:
In cases where both `jsii.tsc.outDir` and `jsii.tsc.rootDir` are set, and `rootDir` happens to be a parent directory of `outDir`, the generated `exclude` would match all files from `rootDir`, resulting in an empty input set.

This makes sure to not emit an `exclude` declaration for `outDir` in this case, so that input files do not get excluded.

---


packages/jsii/lib/compiler.ts
```diff
@@ -289,8 +289,12 @@ export class Compiler implements Emitter {
       ],
       exclude: [
         'node_modules',
-        ...pi.excludeTypescript,
```
eladb, line 292: can we write a test for this?

eladb (commented):
Ping

RomainMuller:
The PR includes a test by way of jsii-calc-base-of-base... Are you looking for something else?

eladb (approved):

eladb:
What's up with this PR?

PR #1766, opened 2020-06-30
fix(kernel): tarball unpacking does not behave like 'npm install'

### LINKED ISSUE #1765: Kernel "load" results in surprising file permissions
### Affected Languages

- [X] `Python`
- [X] `Java`
- [X] .NET (`C#`, `F#`, ...)

### General Information
* **JSII Version:** *all existing versions*
* **Platform:**     *UNIX-like with a non-default `umask` such as `0027`*

### What is the problem?
Libraries loaded in the jsii kernel are not installed by `npm`, but directly unpacked. The `umask` is not overridden, leading to the current process configuration to be applied. This leads to installs that differ in file permissions from their `npm install`ed counterparts.

This has been found to cause aws/aws-cdk#8233

RomainMuller:
In it's wisdom, `npm install` does override the process' `umask` to
`0o022` before unpackging the tarball, to ensure the produced install
has the kind of permissions that one would expect, regardless of the
system-configured `umask`.

Because `@jsii/kernel` did not reproduce this behavior, loaded libraries
could be unpacked with unexpectedly tight permissions, leading to weird
issues when those files were used in contexts that required those
permissions. For example, this is the cause of aws/aws-cdk#8233.

Fixes #1765

---

### DIFF
--- a/packages/@jsii/kernel/lib/kernel.ts
+++ b/packages/@jsii/kernel/lib/kernel.ts
@@ -109,7 +109,7 @@ export class Kernel {
     fs.mkdirpSync(packageDir);
 
     // Force umask to have npm-install-like permissions
-    const umask = process.umask(0o022);
+    const originalUmask = process.umask(0o022);
     try {
       // untar the archive to its final location
       tar.extract({
@@ -122,7 +122,7 @@ export class Kernel {
       });
     } finally {
       // Reset umask to the initial value
-      process.umask(umask);
+      process.umask(originalUmask);
     }
 
     // read .jsii metadata from the root of the package
--- a/packages/@jsii/kernel/lib/kernel.ts
+++ b/packages/@jsii/kernel/lib/kernel.ts
@@ -107,14 +107,23 @@ export class Kernel {
 
     // Create the install directory (there may be several path components for @scoped/packages)
     fs.mkdirpSync(packageDir);
-    // untar the archive to its final location
-    tar.extract({
-      strict: true,
-      file: req.tarball,
-      cwd: packageDir,
-      strip: 1, // Removes the 'package/' path element from entries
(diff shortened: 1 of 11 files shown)


packages/@jsii/kernel/lib/kernel.ts
```diff
@@ -107,14 +107,23 @@ export class Kernel {
...
-      cwd: packageDir,
-      strip: 1, // Removes the 'package/' path element from entries
-      sync: true,
-    });
+
+    // Force umask to have npm-install-like permissions
+    const originalUmask = process.umask(0o022);
+    try {
+      // untar the archive to its final location
+      tar.extract({
```
eladb, line 115: You should use the built-in support in the [`tar`](https://www.npmjs.com/package/tar#tarcoptions-filelist-callback-alias-tarcreate) library.

One of these options should help:

* `portable` Omit metadata that is system-specific: ctime, atime, uid, gid, uname, gname, dev, ino, and nlink. Note that mtime is still included, because this is necessary for other time-based operations. Additionally, mode is set to a "reasonable default" for most unix systems, based on a umask value of 0o22.
* `mode` The mode to set on the created file archive

eladb: I feel `portable` is the right one to use here.

RomainMuller: Those options are specific to the `create` operation and this is the `extract` operation.

The only "built-in" option available is the `umask` option to `tar.extract`, however that is "internal" (and not exposed in the `d.ts` which is why I **decided** against using that).

eladb: Okay sorry my mistake.

PR #1779, opened 2020-07-07
fix(jsii-reflect): TypeSystem.methods and .properties ignores submodules

RomainMuller:
Contrary to their desired behavior, the `properties` and `methods`
accessors on `TypeSystem` do no list *all* members in the loaded
assemblies, as they do no traverse `submodules`.

This commit changes the behavior so submodules are correctly traversed
when collecting methods and properties.

---


packages/jsii-reflect/lib/type-system.ts
```diff
@@ -238,23 +238,35 @@ export class TypeSystem {
...
   public get methods() {
-    const out = new Array<Method>();
-    this.assemblies.forEach((a) => {
-      a.interfaces.forEach((t) => out.push(...t.ownMethods));
-      a.classes.forEach((t) => out.push(...t.ownMethods));
-    });
-    return out;
+    const getMethods = (mod: ModuleLike): readonly Method[] => {
+      return [
+        ...flatMap(mod.submodules, getMethods),
```
rix0rrr, line 243: Wow this is loco!

PR #1782, opened 2020-07-09
fix(pacmak): python pack fails when installing 'black' via pip

nija-at:
The failure is because of intermittent unavailability of black or one of
its packages from the PyPI index.

Instead, install back into the environment, i.e., superchain, and use
that.

---


eladb:
Feels a bit weird to include this in superchain, but I see the pragmatic reasoning.

PR #1783, opened 2020-07-13
fix(pacmak): race condition in python packing when run over multiple packages

nija-at:
The root cause seems to be that the async APIs in Node's fs modules
isn't strictly single threaded.

When packed over several packages, pacmak schedules these in parallel,
creating a race condition on the local cache for Python Black.
All threads thrash to try and set up the package directory and one of
them fails.

The fix is to introduce a process wide promise that awaits until the
Python Black is fully set up.

Testing
Unit tests for such race conditions are hard to write correctly. This
has been verified manually by running against 10 CDK packages at a time.
I was able to confirm that the issue does show up consistently without
this fix and the fix works correctly.

---

### DIFF
--- a/packages/jsii-pacmak/lib/targets/python.ts
+++ b/packages/jsii-pacmak/lib/targets/python.ts
@@ -34,7 +34,7 @@ import { die, toPythonIdentifier } from './python/util';
 const spdxLicenseList = require('spdx-license-list');
 
 export default class Python extends Target {
-  private static BLACK_PATH: Promise<string> | undefined;
+  private static BLACK_PATH?: Promise<string>;
 
   protected readonly generator: PythonGenerator;
 
--- a/packages/jsii-pacmak/lib/targets/python.ts
+++ b/packages/jsii-pacmak/lib/targets/python.ts
@@ -34,6 +34,8 @@ import { die, toPythonIdentifier } from './python/util';
 const spdxLicenseList = require('spdx-license-list');
 
 export default class Python extends Target {
+  private static BLACK_PATH: Promise<string> | undefined;
+
   protected readonly generator: PythonGenerator;
 
   public constructor(options: TargetOptions) {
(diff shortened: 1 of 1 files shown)


packages/jsii-pacmak/lib/targets/python.ts
```diff
@@ -45,8 +47,13 @@ export default class Python extends Target {
   public async generateCode(outDir: string, tarball: string): Promise<void> {
     await super.generateCode(outDir, tarball);
 
+    // Using a static variable as a lock to prevent racing. Since blackPath() uses
+    // Promise APIs from fs and os modules (that use libuv), an additional lock is required.
+    if (Python.BLACK_PATH === undefined) {
+      Python.BLACK_PATH = this.blackPath();
+    }
```
RomainMuller, line 54: I'd have wrapped this caching within `this.blackPath()` instead, possibly even making `blackPath()` static while there. This would "guard" against incorrectly calling `blackPath` multiple times?

packages/jsii-pacmak/lib/targets/python.ts
```diff
@@ -34,6 +34,8 @@ import { die, toPythonIdentifier } from './python/util';
 const spdxLicenseList = require('spdx-license-list');
 
 export default class Python extends Target {
+  private static BLACK_PATH: Promise<string> | undefined;
```
RomainMuller, line 37:
```suggestion
  private static BLACK_PATH?: Promise<string>;
```

packages/jsii-pacmak/lib/targets/python.ts
nija-at: I like the layering that I have already. It lets me test `blackPath()`. I'll have to possibly mock a few extra things to get the correct test behaviour of current `blackPath()` (see unit tests that are already there).

I don't expect this to be called again from anywhere else. The method organization is simply readability and testability.

How strongly do you feel?

RomainMuller: Not very :) And you've convinced me there was enough reason to leave it like so.

nija-at: 😊

PR #1802, opened 2020-07-20
chore: add workflow to run "yarn upgrade" periodically

RomainMuller:
This makes sure our lockfile remains up-to-date, including for
transitive dependencies, and reduces the likelihood of receiving a
sevurity notification from GitHub for one of those.

---

### DIFF
--- a/.github/workflows/yarn-upgrade.yml
+++ b/.github/workflows/yarn-upgrade.yml
@@ -1,0 +1,62 @@
+name: Yarn Upgrade
+
+on:
+  schedule:
+    # Every wednesday at 13:37 UTC
+    - cron: 37 13 * * 3
+  workflow_dispatch: {}
+
+jobs:
+  upgrade:
+    name: Yarn Upgrade
+    runs-on: ubuntu-latest
+    steps:
+
+      - name: Check Out
+        uses: actions/checkout@v2
+
+      - name: Set up Node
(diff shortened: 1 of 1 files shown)


.github/workflows/yarn-upgrade.yml
```diff
@@ -0,0 +1,62 @@
...
+      - name: Locate Yarn cache
+        id: yarn-cache
+        run: echo "::set-output name=dir::$(yarn cache dir)"
+
+      - name: Restore Yarn cache
+        uses: actions/cache@v2
+        with:
+          path: ${{ steps.yarn-cache.outputs.dir }}
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |-
```
njlynch, line 32: I don't see a reference to this trailing dash anywhere in the docs or other examples. yml magic I don't know about, or typo? Should this be:
```suggestion
          restore-keys: |
```

RomainMuller: The `-` is the [block chomping indicator](https://yaml.org/spec/1.2/spec.html#id2794534) for stripping the final break and any trailing empty lines from the multiline string.

njlynch: Ah, TIL. I hate Googling for odd syntax "yaml pipe dash" didn't really do a lot for me. :D

PR #1817, opened 2020-07-24
chore(pacmak): improve error messaging

RomainMuller:
When a shelled-out process fails (either on a non-0 status, or killed by
a signal), the error message did not include the command that just
failed, causing ambiguity with respects to what happened. Additionally,
the output included STDOUT and STDERR concatenated (in this order)
without any way to distinguishe which is which... This adds a prefix
string to each line of those to denote whether STDOUT or STDERR.

---


packages/jsii-pacmak/lib/util.ts
```diff
@@ -69,14 +69,23 @@ export async function shell(
...
         const err = Buffer.concat(stderr).toString('utf-8');
-        if (code != null) {
-          return ko(
-            new Error(`Process exited with status ${code}\n${out}\n${err}`),
-          );
-        }
         return ko(
-          new Error(`Process terminated by signal ${signal}\n${out}\n${err}`),
+          new Error(
+            `Process exited with ${
```
nija-at, line 74: nit: I still like 'terminated' over 'exited' to make it clear it was unexpected.

RomainMuller: I've actually seen "failed" used, which I like a little more than "terminated" (because termination - to me - is pertaining to SIGKILL / SIGTERM. But that is a jolly good suggestion!

PR #1842, opened 2020-07-31
fix(rosetta): OOpsie -- couldn't find root file back on Windows

RomainMuller:
The root file was located by comparing file paths in a way that was not
adjusting path separators for Windows, and the result was inability to
match normalized paths with Unix-style ones.

This normalizes paths before attempting to match them, so that the match
can succeed.

---

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -152,7 +152,7 @@ jobs:
           name: release-package
           path: ${{ github.workspace }}/dist/
   test:
-    name: Test (${{ runner.os }} / java ${{ matrix.java }} / node ${{ matrix.node }} / python ${{ matrix.python }})
+    name: Test (${{ matrix.os }} / java ${{ matrix.java }} / node ${{ matrix.node }} / python ${{ matrix.python }})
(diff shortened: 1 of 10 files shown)


.github/workflows/main.yml
```diff
@@ -75,6 +80,60 @@ jobs:
...
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.6'
+      - name: 'Linux: Install python3-venv'
+        if: runner.os == 'Linux'
+        run: sudo apt install -y python3-venv
+      - name: 'Windows: Expose python3 command'
+        if: runner.os == 'Windows'
+        shell: bash
+        run: cp ${pythonLocation}/python.exe ${pythonLocation}/python3.exe
```
njlynch, line 111: When will this evaluate to true, given this all "runs-on" ubuntu-latest?

RomainMuller: You're correct. I've kinda standardized on using the same setup process everywhere, but this never runs on Windows, so I should probably just remove that...

PR #1843, opened 2020-07-31
chore: fix running tests on Windows

RomainMuller:
Trying to remove the elements that make tests not pass on Windows,
or tweaking the GitHub workflow so windows can pass (proving there
is a way to develop jsii on Windows).

Important changes include:
- Workflow uploads tarball instead of set of files (it's much faster)
- Replace diff-test.sh with a snapshot-based jest test
- Making `python` stuff Windows-friendly (`bin` vs. `Scripts`)
- Fixing some unit tests that turned out flaky

---


RomainMuller:
Starting to guess this error is because of a `\` versus `/` again... Windows, amirite?

RomainMuller:
Yay - it now works. but there's an issue with `jsii-tree` tests on Windows (that's what we are here for, I guess!)

PR #1850, opened 2020-08-05
fix(rosetta): use of keyword 'lambda' produces invalid Python code

nija-at:
Replace all 'lambda' identifiers with 'awslambda'

fixes https://github.com/aws/aws-cdk/issues/9169

---


packages/jsii-rosetta/lib/languages/python.ts
```diff
@@ -833,6 +839,10 @@ const TOKEN_REWRITES: { [key: string]: string } = {
   false: 'False',
 };
 
+const IDENTIFIER_KEYWORDS: { [key: string]: string } = {
+  lambda: 'awslambda',
```
rix0rrr, line 843: This is too CDK-specific for JSII. Why not `lambda_` ? It's not great but a simple convention.

In fact, any keyword can be non-keyworded by appending `_`.

PR #1861, opened 2020-08-07
chore(jsii): better errors around use of hidden types

### LINKED ISSUE #1860: How to keep base classes private?
### The Question

This is in response to [this post](https://github.com/aws/jsii/issues/1818#issuecomment-665561112) of @RomainMuller:

> [...] those types you are not exporting from the module's entry point (`lib/index.ts`) cannot be used anywhere in exported declarations... But their TypeScript AST nodes have the `Export` modifier (since they are internally exported). The `jsii` compiler normally erases unexpected types from the exported API definition (so you can have "private base classes" and it "just works", however for those cases it requires the declarations to have `@internal` as part of their JSDoc.

I tried to keep my base classes private by adding the `@internal` tag to the tsdoc, but I'm having no luck. Still seeing lots of errors like `error TS0: Exported APIs cannot use un-exported type my-package.Foo`.

Is there an example available or could you have a look at my project again? 🐱 

This is what I'm getting in my [current attempt](https://github.com/udondan/iam-floyd/pull/23):

```
error TS0: Exported APIs cannot use un-exported type iam-floyd.PolicyStatementWithActions
error TS0: Exported APIs cannot use un-exported type iam-floyd.PolicyStatementWithPrincipal
error TS0: Exported APIs cannot use un-exported type iam-floyd.PolicyStatementWithEffect
error TS0: Exported APIs cannot use un-exported type iam-floyd.PolicyStatementWithCondition
error TS0: Exported APIs cannot use un-exported type iam-floyd.PolicyStatementWithResources
```

All of those classes have the `@internal` flag set, e.g. https://github.com/udondan/iam-floyd/blob/e308174fe9afaeb7ed881a3da59fba37c241ca0b/lib/shared/policy-statement/3-actions.ts#L22

RomainMuller:
When a hidden (that is, `@internal` or not exported) type is used in a
visible (`public` or `protected` on an exported type) API, the error
produced would refer to the unusable type, but would not give any
indication of where it was being used from.

This makes several enhancements to this process:
- Qualify the kind of use for the type (return, parameter, ...)
- Attach the error to the resolving node (usage location)
- Provide a related message with the unusable type's declaration
- Specifically message around "this" (used or inferred as a return type)

This is going to particularly enhance the experience of folks extending
internal base types, where those internal base types declare members
that return hidden types (or "this").

Fixes #1860

---

### DIFF
--- a/packages/jsii/lib/assembler.ts
+++ b/packages/jsii/lib/assembler.ts
@@ -621,7 +621,7 @@ export class Assembler implements Emitter {
         symbol.name !== Case.snake(symbol.name)
       ) {
         this._diagnostic(
-          declaration,
+          declaration.name,
           ts.DiagnosticCategory.Error,
           `Submodule namespaces must be camelCased or snake_cased. Consider renaming to "${Case.camel(
             symbol.name,
@@ -706,25 +706,39 @@ export class Assembler implements Emitter {
         if (currNs.name !== ns.name) {
           const currNsDecl = currNs.valueDeclaration ?? currNs.declarations[0];
           const nsDecl = ns.valueDeclaration ?? ns.declarations[0];
+
+          // Make sure the error message always lists causes in the same order
+          const refs = [
+            [currNs.name, currNsDecl] as const,
+            [ns.name, nsDecl] as const,
+          ].sort(([l], [r]) => l.localeCompare(r));
+
           this._diagnostic(
-            symbol.valueDeclaration,
+            (symbol.valueDeclaration as { name?: ts.Node }).name ??
+              symbol.valueDeclaration,
             ts.DiagnosticCategory.Error,
-            `Symbol is re-exported under two distinct submodules (${currNs.name} and ${ns.name})`,
+            `Symbol is re-exported under two distinct submodules (${refs
+              .map(([name]) => name)
+              .join(' and ')})`,
             [
               {
                 category: ts.DiagnosticCategory.Warning,
-                file: currNsDecl.getSourceFile(),
-                length: currNsDecl.getStart() - currNsDecl.getEnd(),
-                messageText: `Symbol is exported under the "${currNs.name}" submodule`,
-                start: currNsDecl.getStart(),
+                file: refs[0][1].getSourceFile(),
+                length:
+                  refs[0][1].getEnd() -
+                  refs[0][1].getStart(refs[0][1].getSourceFile()),
+                messageText: `Symbol is exported under the "${refs[0][0]}" submodule`,
+                start: refs[0][1].getStart(refs[0][1].getSourceFile()),
                 code: JSII_DIAGNOSTICS_CODE,
               },
               {
                 category: ts.DiagnosticCategory.Warning,
-                file: nsDecl.getSourceFile(),
-                length: nsDecl.getStart() - nsDecl.getEnd(),
-                messageText: `Symbol is exported under the "${ns.name}" submodule`,
-                start: nsDecl.getStart(),
+                file: refs[1][1].getSourceFile(),
+                length:
+                  refs[1][1].getEnd() -
+                  refs[1][1].getStart(refs[1][1].getSourceFile()),
+                messageText: `Symbol is exported under the "${refs[1][0]}" submodule`,
+                start: refs[1][1].getStart(refs[1][1].getSourceFile()),
                 code: JSII_DIAGNOSTICS_CODE,
               },
             ],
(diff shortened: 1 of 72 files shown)


RomainMuller:
Added another commit to actually switch the way we test for errors in the "negatives" to actually look at the rendered error entirely (thanks to jest snapshots) in order to have a better picture of the error UX we are offering.

packages/jsii/lib/assembler.ts
```diff
@@ -397,23 +405,49 @@ export class Assembler implements Emitter {
...
+      hasError = true;
+    }
+
+    const groups = /^"([^"]+)"\.(.*)$/.exec(tsName);
+    if (!groups) {
+      if (!hasError) {
+        this._diagnostic(
+          resolvingNode,
+          ts.DiagnosticCategory.Error,
+          `Cannot internal type ${tsName} as a ${typeUse} in exported declarations`,
```
njlynch, line 432:
```suggestion
          `Cannot use internal type ${tsName} as a ${typeUse} in exported declarations`,
```

packages/jsii/lib/assembler.ts
```diff
@@ -397,23 +405,49 @@ export class Assembler implements Emitter {
...
+      const commonMessage = `cannot be used as the ${typeUse} because it is private or @internal`;
       this._diagnostic(
-        node,
+        resolvingNode,
         ts.DiagnosticCategory.Error,
-        `Cannot use private type ${tsName} in exported declarations`,
+        isThisType
+          ? `Type "this" (aka: "${type.symbol.name}") ${commonMessage}`
+          : `Type "${type.symbol.name}" ${commonMessage}`,
+        makeCause(node),
```
njlynch, line 420: I'm not sure I understand what the difference is between `resolvingNode` and `node`; would you want to use `resolvingNode` here in the `makeCause` call?

packages/jsii/lib/assembler.ts
```diff
@@ -487,7 +546,7 @@ export class Assembler implements Emitter {
   ): Promise<void> {
     const declaration = symbol.valueDeclaration ?? symbol.declarations[0];
     if (declaration == null) {
-      // Nothing to do here...
+      // Nothing to do here..
```
njlynch, line 549: :/  I vote either zero, one, or three; two periods/full stops is just weird.
```suggestion
      // Nothing to do here
```

packages/jsii/lib/assembler.ts
```diff
@@ -506,7 +565,7 @@ export class Assembler implements Emitter {
       return;
     }
     if (!ts.isNamespaceExport(declaration)) {
-      // Nothing to do here...
+      // Nothing to do here..
```
njlynch, line 568:
```suggestion
      // Nothing to do here
```

packages/jsii/lib/assembler.ts
```diff
@@ -645,25 +704,38 @@ export class Assembler implements Emitter {
         if (currNs.name !== ns.name) {
           const currNsDecl = currNs.valueDeclaration ?? currNs.declarations[0];
           const nsDecl = ns.valueDeclaration ?? ns.declarations[0];
+
+          const refs = [
+            [currNs.name, currNsDecl] as const,
+            [ns.name, nsDecl] as const,
+          ].sort(([l], [r]) => l.localeCompare(r));
```
njlynch, line 711: A comment here would be nice. Was this just to ensure the names are always in the same order, or something more subtle?

packages/jsii/test/negatives/neg.behavior-requires-iprefix.ts
```diff
@@ -1,5 +1,3 @@
-///!MATCH_ERROR: Interface contains behavior: name should be "ISomething"
```
njlynch, line 1: I'm not seeing this error (or equivalent) in the snapshot. Am I missing it, or did this get dropped somehow?

packages/jsii/lib/assembler.ts
RomainMuller: Good catch!

packages/jsii/lib/assembler.ts
RomainMuller: Let me try a different name that just occurred to me...

packages/jsii/test/negatives/neg.behavior-requires-iprefix.ts
RomainMuller: 😨 OMG good catch. I guess I was never asserting there is > 0 problems found...

RomainMuller: Oh and the issue turned out to be essentially that location-less problem markers are NOT rendered with the API style I was going for... Dun dun dun!

PR #1864, opened 2020-08-10
chore: stop running generated Python though `black`

### LINKED ISSUE #1856: jsii-pacmak very slow running python in fresh pipenv
### Affected Languages

- [X] `Python`
- [ ] `Java`
- [ ] .NET (`C#`, `F#`, ...)

### General Information
* **JSII Version:** 1.7.0
* **Platform:**     Linux docker-desktop 4.19.104-microsoft-standard

### What is the problem?
jsii-pacmak is very slow when python is compiling 700+ files (https://github.com/hashicorp/terraform-cdk/pull/276) in a fresh pipenv. On the order of 10 minutes.

I managed to track the main slowness of trying to integrate jsii-srcmak into cdktf to running python setup. I'm not certain how much of the time is sdist vs wheel though.
Subsequent runs in the same environment are much faster. Still slower than ideal, but #1854 and #1855 are more relevant in that case.

@skorfmann also observed "one python process takes up the majority of the time and consumes up to 2.8 GB of memory  at 100% cpu" while building the cdk tf aws provider (https://github.com/skorfmann/cdktf-provider-aws)

RomainMuller:
Running `black` on the generated source is slow... And generating
*properly* formatted code is faster. This removes `black` invocations
and instead tries to generate Python that is closer to the prescriptions
of PEP8 (more specifically, closer to the output of `black`).

There are some minor changes introduced by this change in the generated
Python files, compared to when `black` was post-processing, which are
hopefully acceptable degradations:

- Some long type signatures are single-lined now where `black` would
  have broken them down into multiple lines. Reproducing the `black`
  behavior is too close to requiring a full blown python parser.
- Certain lines that `black` chose not to break down into multiple lines
  are now split. I could not understand the mechanisms `black` uses to
  make it's choice and decided the broken down lines are readable
  enough.
- Some `pass` statements used to be generated in proxies when there were
  no members, although there existed a synthetic member. Those
  statements were not sparking joy and were gently disposed of.
- Alphanumerically ordered elements in piecemeal import declarations, so
  as to reduce the likelyhood of fake changes if the order of processing
  changes (resulting in a different creation/registration order).

Fixes #1856

---


jsteinich:
Thanks for working on this. Any estimate when a version will be released with the change?

PR #1866, opened 2020-08-10
fix(jsii): unable to use nested types from dependencies

RomainMuller:
Using `--project-references`, nested types from dependencies could not
be used as they would result in the following `jsii` error:

```
Type "<type name>" cannot be used as the return type because it is private or @internal
```

THis is because in `--project-references` mode, the ambient declaration
for the nested type is the declaration that gets resolved to, and those
cannot have the `export` keyword: instead, the surrounding module
declaration is annotated with `export ambient`.

This adds the necessary code path to actually identify this scenario and
to appropriately detect the type is actually visible and exported.

The check that was failing had been added in #1861 as a way to provide a
more helpful error message when private or `@internal` types are
mistakenly used on exported APIs; which explains why this situation did
not previously occur.

> Hidden gem: we were previously adding the `static` keyword when
> generating nested classes in **C#**. This is a **Java**-ism and results
> in code that won't compile, since **C#** does not allow `static` nested
> classes to declare `protected` members (which we actually emit some
> of).

---


packages/jsii-pacmak/lib/targets/dotnet/dotnetgenerator.ts
```diff
@@ -294,7 +293,7 @@ export class DotNetGenerator extends Generator {
     this.dotnetRuntimeGenerator.emitAttributesForClass(cls);
 
     this.code.openBlock(
-      `public${inner}${absPrefix} class ${className}${implementsExpr}`,
+      `public${absPrefix} class ${className}${implementsExpr}`,
```
RomainMuller, line 296: C# does not allow static member classes to declare protected members (our contract includes some, so that was bad). The presence of the `static` keyword here was a Java-ism, since nested classes in C# (with or without `static`) are actually semantically related to Java static nested classes (the syntactic sugar that Java has for non-static nested classes does not exist in C#).

rix0rrr: This should have been a separate PR, don'tcha agree?

RomainMuller:
> Approved but you know you've been a naughty boy and you should really split this up into 2 PRs before you merge it 😎

How so? I need both changes in order for the test around either to pass...

packages/jsii-pacmak/lib/targets/dotnet/dotnetgenerator.ts
RomainMuller: Except that separate PR's test would not have been able to pass without the other changes in this PR, and vice-versa...

PR #1871, opened 2020-08-11
fix(jsii): README.md filename is case sensitive

### LINKED ISSUE #1862: Readme.md Filename Case Sensitive for Python
### Affected Languages

- [x] `Python`
- [ ] `Java`
- [ ] .NET (`C#`, `F#`, ...)

### General Information
* **JSII Version:** 1.10.0 (build 27cfbed), typescript 3.9.7
* **Platform:**  Darwin Kernel Version 19.5.0: Tue May 26 20:41:44 PDT 2020; root:xnu-6153.121.2~2/RELEASE_X86_64 x86_64

### What is the problem?

When publishing a package to Pypi, it gets rejected with the following error message when the readme file is named like this `Readme.md`. 

```
Uploading skorfmann_cdktf_provider_aws-0.0.9-py3-none-any.whl

  0%|          | 0.00/2.73M [00:00<?, ?B/s]
  0%|          | 8.00k/2.73M [00:00<00:50, 56.0kB/s]
 10%|█         | 288k/2.73M [00:00<00:32, 79.2kB/s] 
 67%|██████▋   | 1.83M/2.73M [00:00<00:08, 113kB/s]
HTTPError: 400 Bad Request from https://upload.pypi.org/legacy/
100%|██████████| 2.73M/2.73M [00:00<00:00, 4.14MB/s]
The description failed to render for 'text/markdown'. See https://pypi.org/help/#description-content-type for more information.
Content received from server:
<html>
 <head>
  <title>400 The description failed to render for 'text/markdown'. See https://pypi.org/help/#description-content-type for more information.</title>
 </head>
 <body>
  <h1>400 The description failed to render for 'text/markdown'. See https://pypi.org/help/#description-content-type for more information.</h1>
  The server could not comply with the request since it is either malformed or otherwise incorrect.<br/><br/>
The description failed to render for &#x27;text/markdown&#x27;. See https://pypi.org/help/#description-content-type for more information.
```

I was able to fix it by changing the filename to `README.md`

It was working on my local machine - I guess due to the case insensitive filesystem on OSX. - but failing in CI.

RomainMuller:
The `README.md` file was processed in a case sensitive manner, but it
does not really matter what casing it uses. Changed the flow to look for
the file in a case-insensitive manner instead.

Fixes #1862

---


packages/jsii/lib/assembler.ts
```diff
@@ -226,7 +226,12 @@ export class Assembler implements Emitter {
     }
 
     async function _loadReadme(this: Assembler) {
-      const readmePath = path.join(this.projectInfo.projectRoot, 'README.md');
+      // Search for `README.md` in a case-insensitive way
+      const fileName =
+        (await fs.readdir(this.projectInfo.projectRoot)).find(
+          (file) => file.toLocaleLowerCase() === 'readme.md',
+        ) ?? 'README.md';
+      const readmePath = path.join(this.projectInfo.projectRoot, fileName);
```
njlynch, line 234: It feels a bit weird you read the files in the directory, then check to see if the path exists. I think the below reads a bit cleaner. Also, why the `await fs.readdir` vs `fs.readdirSync`?

```suggestion
      // Search for `README.md` in a case-insensitive way
      const fileName =
        fs.readdirSync(this.projectInfo.projectRoot).find(
          (file) => file.toLocaleLowerCase() === 'readme.md',
        );
      if (!fileName) { return undefined; }
      const readmePath = path.join(this.projectInfo.projectRoot, fileName);
      // Skip the fs.pathExists... check below
```

RomainMuller: You're right. I guess I've been lazy 👍

packages/jsii/lib/assembler.ts
```diff
@@ -226,7 +226,15 @@ export class Assembler implements Emitter {
...
+        (await fs.readdir(this.projectInfo.projectRoot)).find(
+          (file) => file.toLocaleLowerCase() === 'readme.md',
+        );
+      if (fileName == null) {
+        return undefined;
+      }
+      const readmePath = path.join(this.projectInfo.projectRoot, fileName);
       if (!(await fs.pathExists(readmePath))) {
         return undefined;
       }
```
njlynch, line 240: Are these lines still needed? Trying to think of a scenario where `readdir` matches a file but `pathExists` returns false.

RomainMuller: 🤦🏻‍♂️

PR #1877, opened 2020-08-12
chore: sanitize the .NET codebase (a little)

RomainMuller:
Configured a bunch of Roslyn analyzers to ensure code quality, and made the necessary corrections:

- Correct usage of `ArgumentException`
- Use `Array.Empty<T>()` to spare allocations of empty objects
- Override equality on data types
- Specify `InvariantCulture` where needed (could be the source of weird bugs)
- Some other miscellaneous tweaks

Additionally, folded all the contents of `Amazon.JSII.JsonModel` under the `Amazon.JSII.Runtime`
assembly, and issued type frowards in `Amazon.JSII.JsonModel` to maintain backwards compatibility.

---


packages/@jsii/dotnet-runtime/src/Amazon.JSII.Analyzers.UnitTests/Helpers/DiagnosticResult.cs
```diff
@@ -54,10 +89,50 @@ public DiagnosticResultLocation[] Locations
...
+        }
+
+        public bool Equals(DiagnosticResult other)
+        {
+            return Id == other.Id
+                   && Severity == other.Severity
+                   && Message == other.Message
+                   && Path == other.Path
+                   && Line == other.Line
+                   && Column == other.Column;
```
njlynch, line 125: This is making the assumption that two `DiagnosticResults` with the same first `DiagnosticResultLocation` are the same, regardless of the other `DiagnosticResultLocation` in `Locations`. I lack context to know if that's valid or not; I _suspect_ it's ok, but worth calling it out.

packages/@jsii/dotnet-runtime/src/Amazon.JSII.Analyzers/JsiiOptionalAnalyzer.cs
```diff
@@ -15,8 +16,8 @@ public class JsiiOptionalAnalyzer : DiagnosticAnalyzer
         private const string Title = "A required property is missing or null";
         private const string MessageFormat = "The property is required and cannot be null";
         private const string MessageFormatWithPropertyName = "The property {0} is required and cannot be null";
```
njlynch, line 18: You added periods to the two below messages; any reason not to add them to these?

packages/@jsii/dotnet-runtime/src/Amazon.JSII.Analyzers/JsiiOptionalAnalyzer.cs
```diff
@@ -15,8 +16,8 @@ public class JsiiOptionalAnalyzer : DiagnosticAnalyzer
...
-        private const string Description = "The property is required and cannot be null";
-        private const string DescriptionWitPropertyName = "The property {0} is required and cannot be null";
+        private const string Description = "The property is required and cannot be null.";
+        private const string DescriptionWitPropertyName = "The property {0} is required and cannot be null.";
```
njlynch, line 20:
```suggestion
        private const string DescriptionWithPropertyName = "The property {0} is required and cannot be null.";
```

packages/@jsii/dotnet-runtime/src/Amazon.JSII.Analyzers/JsiiOptionalAnalyzer.cs
RomainMuller: Basically it's the following rules from the Roslyn analyzer for Roslyn analyzers (so meta):
- [RS1033](https://github.com/dotnet/roslyn-analyzers/blob/master/src/Microsoft.CodeAnalysis.Analyzers/Microsoft.CodeAnalysis.Analyzers.md#rs1033-define-diagnostic-description-correctly)
  > The diagnostic description should be one or multiple sentences ending with a punctuation sign
- [RS1032](https://github.com/dotnet/roslyn-analyzers/blob/master/src/Microsoft.CodeAnalysis.Analyzers/Microsoft.CodeAnalysis.Analyzers.md#rs1032-define-diagnostic-message-correctly)
  > The diagnostic message should not contain any line return character and should either be a single sentence without a trailing period or a multi-sentences with a trailing period

PR #1889, opened 2020-08-14
chore: use pep517 tool to build wheels and sdist

RomainMuller:
The `pep517` tool offers proper support for PEP 517, including
build isolation, while it is able to produce both a wheel and a
source distribution, making it fit for the purpose. It reduces
the amount of things that must be manually installed in the
process.

Bonus content (tangentially related to the original intent):
- Allowing all the `venv`s to use system site packages, as this
  can make the process faster if requirements are already
  satisfied by "globally" available distributions
- Some version bumps in required python tooling that gets
  installed "by hand" at some points
- Migrating the `@jsii/python-runtime` package to use the
  `setup.cfg` format instead of all `setup.py`, so it actually
  works well with `pep517` (it might have been technically
  possible to make it work without this change, but the
  documentation didn't seem to discuss this as an option at
  all, so I preferred not to)

---


.github/workflows/main.yml
```diff
@@ -194,19 +195,19 @@ jobs:
             dotnet: '3.1.x'
             java: '8'
             node: '10'
-            python: '3.6'
+            python: '3.8'
```
rix0rrr, line 198: Does this coincide with us raising our minimum Python requirements?

If not, how are we sure what we vend continues to support Python 3.6?

rix0rrr (commented):
Seems there are a bunch of conflated changes in here that aren't even described or motivated in the PR body, which makes it rather hard to review.

* Ignores
* Version changes
* CHANGELOG removal

Now I'm no saint in the PR purity department, but please at least give me a sense of what to expect when I'm starting to review

.github/workflows/main.yml
RomainMuller: This was actually me trying to understand if the intermittent pip install failures were caused by "old" versions of pip (as provided by venv). But it si not, and this will be reverted (I'd have sworn I already did!).

And so this only changed the default version of Python, but we'd still have tested with 3.6 elsewhere (maybe on a different OS, though).

RomainMuller:
@rix0rrr I've cleaned up this PR by shipping the unrelated content separately... I think now it focuses much more on the actual value proposition for this PR :)

PR #1891, opened 2020-08-14
fix(pacmak): `EMFILE` error when running `jsii-pacmak`

RomainMuller:
The `findLocalBuildDirs` function did not have a protection against
inspecting the same `package.json` file multiple times, and
asynchronously traverses the whole dependency tree under the built
package. In certain pathological cases, dependencies could be processed
many times around (think about how `@aws-cdk/aws-iam` is a direct or
transitive dependency of nearly every other `@aws-cdk/aws-*` module).
The asynchronous nature of the process means that *many* instances of
the same file could be opened at the same time, occasionally inching
above the maximum file descriptor count limit, hence causing `EMFILE` (
which is the standard error code for "too many files open").

This change adds a `visitedDirectories` set to prevent re-visiting the
same dependency instance multiple times (based on the absolute path of
the package root). This incidentally also improves the performance of
the process, since making promises incurs overhead, and not
re-processing directories multiple times cut out a significant chunk of
the promises made in extreme cases.

Special thanks to @richardhboyd for having me look into this particular
problem.

---


nija-at (approved):
Nice!

PR #1903, opened 2020-08-17
fix(jsii): class members named after the class result in illegal C#

### LINKED ISSUE #1880: jsii should reject a member name with the same name as the class
## :bug: Bug Report

This has just blocked our pipelines and wasted a lot of Elad's work.

In https://github.com/aws/aws-cdk/pull/9557 and https://github.com/aws/aws-cdk/pull/9584 we added a property called `construct` to the class `Construct`.

In C#, this property name gets renamed to `Construct`, so that the member is now called `Construct::Construct`. However, that name is not allowed because it is reserved for the constructor (which this is not).

In Java this might be a theoretical issue, but it's not a practical one because class names are always pascal-cased while member names are always camel-cased.

The C# generator thought about this and tries to do something "smart": it renames the *class* to `Construct_`, which makes the property name allowed again. However, we now broke backwards compatibility with existing code by renaming a class, which is definitely *not* the intended behavior.

The only solution is to completely disallow this occurrence in the first place:

* Properties or methods with the same (pascal-cased) name as the class should be disallowed by jsii.

This also means that there is a danger in specifying interfaces -- any of the member names in an interface cannot be used as the class name of the implementing class. Yes, there is a way in C# to still properly implement that interface, but it comes with other undesirable consequences attached:

```
using System;
					

interface IConstructHaver {
	int Construct { get; set; }
}

class Construct : IConstructHaver {
	int IConstructHaver.Construct { get; set; }
}

public class Program
{
	public static void Main()
	{
		var c = new Construct();
                c.Construct = 3;  // NOT ALLOWED

                // This works, but is weird
		(c as IConstructHaver).Construct = 3;
	}
}
```

It's probably just pragmatic to disallow this occurrence, regardless of whether the member comes from an interface. It will simplify our analysis to not have to deal with the special case, and devs can always work around any potential future limitations on their own (which should be rare anyway).

### Affected Languages
- [ ] `Python`
- [ ] `Java`
- [x] .NET (`C#`, `F#`, ...)

RomainMuller:
It is illegal in C# to name a class' member (method, property) with the
same name as the class itself (even if the member is static). Since both
class and member names are pascal-cased in C#, this means a class'
members cannot share the same PascalCased representation.

This adds a compile-time validation for this condition, effectively
calling out member names to have the same PascalCased representation
as their declaring class' name.

This is currently only a warning, as this is "made to work" by altering
the type name by appending `_` at the end of it, which is ugly and
dangerous but works, and is currently done in several places).

As all warnings, this turns into an error when operating under `--strict`,
and future work (i.e: #1931) will allow more granular configuration
of these errors, so we can hopefully opt all new codebases into the
strict behavior and eventually drop the slugification.

Fixes #1880

---


docs/typescript-restrictions.md
```diff
@@ -75,6 +75,58 @@ users in languages with conflicts.
...
+export class Base {
+  public method(): any { /* ... */ }
+}
+
+export class Child {
+  // ⛔️ Illegal change of signature when overriding a member
+  public method(): string { /* ... */ }
+}
+```
+
```
RomainMuller, line 129: Those restrictions already existed, but were not documented...

PR #1906, opened 2020-08-17
fix(java): module "<module name>" not found error

### LINKED ISSUE #1904: 'Module not found' error from JSII kernel in non-typescript languages
### Affected Languages

- [ ] `Python`
- [X] `Java`
- [ ] .NET (`C#`, `F#`, ...)

### General Information
* **JSII Version:** 1.9.0
* **Platform:**     OSX

### What is the problem?
The following code throws the error - `Module '@aws-cdk/aws-lambda' not found` - with a stack trace from the kernel. This occurs exactly at the line where a `Fn.conditionIf()` is called and assigned to a variable.

```java
package com.myorg;
import software.amazon.awscdk.core.App;
import software.amazon.awscdk.core.ICfnConditionExpression;
import software.amazon.awscdk.core.Fn;
import software.amazon.awscdk.core.Stack;
import software.amazon.awscdk.services.lambda.*;
import java.util.List;
import java.util.concurrent.locks.Condition;
import java.util.ArrayList;
public class HelloCdkJavaApp {
    public static void main(final String[] args) {
        final App app = new App();
        final Stack stack = new Stack(app, "mystack");
        final List<String> groups = new ArrayList<String>();
        groups.add("securitygroup1");
        final List<String> subnets = new ArrayList<String>();
        subnets.add("subnet1");
        final CfnFunction.VpcConfigProperty prop = new CfnFunction.VpcConfigProperty.Builder()
            .securityGroupIds(groups)
            .subnetIds(subnets)
            .build();
        final ICfnConditionExpression conditional = Fn.conditionIf("some condition", prop, "AWS::NoValue");
        new CfnFunction(stack, "fn", CfnFunctionProps.builder()
            .code(CfnFunction.CodeProperty.builder().zipFile("zipFile").build())
            .role("role")
            .handler("index.handler")
            .runtime("runtime")
            .vpcConfig(conditional)
            .build());
        app.synth();
    }
}
```

### Verbose Log
https://gist.github.com/nija-at/cf1f83a3aa55894127e1766fd1fe6fba

RomainMuller:
Struct proxies did not trigger loading of their parent module when
passed as a parameter, which then causes the `@jsii/kernel` type checker
to be unable to locate the struct type definition for checking pruposes,
leading to a crash.

This updatates the `JsiiSerializable` serialization process to
proactively load the relevant modules (the call is idempotent and short
circuits if the module is known to already be loaded).

Fixes #1904

---


RomainMuller:
Labeled with `pr/do-not-merge` as I still want to add a similar test in Python & .NET to confirm the bug does not affect them.

packages/@scope/jsii-calc-lib/lib/submodule/index.ts
```diff
@@ -30,4 +30,11 @@ export namespace NestingClass {
   export class NestedClass {
     public readonly property: string = 'property';
   }
+
+  /**
+   * This is a struct, nested within a class. Normal.
+   */
+  export interface NestedStruct {
```
MrArnoldPalmer, line 37: ew... what.

MrArnoldPalmer:
Was this only triggered by nested structs for some reason?

RomainMuller:
@MrArnoldPalmer yeah - the bottom line is that Java does not issue a `create` call to the kernel upon initialization, which is where the `load` for the hosting module happens for other objects (the struct proxies are "pure data" so they don't appear in the JS process before they're actually sent across the boundary).

The nesting *might* not add to the problem (but I had it to reproduce the original issue discovered, and I reckon this makes it a slightly more hostile pattern - meaning I have more chances to catch more issues with that :D).

PR #1931, opened 2020-08-21
feat(jsii): standardized error messages

RomainMuller:
All errors that jsii can emit now have a dedicated jsii error code (and
symbolic name), which will make it easier to track bugs related to those
as well as to provide a better experience for the documentation of the
errors, their cause, and what users should do to fix them.

This lays the foundation for being able to customize the severity of
diagnostics (non-error ones, like warnings and suggestions) individually
in order to be able to have flexibility with breaking changes.

---


packages/jsii/lib/jsii-diagnostic.ts
```diff
@@ -0,0 +1,791 @@
...
+     */
+    public static lookup(codeOrName: string | number): Code | undefined {
+      if (typeof codeOrName === 'number') {
+        return this.byCode[codeOrName];
+      }
+      return this.byName[codeOrName];
+    }
+
+    // eslint-disable-next-line @typescript-eslint/explicit-member-accessibility
+    readonly #defaultCategory: ts.DiagnosticCategory;
```
rix0rrr, line 701: This `#` is throwing me off and I can't quickly find what it means. Do you have a reference?

packages/jsii/lib/jsii-diagnostic.ts
```diff
@@ -0,0 +1,791 @@
...
+    }
+
+    /**
+     * Creates a new `JsiiDiagnostic` message with source code location denoted
+     * by the provided `location` node.
+     *
+     * @param location the source code location attachment of the message.
+     * @param args     the arguments to the message formatter.
+     */
+    public create(
```
rix0rrr, line 784: Why not 2 orthogonal methods?

```ts
Code.create(...).attach(...);
```

packages/jsii/lib/assembler.ts
```diff
@@ -335,46 +329,29 @@ export class Assembler implements Emitter {
...
       if (assembly) {
         if (!(assembly.name in this.projectInfo.peerDependencies)) {
-          this._diagnostic(
-            referencingNode,
-            ts.DiagnosticCategory.Warning,
-            `The type '${ref}' is exposed in the public API of this module. ` +
-              `Therefore, the module '${assembly.name}' must also be defined under "peerDependencies". ` +
-              'This will be auto-corrected unless --no-fix-peer-dependencies was specified.',
+          this._diagnostics.push(
+            JsiiDiagnostic.Code.JSII_0005_MISSING_PEER_DEPENDENCY.create(
```
rix0rrr, line 333: `JsiiDiagnostic.Code` feels rather verbose. Why not one top-level class? Ultimately we only care about the predefined `JSII_0005_MISSING_PEER_DEPENDENCY` (et al) constants, no? Why not make them easier to access?

packages/jsii/lib/assembler.ts
```diff
@@ -416,14 +393,18 @@ export class Assembler implements Emitter {
...
-          : `Type "${type.symbol.name}" ${commonMessage}`,
-        makeCause(typeDeclaration),
+      this._diagnostics.push(
+        JsiiDiagnostic.Code.JSII_3001_EXPOSED_INTERNAL_TYPE.create(
+          typeAnnotationNode,
+          type.symbol,
+          isThisType,
+          typeUse,
+        ).addRelatedInformation(
+          JsiiDiagnostic.Code.JSII_9999_RELATED_INFO.create(
```
rix0rrr, line 403: `JSII_9999_RELATED_INFO `: this feels like cheating. Are "RelatedInformation"s *really* a type of Diagnostic?

Or do they just happen to share some features like formatting and a location and was this a convenient way to get that? Some remodeling might be in order.

MrArnoldPalmer:
@RomainMuller What is the convention for the error code numbers? Should we add an explanation for those to the docs?

RomainMuller:
There isn't a general convention at this point. I've laid out some basic categories and allocated sequentially in those...

I'd like to document that the error codes are to be handled as experimental at this point and that we may re-allocate them at a later point once we have a better idea of whether there is something meaningful we should use?

packages/jsii/lib/jsii-diagnostic.ts
RomainMuller: https://github.com/tc39/proposal-class-fields/blob/master/PRIVATE_SYNTAX_FAQ.md

packages/jsii/lib/assembler.ts
RomainMuller: You're probably right. I've been self-debating this myself, too.

packages/jsii/lib/jsii-diagnostic.ts
RomainMuller: I actually want to eventually __require__ all diagnostic messages have a source location attached, so I mean for the node to be provided upfront, always... This is why the other method is deprecated.

packages/jsii/lib/assembler.ts
RomainMuller: The `ts.DiagnosticRelatedInformation` interface is actually the base interface of `ts.Diagnostic` (the later only adds properties we don't care about using - `reportsUnnecessary`, `reportsDeprecated`, `source` - as well as the `relatedInformation` property that is an optional array of `ts.DiagnosticRelatedInformation`).

The consequence of this is that a `code` is needed on the related information in the exact same way that for top-level diagnostics... And those codes are actually visible to programmatic users.

PR #1940, opened 2020-08-24
feat: upgrade to typescript ~4.0.2

RomainMuller:
This upgrades the version of the TypeScript compiler to the `4.0` tree.
More information about the content of the upgrade can be found in the
[Announcing TypeScript 4.0] post. In particular, this release includes a
few [breaking changes] in the TypeScript language from the previous
release, including (but not limited to) the following:

- Properties Overriding Accessors (and vice versa) is an Error
- Operands for `delete` must be optional

Most codebases will be unaffected by those changes, or the path to fix
a codebase across those is well documented and should not require
significant effort to follow.

[announcing typescript 4.0]: https://devblogs.microsoft.com/typescript/announcing-typescript-4-0
[breaking changes]: https://devblogs.microsoft.com/typescript/announcing-typescript-4-0/#breaking-changes

BREAKING CHANGE: This release updates the TypeScript compiler from 3.9.x to 4.0.x. Refer to Microsoft's [Announcing TypeScript 4.0] post for more information.

---


nija-at (commented):
I vaguely there being customer complaints or breakage the last time we did a typescript upgrade as part of a minor version bump. Am I mis-remembering?

RomainMuller:
This was due to the breaking changes on the TypeScript language specification, particularly around places where the type checker had tightened behavior (used to let things pass it shouldn't have).

We decided to consider those as "breaking bug fixes" where the cost of the breakage is worth taking (as it results in safer code overall).

MrArnoldPalmer (approved):
This seems relatively safe with the integ test passing. Approved with "do-not-merge" just in case there are more dissenters.

eladb (changes requested):
Isn't this a breaking change to jsii?

skorfmann:
Just stumbled upon this when trying to build a jsii project which was using a bundled dependency with Typescript types of version `> 4`

Do you have plans to move forward with Typescript 4?

webratz:
Also was hit by jsii still using typescript 3. I wanted to use `ajv` as a bundled dependency which uses typescript 4. This lead to a lot of very confusing errors with no clear indication of what is wrong.
In this particular case i was able to downgrade to an old (v6) version of the library, but in general typescript 4 support would be super useful for jsii.

skorfmann:
This has been impacting us in the cdktf project recently as well. We had to build a significant [workaround](https://github.com/hashicorp/terraform-cdk/pull/1248) for a Typescript [bug](https://github.com/microsoft/TypeScript/issues/38691) which apparently won't be fixed in 3.9.

We would be up for helping to make this happen as well.

eladb:
@RomainMuller looks stale. Should we close for now?

PR #1941, opened 2020-08-25
feat: add support for bin-scripts (python only)

kozlove-aws:
### Problem
JSII is missing functionality for packing and using bin scripts from Python package.

### Solution
Were extended assembly and project-info properties for bin-scripts.
Added generation scripts for each script in bin section.
Added section `scripts` to `pyhon.py` script.

### Testing

Was build project with changed scripts and checked that file `.jsii` contains `bin` section.
Also was checked that without `bin` section in `package.json` empty `bin` section was not created in `.jsii` file.
Was created package and verified that scripts were created and paths to it were added to `setup.py`.

---


packages/@jsii/spec/lib/assembly.ts
```diff
@@ -143,6 +143,13 @@ export interface Assembly extends AssemblyConfiguration, Documentable {
    * @default none
    */
   readme?: { markdown: string };
+
+  /**
+   * List of bin-scripts
+   *
+   * @default none
+   */
+  bin?: { [script: string]: string };
```
RomainMuller, line 152: Let's tighten this one up from the start (cannot fix the older fields to `readonly` because... backwards-compatibility... but we can be nice with the new ones!)

```suggestion
  bin?: { readonly [script: string]: string };
```

packages/jsii-pacmak/lib/targets/python.ts
```diff
@@ -1429,6 +1429,13 @@ class PythonModule implements PythonType {
       code.line(`from ${'.'.repeat(distanceFromRoot + 1)}_jsii import *`);
 
       this.emitRequiredImports(code, context);
+      if (this.assembly.bin) {
```
RomainMuller, line 1432: I like being explicit here (JS' truthiness/falsiness is a little awkward)

```suggestion
      if (this.assembly.bin != null) {
```

packages/jsii-pacmak/lib/targets/python.ts
```diff
@@ -1429,6 +1429,13 @@ class PythonModule implements PythonType {
...
+        code.line(`print('BIN-scripts were found.');`);
+        const scripts = Object.keys(this.assembly.bin);
+        for (const script of scripts) {
+          code.line(`print('${script}: ${this.assembly.bin[script]}');`);
+        }
```
RomainMuller, line 1437:
```suggestion
        const scripts = Object.keys(this.assembly.bin);
        for (const [name, path] of Object.entries(this.assembly.bin)) {
          code.line(`print('${name}: ${path}');`);
        }
```

packages/@jsii/kernel/lib/api.ts
```diff
@@ -103,13 +103,13 @@ export interface LoadResponse {
 export interface InvokeScriptRequest {
   readonly pkgname: string;
   readonly script: string;
-  readonly args?: any[];
+  readonly args?: string[];
```
RomainMuller, line 106:
```suggestion
  readonly args?: readonly string[];
```

packages/@jsii/kernel/lib/kernel.ts
```diff
@@ -140,8 +140,11 @@ export class Kernel {
     req: api.InvokeScriptRequest,
   ): api.InvokeScriptResponse {
     const result = cp.spawnSync(
-      path.join(this._getPackageDir(req.pkgname), req.script),
-      req.args,
+      'node',
+      [path.join(this._getPackageDir(req.pkgname), req.script)].concat(
+        req.args ?? [],
+      ),
```
RomainMuller, line 146: It's probably going to be more efficient to use the same `node` runtime that is used by the Kernel process (and it's also going to be safer in Windows):
- `process.execPath` contains the path to the `node` binary that runs the current process.
- `process.execArgv` contains the `node` arguments that the current process received.

The `--` indicates the end of `node` options and guarantees that none of your own command options are possibly interpreted by node itself.

```suggestion
      process.execPath,
      [
        ...process.execArgv,
        '--',
        path.join(this._getPackageDir(req.pkgname), req.script),
        ...req.args ?? [],
      ],
```

packages/@jsii/kernel/test/kernel.test.ts
```diff
@@ -2133,6 +2133,17 @@ defineTest('Override transitive property', (sandbox) => {
   expect(propValue).toBe('N3W');
 });
 
+defineTest('invokeBinScript() return output', (sandbox) => {
+  sandbox.create({ fqn: 'jsii-calc.Calculator' });
+
```
RomainMuller, line 2138: I don't think this is necessary

```suggestion
```

packages/@jsii/kernel/lib/api.ts
```diff
@@ -100,6 +100,19 @@ export interface LoadResponse {
   readonly types: number;
 }
 
+export interface InvokeScriptRequest {
+  readonly pkgname: string;
```
RomainMuller, line 104: I would rather try to make sure we are consistent with how we refer to those concepts... in this case, we refer to *assemblies* (since the value that ought to be passed here is the same that was passed to `LoadRequest.assembly`).

```suggestion
  readonly assembly: string;
```

packages/@jsii/kernel/lib/kernel.ts
```diff
@@ -148,6 +136,25 @@ export class Kernel {
     };
   }
 
+  public invokeBinScript(
+    req: api.InvokeScriptRequest,
+  ): api.InvokeScriptResponse {
+    const result = cp.spawnSync(
+      'node',
+      [path.join(this._getPackageDir(req.pkgname), req.script)].concat(
```
RomainMuller, line 144: I think the `script` value provided should not be a path, instead it should be a key from the `bin` map that was gathered from `package.json` during compilation.

You need to validate the script name was in the map, and fail if not (you don't want to allow a request to run *anything from anywhere*, that's kind of a bad security posture to have. Then, you need to use the *value* that corresponds to the `bin` key that was requested in order to compose the final executable path.

*Maybe* we shouldn't actually use `node` shell out here... Your NPM package's `bin` script might not actually be a node script (could be shell script, could be something else). On the other hand it might be tricky to make this work smooth on Windows if we skip the shell out (I guess it's a good first step?)

packages/@jsii/kernel/lib/api.ts
```diff
@@ -100,6 +100,19 @@ export interface LoadResponse {
...
+  readonly script: string;
+  readonly args?: string[];
+}
+
+export interface InvokeScriptResponse {
+  readonly status: number | null;
+  readonly stdout: Buffer;
+  readonly stderr: Buffer;
+  readonly output: string[];
+}
```
RomainMuller, line 114: What is `output` for? Seems to be the same as `stdout`?

Also, I would make `status` be able to represent either a code or a signal, since these are the two ways a process can finish.

```suggestion
export interface InvokeScriptResponse {
  readonly status:
    | { code: cp.SpawnSyncReturns<unknown>['status'] }
    | { signal: cp.SpawnSyncReturns<unknown>['signal'] };
  readonly stdout: Buffer;
  readonly stderr: Buffer;
}
```

packages/@jsii/kernel/lib/kernel.ts
```diff
@@ -148,6 +136,25 @@ export class Kernel {
...
+        req.args ?? [],
+      ),
+      { encoding: 'utf-8' },
+    );
+
+    return {
+      output: result.output,
+      stdout: result.stdout,
+      stderr: result.stderr,
+      status: result.status,
```
RomainMuller, line 154: This might be `null` if the app was killed by a signal... Might want to also pass `signal` though.
The contract for `spawnSync` guarantees that exactly one of `result.status` and `result.signal` will be non-`null`.

Following my suggestion for the type change above... This would be something like:
```suggestion
      status: result.status != null
        ? { code: result.status }
        : { signal: result.signal! }, // <- The `!` is to signal we "know" `result.signal` cannot be `null` here
```

packages/jsii/lib/assembler.ts
```diff
@@ -201,6 +201,7 @@ export class Assembler implements Emitter {
       readme,
       jsiiVersion,
       fingerprint: '<TBD>',
+      bin: this.projectInfo.bin,
```
RomainMuller, line 204: Let's try to keep `fingerprint` last here, at least for now.

```suggestion
      bin: this.projectInfo.bin,
      fingerprint: '<TBD>',
```

packages/jsii-pacmak/lib/targets/python.ts
```diff
@@ -1642,6 +1694,12 @@ class Package {
       code.openFile(filename);
       mod.emit(code, context);
       code.closeFile(filename);
+
+      for (const script of mod.emitBinScripts(code)) {
+        if (scripts.indexOf(script) < 0) {
+          scripts.push(script);
+        }
```
RomainMuller, line 1701: I think the duplication defense here is not necessary (it also does not hurt!)

packages/jsii-pacmak/lib/targets/python.ts
```diff
@@ -1631,6 +1681,8 @@ class Package {
       a.pythonName.localeCompare(b.pythonName),
     );
 
+    const scripts: string[] = [];
```
RomainMuller, line 1684: We tend to prefer this form for some reason:

```suggestion
    const scripts = new Array<string>();
```

packages/jsii-pacmak/lib/targets/python.ts
```diff
@@ -1478,6 +1478,56 @@ class PythonModule implements PythonType {
...
+          code.line();
+          emitList(
+            code,
+            '__jsii_assembly__ = jsii.JSIIAssembly.load(',
+            [
+              JSON.stringify(this.assembly.name),
+              JSON.stringify(this.assembly.version),
+              JSON.stringify(this.pythonName.replace('._jsii', '')),
+              `${JSON.stringify(this.assemblyFilename)}`,
+            ],
```
RomainMuller, line 1509: You will probably run into issues doing that, because this flow never loads the current package's dependencies into the kernel...

Instead of re-doing this, you should import the module that actually initializes things correctly:
```py
import ._jsii # <- And voilà, dependencies and the current module are loaded!
```

RomainMuller (changes requested):

packages/jsii-pacmak/lib/targets/python.ts
kozlove-aws: It does not hurt but looks strange when the same script appeared several times.

packages/jsii-pacmak/lib/targets/python.ts
kozlove-aws: I just to try keep style of this file. Such structure appear 5 times in this script.

packages/jsii-pacmak/lib/targets/python.ts
RomainMuller: But... it shouldn't... I mean a script can only be registered once... so if we have duplicates, it could mean something wrong happened somewhere else?

packages/@jsii/kernel/lib/kernel.ts
kozlove-aws: I do not think that it is a good idea to add as result one of property. It could be really inconvenient to parse such property.
What do you think if we add signal as additional property?

packages/jsii-pacmak/lib/targets/python.ts
kozlove-aws: Do not really understand what do you mean here?

packages/jsii-pacmak/lib/targets/python.ts
kozlove-aws: You are right. I have made some tests and have not found any dublicates.

RomainMuller:
So sorry - this was closed in error as the `master` branch was renamed to `main`. Apologies!

packages/@jsii/kernel/lib/kernel.ts
```diff
@@ -139,17 +139,40 @@ export class Kernel {
...
+      const result = cp.spawnSync(
+        process.execPath,
+        [
+          ...process.execArgv,
+          '--',
+          path.join(packageDir, scriptPath),
+          ...(req.args ?? []),
+        ],
+        { encoding: 'utf-8' },
+      );
```
RomainMuller, line 164: The script *might* not be a node script (could be bash, etc...), so this feels a little unsafe.

```suggestion
        const result = cp.spawnSync(
          path.join(packageDir, scriptPath),
          req.args ?? [],
          {
            encoding: 'utf-8',
            env: {
              ...process.env,
              // Make sure the current NODE_OPTIONS are honored if we shell out to node
              NODE_OPTIONS: process.execArgv.join(' '),
              // Make sure "this" node is ahead of $PATH just in case
              PATH: `${path.dirname(process.execPath)}:${process.env.PATH}`,
            },
            shell: true,
          },
        );
```

packages/@jsii/kernel/test/kernel.test.ts
```diff
@@ -2133,6 +2133,15 @@ defineTest('Override transitive property', (sandbox) => {
   expect(propValue).toBe('N3W');
 });
 
+defineTest('invokeBinScript() return output', (sandbox) => {
+  const result = sandbox.invokeBinScript({
+    assembly: 'jsii-calc',
+    script: 'calc',
+  });
+
+  expect(result.stdout).toEqual('Hello World!\n');
```
RomainMuller, line 2142: You probably should also asset the value of `stderr`, `status` and `signal`.

packages/@jsii/python-runtime/tests/test_compliance.py
```diff
@@ -85,6 +85,7 @@
 from scope.jsii_calc_lib import IFriendly, EnumFromScopedModule, Number
 from scope.jsii_calc_lib.custom_submodule_name import IReflectable, ReflectableEntry
 
+from subprocess import Popen, PIPE, STDOUT
```
RomainMuller, line 88: That doesn't appear to be necessary.

packages/jsii-pacmak/lib/targets/python.ts
```diff
@@ -1695,11 +1695,7 @@ class Package {
       mod.emit(code, context);
       code.closeFile(filename);
 
-      for (const script of mod.emitBinScripts(code)) {
-        if (scripts.indexOf(script) < 0) {
-          scripts.push(script);
-        }
-      }
+      Array.prototype.push.apply(scripts, mod.emitBinScripts(code));
```
RomainMuller, line 1698: Why not `scripts.push(...mod.emitBinScripts(code))`?

packages/jsii-calc/test/assembly.jsii
```diff
@@ -7,6 +7,9 @@
     ],
     "url": "https://aws.amazon.com"
   },
+  "bin": {
+    "calc": "bin/calc"
```
RomainMuller, line 11: Might be interesting to rename `bin/calc` to something else to remove the "coincidence" that it matches the script's name (the key here).

For example, you could call it `bin/run` instead.

There's also a chance this does not test fine on Windows (because Windows is hard 🤓). In such cases you'd need to drop in a CMD entry point (I'll give you more info on this once we know this is necessary, hopefully it won't be).

kozlove-aws: Good idea.

packages/jsii-pacmak/lib/targets/python.ts
kozlove-aws: I just not very familiar with typescript so had no idea about '...' operator.

RomainMuller (approved):

RomainMuller:
@mergifyio update

PR #1951, opened 2020-08-26
feat(jsii): reflect stabilities in doc comment summaries

rix0rrr:
In order to make the stability of certain API elements more clear, note
any non-standard API stabilities in the APIs summary docstring.

This makes it clearly visible in the IDEs autocomplete.

IMPLEMENTATION NOTES

We put the onus of rendering of the `(experimental)/(deprecated)` prefixes on the code generators for
 every language. We played with the idea of sticking it directly into the manifest but then the API reference
 would also have had it, and the API reference already has color coding and graphical symbols to indicate
 these statuses.

To modify the `.js/.d.ts` generation:

* Reversed the order of `program.emit()` and `assembler.emit()`. `assembler.emit()` now comes first,
  so that we know the correct stability before we start emitting code.
* For every node we encounter, we regenerate the doc comment block to include stability in the summary 
   line if appropriate. We remember the desired doc comment block for later.
* The annotation is being done by TypeScript transforms. The original comments are stripped out of the
  SourceFile (in-memory) and replaced with synthetic comments containing the generated doc comments.

---


RomainMuller:
1. I thought out goal was to have "experimental" be the FIRST THING in the documentation? Not the last one?
2. Does it also render in the TypeScript documentation? (In the `.d.ts` files)

packages/@scope/jsii-calc-lib/test/assembly.jsii
```diff
@@ -179,7 +182,7 @@
       "docs": {
         "remarks": "These classes can be greeted with\na \"hello\" or \"goodbye\" blessing and they will respond back in a fun and friendly manner.",
         "stability": "deprecated",
-        "summary": "Applies to classes that are considered friendly."
+        "summary": "Applies to classes that are considered friendly. (deprecated)"
```
eladb, line 185: I think deprecation should be a prefix, not a postfix:

    (DEPRECATED) Applies to classes to are considered friendly

Also, where is the deprecation message going?

packages/jsii/lib/assembler.ts
```diff
@@ -2340,6 +2374,39 @@ export class Assembler implements Emitter {
...
   }
+
+  /**
+   * From the given JSIIDocs, re-render the TSDoc comment for the Node
+   *
+   * For most normal cases, this yields the same output back as the one that
+   * we originally saw (modulo whitespace), but if the JSIIDocs got changed upon
+   * parsing we'll render the JSIIDocs back into the .js/.d.ts output.
+   */
+  private regenerateDocString(
```
eladb, line 2386: do we have any tests for this?

packages/jsii/lib/docs.ts
```diff
@@ -220,7 +285,7 @@ const SUMMARY_MAX_WORDS = 20;
  *
  * In principle we'll take the first paragraph, but if there are no paragraphs
  * (because people don't put in paragraph breaks) or the first paragraph is too
- * lang, we'll take the first sentence (terminated by a punctuation).
+ * long, we'll take the first sentence (terminated by a punctuation).
```
RomainMuller, line 288: 👌🏻

packages/jsii/lib/ts-comment-replacer.ts
```diff
@@ -0,0 +1,132 @@
...
+   */
+  private addTsdocComment(node: ts.Node, text: string) {
+    const lines = text.trim().split('\n');
+
+    // eslint-disable-next-line prettier/prettier
+    const commentContents = [
+      '*\n',
+      ...lines.map((l) => ` * ${l}\n`),
+      ` `,
+    ].join('');
```
RomainMuller, line 102: Why not

```suggestion
    const commentContents = [
      '*',
      ...lines.map((l) => ` * ${l}`),
      ' ',
    ].join('\n');
```

packages/jsii/test/docs.test.ts
```diff
@@ -1,6 +1,8 @@
+/* eslint-disable prettier/prettier */
```
RomainMuller, line 1: No. Please don't do that. 🙅🏻‍♂️

packages/jsii/lib/compiler.ts
```diff
@@ -215,43 +215,63 @@ export class Compiler implements Emitter {
...
+      hasErrors = true;
+      LOG.error(
+        'Compilation errors prevented the JSII assembly from being created',
+      );
+    }
+
     return {
       emitSkipped: hasErrors,
-      diagnostics,
+      diagnostics: ts.sortAndDeduplicateDiagnostics(diagnostics),
```
RomainMuller, line 274: 😲

RomainMuller (approved):

RomainMuller:
BTW - I ❤️ how the PR description represents the decisions made and why. That is inspiring 😁

PR #1985, opened 2020-09-08
fix: pip dependency resolution failure in CI

MrArnoldPalmer:
Attempts to fix intermittent pip module dependency resolution failure by
using the new `2020-resolver` feature flag.

NOTE: this may or may not actually work. Testing with PR verification
workflows.

---

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -194,19 +194,19 @@ jobs:
             dotnet: '3.1.x'
             java: '8'
             node: '10'
-            python: '3.6'
+            python: '3.7'
           # Test using macOS
           - os: macos-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -194,19 +194,19 @@ jobs:
             dotnet: '3.1.x'
             java: '8'
             node: '10'
-            python: '3.6'
+            python: '3.7'
```
RomainMuller, line 197: I'd rather we keep on testing with the **minimum supported** version of python outside of the python-specific sub-matrix.

MrArnoldPalmer:
Closing in favor of future fixes

PR #2013, opened 2020-09-15
feat(pacmak): retry select external command invocations

RomainMuller:
When generating packages, `jsii-pacmak` will defer to external tools for
compilation (`dotnet`, `mvn`, `pip`, ...). Those commands will typically
require some kind of network transfer (e.g: to resolve dependencies from
their standard package registry), which is susceptible to fail (due to
poor network conditions, getting throttled in CI/CD, etc...).

In order to reduce the likelihood that such failures will interrupt the
user workflow abruptly (forcing a manual retry, and possibly breaking
CI/CD workflows), back-off and retries will be performed when these
commands fail (up to 5 tentatives in total, with exponential back-off
using a 150ms base cooldown and a factor of 2).

---


packages/jsii-pacmak/lib/util.ts
```diff
@@ -30,12 +21,132 @@ export function resolveDependencyDirectory(
   );
 }
 
+export interface RetryOptions {
+  /**
+   * The maximum amount of attempts to make.
+   *
+   * @default 5
+   */
+  maxTries?: number;
```
rix0rrr, line 30: `maxAttempts` then? This word grates on me for some reason.

rix0rrr (changes requested):
Why did the retry mechanism need to be rewritten instead of simply turned on?

RomainMuller:
For some reason I only noticed there was already support for retries **after** writing the new retry wrapper. And the old retry mechanism lacked support for back-off, making it questionnable for working around throttles / DDOS protections.

rix0rrr (approved):

PR #2045, opened 2020-09-25
chore: migrate from yargs to OCLIF

RomainMuller:
> ℹ️ This is part PR, part RFC in the sense that I have researched [OCLIF]
> a little bit at this point and have liked what I saw; and this would be a
> first step towards migrating all `jsii` command lines into this framework;
> but I'm also looking to hear what y'all think about it and whether you're
> into this or not at all.
>
> [OCLIF]: https://oclif.io

Switching away from `yargs` to benefit some of the advantages of `oclif`:
- Ability to model comamnds in separate packages
- Async-friendly, simpler API
- API to easly run CLI commands programmatically

This will eventually allow for one central `jsii` command, with several
subcommands (`jsii [compile]`, `jsii package`, `jsii reflect`, ...),
providing a single entry point, which is better for discoverability of
available commands (for example, via `jsii --help`).

Some interesting tibits that may seem unusual from this PR:
- `oclif` requires use of the `files` pragma in `package.json`, which
  effectively replaces the `.npmignore` file (it's an allow-list of what
  can ship in the tarball)
- a single `bin/run` endpoint is used by `oclif`, and the bin mapping
  makes it be symlinked as `jsii` in the `node_modules/.bin` directory
  on install. In development, the command can be run "intuitively" (once
  `yarn install` was done) using `yarn jsii` or `./bin/run`
- an `oclif.manifest.json` file is created at build time (and shipped in
  the tarball), and helps `oclif` initialize faster (ideal for released
  packages)

---


MrArnoldPalmer:
@RomainMuller is the plan to support all of the current CLI commands for the stuff that is separate (pacmak/diff/jsii) in addition to adding all of these as subcommands to `jsii` so as to not break existing users? Or is this part of JSII 2.0?

RomainMuller:
> @RomainMuller is the plan to support all of the current CLI commands for the stuff that is separate (pacmak/diff/jsii) in addition to adding all of these as subcommands to `jsii` so as to not break existing users? Or is this part of JSII 2.0?

My idea was to migrate all existing CLIs (`jsii-pacmak`, `jsii-config`, `jsii-diff`, `jsii-reflect`, ...) to OCLIF in a non-breaking way, so `jsii pacmak` becomes an alternative to `jsii-pacmak`, etc...). Then later (as part of jsii 2.0), considering making the new form the only possible form.

Issue #2039, opened 2020-09-23
Go - Define Runtime Object Creation Lifecycle


MrArnoldPalmer:
Now that go code generation is generating all types with correct type signatures, the generated constructor methods can call to the runtime library to instantiate objects within the JSII node runtime.

The first step in this process is defining runtime functions for object creation that are called within code generated "class" constructors. This will create a `CreateRequest` struct, serialize it, and pass it to the runtime client. It also will handle the response and deserializing any data within the returned JSON `CreateResponse` struct to the expected type of the generated function.

PR #2046, opened 2020-09-26
feat(go): runtime load and object creation

MrArnoldPalmer:
Defines runtime functions for loading jsii libraries, instantiating jsii
objects, calling object methods, and accessing object properties.
Changes go code generation to call these new functions where
appropriate.

Adds the `JsiiObj` interface type in `types.go` for implementation by
all generated data structures. This allows these data structures to be
tied to their corresponding instances within the kernel runtime which
allows for passing values by reference over the wire where necessary,
such as when receiving a `CreateResponse` or when sending
`InvokeMethod`.

Make client data structure private and only expose the necessary
functions in `runtime.go` for usage in generated code.

Adds new classes in the jsii-pacmak go target for making calls to the
new runtime methods. Also generates a `Load` function on each package
that can be called to bootstrap the library within the runtime process
and make sure all dependencies are loaded. This functionality is wrapped
in a `sync.Once.Do` function to ensure that its only ever called once
per library. All calls to runtime functions are preceded with a call to
load.

Changes embedding of both the runtime and generated modules `.js` files.
This was necessary because you cannot reference files relative to go
module source files reliably. This made our previous strategy of just
putting the js packages `.tgz` file within the source tree output not
work when attempting to load the library into the kernel process.
Similarly creates a separate `assets` package in `go-runtime` that is
gitignored and bundles the `@jsii/runtime` js file for reference in
`client.go`.

Because of the embedding of assets within go source files, excludes
files with names known to be embedded assets from snapshot pacmak output
snapshot testing. This prevents test failures from different node
versions, dependencies, etc.

Fixes incorrect tagging of properties in api definitions that caused
json encoding/decoding to break. Tags were missing wrapping quotes.

Removes usage of `jsii.Any` in api.go definitions in favor of
`interface{}`. This is a more idiomatic go type that covers the same
cases and makes for easier encoding/decoding of these data structures.

Removes the runtime noop module because its no longer needed.

fixes: #2039

---


packages/@jsii/go-runtime/build-tools/gen.ts
```diff
@@ -1,14 +1,35 @@
...
-  );
+const RT_DIR = resolve(__dirname, '..', 'assets');
+const RT_PATH = resolve(RT_DIR, 'assets.go');
+
+fs.removeSync(RT_PATH);
+
+fs.open(RT_PATH, 'w', async (err, fd) => {
+  if (err) {
+    console.error(err);
+    return;
```
RomainMuller, line 18: Should this has thrown an error instead? In case of error, we'd appear to have succeeded (short of parsing STDERR).

packages/@jsii/go-runtime/jsii-experimental/types.go
```diff
@@ -0,0 +1,13 @@
...
+
+type Any interface{}
+
+// JsiiObj is a trait implemented by every type that can have a corresponding
+// instance within the kernel runtime. getInstanceId is needed for all generated
+// types that may be passed by reference over the wire. For classes,
+// getInstanceId returns a private property that is a jsii object instance ID so
+// that calls to the runtime referencing this object can correctly reference the
+// corresponding instance within the kernel.
+type JsiiObj interface {
```
RomainMuller, line 11: It would be best if we can avoid this; although I guess Go's composition model means this name will not conflict with an actual user name?

Ideally, I'd rather follow a model where all the "magic" and bookkeeping happens *strictly outside* of the managed objects. The model where we attach info on the object itself is actually somewhat problematic in Java & Go IMHO.

RomainMuller (commented):
Would also be nice if we have a new `.md` document describing the runtime components, how they interact with each other, etc... This could be a good place to also investigate the whole "out-of-instance-bookkeeping" I was talking about earlier.

packages/@jsii/go-runtime/jsii-calc-test/main.go
```diff
@@ -1,21 +1,17 @@
...
-	jsiicalc "github.com/aws-cdk/jsii/jsii-calc/golang/jsiicalc"
-	"github.com/aws-cdk/jsii/jsii-experimental"
+	calc "github.com/aws-cdk/jsii/jsii-calc/golang/jsiicalc"
+	baseofbase "github.com/aws-cdk/jsii/jsii-calc/golang/scopejsiicalcbaseofbase"
 )
 
 func main() {
 	fmt.Println("Hello, JSII")
-	// client, err := jsii.InitClient()
+	fmt.Printf("JSII_AGENT=%s\n", calc.NewJsiiAgent().GetValue())
```
RomainMuller, line 11: Note: `GetValue` is supposed to be `static`, which I suspect is why this does not currently work.

RomainMuller:
I started slicing this PR down into smaller chunks, so they can be reviewed easier:
- [ ] #2065 - changes indentation to be tab-based (consistent with `gofmt`)
- [ ] #2066 - cleanly embeds the `@jsii/runtime` application within the go runtime library

We can expect those will cause merge conflicts under this PR (I've implemented or named certain things a little differently). Those should be easy enough to deal with (and I should be able to take care of this).

PR #2055, opened 2020-09-28
chore: stop using the delete operator

RomainMuller:
It can adversely impact the performance of software through causing extraneous
hidden class transitions on the object. Those can be saved by never deleting
properties, instead assigning a blank/empty value.

For some cases, replaced object stores with Map instances, as the constant
mutations of the object store cause hidden class pollution, which Map does
avoid.

---


rix0rrr:
Got any concrete numbers on a PR that seems to be designed to increase performance?

eladb (commented):
Curious about this... Also interested in what performance gain we get from it. Do we need to apply this to AWS CDK as well? 

Also, is there some eslint rule that will ensure we don't introduce `delete`s in the future?

PR #2058, opened 2020-09-29
fix(dontet): iso-8601 date strings get turned into DateTime

RomainMuller:
When a ISO-8601-encoded date is passed from JavaScript to the .NET app,
the JSON deserializer would interpret that as a `DateTime` instead of
passing the string un-touched. This is a problem, since the jsii kernel
protocol has a dedicated wrapper key for DateTime values (`$jsii$date`).

This PR adds a compliance test around this particular scenario, and
disabled `DateTime` handling from the standard deserializer (the wrapped
values are still processed correctly).

> This PR also changes the way the .NET test package generates the
> tested libraries (`jsii-calc` & dependencies) so that instead of
> generating NuGet packages, it only generates C# projects. This makes
> the experience of debugging the tests **much** nicer, and reduces the
> likelihood of cached build artifacts getting in the way.

Fixes aws/aws-cdk#10513

---

### DIFF
--- a/packages/@jsii/dotnet-runtime-test/generate.sh
+++ b/packages/@jsii/dotnet-runtime-test/generate.sh
@@ -1,9 +1,20 @@
 #!/bin/bash
 set -euo pipefail
 test="./test"
+genRoot="${test}/generated"
 
-# Generate NuGet packages for jsii-calc and its dependencies.
-jsii-pacmak -t dotnet --recurse -o ${test}/generated ../../jsii-calc
+# Clean up before we start working
+rm -rf genRoot
+
+# Generate .NET projects for jsii-calc and its dependencies.
+jsii-pacmak -t dotnet --code-only --recurse -o ${genRoot} ../../jsii-calc
+
+# Hack around project references to de-duplicate Amazon.JSII.Runtime in generated code.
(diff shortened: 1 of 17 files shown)


rix0rrr (approved):

packages/@jsii/dotnet-runtime-test/generate.sh
```diff
@@ -1,9 +1,20 @@
 #!/bin/bash
 set -euo pipefail
 test="./test"
+genRoot="${test}/generated"
 
-# Generate NuGet packages for jsii-calc and its dependencies.
-jsii-pacmak -t dotnet --recurse -o ${test}/generated ../../jsii-calc
+# Clean up before we start working
+rm -rf genRoot
```
eladb, line 7:
```suggestion
rm -rf ${genRoot}
```

eladb (changes requested):

packages/@jsii/dotnet-runtime-test/generate.sh
RomainMuller: Good catch! (I had made this mistake in a bunch of other places, but apparently this one survived the fixing :)

PR #2059, opened 2020-09-29
chore: publish Docker image from GitHub actions instead of Travis

RomainMuller:
This gets rid of the last use-case we had for TravisCI, which can hence
be disabled. The general experience ought to be somewhat better as there
is no only a single master for all meaningful task executions, instead
of being scattered on two distinct platforms.

Since TravisCI was also the only place where "working directory clean"
was being asserted, this also adds that check to the "main" workflow.

---


.mergify/config.yml
```diff
@@ -29,7 +29,6 @@ pull_request_rules:
       - "#review-requested=0"
       - "#changes-requested-reviews-by=0"
       - status-success~=AWS CodeBuild us-east-1
-      - status-success=continuous-integration/travis-ci/pr
```
iliapolo, line 32: Shouldn't this be replaced with a `status-success` for the GitHub action that builds the image on PR's?

RomainMuller: Mergify does not allow me to specify conditional checks... I don't want to re-build the image in ALL pr validations (only those that touch the Superchain file)

iliapolo: Is it really that bad building it for all PR's? I tend to usually accept the penalty in favor of better validation. 

You could also do the path detection from inside the github action rather than in the workflow file, and just no-op it when needed.

iliapolo (approved):
Added `do-not-merg` in case you do want to consider my comment. Im fine either way.

.mergify/config.yml
RomainMuller: > Is it really that bad building it for all PR's? I tend to usually accept the penalty in favor of better validation.

The "penalty" is getting throttled by Docker and having PR validations fail...

But I like where you're going with doing in-action detection... I'll give that a shot.

iliapolo (approved):

PR #2061, opened 2020-09-29
fix(dotnet): abundant nullability warnings in generated code

RomainMuller:
The generated code had a number of quirks related to nullability
annotations (`?` and `!`) that resulted in sometimes abundant warnings
emitted during the compilation of the generated code. This makes the
necessary changes to correctly reflect that method arguments can be
`null` in the `DeptyBase` methods; and adds the necessary `?` and `!`
where they were previously missing.

This results in much fewer compilation warnings being emitted while
compiling `jsii-calc` and dependencies (the remaining warnings are
actually "legitimate" in the sense that they are the result of
us/users making questionable naming choices).

---

### DIFF
--- a/packages/@jsii/dotnet-runtime/src/Amazon.JSII.Runtime/Deputy/DeputyBase.cs
+++ b/packages/@jsii/dotnet-runtime/src/Amazon.JSII.Runtime/Deputy/DeputyBase.cs
@@ -23,12 +23,12 @@ namespace Amazon.JSII.Runtime.Deputy
         /// </summary>
         protected sealed class DeputyProps
         {
-            public DeputyProps(object[]? arguments = null)
+            public DeputyProps(object?[]? arguments = null)
             {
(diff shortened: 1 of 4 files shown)


packages/@jsii/dotnet-runtime/src/Amazon.JSII.Runtime/Deputy/DeputyBase.cs
```diff
@@ -23,12 +23,12 @@ public abstract class DeputyBase : IConvertible
         /// </summary>
         protected sealed class DeputyProps
         {
-            public DeputyProps(object[]? arguments = null)
+            public DeputyProps(object?[]? arguments = null)
```
rix0rrr, line 26: This reads like 🤢  to me but I guess this is what we have to do...?

rix0rrr (approved):

PR #2070, opened 2020-10-02
fix(jsii-diff): external structs returned from methods cannot be changed

### LINKED ISSUE #2064: jsii-diff: errors caused by L1 props types returned from `render()` methods
## :bug: Bug Report

Found in https://github.com/aws/aws-cdk/pull/10493

Root cause for compatibility check failures:

```
 formerly required property 'containerPort' is optional in @aws-cdk/aws-ecs.CfnService.LoadBalancerProperty [changed-type:@aws-cdk/aws-ecs.BaseService.loadBalancers]
 formerly required property 'subnets' is optional in @aws-cdk/aws-ecs.CfnService.AwsVpcConfigurationProperty [changed-type:@aws-cdk/aws-ecs.BaseService.networkConfiguration]
 formerly required property 'containerPort' is optional in @aws-cdk/aws-ecs.CfnService.LoadBalancerProperty [changed-type:@aws-cdk/aws-ecs.Ec2Service.loadBalancers]
 formerly required property 'subnets' is optional in @aws-cdk/aws-ecs.CfnService.AwsVpcConfigurationProperty [changed-type:@aws-cdk/aws-ecs.Ec2Service.networkConfiguration]
 formerly required property 'containerPort' is optional in @aws-cdk/aws-ecs.CfnService.LoadBalancerProperty [changed-type:@aws-cdk/aws-ecs.FargateService.loadBalancers]
 formerly required property 'subnets' is optional in @aws-cdk/aws-ecs.CfnService.AwsVpcConfigurationProperty [changed-type:@aws-cdk/aws-ecs.FargateService.networkConfiguration]
 formerly required property 'containerPort' is optional in @aws-cdk/aws-ecs.CfnService.LoadBalancerProperty [changed-type:@aws-cdk/aws-ecs.CfnService.loadBalancers]
```

Changes originating in L1s are marked as breaking failures, even though they technically shouldn't be.

Probably because L1s are being returned from `render()`/`bind()` functions, making the required->optional change breaking.

But if feels like the `@stability external` classification should weigh more heavily.

This is a good opportunity to refactor `jsii-diff` a little to have it do less work as well: when crawling complex types: separate the "weakening-ok/strengthening-ok" classifications from doing the actual checks themselves.

rix0rrr:
We used to be checking the compatibility of structs in-line, as
a part of checking whether the call signature of a function is
compatible with its previous signature (i.e., checking that a
return type is only strengthened and an input type is only weakened).

This had a couple of downsides:

- Performance: we would check the same type multiple times if it
  occurred (potentially recursively) in multiple call signatures.
- Locality: we would report the mismatch against the function signature
  instead of the changed type.
- Locality(2): if the change occurred in an `external` type but the
  consuming function was `stable`, the error ought to be suppressed but
  was reported instead.

The last reason recently cropped up in real life so that was a good
opportunity to rectify this long-standing architectural defect. I'm
afraid the result is a rather large refactoring; I apologize in advance,
but rest assured that everything will be better in the future, dear
reader.

New way of working:

* First identify all types in the assembly
* Crawl all types that have callables (methods, properties,
  initializers) and identify the types they use and whether they
  occur in input or output position.
* Validate all types, taking extra care to structurally compare
  structs based on whether they appear in input or output positions.

This only checks every struct once, report errors against the
struct itself and allows ignoring of structs that are marked `external`
or `experimental`.

While working on this change, it occurred to me that we were missing
a couple of validations that I added at the same time:

- In a `@subclassable` type, no method or property types can be changed
  *at all* (because C# won't allow it).
- A mutable property's type cannot be changed at all (it can
  both not be strengthened because it can be written, and not be weakened
  because it can be read).

Fixes #2064.

---


rix0rrr:
Implementation-wise:

`index.ts` | used to contain collection logic | moved to `type-comparison.ts`, is now OO
`classes-ifaces.ts` | used to contain rules logic | moved to `type-comparison.ts`, is now OO
 | and validation logic | moved to `validations.ts`
`enums.ts` | used to contain validation logic | moved to `validations.ts`

packages/jsii-diff/test/classes.test.ts
```diff
@@ -359,20 +359,83 @@ describe('implement base types need to be present in updated type system', () =>
...
+    'name: string',
+    'name: string | number',
+    /string \| number is not assignable to string/,
+  ],
+  [
+    'name: string | number',
+    'name: string',
+    undefined, // Strengthening is okay
+  ],
+])('change class property ', (oldDecl, newDecl, error) =>
```
RomainMuller, line 383: I kinda would have preferred using objects rather than tuples for this (since this is string-abundant).

```suggestion
test.each([
  {
    oldDecl: 'name: string',
    newDecl: 'name?: string',
    error: /type Optional<string> \(formerly string\): output type is now optional/,
  },
  {
    oldDecl: 'name?: string',
    newDecl: 'name: string',
    error: undefined, // Strengthening is okay
  },
  {
    oldDecl: 'name: string',
    newDecl: 'name: string | number',
    error: /string \| number is not assignable to string/,
  },
  {
    oldDecl: 'name: string | number',
    newDecl: 'name: string',
    error: undefined, // Strengthening is okay
  },
])('change class property ', ({oldDecl, newDecl, error}) =>
```

The same applies to subsequent similar blocks.

packages/jsii-diff/test/util.ts
```diff
@@ -12,10 +12,15 @@ export async function expectNoError(original: string, updated: string) {
 }
 
 export async function expectError(
-  error: RegExp,
+  error: RegExp | undefined,
   original: string,
   updated: string,
 ) {
+  if (error === undefined) {
```
RomainMuller, line 19: That's purely cosmetic, but I've been trying to be consistent with this particular notation:

```suggestion
  if (error == null) {
```

packages/jsii-diff/test/util.ts
```diff
@@ -12,10 +12,15 @@ export async function expectNoError(original: string, updated: string) {
...
+    await expectNoError(original, updated);
+    return;
```
RomainMuller, line 21: I reckon this is equivalent (but saves 1 promise un-wrapping, then re-wrapping):

```suggestion
    return expectNoError(original, updated);
```

RomainMuller (commented):
I was about to express some concerns with how this focuses solely on the type uses *within* this assembly, but re-considered a couple of times and right now I'm thinking it makes sense for the purpose of this tool (`@subclassable` is how the external usage is gated/protected).

This looks great as always, and I only have pretty minimal comments at this stage (basically - minimal style tuning). I'll be happy to approve once you've had a chance to go through those and decide on which you want to take...

packages/jsii-diff/test/classes.test.ts
rix0rrr: Cool, never even considered to do this.

packages/jsii-diff/test/util.ts
rix0rrr: I had that originally, but if I do that then eslint starts complaining about the regular return path:

```
{
	"resource": "/Users/huijbers/Workspaces/PublicCDK/jsii/packages/jsii-diff/test/util.ts",
	"owner": "eslint",
	"code": {
		"value": "consistent-return",
		"target": {
			"$mid": 1,
			"external": "https://eslint.org/docs/rules/consistent-return",
			"path": "/docs/rules/consistent-return",
			"scheme": "https",
			"authority": "eslint.org"
		}
	},
	"severity": 8,
	"message": "Expected to return a value at the end of async function 'expectError'.",
	"source": "eslint",
	"startLineNumber": 14,
	"startColumn": 23,
	"endLineNumber": 14,
	"endColumn": 23
}
```

And putting an empty `return` at the end of the function doesn't fix it.

Rather take the path of least resistance here and it's not like it's in a hot path or anything :).

RomainMuller: Makes sense!

RomainMuller (approved):

PR #2072, opened 2020-10-03
chore(java): use Jackson 2.11.3

sullis:
---


RomainMuller (approved):
I've restored the version range notation here in order to have more predictable resolution behavior.

I've opened up the upper bound so that it is effectively major-version pinned (instead of minor), since Jackson [claims](https://github.com/FasterXML/jackson/wiki/Jackson-Releases#general) to follow [Apache versioning](https://apr.apache.org/versioning.html), which from a user standpoint is close enough to [Semantic Versioning](http://semver.org/)

PR #2076, opened 2020-10-05
feat(pacmak/java): emit default interface implementations

### LINKED ISSUE #2014: [jsii-pacmak] Java code generation - extend a base class instead of generating the properties/methods
Given a TypeScript implementation:

```ts
export interface IResource  {
  readonly stack: Stack;
}
```

The Java generated code will include a `Jsii$Proxy` class that Implements all (including inherited) members obtained from `.jsii`:
```java
final static class Jsii$Proxy extends software.amazon.jsii.JsiiObject implements software.amazon.awscdk.core.IResource {
    public @org.jetbrains.annotations.NotNull software.amazon.awscdk.core.Stack getStack() {
        return this.jsiiGet("stack", software.amazon.awscdk.core.Stack.class);
    }
}
```
In the above example, the `IResource` interface includes a single property `Stack` which translates to an implementation of a get method -  `getStack()`, satisfying the interface requirement.
 
A common pattern in the AWS CDK is to create an interface and an abstract class which implement the interface, in the `IResource` example we will have the abstract class `Resource` which implements `IResource`:

```ts 

export abstract class Resource implements IResource {
...
}
``` 

Every class that wishes to implement `IResource` will extends `Resource`. 

We can use the same logic when generating `Jsii$Proxy` class, instead of implementing all of the memebers required by the interface,  `Jsii$Proxy` can extend the base class:

```java
final static class Jsii$Proxy extends software.amazon.jsii.JsiiObject implements software.amazon.awscdk.core.IResource extends software.amazon.awscdk.core.Resource {
  public @org.jetbrains.annotations.NotNull software.amazon.awscdk.core.Stack getStack() {
      return this.jsiiGet("stack", software.amazon.awscdk.core.Stack.class);
   }
}
```
#### Why should we do it?
First it will reduce the amount of generated code. 

It will also prevent a specific situation in which `Jsii$Proxy` class that was generated from a `.jsii` file of an old AWS CDK version might break when compiled against java class files that were generated with a newer version of the AWS CDK. 

To make it clearer lets look at an example, assume the above `Jsii$Proxy` was created using a `.jsii` file generated for AWS CDK  version `1.59.0`. In version `1.60.0` of the AWS CDK we add a property to `IResource` - `env`  and we also add it to `Resource`.  Since the  `Jsii$Proxy` class was created with no knowledge of the newly added `env` property, if we try to compile it with Java class files generated by version `1.60.0`  we will get a compilation error as it no longer implements `IResource`. We can avoid this by having `Jsii$Proxy` extending `Resource`.

RomainMuller:
It should be possible to add properties or methods to interfaces that
are only meant to be returned by APIs (i.e: non-`@sublassable`) without
breaking extensions of those interfaces in other modules. Unfortunatley,
when such an interface is extended in another module, it's generated
proxy class would previously not have implementations for the new
member, creating the potential for a `NoSuchMethodException` at runtime.

This PR changes the code generation so that a new `Jsii$Default` nested
interface is generated for each interface, which contains default
implementations (that forward calls to the *jsii kernel*) for all of its
members.

> *Note:* members whose name collide with members declared on
> `java.lang.Object` cannot have `default` implementations, since these
> would always be overridden by `java.lang.Object` implementations.

The new `Jsii$Default` interface is implemented (and the default
implementations used) by the `Jsii$Proxy` class. When an interface
extends one or more bases, its `Jsii$Default` implementation extends
those of it's parent interfaces (and inherits their implementations).

In order to support backwards compatibility with libraries generated
with versions of `jsii-pacmak` that do not implement this feature, a
new `metadata.jsii.pacmak.hasDefaultInterfaces` entry is added to the
*jsii assembly* at compilation time. It's value is used to determine
whether `jsii-pacmak` can rely on the existence of the `Jsii$Default`
interface being present.

Finally, new elements were added to the Java runtime library for jsii:
- The `@Internal` annotation is used to designate API elements that are
  not meant for public usage
- The `IJsiiObject` interface is the base interface of all
  `Jsii$Default` interfaces, and is implemented by `JsiiObject`. Since
  interface members need be `public`, the methods declared by this
  interface have been maked `@Internal`, and were renamed from `jsiiFoo`
  to `$jsii$foo` (in order to completely remove the risk for name
  clashes with user-defined APIs)

Fixes #2014

---


rix0rrr:
I guess you didn't go default-implementations-in-interfaces on purpose?

Is there still a point to this change if we don't go all the way? Aren't we setting ourselves up to fail in the future?

packages/jsii-pacmak/lib/targets/java.ts
```diff
@@ -1849,18 +1874,33 @@ class JavaGenerator extends Generator {
     const props = Object.values(propsByName);
     this.emitInterfaceBuilder(ifc, INTERFACE_PROXY_CLASS_NAME, props);
 
+    // If the data type has exactly 1 base datatype, we'll extend the base's proxy
+    const baseInterface =
+      ifc.interfaces?.length === 1 ? ifc.interfaces[0] : undefined;
+    const ownProps = props.filter(
+      (prop) => baseInterface == null || !prop.inherited,
```
rix0rrr, line 1881: This is stretching the definition of `ownProps` somewhat, in my opinion.

I mean, I guess it's "own" for *someone* (the generated class or the *generatee* interface), but this is not the someone I would have expected. I would have expected "own" to mean "as declared", not "as generated".

Maybe call it `hereProps`/`thereProps` instead? (For funzies)

Or `declaredProps`/`inheritedProps` ?

rix0rrr (approved):
Okay with suggestion and questions about long-term plans

packages/jsii-pacmak/lib/targets/java.ts
RomainMuller: Yeah I guess it's "contextually own props" sort of... Can definitely be misleading when looking only at usage sites...

RomainMuller:
> I guess you didn't go default-implementations-in-interfaces on purpose?
>
> Is there still a point to this change if we don't go all the way? Aren't we setting ourselves up to fail in the future, again, just in more complicated cases?

"Default implementation in interface" is an interesting approach however it will require introducing a whole additional set of interfaces: if the default behavior on the currently generated interface is "forward call to the jsii kernel", then a user implementing the interface (in Java-land) who "forgot" to implement one method would get a crash on any attempt to call that method (java -> asks kernel to call method -> method does not exist in kernel-owned instance -> 💥).

The current change improves the situation in a good chunk of uses; and which must be done anyway, because I cannot just assume all existing Java interfaces - including from dependencies - have been generated WITH default implementations; and might even be a breaking change if we happen to _actually_ have to emit additional interfaces (cannot assume those are PRESENT in dependencies).

RomainMuller:
After further conversation, even the current change is a breaking change, since this assumes the `Jsii$Proxy` classes are non-`private`; which would result in compilation failures when extending interfaces generated with previous `jsii-pacmak` releases.

So I guess since this *must* break, we should probably go "all in" with the default implementation-having interfaces... And we'll need to gate this behind a feature flag.

packages/@jsii/java-runtime/project/src/main/java/software/amazon/jsii/IJsiiObject.java
```diff
@@ -0,0 +1,59 @@
...
+     * Calls a JavaScript method on the object.
+     * @param method The name of the method.
+     * @param nativeType The return type.
+     * @param args Method arguments.
+     * @param <T> Java type for the return value.
+     *
+     * @return A return value.
+     */
+    @Nullable
+    <T> T $jsii$call(final String method, final NativeType<T> nativeType, @Nullable final Object... args);
```
rix0rrr, line 28: Are people going to see this in their IDEs?

rix0rrr: Should these be `@Internal`?

packages/@scope/jsii-calc-base-of-base/test/assembly.jsii
```diff
@@ -11,6 +11,13 @@
   "homepage": "https://github.com/aws/jsii",
   "jsiiVersion": "0.0.0",
   "license": "Apache-2.0",
+  "metadata": {
+    "jsii": {
+      "pacmak": {
+        "hasDefaultInterfaces": true
```
rix0rrr, line 17: Mmmmm! I likey!

packages/jsii-pacmak/lib/targets/java.ts
```diff
@@ -665,7 +667,8 @@ class JavaGenerator extends Generator {
 
   protected onEndClass(cls: spec.ClassType) {
     if (cls.abstract) {
-      this.emitProxy(cls);
+      const type = this.reflectAssembly.findType(cls.fqn) as reflect.ClassType;
```
rix0rrr, line 670: Wow this whole file is some hairy bull@#$%.

I don't think I can even start reviewing this maze of `if` statements in any kind of meaningful sense. Guess I'll trust you and the test set. :)

packages/jsii/lib/project-info.ts
```diff
@@ -137,6 +137,18 @@ export async function loadProjectInfo(
...
+  const metadata = mergeMetadata(
+    {
+      jsii: {
+        pacmak: {
+          // When `true`, `jsii-pacmak` will use the `Jsii$Default` implementation in code generation even for dependencies.
+          hasDefaultInterfaces: true,
+        },
+      },
+    },
+    pkg.jsii?.metadata,
```
rix0rrr, line 149: Sooo, you expect package.json to look like:

```js
{
  "jsii": {
    "metadata": {
      "jsii": {
         "pacmak": { ... }
      }
    }
  }
}
```

Feels like a bit of a stutter to repeat `jsii`. I can sorta see how it makes sense, but what is `metadata` expected to be used for?

rix0rrr (commented):
Love the metadata annotation solution! 

One question: the devil is usually in the transitive dependencies. Will the following daisy-chain of linking work properly?

`cdk-without-$Default`
used by
`library-with-$Default`
used by
`library-with-$Default`

I'll trust you on the implementation front :)

rix0rrr (approved):
Provisional approval after some requested deep soul searching :P.

Also:

4j0m1v

packages/@jsii/java-runtime/project/src/main/java/software/amazon/jsii/IJsiiObject.java
RomainMuller: The interface is `@Internal` as a whole, so all members are implicitly `@Internal`.

This might show up in IDE's, yeah... This isn't *entirely* unheard of, but at least this gives a pretty clear signal that though shalt not touch this.

packages/jsii-pacmak/lib/targets/java.ts
RomainMuller: I initially tried cleaning up this mess... and tracked back after sinking almost a day into it and not being close to done. I feel you, and I am genuinely itching to re-write this whole piece entirely at this point... maybe someday...

packages/jsii/lib/project-info.ts
RomainMuller: Metadata is used to carry arbitrary information into the assembly... For example if you have a custom documentation generator (*cough cough*) and need to provide it with specific configuration (say... a marketing-blessed name for cloud services?) so it renders the content you need... Then you can add whatever you please in there.

So the repeated `jsii` in there is mostly just because it's `jsii`'s own metadata namespace...

An alternative I have considered was to drop down and key up as `jsii-pacmak` instead (so it's one fewer level deep)... I don't know if this would look better or not to you? It didn't make much of a difference to me; and I felt the "top-level key" thing was perhaps a good thing to have in the long run?

RomainMuller:
And so with respects to the "daisy chaining", I believe I have tested this to work properly. The usage of the interfaces is conditioned to the dependency that owns a class having the flag... Worst case scenario we'll be generating sub-par code if the dependency closure is "unfortunate" here; but it *should* work (at least - as good as it did before this change, no worse).

packages/jsii/lib/project-info.ts
rix0rrr: Yeah it's fine. Just looked weird at first.

packages/@jsii/java-runtime/project/src/main/java/software/amazon/jsii/IJsiiObject.java
rix0rrr: Well if this is going to show up in IDE's, on  *all objects*, I'm less of a fan of putting the default implementation on the interface.

We only really need `$jsii$objectHandle()` or whatever you want to call it from the state of the interface. The implementation of `$jsii$call` could just be a static somewhere, so at least we don't pollute the autocomplete of every single object out there (too much).

packages/jsii-pacmak/lib/targets/java.ts
rix0rrr: I get it, and I don't mind.

rix0rrr:
Not too keen on the `$jsii$get` etc. appearing in the autocomplete of every single object. I'd rather have those as static methods somewhere. (https://github.com/aws/jsii/pull/2076#discussion_r507692899)

RomainMuller:
> Not too keen on the `$jsii$get` etc. appearing in the autocomplete of every single object. I'd rather have those as static methods somewhere. ([#2076 (comment)](https://github.com/aws/jsii/pull/2076#discussion_r507692899))

Good idea! Implemented!

packages/@jsii/java-runtime/project/src/main/java/software/amazon/jsii/IJsiiObject.java
RomainMuller: I've even inaugurated an external instance-to-objectID map... Would be nice if this could be "the standard", but I run into problems as this would require to be able to call `.hashCode` on instances *before* they are initialized, which isn't... something one should do.

But "pure" instances can now be used without any kind of inheritance from `JsiiObject`, which is a nice bonus, I think.

PR #1691, opened 2020-05-19
chore: remove @jsii/ruby-runtime

MrArnoldPalmer:
Removes the ruby-runtime package since its currently not a supported
target and not being actively developed in order to simplify and speed
up builds. Remove ruby dependencies from superchain Dockerfile.

---

By submitting this pull request, I confirm that my contribution is made under the terms of the [Apache 2.0 license].

[Apache 2.0 license]: https://www.apache.org/licenses/LICENSE-2.0

PR #2092, opened 2020-10-08
chore: removes unused software from superchain to reduce image size

udondan:
As previously discussed in https://github.com/udondan/jsii-publish/issues/1, this PR removes gcc and ruby to reduce the image size by ~180mb.

BREAKING CHANGE: The `jsii/superchain` Docker image no longer includes `ruby` and `gcc`. Users who need those should install them as part of their initialization script.

---


superchain/Dockerfile
```diff
@@ -47,14 +47,14 @@ RUN yum -y install python3 python3-pip
...
-  && gem install 'bundler:~>1.17.3' 'bundler:~>2.1.4'
-ENV PATH="$GEM_HOME/bin:$GEM_HOME/gems/bin:$PATH"
+#RUN amazon-linux-extras install ruby2.6                                                                                \
+#  && yum -y install ruby-devel rubygem-rdoc                                                                            \
+#  && yum clean all && rm -rf /var/cache/yum                                                                            \
+#  && echo 'install: --no-document' > /usr/local/etc/gemrc                                                              \
+#  && echo 'update: --no-document' >> /usr/local/etc/gemrc                                                              \
+#  && mkdir -p "$GEM_HOME"                                                                                              \
+#  && gem install 'bundler:~>1.17.3' 'bundler:~>2.1.4'
+#ENV PATH="$GEM_HOME/bin:$GEM_HOME/gems/bin:$PATH"
```
RomainMuller, line 57: Can we simply remove this instead of commenting out? GitHub is our memory if we even need to pull this back from the grave (plus it's not so hard, since it's an `amazon-linux-extras` install).

Can you also ensure you amend the README.md file in the `superchain` directory to remove `ruby` (and `gcc` if it so happens to be there) from there?

RomainMuller (commented):
@all-contributors add @udondan for maintenance

RomainMuller:
@all-contributors add @udondan for maintenance

superchain/Dockerfile
udondan: Good points, thanks, all taken care of.

RomainMuller (dismissed):
Great! Thank you @udondan.

I'll temporarily mark this `pr/do-not-merge` as I run some manual sanity tests just to be absolutely certain `gcc` is not needed for certain use-cases I know about which may require compiling code at install (i.e.: native components of node packages, etc...)

I'm fairly confident the `ruby` disappearance is not going to cause us problems.

RomainMuller:
LOL turns out CDK checks for `ruby` despite actually not using it (because it was listed in the project's requirements, and we have a requirements check as part of the build).

I need to actually remove/fix this before we merge this PR...

udondan:
Not sure what's going on, for some reason GitHub does not notify me by mail about updates in this issue. 😿 Just seeing updates by actively looking here. If I don't reply, It is not because I'm not interested.

RomainMuller:
@udondan - I haven't noticed a delay in your answering 🤣

But - do check what's in the `Notifications` section in the right bar in GitHub? Maybe somehow you're unsubscribed?

udondan:
Good point with gcc, I forgot about installing packages. I remember this is also a very common situation with Python packages.

The question for me now is this: If it turns out, jsii and the required core packages all can be installed w/o gcc, should we still remove or keep it? Or in other words, if a user's application requires additional packages that require compiling, should superchain support this right out of the box or is it the responsibility of the user to install gcc?

gcc weighs 135mb in the image, ruby only 45mb.

A little bit about my motivation of reducing the image size. It's not only the second longer it takes for downloading while building my jsii package. It's about decreasing the overall ecological footprint. 180mb is basically nothing, but things sum up.

- It takes longer to build
- It takes longer to upload to Docker Hub
- It takes more space on Docker Hub (not only a single copy, probably with many additional backups for every version)
- It will increase traffic for downloads
- Downloads take longer
- Downloads produce more traffic
- Downloads will be cached, requiring more disk space
- Running the image requires more disk space
- All the above increasing CPU time, waiting for longer builds, downloads etc

On the modern web it feels like all these resources (disk space, memory, traffic, CPU time, power consumption) are free - because Docker, GitHub etc offer them for free. But they're not. They produce carbon emissions due to power consumption and additional hardware orders. So yeah, 180mb or 135mb is not much, but things sum up and every tiny bit is worth trying to optimize. 😸

RomainMuller:
I mean I also see value in having people "assume less" about what is or is not in the image. If I have people assume something is in there at a particular version, I might end up in an awkward spot when I must change to an incompatible version in order to get my business going...

I'm all for including stuff that *the majority* of folks will need (because this will likely improve the odds of that being cached in; which would result in overall fewer bytes transferred on each run, particularly as this image is intended to be used in CI/CD contexts, where all runs would start on a fresh container).

RomainMuller (approved):

MrArnoldPalmer:
Just for context, we originally kept ruby in the superchain Dockerfile intentionally. https://github.com/aws/jsii/pull/1691#discussion_r427460164

RomainMuller:
> Just for context, we originally kept ruby in the superchain Dockerfile intentionally. [#1691 (comment)](https://github.com/aws/jsii/pull/1691#discussion_r427460164)

Fair. But I think this ship has sailed :)

RomainMuller (approved):

RomainMuller:
@Mergifyio update

PR #2100, opened 2020-10-09
fix(python): leftover jsii-kernel-* directories in TMPDIR

RomainMuller:
The Python runtime did not properly signal shutdown to the `node`
process hosting the *jsii kernel*, causing it to be abruptly killed on
Python interpreter exit. This means the *jsii kernel*'s own clean-up
code was not allowed to run.

This adds an `atexit` hook which will properly close the `node` process'
`STDIN` file descriptor, effectively signaling it should shut down. This
then gives the process 5 seconds to clean up after itself and finish
before sending it `SIGTERM` (at which point it is given 5 seconds to
terminate after which it will be terminated).

---


RomainMuller:
@all-contributors add @edsenabr for bug since he's the one who mentioned this one to me!

PR #2132, opened 2020-10-15
feat(rosetta): add default fixture for typical usage

RomainMuller:
Keeps track of the necessary information so that each example that does
not use a fixture automatically has the following transparently
available:

1. A named import of the current module, where the import name is the
   alphanumeric tail of the module's name (e.g: `@aws-cdk/aws-s3` is
   imported as `s3`)
2. For documentation declared on a type or type member, an instance of
   the type in question, as a constant named as the camelCased
   unqualified class name (e.g: for type `CfnBucket.LifecyclePolicy`,
   the constant is named `lifecyclePolicy`)

Additionally, in order to facilitate support for additional `import`
statements with the context of fixtures, those are transparently hoisted
above the first fixture line, so they are guaranteed to be at the
module's top-level scope (as this is the only place where they are
allowed).

Finally, added a CLI option to disable the use of worker threads, which
can be extremely helpful when trying to troubleshoot rosetta.

---


rix0rrr:
@nija-at recently opined to me that the `import * as s3 from '@aws-cdk/aws-s3'` and subsequently having to write `new s3.Bucket` is a bit cumbersome (his ire was directed towards `cognito` at that point, which takes up more visual space than just `s3`).

Since we have complete information, couldn't we just import all exported types by name?

```ts
import { Bucket, BucketProps, LifecycleTransition, ... } from '@aws-cdk/aws-s3';
```

So that the example code could reference simply `new Bucket()` ?

RomainMuller:
> Since we have complete information, couldn't we just import all exported types by name?
> 
> ```ts
> import { Bucket, BucketProps, LifecycleTransition, ... } from '@aws-cdk/aws-s3';
> ```
> 
> So that the example code could reference simply `new Bucket()` ?

This is brittle... New types can be added, which will expand the import list, and potentially conflict with a declaration/import in one of the example snippets... At which point it will no longer compile.

I kinda think we should rather look into manually writing fixtures that import all necessary things... This feels safer.

nija-at:
I'm ok with creating manual fixtures at least to begin with. We should automate these away over time.

I see two major gaps at the moment - 

(1) I see in this space is lack of visibility into invalid rosetta snippets. We need to do something to build quicker feedback than having to wait and look at the docs page to realize something is broken.

> New types can be added, which will expand the import list, and potentially conflict with a declaration/import in one of the example snippets... At which point it will no longer compile.

If I had the above, then I can be notified of this lack of compilation and adjust the snippet. The feedback loop is missing.

(2) adopt rosetta across the cdk repo - if we want this to be successful, **most** packages must start using rosetta. IIUC, this is not the case today.

----

Can we do something to address (1)

rix0rrr:
> How can we address (1)?

It is 100% feasible to add an `yarn XXX` script to do a rosetta-compile on just the current package.

This is not a default part of `yarn build` as it is fantastically more efficient to do Rosetta for the entire repo at once than per package, but for a dev workflow a single-package build should be sufficient.

> quicker feedback than having to wait and look at the docs page 

If we finish the migration it can become a build blocker. Gets the iteration time down from 1 week to 1 hour (still not perfect but better...)

rix0rrr:
@RomainMuller  If you don't feel like it I don't mind, but these arguments sound a little FUDdy:

> This is brittle... New types can be added, which will expand the import list, 

Which is why we would autogenerate it. That's not so brittle.

> and potentially conflict with a declaration/import in one of the example snippets... 

Snippets have no business declaring types that come from the library. If they DO and MUST (we are talking <1% of snippets here), they can switch off the autofixture.

I don't really see the problem.

nija-at:
> This is not a default part of yarn build as it is fantastically more efficient to do Rosetta for the entire repo at once than per package, but for a dev workflow a single-package build should be sufficient.

I like this and we should do this. Maybe we can make this an opt-in at the `package.json` level and an env variable.
The env var can be useful to skip this so as to not slow down `buildup` and such.

rix0rrr (approved):

RomainMuller:
This has been superseded by several other pull requests and is no longer needed (some features here are not replicated in the other work, but may not be necessary - we can revisit in the future if needed).

PR #2138, opened 2020-10-18
fix(pacmak/python): escape character escapes in python docs

### LINKED ISSUE #2137: Comments don't escape character escapes
### Affected Languages

- [X] `Python`
- [ ] `Java`
- [ ] .NET (`C#`, `F#`, ...)

### General Information
* **JSII Version:** 1.1.0

### What is the problem?
If a comment contains what looks like a character escape, for example `\N`, it remains unescaped in the generated python comments. If it happens to be a valid escape, then it isn't that big of a deal, but if it isn't, then referencing the generated code results in an error of the form `malformed character escape`.

### Verbose Log
https://github.com/hashicorp/terraform-cdk/issues/394 has some more details.

jsteinich:
Fixes #2137 

---


packages/jsii-pacmak/test/python.test.ts
```diff
@@ -113,6 +113,17 @@ test('list with multiline text', () => {
   );
 });
 
+test('escape character escaping', () => {
+  converts(
+    [
+      'For example, if you specify \\N, BigQuery interprets \\N as a null value',
```
jsteinich, line 119: I'm not certain if I'm doing something wrong with my test or the markdown parser is splitting into multiple nodes and the change doesn't actually work.
The literal string I'm hoping to test is `For example, if you specify "\N", BigQuery interprets "\N" as a null value`

jsteinich: Seems like it is a separate node.

packages/jsii-pacmak/lib/markdown.ts
```diff
@@ -20,7 +20,17 @@ export function md2rst(text: string) {
...
   function textOf(node: commonmark.Node) {
-    return node.literal ?? '';
+    if (node.literal) {
+      if (
+        node.prev?.literal === '\\' && // following a \ (might need escaping)
+        !/^\\.*/.exec(node.literal) && // doesn't start with a \ (not being escaped)
+        node.prev.prev?.literal !== '\\' // not two \ in a row (not already escaped)
+      ) {
+        return `\\${node.literal}`;
+      }
```
jsteinich, line 30: This isn't a perfect test, but seems unlikely to need expansion in practice.

RomainMuller: At least it's a start and it appears to be sufficient for the test... I don't see a reason to over-complicate this...

packages/jsii-pacmak/test/python.test.ts
```diff
@@ -113,6 +113,17 @@ test('list with multiline text', () => {
...
+  converts(
+    [
+      `For example, if you specify \\N, BigQuery interprets \\N as a null value`,
+    ],
+    [
+      `For example, if you specify \\\\N, BigQuery interprets \\\\N as a null value`,
+    ],
+  );
+});
+
```
RomainMuller, line 126: This omits to test for the "must not escape" case.

packages/jsii-pacmak/lib/markdown.ts
```diff
@@ -20,7 +20,17 @@ export function md2rst(text: string) {
...
+      return node.literal;
+    }
+    return '';
```
RomainMuller, line 33: I wonder if the *correct* thing to do isn't to simply *always* escape back-slashes... If the Markdown literal has an escaped backslash, then the Python documentation must show an escaped backslash, too... Ergo we must escape even escaped escape sequences properly...

This would mean the function could basically be:

```suggestion
    return node.literal?.replace(/\\/g, '\\\\') ?? '';
```

Don't you think so?

RomainMuller (approved):

RomainMuller (approved):

PR #2146, opened 2020-10-19
feat(pacmak): prerelease identifier support

### LINKED ISSUE #2114: [pacmak] support pre-release versions
### Affected Languages
- [x] `Python`
- [x] `Java`
- [x] .NET (`C#`, `F#`, ...)

### Description

We would like to release the the next major version of `aws-cdk` and the `constructs` libraries
initially as pre-release. See [cdk pre-release identifiers](https://github.com/aws/aws-cdk-rfcs/blob/master/text/0079-cdk-2.0.md#releasing-to-package-managers) and [`constructs` pre-release identifier](https://github.com/aws/constructs/commit/c9b0726008dd9b5148c6efea5341a0f1c1f62715#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R126).

JSII packing needs to natively support this.

### Proposed Solution

If the `version` field in `package.json` contains a pre-release suffix, JSII should treat the suffix as a pre-release identifier and create pre-release versions appropriate for the package managers of the target languages.

For example, if `version` is specified as `2.0.0-alpha.1`, it must automatically detect that this is a pre-release version and create target versions as `2.0.0-alpha.1` for Maven Central, `2.0.0-alpha.1` for Nuget and `2.0.0.a1` for PyPI.

Some package managers are restrictive in what is acceptable pre-release versions. For example, PyPI [only allows](https://packaging.python.org/guides/distributing-packages-using-setuptools/#pre-release-versioning) `devN`, `aN`, `bN` and `rcN` (where `N` is the pre-release version number) as valid pre-release versions. 
For this reason, when enabled for these package managers, it must make reasonable assumptions about allowed pre-release identifiers, erring if a pre-release identifier cannot be translated and allow overriding them in the `jsii` section.

RomainMuller:
When an NPM package built with `jsii` is tagged with a version that
includes a prerelease identifier (e.g: `1.2.3-alpha.1`), automatically
represent this as a pre-release version in generated packages, too. For
most currently supported languages, this simply means forwarding the
version untouched, however for Python, prerelease identifiers need to
be mapped into one of the supported tokens in PyPI (`.dev#`, `.a#`,
`.b#` or `.rc#`).

Fixes #2114

---


packages/jsii-pacmak/lib/targets/version-utils.ts
```diff
@@ -55,10 +67,90 @@ export function toPythonVersionRange(semverRange: string): string {
...
 }
 
+/**
+ * Converts an original version number from the NPM convention to the target
+ * language's convention for expressing the same. For versions that do not
+ * include a prerelease identifier, this always returns the assembly version
+ * unmodified.
+ *
+ * @param assemblyVersion the assembly version being released
+ * @param target          the target language for which the verison is destined
```
nija-at, line 77:
```suggestion
 * @param target          the target language for which the version is destined
```

packages/jsii-pacmak/lib/targets/version-utils.ts
```diff
@@ -55,10 +67,90 @@ export function toPythonVersionRange(semverRange: string): string {
...
+        case 'beta':
+          return `${version.major}.${version.minor}.${version.patch}.b${sequence}`;
+        case 'rc':
+          return `${version.major}.${version.minor}.${version.patch}.rc${sequence}`;
+        default:
+          info(
+            `Unable to map prerelease identifier  (in: ${assemblyVersion}) to python, as label ${inspect(
+              label,
+            )} is not mapped (only "dev", "alpha", "beta" and "rc" are)`,
+          );
```
nija-at, line 129: throw an error here

packages/jsii-pacmak/lib/targets/version-utils.ts
```diff
@@ -55,10 +67,90 @@ export function toPythonVersionRange(semverRange: string): string {
...
+    case 'java':
+    case 'js':
+      // Not touching - the NPM version number should be usable as-is
+      break;
+    default:
+      info(
+        `Unknown target ${inspect(
+          target,
+        )} for ${assemblyVersion}. Returning version as-is.`,
+      );
```
nija-at, line 142: throw an error here as well.

packages/jsii-pacmak/lib/targets/version-utils.ts
```diff
@@ -137,6 +229,6 @@ function toBracketNotation(
     if (trimDashZero) {
       str = str.replace(/-0$/, '');
     }
-    return suffix ? `${str}${suffix}` : str;
+    return suffix ? `${str}${suffix}` : toReleaseVersion(str, target);
```
nija-at, line 232: The existing implementation looks as though one form of version string (semver) is being converted to another form (bracket form?). 
Perhaps leave this translation 'pure' and move the call to `toReleaseVersion()` one level above?

packages/jsii-pacmak/lib/targets/dotnet.ts
```diff
@@ -274,7 +275,7 @@ export default class Dotnet extends Target {
     assm: spec.Assembly,
   ): { [language: string]: PackageInfo } {
     const packageId = assm.targets!.dotnet!.packageId;
-    const version = assm.version;
+    const version = toReleaseVersion(assm.version, 'dotnet');
```
nija-at, line 278: Something is off here in terms of code cleanliness. 

Ideally, besides the jsii `Assembly` context, there's also a pacmak specific context that gets passed around, which converts `assembly.version` into the release version identifier. So the conversion of one version to the other happens only once. Something like -

```ts
interface PacMakContext {
  public readonly assm: spec.Assembly;
  // ...
  public readonly releaseVersion;
}
```

This would "home" the translation in the correct place.

Do you see value in something like that?

packages/jsii-pacmak/lib/targets/version-utils.ts
```diff
@@ -55,10 +67,90 @@ export function toPythonVersionRange(semverRange: string): string {
...
+  if (version == null) {
+    throw new Error(
+      `Unable to parse the provided assembly version: "${assemblyVersion}"`,
+    );
+  }
+  if (version.prerelease.length === 0) {
+    return assemblyVersion;
+  }
+  switch (target) {
+    case 'python':
```
nija-at, line 95: Hmmm. Do we not have an enum or something that represents the list of supported target langs?

packages/jsii-pacmak/lib/targets/version-utils.ts
```diff
@@ -55,10 +67,90 @@ export function toPythonVersionRange(semverRange: string): string {
...
+      }
+      switch (label) {
+        case 'dev':
+          return `${version.major}.${version.minor}.${version.patch}.dev${sequence}`;
+        case 'alpha':
+          return `${version.major}.${version.minor}.${version.patch}.a${sequence}`;
+        case 'beta':
+          return `${version.major}.${version.minor}.${version.patch}.b${sequence}`;
+        case 'rc':
+          return `${version.major}.${version.minor}.${version.patch}.rc${sequence}`;
```
nija-at, line 123: We should allow user overrides of the pre-release suffix, per language, in the `package.json`? 

Useful specifically for target package managers, like Python, that don't allow *any* pre-release identifier.
*use case:* customer releases pre-release version of their library in 1 language and now wants to expand to all, but has used a pre-release identifier that doesn't automatically map.

nija-at: Another reason this might be worth adding is that it's not clear what Python allows or disallows. [This spec](https://packaging.python.org/guides/distributing-packages-using-setuptools/#pre-release-versioning) states "may be published as" which is ambiguous.

And [PEP-440](https://www.python.org/dev/peps/pep-0440/#pre-release-spelling) states 'alpha' and 'beta' are allowed and, 'pre' represents release candidate. Is PEP-440 active?

nija-at: Might have found the first use case 😅

https://github.com/aws/constructs/blob/2a6bb57f02a22d844b71b97428f2545afb4f87ea/package.json#L126

packages/jsii-pacmak/lib/targets/version-utils.ts
RomainMuller: The bracket form is a *range* transformation... And the range has versions on both sides, which need correct handling.

packages/jsii-pacmak/lib/targets/dotnet.ts
RomainMuller: Could be. I mean this is also why long term I'm interested in having code-gen happen within `jsii` directly...

packages/jsii-pacmak/lib/targets/version-utils.ts
RomainMuller: Not an enum; but the `TargetName` is a string-literal-union type, so one can't make typos here without the compiler noticing.

packages/jsii-pacmak/lib/targets/version-utils.ts
RomainMuller: I would map `pre` to `dev` and the `constructs` problem would be gone?

I'm ambivalent about adding more knobs to override behaviors in the configuration files; because the onus is then on the user to know what is acceptable or not (so now, they have to go look at PEP-440, and get confused by how it seems to say something different from `pip`'s documentation....). I would rather add such a feature only once it has been demonstrated to be the only way to unblock a reasonable use-case.

packages/jsii-pacmak/lib/targets/dotnet.ts
RomainMuller: Also - the problem with your suggestion is that it would need to carry a context... for *each* dependency, too. Because we also need to perform the prerelease mapping for dependency ranges (otherwise we'll have resolution issues)

packages/jsii-pacmak/lib/targets/version-utils.ts
```diff
@@ -55,10 +67,93 @@ export function toPythonVersionRange(semverRange: string): string {
...
+ * language's convention for expressing the same. For versions that do not
+ * include a prerelease identifier, this always returns the assembly version
+ * unmodified.
+ *
+ * @param assemblyVersion the assembly version being released
+ * @param target          the target language for which the version is destined
+ *
+ * @returns the version that should be serialized
+ */
+export function toReleaseVersion(
```
nija-at, line 81: We need this entire pre-release support in pacmak and the python quirks documented somewhere.

I see this package doesn't have a proper README.
Can we start by adding this?

nija-at: As a side note, I recommend adding something similar to 'prlinter' for jsii, so that all `feat` require a README update. Pretty useful IMO.

packages/jsii-pacmak/lib/targets/dotnet.ts
nija-at: I'll let you decide how useful or not it is. I've barely worked in this side of the code base.
Just feels like something we could centralize `toReleaseVersion()` and make it available via a context.

packages/jsii-pacmak/test/targets/version-utils.test.ts
```diff
@@ -102,3 +104,60 @@ describe(toPythonVersionRange, () => {
...
+      java: '1.2.3',
+      js: '1.2.3',
+      python: '1.2.3',
+    },
+    '1.2.3-pre': {
+      dotnet: '1.2.3-pre',
+      go: '1.2.3-pre',
+      java: '1.2.3-pre',
+      js: '1.2.3-pre',
+      python: /Unable to map prerelease identifier \(in: 1\.2\.3-pre\) components to python: \[ 'pre' \]/,
```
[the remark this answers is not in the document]
RomainMuller: Only `-pre` is not supported nor supportable.
Python *requires* a sequence identifier (so `-pre.#`).

If I make `-pre` map to `.dev0`, then how do I distinguish it form the (also valid) `-pre.0`?

packages/jsii-pacmak/test/targets/version-utils.test.ts
```diff
@@ -102,3 +104,60 @@ describe(toPythonVersionRange, () => {
...
+      go: '1.2.3-rc.9',
+      java: '1.2.3-rc.9',
+      js: '1.2.3-rc.9',
+      python: '1.2.3.rc9',
+    },
+  };
+
+  for (const [version, targets] of Object.entries(examples)) {
+    test(`"${version}" translations`, () => {
+      for (const [target, expectedResult] of Object.entries(targets)) {
```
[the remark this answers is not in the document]
RomainMuller: This is, in fact, guaranteed by the TypeScript compiler, due to how expectations are defined as:

```ts
type Expectations = { readonly [K in TargetName]: string | RegExp };
```

This makes valid values require to have keys in `TargetName` (and nothing else), with none of them optional (so all of them required). I verified this while authoring the PR (I was pleasantly surprised)

packages/jsii-pacmak/test/targets/version-utils.test.ts
```diff
@@ -102,3 +104,60 @@ describe(toPythonVersionRange, () => {
       expect(toPythonVersionRange(semver)).toEqual(python));
   }
 });
+
```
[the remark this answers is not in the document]
RomainMuller: That is precisely why this PR is still a *Draft*. I don't quite have a good way to test for this right now... And I have to come up with something.

packages/jsii-pacmak/lib/targets/version-utils.ts
RomainMuller: > As a side note, I recommend adding something similar to 'prlinter' that we have in the cdk for the jsii repo. Enforce that all feat require a README update. Pretty useful IMO.

I have mixed feelings on this based on how often I have needed to apply labels to bypass this validation in my CDK contributions (for good reasons) in the past. But you're probably right, still... At least it ensures people have fought about it.

Anyhow - the fact there is no README.md entry yet is because this is a draft and I didn't want to have to come back to it multiple times if the implementation changes dramatically as I get into testing it properly.

docs/configuration.md
```diff
@@ -1,8 +1,7 @@
 # Configuration
 
-The configuration for `jsii` is recorded in the `package.json` file, which is
-the standard package manifest for NPM packages. This document describes the
-constraints and extensions `jsii` adds to the [package.json schema].
+The configuration for `jsii` is recorded in the `package.json` file, which is the standard package manifest for NPM
+packages. This document describes the constraints and extensions `jsii` adds to the [package.json schema].
```
nija-at, line 4: (for the future) prefer wholesale adjustments and refactors to be separate PRs, hence separate commits 😉

packages/jsii-pacmak/lib/targets/dotnet/filegenerator.ts
```diff
@@ -183,6 +184,6 @@ export class FileGenerator {
       // suffix is guaranteed to start with a leading `-`
       return `${assembly.version}${suffix}`;
     }
-    return assembly.version;
+    return toReleaseVersion(assembly.version, TargetName.DOTNET);
```
nija-at, line 187: Non-blocking: It's a consolation that in most languages (besides .NET), `toReleaseVersion()` is called once or from a single file. Is it simple to do the same with .NET?

packages/jsii-pacmak/lib/targets/js.ts
```diff
@@ -2,26 +2,38 @@ import * as spec from '@jsii/spec';
...
 export default class JavaScript extends Target {
   public static toPackageInfos(
     assm: spec.Assembly,
   ): { [language: string]: PackageInfo } {
     const packageInfo: PackageInfo = {
       repository: 'NPM',
-      url: `https://www.npmjs.com/package/${assm.name}/v/${assm.version}`,
+      url: `https://www.npmjs.com/package/${assm.name}/v/${toReleaseVersion(
+        assm.version,
+        TargetName.JAVASCRIPT,
```
nija-at, line 17: Call `toReleaseVersion()` once in this code block.

```ts
const releaseVersion = toReleaseVersion(assm.version, TargetName.JAVASCRIPT);
```

packages/jsii-pacmak/lib/targets/js.ts
```diff
@@ -2,26 +2,38 @@ import * as spec from '@jsii/spec';
...
+      )}`,
       usage: {
         'package.json': {
           language: 'js',
           code: JSON.stringify({ [assm.name]: `^${assm.version}` }, null, 2),
```
nija-at, line 22: just checking - need to call `toReleaseVersion()` here?

packages/jsii-pacmak/lib/targets/version-utils.ts
nija-at: I understand it's annoying to add the label, but the value outweighs it, IMO.

packages/jsii-pacmak/lib/targets/version-utils.ts
```diff
@@ -55,10 +67,98 @@ export function toPythonVersionRange(semverRange: string): string {
...
+      }
+      if (!Number.isInteger(sequence)) {
+        throw new Error(
+          `Unable to map prerelease identifier (in: ${assemblyVersion}) to python, as sequence ${inspect(
+            sequence,
+          )} is not an integer`,
+        );
+      }
+      const baseVersion = `${version.major}.${version.minor}.${version.patch}`;
+      switch (label) {
```
nija-at, line 114: Link to the pep spec would be useful here

docs/configuration.md
```diff
@@ -194,21 +180,40 @@ Example:
...
+are accepted by `jsii-pacmak`, and are translated according to the following table:
+
+| Source Version (`npm`) | Python Version ([PEP-440]) | Notes                                  |
+| ---------------------- | -------------------------- | -------------------------------------- |
+| `X.Y.Z.dev.N`          | `X.Y.Z.devN`               | Development, iteration `N`.            |
+| `X.Y.Z.pre.N`          | `X.Y.Z.devN`               | Development, iteration `N`             |
+| `X.Y.Z.alpha.N`        | `X.Y.Z.aN`                 | Alpha release, iteration `N`           |
+| `X.Y.Z.beta.N`         | `X.Y.Z.bN`                 | Beta release, iteration `N`            |
+| `X.Y.Z.rc.N`           | `X.Y.Z.rcN`                | Release candidate, iteration `N`       |
+| `X.Y.Z.prN.M`          | `X.Y.Z.postN.devM`         | Post-release number `N`, iteration `M` |
```
nija-at, line 201: I don't see a use case for why we should support post releases or development versions of post releases. They should be new patch/minor versions, or new pre-release identifiers.

I can only find this in Python and even PEP-440 discourages them -

> The use of post-releases to publish maintenance releases containing actual bug fixes is strongly discouraged. In general, it is better to use a longer release number and increment the final component for each maintenance release.

> Creating post-releases of pre-releases is strongly discouraged, as it makes the version identifier difficult to parse for human readers. In general, it is substantially clearer to simply create a new pre-release by incrementing the numeric component.

nija-at (changes requested):
My main concern is about post releases.

docs/configuration.md
RomainMuller: Yeah this is auto-format-on-save... -_-

packages/jsii-pacmak/lib/targets/dotnet/filegenerator.ts
RomainMuller: Maybe it would be but this could entail a significant refactor... And perhaps would reduce the overall quality of the .NET generator's code quality 🙃

packages/jsii-pacmak/lib/targets/js.ts
RomainMuller: I guess yes (for consistency). But `toReleaseVersion` is basically no-op for JavaScript.

docs/configuration.md
RomainMuller: I mean I don't intend on those post-releases being published at all. I conventionally use `-pr<pr-id>.0` version suffix for PR-bound pre-releases (i.e: the artifacts that would be produced if that PR was to be merged).

So I found it actually mapped kind of neatly (this is the post-release of this thing for pr `<pr-id>`, sequence `0`).

Would it be "better" to you if I'd simply map `X.Y.Z-prN.0` to `X.Y.Z.devN` (and fail if the sequence is non-0)?

nija-at: > Would it be "better" to you if I'd simply map X.Y.Z-prN.0 to X.Y.Z.devN (and fail if the sequence is non-0)?

Better yes, but still doesn't work. 

Do we have a use case for PR-bound pre-releases? What's the value over using `dev` prefix with an auto-inc sequence instead?

packages/jsii-pacmak/test/targets/version-utils.test.ts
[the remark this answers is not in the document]
nija-at: Makes sense. I mis-read.

packages/jsii-pacmak/lib/targets/dotnet/filegenerator.ts
nija-at: Alright-y. Not ideal but ok :)

docs/configuration.md
```diff
@@ -226,30 +230,28 @@ Example:
...
 ```
 
-The resulting artifact is a **Maven** package that can be deployed to
-[Maven Central] using the `deploy-staged-repository` command of the
-[nexus-staging-maven-plugin].
+The resulting artifact is a **Maven** package that can be deployed to [Maven Central] using the
+`deploy-staged-repository` command of the [nexus-staging-maven-plugin].
 
-[Maven Centra]: https://search.maven.org
+[maven centra]: https://search.maven.org
```
nija-at, line 236:
```suggestion
[maven central]: https://search.maven.org
```

nija-at (approved):

docs/configuration.md
RomainMuller: Alright - dropping support for this, will re-instate if actually needed.

PR #2154, opened 2020-10-20
chore(tests): migrate jest config to jest.config.ts

RomainMuller:
This enables customizing behavior when `CI=true`, which is
conventionally the case on continuous integration systems, where
parallelization can be detrimental to performance due to use of budget
compute platforms.

Additionally, switched `jest` runner to `jest-circus` (which will
eventually become the default `jest` runner), as it promises to be
somewhat faster (at least, it did not degrade performance as far as I
could tell).

The `jest.config.ts` file is naturally excluded from all `npm` packages,
as it is a `.ts` file. It has also been excluded from TypeScript
compiler inputs (via `exclude` in `tsconfig.json` files) as it does not
need to be compiled with the projects (which would additionally cause
spurious `jest.config.d.ts` and `jest.config.js` files to be created).

---


packages/@jsii/runtime/jest.config.ts
```diff
@@ -0,0 +1,17 @@
+import type { Config } from '@jest/types';
+import base from '../../../jest.config';
+
+const config: Config.InitialOptions = {
+  ...base,
+
+  coverageThreshold: {
+    ...base.coverageThreshold,
```
MrArnoldPalmer, line 8: NIT: We could use something like [deep-merge](https://www.npmjs.com/package/deepmerge) or write our own little recursive object merge function. I feel like we probably already have one somewhere.

packages/jsii/lib/compiler.ts
```diff
@@ -433,8 +433,10 @@ export class Compiler implements Emitter {
         continue;
       }
 
-      // eslint-disable-next-line @typescript-eslint/no-var-requires, @typescript-eslint/no-require-imports
-      const tsconfig = require(tsconfigFile);
+      const { config: tsconfig } = ts.readConfigFile(
+        tsconfigFile,
+        ts.sys.readFile,
+      );
```
MrArnoldPalmer, line 439: What's this for? Just better form to use `readConfigFile` instead of a json `require()`? I agree with that just wondering if its necessary for some reason.

jest.config.ts
```diff
@@ -0,0 +1,36 @@
...
+      statements: 60,
+    },
+  },
+  errorOnDeprecated: true,
+  // When in Continuous Integration, use only 1 worker (assuming "frugal" runner type)
+  maxWorkers: env.CI === 'true' ? 1 : defaults.maxWorkers,
+  setupFilesAfterEnv: ['jest-expect-message'],
+  testEnvironment: 'node',
+  testMatch: ['**/?(*.)+(spec|test).ts'],
+  testRunner: 'jest-circus/runner',
```
MrArnoldPalmer, line 30: Since `jest-circus` is included in the top level config, shouldn't it be a devDependency in the repo root package.json or within each individual module using jest? All modules inheriting their config from here will use it as their test runner and only `jsii-pacmak` has it in its package.json?

MrArnoldPalmer (changes requested):

packages/jsii/lib/compiler.ts
RomainMuller: The thing here is that `tsconfig.json` files are not "standard" JSON - they allow comments and trailing commas. `require` and `readJson` will choke on that and fail, which is undesirable. This uses the same code that `tsc` itself uses, so we are "guaranteed" to succeed at loading that if it's a valid configuration file.

jest.config.ts
RomainMuller: Oh wow, so I ran `yarn add -WD jest-circus` in the wrong directory! Good catch!

MrArnoldPalmer (approved):

PR #2167, opened 2020-10-22
chore: in-process generated code regression tests

RomainMuller:
Instead of shelling out to `jsii-pacmak`, this runs the regression tests
in-process, enabling coverage collection from these. This provides much
better visibility in terms of the test coverage (which is >80% overall),
while previously it was ridiculously under-estimated.

In order to achieve this feat, I moved the bits of logic in
`bin/jsii-pacmak.ts` into the `lib/index.ts` file, so that this code can
conveniently be invoked from a test. This would constitute a first step
towards a programmatic API for `jsii-pacmak` in some future.

This happens to also make `jsii-pacmak` a tiny bit faster (through
some improvements I made in the `async` handling code while
I was there), although noting really material.

---


eladb (commented):
Haven't reviewed the PR but from the description it sounds like we are losing some coverage by not including the CLI. Is that correct or are we now doing both?

RomainMuller:
> Haven't reviewed the PR but from the description it sounds like we are losing some coverage by not including the CLI. Is that correct or are we now doing both?

The CLI is still being used by the build tests (which are orchestrated outside of `jest`). So it's actually still covered as good as it's always been.

MrArnoldPalmer (approved):

Issue #1830, opened 2020-07-29
[jsii] Confusing error message when using internal types in exported declarations


RomainMuller:
### Affected Languages

- [X] `TypeScript` or `Javascript`
- [ ] `Python`
- [ ] `Java`
- [ ] .NET (`C#`, `F#`, ...)

### General Information
* **JSII Version:** `1.9.0`
* **Platform:**     *any*

### What is the problem?
It can be useful to have internal base types that are intentionally not exported out of the module. In many cases, developers want to organize their code in such a way that those types are *internally exported* within the module, but not from the module's entry point:
- Declaring these types in *private* files, unexpected from the index
- Declaring these types in *public* files, but only exporting specific symbols in the index (using `export { ... } from "...";`)

When such a type is extended/implemented, the `jsii` compiler does not automatically erase those (as it would for un-exported types, or those documented as `@internal`), and instead produces the following errors (see #1818 for more context of this particular one):

```
lib/shared/policy-statement/9-final.ts:6:1 - error TS9999: JSII: Unable to resolve referenced type 'iam-floyd.PolicyStatement'. Type may be @internal or unexported

  6 export class PolicyStatement extends PolicyStatementWithPrincipal {
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  7   protected cdkApplied = false; // internally used to check if resources, actions and conditions have already been applied to the policy
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
... 
 15   }
    ~~~
 16 }
    ~
lib/generated/awsdeeplens.ts:8:1 - error TS9999: JSII: Unable to resolve referenced type 'iam-floyd.PolicyStatement'. Type may be @internal or unexported

  8 export class Deeplens extends PolicyStatement {
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  9   public servicePrefix = 'deeplens';
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
... 
514   }
    ~~~
515 }
    ~
```

These are misleading:
[...]

PR #2172, opened 2020-10-23
fix(jsii): annotation '@internal' causes missing type declarations downstream

### LINKED ISSUE #1947: @internal API elements are erased for all languages except TypeScript
### Affected Languages
- [ ] `Python`
- [ ] `Java`
- [ ] .NET (`C#`, `F#`, ...)

### What is the problem?

Currently in `compliance.ts`:

```ts
/**
 * @internal
 */
export interface InternalInterface {
  readonly prop: string;
}

export interface ExtendsInternalInterface extends InternalInterface {
  readonly boom: boolean;
}
```

We made this work (with lots of effort) for jsii client languages... but then forget to do it for TypeScript. The `.d.ts` looks like this:

```ts
// (...nothing about InternalInterface)

export interface ExtendsInternalInterface extends InternalInterface {   // <--- asplodes!
    readonly boom: boolean;
}
```

This leads to compilation errors for TypeScript.

RomainMuller:
Stop setting `stipInternal: true` in `tsconfig.json` files, as this makes it impossible
to mark intentionally unexported types as `@internal` without causing downstream
compilation failures due to missing type declarations.

Fixes #1947
Related to #1830 

---

### DIFF
--- a/packages/jsii/lib/compiler.ts
+++ b/packages/jsii/lib/compiler.ts
@@ -32,7 +32,7 @@ const BASE_COMPILER_OPTIONS: ts.CompilerOptions = {
   strict: true,
   strictNullChecks: true,
   strictPropertyInitialization: true,
-  stripInternal: true,
+  stripInternal: false,
   target: ts.ScriptTarget.ES2018,
 };
 


nija-at (approved):
feat? Should be a 'fix', no?

RomainMuller:
> feat? Should be a 'fix', no?

I hesitated... changed it up :)

PR #2192, opened 2020-10-27
feat(dotnet): introduce `UnsafeCast<T>()` method

RomainMuller:
In order to allow converting an opaque object instance to an arbitrary
(jsii-managed) interface type, the `UnsafeCase` operation may be used to
instantiate a proxy while bypassing all type consistency checks.

This is similar to using `as any` or `as T` in TypeScript, meaning that
if the user performs a cast to an incorrect/unsupported type, undefined
behavior follows.

This would unblock certain specific use-case scenarios that static
languages render difficult to enact, such as the one described in
aws/aws-cdk#3284.

In particular, this is the only way out until aws/aws-cdk-rfcs#193
is delivered, for dealing with instances of interfaces returned through
a type union API.

---


MrArnoldPalmer (approved):
I'm into this. 👍

PR #2211, opened 2020-11-02
feat(rosetta): hoist imports above fixtures

RomainMuller:
When `import` statements are used in a code example, they must be
hoisted at the top of the fixture (if any fixture is used), as they
otherwise may occur at locations where such statements are illegal (they
are only accepted at the top-level scope).

Additionaly, added some (hidden) comments to highlight where the hoisted
imports begin and end, as well as where the snipped code begins and ends
in an attempt to make the "complete source code" a little easier to
navigate, especially in the context of larger fixtures.

---


packages/jsii-rosetta/lib/snippet.ts
```diff
@@ -66,7 +66,7 @@ export function completeSource(snippet: TypeScriptSnippet) {
 function parametersFromSourceDirectives(
   source: string,
 ): [string, Record<string, string>] {
-  const [firstLine, rest] = source.split('\n', 2);
+  const [firstLine, ...rest] = source.split('\n');
```
RomainMuller, line 69: Note - there was a bug right here that my test highlighted (tests with only 1 line of snippet have missed this entirely).

The culprit being that JavaScript's `String.split` function doesn't work like most folks expect it to when it comes to the limit: instead of limiting the amount of cuts that will be made in the source (like in java) and having the last slice be "the rest", JavaScript cuts on all separators, and the limit restricts the amount of slices returned.

In other words:
```java
// Example.java
final String abc = "a,b,c";
abc.split(",", 2);
//=> ["a", "b,c"]
```

Versus
```js
// Example.js
const abc = 'a,b,c';
abc.split(',', 2);
//=> ['a', 'b'] // Omitting slice #3, 'c'.
```

packages/jsii-rosetta/test/jsii/fake-assembly.ts
```diff
@@ -0,0 +1,17 @@
+import { Assembly, SchemaVersion } from '@jsii/spec';
```
RomainMuller, line 1: This was moved from `jsii/assemblies.test.ts`, because `rosetta.test.ts` imports this function; but the importing actually caused all tests in `jsii/assemblies.test.ts` to be run twice as a collateral.

rix0rrr (approved):

PR #2226, opened 2020-11-04
chore(rosetta): de-duplicate diagnostics early

RomainMuller:
Making sure diagnostics are always sorted and de-duplciated reduces
the amount of overhead pertaining to keeping diagnostics in memory
and could result in better overall performance (due to lower memory
pressure)

---


packages/jsii-rosetta/lib/translate.ts
```diff
@@ -44,7 +44,8 @@ export function translateTypeScript(
  */
 export class Translator {
   private readonly compiler = new TypeScriptCompiler();
-  public readonly diagnostics: ts.Diagnostic[] = [];
+  // eslint-disable-next-line @typescript-eslint/explicit-member-accessibility
```
nija-at, line 47: Oh gosh, typescript eslint is not catching up with newer typescript features

packages/jsii-rosetta/lib/translate.ts
```diff
@@ -107,7 +114,7 @@ export interface SnippetTranslatorOptions extends AstRendererOptions {
 
 export interface TranslateResult {
   translation: string;
-  diagnostics: ts.Diagnostic[];
+  diagnostics: readonly ts.Diagnostic[];
```
nija-at, line 117: What does it mean to have `readonly` after the `:`?

Also mark `translation` as `readonly`?

nija-at (approved):

packages/jsii-rosetta/lib/translate.ts
RomainMuller: The culprit appears to be that Babel does not support this just yet 🤷🏻‍♂️ or not in a stable way (couldn't completely understand what I read on the subject)

packages/jsii-rosetta/lib/translate.ts
RomainMuller: `readonly` here is a modifier applied to the array type... Actually means I promise (under the type checker's supervision) I will not modify the array provided as input.

PR #2229, opened 2020-11-04
chore: enable additional eslint rules

RomainMuller:
Enabled the following rules as they typically catch programming errors
that are susceptible to be the causes of difficult to diagnose bugs or
performance issues:

- `@typescript-eslint/no-misused-promises` helps ensure not-`await`ed
  promises are not used in conditions, etc... as those are typically not
  the intention of the developer and constitutes a bug that is difficult
  to identify when just reading the code (i.e: likely missed in code
  reviews).
- `@typescript-eslint/prefer-nullish-coalescing` requires consistent use
  of the `??` operator instead of `||` for nullish-coalescing, as this
  operator is safer in this context (`||` is falsey-coalescing, which
  has a broader reach).
- `@typescript-eslint/return-await` ensures consistency in how promises
  are returned from `async` functions, in particular making sure `await`
  is present when a promise is returned from a *try-catch-finally*
  block, as failure to `await` in this context is usually a bug.
- `@typescript-eslint/switch-exhaustiveness-check` requires all `switch`
  are exhaustive, making sure no value is silently ignored, which often
  is the source of silent bugs that are difficult to root cause.

One of the code changes made branch coverage in `jsii-reflect` drop
below the threshold (it was _just above_ and fell _just below_), so
instead of lowering the threshold I added new tests to improve the
coverage of the library.

---


packages/@jsii/kernel/test/kernel.test.ts
```diff
@@ -95,7 +95,7 @@ defineTest.skip = function (
       });
       expect(result.result).toBe(0o644);
 
-      return closeRecording(kernel);
+      return await closeRecording(kernel);
```
RomainMuller, line 98: There was a bug right there that this change allows to prevent! Thankfully, this bug was in a test so it did not get to cause actual harm.

packages/jsii-reflect/test/__snapshots__/tree.test.ts.snap
```diff
@@ -0,0 +1,3178 @@
...
+ │   │   │ └── [4mreturns[24m: void
+ │   │   ├─┬ for() [90mmethod[39m
+ │   │   │ └── [4mreturns[24m: void
+ │   │   ├─┬ from() [90mmethod[39m
+ │   │   │ └── [4mreturns[24m: void
+ │   │   ├─┬ global() [90mmethod[39m
+ │   │   │ └── [4mreturns[24m: void
+ │   │   ├─┬ if() [90mmethod[39m
+ │   │   │ └── [4mreturns[24m: void
+ │   │   ├─┬ import() [90mmethod[39m
```
nija-at, line 1615: Ahhhh.... what's all this being sneaked in? 😉

RomainMuller: Oh it's improved test coverage... Some of the code tweaks (to be safer) caused the branch coverage to go down a little... And this basically improves coverage of previously only uncovered code.

nija-at (approved):

packages/jsii-reflect/test/__snapshots__/jsii-tree.test.ts.snap
```diff
@@ -1712,10 +1712,10 @@ exports[`jsii-tree --all 1`] = `
  │   │   ├── <initializer>() initializer (stable)
  │   │   └─┬ youSeeMe property (stable)
  │   │     └── type: string
- │   ├─┬ class StructPassing
+ │   ├─┬ class StructPassing (external)
```
RomainMuller, line 1715: Example of a "fixed bug" (really a feature gap) caused by a non-exhaustive `switch`.

PR #2230, opened 2020-11-04
feat(rosetta): support computed properties

RomainMuller:
When in the context of a dictionary, computed properties are perfectly valid,
and they can be supported relatively easily.

Additionally, re-wrote how templated strings are converted to Java in order
to make the transform better Java through use of String.format instead of
resorting to String concatenation.

---


packages/jsii-rosetta/test/translations/expressions/double_quoted_dict_keys.java
```diff
@@ -1 +1 @@
-Map<String, String> x = Map.of("key", "value");
+final Map<String, String> x = Map.of("key", "value");
```
rix0rrr, line 1: I'm not sure that I agree with this change.

[Yes, it's technically more correct to translate `const` to `final`.](https://user-images.githubusercontent.com/524162/98555497-4bff4900-22a2-11eb-9fa3-59706593f796.png)


But, the goal of this tool is not to produce perfect translations -- the goal is to produce acceptable sample code, and the primary purpose of example code is that it is easy to read.

I find the additional `final` to be distracting, when what we're really trying to get consumers to understand is that they're declaring a variable.

rix0rrr (changes requested):

packages/jsii-rosetta/test/translations/expressions/double_quoted_dict_keys.java
RomainMuller: Yeah you've got a point right here....

rix0rrr (approved):

PR #2233, opened 2020-11-06
feat(jsii): configure diagnostics in package.json

nija-at:
Diagnostic codes and categories already exist in JSII. This change adds
configurability via `package.json`.

---


packages/jsii/lib/jsii-diagnostic.ts
```diff
@@ -816,3 +816,42 @@ export class JsiiDiagnostic implements ts.Diagnostic {
...
+    case 'error':
+      return ts.DiagnosticCategory.Error;
+    case 'warning':
+      return ts.DiagnosticCategory.Warning;
+    case 'suggestion':
+      return ts.DiagnosticCategory.Suggestion;
+    case 'message':
+      return ts.DiagnosticCategory.Message;
+    default:
+      throw new Error(`Unrecognized diagnostic category '${str}'`);
```
RomainMuller, line 842: Would be lovely if this can also refer to the key this unrecognized diagnostic was associated to (sometimes you fail to see the typo for wayyyyyyy too long).

packages/jsii/lib/jsii-diagnostic.ts
```diff
@@ -816,3 +816,42 @@ export class JsiiDiagnostic implements ts.Diagnostic {
...
+function diagnosticCode(str: string): string | number {
+  if (str.toLowerCase().startsWith('jsii')) {
+    const num = +str.slice(4);
+    if (Number.isNaN(num)) {
+      throw new Error(
+        `Invalid diagnostic code ${str}. A number must follow code that starts with 'JSII'`,
+      );
+    }
+    return num;
+  }
```
RomainMuller, line 855: Ooof! This looks a little too brittle IMHO.

```suggestion
  const m = /^JSII(\d+)$/i.exec(str);
  if (m) {
    return parseInt(m[1], 10);
  }
```

In case where `JSII` is not followed by digits (where it might end up NAN, but could also have been a float per your test), I'd just use it up as a string, and let that fail.

packages/jsii/lib/project-info.ts
```diff
@@ -44,6 +44,7 @@ export interface ProjectInfo {
   readonly targets: spec.AssemblyTargets;
   readonly metadata?: { [key: string]: any };
   readonly jsiiVersionFormat: 'short' | 'full';
+  readonly diagnostics?: { readonly [code: string]: string };
```
RomainMuller, line 47: The validation of the values for this map could be done in here (same as we validate the other things - it'd actually be more consistent).

Then you could have:
```suggestion
  readonly diagnostics?: { readonly [code: string]: ts.DiagnosticCategory };
```

RomainMuller (approved):

RomainMuller:
@mergifyio refresh

PR #2238, opened 2020-11-09
fix(runtime): "Error: EOF: end of file, read" on Windows

RomainMuller:
NodeJS makes it unsafe to perform syncrhonous (aka blocking) operations
on file descriptors 0 through 2 (STDIN, STDOUT, STDERR), as those are
set to `O_NONBLOCK` by `libuv`. This is particularly problematic on
Windows where there is no easy way to re-open those in blocking mode.

Consequently, we must handle some quirks in cases where blocking
attempts result in conditions typical of non-blocking access: `EAGAIN`
errors must be caught and retried, and `EOF` errors must be "ignored".

This PR adds the necessary code to correctly handle the `EOF` error, and
adds a 50ms sleep before re-trying a `EAGAIN` error that is achieved
thanks to a neat trick suggested by the [`sleep` npm package][sleep]:
using `Atomics.wait` enables syncrhonous sleep to be done in `node`
without having to ship a native module.

[sleep]: https://www.npmjs.com/package/sleep

---


packages/@jsii/runtime/lib/program.ts
```diff
@@ -1,3 +1,5 @@
+import * as process from 'process';
```
RomainMuller, line 1: Added as part of my rumblings here... `process.env` is used here, and I kinda like to standardize on importing this even though it's implicit.

.github/workflows/main.yml
```diff
@@ -217,6 +217,7 @@ jobs:
             python: '3.6'
           # Test alternate .NETs
           - java: '8'
+            # Current release info can be found at https://dotnet.microsoft.com/download/dotnet/5.0
```
RomainMuller, line 220: Not strictly related to this PR, but this can be helpful in the future. This is kinda trivial, too. This got added as I was trying to find a way to reliably cause an EOF to happen on Windows in CI/CD. Spoiler: I did not succeed, for some reason.

packages/@jsii/runtime/lib/sync-stdio.ts
```diff
@@ -65,3 +65,58 @@ export class SyncStdio {
...
+ */
+function readSync(
+  fd: number,
+  buffer: Buffer,
+  offset: number,
+  length: number,
+  position: number | null = null,
+): number {
+  // eslint-disable-next-line no-constant-condition
+  while (true) {
```
RomainMuller, line 90: I used a `while` loop instead of recursively re-entering, as this avoids stacking up in case `EAGAIN` happens *a lot*, which I imagine might be somewhat frequent on resource-constrained systems (where we'd likely also have less memory available for wasting on stack frames).

packages/@jsii/runtime/lib/sync-stdio.ts
```diff
@@ -61,6 +48,62 @@ export class SyncStdio {
...
+    } catch (error) {
+      switch (error.code) {
+        // HACK: node may set O_NONBLOCK on it's STDIN depending on what kind of input it is made
+        // of (see https://github.com/nodejs/help/issues/2663). When STDIN has O_NONBLOCK, calls may
+        // result in EAGAIN. In such cases, the call should be retried until it succeeds. This kind
+        // of polling will result in horrible CPU thrashing, but there does not seem to be a way to
+        // force a O_SYNC access to STDIN in a reliable way within node.
+        // In order to make this stop we need to either stop depending on synchronous reads, or to
+        // provision our own communication channel that can reliably be synchronous. This work is
+        // "tracked" at https://github.com/aws/aws-cdk/issues/5187
```
njlynch, line 90: I see this was copy/pasted; is it still relevant? The referenced issue is closed. Do we have an active/open issue, and/or `jsii` issue to "track" this?

packages/@jsii/runtime/lib/sync-stdio.ts
```diff
@@ -1,14 +1,15 @@
 import * as fs from 'fs';
 
-const STDIN_FD = 0;
-const STDOUT_FD = 1;
-const STDERR_FD = 2;
-const INPUT_BUFFER_SIZE = 1024 * 1024; // not related to max line length
+import { sleep } from './sleep';
 
-const EMPTY_BUFFER = Buffer.alloc(0);
+const STDIN_FD = (process.stdin as any).fd ?? 0;
```
njlynch, line 5: When will these (not) be set? Might deserve a comment.

packages/@jsii/runtime/lib/sleep.ts
```diff
@@ -0,0 +1,14 @@
+// We need a shared buffer array for the purpose of this exercise.
+const array = new Int32Array(new SharedArrayBuffer(4));
+
+/**
+ * Sleeps for a set amount of time, syncrhonously.
```
njlynch, line 5:
```suggestion
 * Sleeps for a set amount of time, synchronously.
```

packages/@jsii/runtime/lib/sync-stdio.ts
njlynch: This might be a nice comment to have in source, rather than just on the PR.

njlynch (commented):
All of my comments are about comments. :)

Also, are there any tests for this section of the code?

njlynch (approved):

PR #2247, opened 2020-11-12
feat(dotnet,java): finalize kernel process on VM shutdown

RomainMuller:
This change adds shutdown hook registrations in the .NET and Java
runtimes so that the `node` process hosting the JSII runtime is
gracefully terminated upon program end. This allows any "on exit" hooks
registered in the JavaScript process to properly run when the host
program ends. This is similar to what was done in #2100 for Python.

---

### DIFF
--- a/packages/@jsii/java-runtime/project/src/main/java/software/amazon/jsii/JsiiRuntime.java
+++ b/packages/@jsii/java-runtime/project/src/main/java/software/amazon/jsii/JsiiRuntime.java
@@ -15,7 +15,6 @@ import java.lang.reflect.InvocationTargetException;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
 import static software.amazon.jsii.JsiiVersion.JSII_RUNTIME_VERSION;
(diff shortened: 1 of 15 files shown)


eladb (commented):
I would recommend separating the two changes (STDERR and EXIT handling).

RomainMuller:
Extracted STDERR-related changes to https://github.com/aws/jsii/pull/2248

packages/@jsii/java-runtime/project/src/main/java/software/amazon/jsii/JsiiRuntime.java
```diff
@@ -178,33 +184,48 @@ protected void finalize() throws Throwable {
...
+            if (stdin != null) {
+                stdin.close();
+                stdin = null;
+            }
 
-        if (stdout != null) {
-            stdout.close();
-            stdout = null;
-        }
+            if (childProcess != null) {
```
eladb, line 194: I feel like I am missing something... Who sends the termination signal to the child process? It seems like we are just waiting for 5 seconds and then killing it?

RomainMuller: The jsii Kernel interprets EOF on it's `STDIN` as a signal it should shut down, and obliges. We give it 5 seconds to clean up after itself, then dispose of it forcibly if it's still not done.

eladb: Worth adding a comment in the relevant places because it's not obvious from the code

eladb (approved):

PR #2248, opened 2020-11-12
feat(dotnet,java): kernel process inherits host's STDERR

RomainMuller:
Removed the code that surrounded handling of the `STDERR` stream from
the Kernel process: there is no need to PIPE it, and simply inheriting
it (i.e: subprocess `STDERR` is the same stream as the parent's) results
in a more intuitive behavior, as `STDERR` output from child processes
surfaces "immediately" even if the parent process somehow cannot
forward it anymore (e.g: if it crashed or deadlocked). This also means
the `STDERR` forwarding thread is no longer needed.

---


packages/@jsii/dotnet-runtime/src/Amazon.JSII.Runtime/Services/NodeProcess.cs
```diff
@@ -38,8 +38,7 @@ public NodeProcess(IJsiiRuntimeProvider jsiiRuntimeProvider, ILoggerFactory logg
                     StandardInputEncoding = utf8,
                     RedirectStandardOutput = true,
                     StandardOutputEncoding = utf8,
-                    RedirectStandardError = true,
-                    StandardErrorEncoding = utf8
+                    UseShellExecute = false,
```
eladb, line 41: I think it's a good change but usually has implications on being able to locate executables. Have you tested this on windows?

packages/@jsii/dotnet-runtime/src/Amazon.JSII.Runtime/Services/NodeProcess.cs
```diff
@@ -58,18 +57,23 @@ public NodeProcess(IJsiiRuntimeProvider jsiiRuntimeProvider, ILoggerFactory logg
...
+
         public TextWriter StandardInput => _process.StandardInput;
 
         public TextReader StandardOutput => _process.StandardOutput;
 
-        public TextReader StandardError => _process.StandardError;
-
         void IDisposable.Dispose()
         {
-            StandardInput.Dispose();
```
eladb, line 69: Why are these not needed anymore ?

packages/@jsii/dotnet-runtime/src/Amazon.JSII.Runtime/Services/Runtime.cs
```diff
@@ -10,19 +9,14 @@ internal sealed class Runtime : IRuntime
         public Runtime(INodeProcess nodeProcess)
         {
             _nodeProcess = nodeProcess ?? throw new ArgumentNullException(nameof(nodeProcess));
-            if (Environment.GetEnvironmentVariable("JSII_DEBUG") != null)
```
eladb, line 13: What functionality do we lose here?

eladb (commented):
Just to make sure I understand: the child process respects JSII_DEBUG and will only output traces to stderr in that case. So it's safe to just INHERIT stderr, correct?

packages/@jsii/dotnet-runtime/src/Amazon.JSII.Runtime/Services/NodeProcess.cs
RomainMuller: * According to [.NET's own documentation](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.processstartinfo.redirectstandardoutput?view=net-5.0#remarks), `UseShellExecute = false` is **required** if you set any `Redirect*` to `true`.
* The *default* value of this property is `false` on `.NET Core` (but `true` on `.NET Framework`). This is only making it explicit

packages/@jsii/dotnet-runtime/src/Amazon.JSII.Runtime/Services/NodeProcess.cs
RomainMuller: `StandardInput`, `StandardOutput` and `StandardError` are actually directly obtained from properties of the `Process` instance... When we call `Dispose()` on the process, it naturally disposes of all resources it owns, including those streams.

Additionally, `Dispose()` might abruptly terminate the stream, possibly without having flushed some buffered data. Calling `Close()` instead tends to ensure the buffer is flushed out before the close happens, so the `node` process gets the signal we are trying to send.

packages/@jsii/dotnet-runtime/src/Amazon.JSII.Runtime/Services/Runtime.cs
RomainMuller: Previously, `STDERR` from anywhere in `node` (i.e: `console.error` from user's apps) were only surfaced to the console of non-JS applications when `JSII_DEBUG` was set to `true`. This means the behavior of a program was *different* if you ran it "pure JavaScript" versus through a jsii binding. Hence I think this was actually a bug and not a feature.

eladb (approved):

PR #2252, opened 2020-11-16
feat(rosetta): make assembly validation opt-in

RomainMuller:
Assembly validation (i.e: checking that the content of the `.jsii` file
correctly implements the `@jsii/spec` schema) can be slow and memory
intensive. It is however typically not of great use, since we should
normally be able to assume assemblies are valid.

Added a `--validate-assemblies` option (default `false`) to control
whether the JSON Schema validation is done or not.

---


packages/jsii-rosetta/lib/jsii/assemblies.ts
```diff
@@ -19,28 +19,32 @@ export interface LoadedAssembly {
...
-        assembly: await loadAssemblyFromFile(loc), // eslint-disable-line no-await-in-loop
-        directory: path.dirname(loc),
-      });
+      return loadAssembly(path.join(location, '.jsii'));
     }
+    return {
+      assembly: await loadAssemblyFromFile(location, validateAssemblies),
+      directory: path.dirname(location),
+    };
   }
```
RomainMuller, line 37: Re-worked this code largely to remove the linter exclusions... It does the same work, except in parallel instead of serially, which could be a little faster, since this involves I/O wait.

MrArnoldPalmer (approved):

PR #2253, opened 2020-11-16
feat(rosetta): support "strict" assemblies

RomainMuller:
A "strict" assembly carries a special metadata entry
(`jsii.rosetta.strict = true`) that instructs `jsii-rosetta` to always
handle that as if `--compile` and `--fail` were provided. This allows
opting specific modules from a mono-repository into "enforced compiling"
operation while retaining the ability to run `jsii-rosetta` only once
from the root of the repository. Concretely this means one can validate
examples from a single package are valid by simply running `jsii-rosetta`
(strict mode will imply `--compile --fail`), while a continuous integration
pipeline can continue to use `jsii-rosetta` on a whole mono-repo in a single
command (which is typically a lot faster).

Also added a `--strict` option to `jsii-rosetta` in order to have
consistent semantics here. And added a sub-command of `jsii-rosetta`
to actually set the metadata entry correctly for the user.

---


nija-at:
> This allows opting specific modules from a mono-repository into "enforced compiling"
> operation while retaining the ability to run `jsii-rosetta` only once
> from the root of the repository.

Can't seem to follow what you're saying here. Can you re-state simply?

packages/jsii-rosetta/README.md
```diff
@@ -24,6 +24,32 @@ If the samples don't compile or don't have full type information:
...
+        }
+      }
+    }
+    // ...
+  },
+  // ...
+}
+```
+
+This effectively enables the `--strict` option when translating code samples from the assembly.
```
nija-at, line 51: Explain what the `--strict` option is.

packages/jsii-pacmak/lib/targets/dotnet/dotnetdocgenerator.ts
```diff
@@ -156,7 +161,11 @@ export class DotNetDocGenerator {
   }
 
   private convertExample(example: string): string {
-    const snippet = typeScriptSnippetFromSource(example, 'example');
+    const snippet = typeScriptSnippetFromSource(
+      example,
+      'example',
+      enforcesStrictMode(this.assembly),
```
nija-at, line 167: Why have you chosen for the DocGenerator classes to now depend on the jsii assembly? Is this intentional?

Can be avoided if the `enforceStrictAssembly` is called at the layer above. This will also avoid code duplication across the different language targets.

packages/jsii-rosetta/bin/jsii-rosetta.ts
```diff
@@ -114,8 +114,24 @@ function main() {
...
           }),
       wrapHandler(async (args) => {
+        // `--strict` is short for `--compile --fail`, and we'll override those even if they're set to `false`, such as
+        // using `--no-(compile|fail)`, because yargs does not quite give us a better option that does not hurt CX.
+        if (args.strict) {
+          args.compile = args.c = true;
+          args.fail = args.f = true;
+        }
+
+        console.error(args);
```
rix0rrr, line 133: Oh no!

packages/jsii-rosetta/bin/jsii-rosetta.ts
```diff
@@ -144,7 +160,11 @@ function main() {
           );
         }
 
-        if (result.diagnostics.some(isErrorDiagnostic) && args.fail) {
+        if (
+          result.diagnostics.some((diag) =>
+            isErrorDiagnostic(diag, { onlyStrict: !args.fail }),
+          )
+        ) {
           process.exit(1);
```
rix0rrr, line 168: Not your change, but since you need to do a pass anyway:

This needs to be:

```suggestion
          process.exitCode = 1;
```

RomainMuller:
> > This allows opting specific modules from a mono-repository into "enforced compiling"
> > operation while retaining the ability to run `jsii-rosetta` only once
> > from the root of the repository.
> 
> Can't seem to follow what you're saying here. Can you re-state simply?
> 
> Do you mean the samples will always be compiled as part of `jsii`?

I've attempted rephrasing (actually added more text).
We still don't run Rosetta as part of `jsii`, no...

packages/jsii-rosetta/bin/jsii-rosetta.ts
```diff
@@ -172,6 +192,41 @@ function main() {
...
+      wrapHandler(async (args) => {
+        const packageJsonPath = (await fs.stat(args.PACKAGE)).isDirectory()
+          ? path.join(args.PACKAGE, 'package.json')
+          : args.PACKAGE;
+        const packageJson = await fs.readJson(packageJsonPath);
+        if (packageJson.jsii == null) {
+          console.error(
+            `The package in ${args.PACKAGE} does not have a jsii configuration! You can set it up using jsii-config.`,
+          );
+          return process.exit(1);
```
rix0rrr, line 215: Same note about the exitCode

packages/jsii-rosetta/bin/jsii-rosetta.ts
```diff
@@ -172,6 +192,41 @@ function main() {
         await readTablet(args.TABLET, args.KEY, args.LANGUAGE);
       }),
     )
+    .command(
+      'configure-strict [PACKAGE]',
```
rix0rrr, line 196: I question the need for this command.

But oh well I guess?

packages/jsii-rosetta/lib/commands/convert.ts
```diff
@@ -10,24 +10,29 @@ export interface TranslateMarkdownOptions extends AstRendererOptions {
...
+   * Whether to operate in `strict` mode or not.
+   */
+  strict?: boolean;
 }
 
 export function translateMarkdown(
   markdown: File,
   visitor: AstHandler<any>,
-  options: TranslateMarkdownOptions = {},
+  { languageIdentifier = '', strict = false }: TranslateMarkdownOptions = {},
```
rix0rrr, line 23: I personally think this is futt-bugly. But to each their own

packages/jsii-rosetta/lib/translate.ts
```diff
@@ -145,14 +145,19 @@ export class SnippetTranslator {
     this.visibleSpans = calculateVisibleSpans(source);
 
     // This makes it about 5x slower, so only do it on demand
-    if (options.includeCompilerDiagnostics) {
+    if (options.includeCompilerDiagnostics || snippet.strict) {
```
rix0rrr, line 148: I thought `strict` implied `includeCompilerDiagnostics`? So this OR is unnecessary?

rix0rrr (approved):
Small changes, but overall okay.

I'd like to implore you to see if this can be arranged differently from pushing down the `strict` parameter all the way into the bowels of the code (can't remember offhand what a better place would be to make the cut though).

If not, then I suppose this needs to be it...

nija-at (commented):
Couple of questions that aren't clear from the code -

If the sample uses a package/code that is not declared as a dependency of this package (either because it's a downstream dependency or a utility package that a customer can choose to use), how does the `--strict` option check this?

> Concretely this means one can validate
> examples from a single package are valid by simply running jsii-rosetta
> (strict mode will imply --compile --fail), while a continuous integration
> pipeline can continue to use jsii-rosetta on a whole mono-repo in a single
> command (which is typically a lot faster).

If I'm interepreting this correctly, the workflow you're suggesting will be that developers can choose to run `jsii-rosetta` on the individual package (in a monorepo) during dev time but it won't be run as part of CI. Instead, the CI system will run it in a single command for the whole monorepo. Did I get that correctly?
Will the single command also interpret the `jsii.rosetta.strict` and fail for packages that have this flag set and don't compile?

packages/jsii-rosetta/lib/translate.ts
RomainMuller: `yargs` gives me no way to encode the `implies` relationship. Their "implies" feature is more like "if you set X, you MUST also set anything implied by X", which is not what I want here...

And then, if I go "manually" mixup the args right next to the entry point, then I also need to figure out all the aliases for that, possibly for a variety of different case translations... I don't quite fancy this.

packages/jsii-pacmak/lib/targets/dotnet/dotnetdocgenerator.ts
RomainMuller: `jsii-pacmak` right now uses a single Rosetta instance for all packages it generates (there could be multiple from `--recurse`). Creating multiple instances of Rosetta means creating multiple TypeScript contexts, which I decided against.

RomainMuller:
> If the sample uses a package/code that is not declared as a dependency of this package (either because it's a downstream dependency or a utility package that a customer can choose to use), how does the --strict option check this?

Not certain I understand the question well, but here's some answers nevertheless:
- The `strict` enabling metadata is recorded on the jsii Assembly. Whenever `jsii-rosetta` translates samples from a repository with that metadata, it will be compiled, and `jsii-rosetta` will fail on compilation errors (as opposed to failing only on attempting to transliterate things it does not support)
- Any dependencies required in order for the compilation to succeed **must** be present and available at the time `jsii-rosetta` is run... This in particular is why it's not possible to perform the "compile check" at time of package compilation - the code examples may depend on packages that are not yet built... Consider the case of `@aws-cdk/core` - it's examples heavily use other libraries such as `@aws-cdk/aws-lambda`, etc... because they are otherwise not really helpful. As a consequence, a "local iteration workflow" would be something along the lines of:
  1. Compile current package & any package needed by code samples... (This could mean compiling a whole monorepo, to simplify)
  2. Edit the documentation you're looking to iterate on
  3. Re-compile (run `jsii` on the package where edits were made)
  4. Run `jsii-rosetta` to validate code samples are all compiling
  5. Goto step 2 ad libitum (until you're happy about the result)
  6. Make pull request

nija-at (approved):

RomainMuller:
@mergifyio refresh

Issue #2047, opened 2020-09-26
Go Runtime Jsii to Native Value Casting


MrArnoldPalmer:
## :rocket: Feature Request

Currently the go runtime loads jsii modules correctly and submits kernel request for object creation, method invocation, and property access. However, it is not Decoding the kernel responses into the corresponding Go data structures when appropriate and dummy types are still being returned.

This issue tracks the handling of `CreateResponse`, `InvokeResponse`, and `GetResponse` in generated go code. This means that the runtime needs methods for handling data structures embedded within these kernel responses, in addition to objects passed by reference via `ObjRef`, and turning them into concrete data structures that the user can treat as normal go structs/enums/etc.

A couple of thoughts on this:
- Does it make sense to tag generated go struct properties with their corresponding json keys, similar to the structs in the runtime api? This would give them a default mapping between objects returned by value over the wire and their go structure and we could just `json.Unmarshal` them?
- Getter implementations for structs should check for the presence of `jsiiObjInstanceId` and if available, always call out to get the value from the kernel api. Local property reference values should only be returned in getters when the struct isn't instantiated within the kernel runtime. Consider what it takes to keep local go struct property values in sync with the corresponding instance within the kernel runtime.

PR #2266, opened 2020-11-19
feat(go): runtime type casting

MrArnoldPalmer:
Adds an implementation map to all code generated getters and method
invocations to pass to the runtime. This allows the runtime to construct
concrete types when the return type of the function is a go interface.

Adds additional tests in `jsii-calc-test` module for various use cases.
Adds logic to correctly deserialize and serialize object references
coming from both the host and kernel.

fix: #2047 

---

### DIFF
--- a/packages/@jsii/go-runtime/jsii-calc-test/main_test.go
+++ b/packages/@jsii/go-runtime/jsii-calc-test/main_test.go
@@ -1,12 +1,15 @@
 package main
 
 import (
-	calc "github.com/aws-cdk/jsii/jsii-calc/go/jsiicalc"
-	"github.com/aws-cdk/jsii/jsii-experimental"
+	"fmt"
 	"math"
 	"os"
+	"reflect"
 	"strings"
 	"testing"
+	calc "github.com/aws-cdk/jsii/jsii-calc/go/jsiicalc"
+	calclib "github.com/aws-cdk/jsii/jsii-calc/go/scopejsiicalclib"
+	"github.com/aws-cdk/jsii/jsii-experimental"
 )
 
 func TestMain(m *testing.M) {
@@ -15,42 +18,80 @@ func TestMain(m *testing.M) {
 	os.Exit(status)
 }
 
-func TestCalculator(t *testing.T) {
-	var initialValue float64 = 10
-	calculatorProps := calc.CalculatorProps{InitialValue: initialValue, MaximumValue: math.MaxFloat64}
-	calculator := calc.NewCalculator(&calculatorProps)
+func initCalculator() calc.CalculatorIface {
+	calculatorProps := calc.CalculatorProps{InitialValue: float64(10), MaximumValue: math.MaxFloat64}
+	return calc.NewCalculator(&calculatorProps)
+}
 
-	t.Run("Primitive Property Access", func(t *testing.T) {
-		actual := calculator.GetValue()
-		if actual != initialValue {
-			t.Errorf("Expected: %f; Actual %f;", initialValue, actual)
+func TestCalculator(t *testing.T) {
+	// Object creation
+	t.Run("Object creation", func(t *testing.T) {
+		calculator := initCalculator()
+		if reflect.ValueOf(calculator).IsZero() {
+			t.Errorf("Expected calculator object to be valid")
 		}
 	})
 
-	t.Run("Method Call Effect", func(t *testing.T) {
-		var factor float64 = 3
-		calculator.Mul(factor)
-		var expected float64 = initialValue * factor
+	t.Run("Property access", func(t *testing.T) {
+		calculator := initCalculator()
+		expected := float64(10)
 		actual := calculator.GetValue()
-
 		if actual != expected {
 			t.Errorf("Expected: %f; Actual %f;", expected, actual)
 		}
 	})
 
-	t.Run("Method Call Returns Any Value", func(t *testing.T) {
+	t.Run("Property mutation", func(t *testing.T) {
+		calculator := initCalculator()
+		var newVal float64 = 12345
+		currentProps := calclib.NewNumber(newVal)
+		calculator.SetCurr(currentProps)
+		actual := calculator.GetValue()
+		if newVal != actual {
+			t.Errorf("Expected: %f; Actual %f;", newVal, actual)
+		}
+	})
+
+	t.Run("Method with side effect", func(t *testing.T) {
+		initial, factor := float64(10), float64(3)
+		calculator := calc.NewCalculator(&calc.CalculatorProps{InitialValue: initial, MaximumValue: math.MaxFloat64})
+		calculator.Mul(factor)
+		expectedProduct := initial * factor
+		actualProduct := calculator.GetValue()
+		if actualProduct != expectedProduct {
+			t.Errorf("Expected quotient: %f; Actual %f;", expectedProduct, actualProduct)
+		}
+	})
+
+	t.Run("Method with interface{} return type", func(t *testing.T) {
+		calculator := initCalculator()
+		expectedTypeName := "Calculator"
+		actualTypeName := calculator.TypeName()
 		// JSII tells us this return value is an "any" type. Therefore the
 		// value received by go is type `interface{}` and can be further
 		// specialized using reflection.
-		expected := "Calculator"
-		actual := calculator.TypeName()
-		switch retType := actual.(type) {
+		switch retType := actualTypeName.(type) {
 		case string:
-			if actual != expected {
-				t.Errorf("Expected: %s; Actual %s", expected, actual)
+			if actualTypeName != expectedTypeName {
+				t.Errorf("Expected type name: %s; Actual %s", expectedTypeName, actualTypeName)
 			}
 		default:
 			t.Errorf("Expected type: %s; Actual type: %s", "string", retType)
+		}
+	})
+
+	t.Run("Method with args and string return type", func(t *testing.T) {
+		calculator := initCalculator()
+		lhs, rhs := 10, 3
+		calculator.SetCurr(calc.NewMultiply(
+			calclib.NewNumber(float64(lhs)),
+			calclib.NewNumber(float64(rhs)),
+		))
+		// expectedString := "(10 * 3)"
+		expectedString := fmt.Sprintf("(%d * %d)", lhs, rhs)
+		actualString := calculator.ToString()
+		if actualString != expectedString {
+			t.Errorf("Expected string: %s; Actual %s;", expectedString, actualString)
 		}
 	})
 }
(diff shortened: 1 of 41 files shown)


SoManyHs (commented):
Looks good to me, some non-blocking comments on the tests but otherwise 👍 

Would be great if @RomainMuller could take a gander as well!

packages/jsii-pacmak/lib/targets/go/package.ts
```diff
@@ -87,6 +87,14 @@ export abstract class Package {
...
   }
 
+  public get importPath(): string {
+    const moduleName = this.root.moduleName;
+    const prefix = moduleName !== '' ? `${moduleName}/` : '';
+    const rootPackageName = this.root.packageName;
+    const suffix = this.filePath !== '' ? `/${this.filePath}` : '';
+    return `${prefix}${rootPackageName}${suffix}`;
+  }
+
```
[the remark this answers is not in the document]
MrArnoldPalmer: Will defer to your method once its merged.

packages/@jsii/go-runtime/jsii-calc-test/main_test.go
```diff
@@ -14,64 +18,137 @@ func TestMain(m *testing.M) {
...
+		if newVal != actual {
+			t.Errorf("Expected: %f; Actual %f;", newVal, actual)
+		}
+	})
 
-		if actual != expected {
-			t.Errorf("Expected: %f; Actual %f;", expected, actual)
+	t.Run("Method with side effect", func(t *testing.T) {
+		initial, factor := float64(10), float64(3)
+		calculator := calc.NewCalculator(&calc.CalculatorProps{InitialValue: initial, MaximumValue: math.MaxFloat64})
```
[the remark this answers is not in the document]
MrArnoldPalmer: definitely can/should/will be.

packages/@jsii/go-runtime/jsii-calc-test/main_test.go
```diff
@@ -14,64 +18,137 @@ func TestMain(m *testing.M) {
...
+		calculator.Mul(factor)
+		expectedProduct := initial * factor
+		actualProduct := calculator.GetValue()
+		if actualProduct != expectedProduct {
+			t.Errorf("Expected quotient: %f; Actual %f;", expectedProduct, actualProduct)
 		}
 	})
 
-	t.Run("Method Call Returns Any Value", func(t *testing.T) {
+	t.Run("Method with interface{} return type", func(t *testing.T) {
```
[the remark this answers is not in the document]
MrArnoldPalmer: We aren't necessarily asserting that. but that is the return type of the method so thats the type that is usable to the user. The test is mostly 'Returns an interface{} but actually has a string that can be extracted via reflection'. I'll change the description here.

packages/@jsii/go-runtime/jsii-calc-test/main_test.go
```diff
@@ -14,64 +18,137 @@ func TestMain(m *testing.M) {
...
+		calculator := initCalculator()
+		expectedTypeName := "Calculator"
+		actualTypeName := calculator.TypeName()
 		// JSII tells us this return value is an "any" type. Therefore the
 		// value received by go is type `interface{}` and can be further
 		// specialized using reflection.
-		expected := "Calculator"
-		actual := calculator.TypeName()
-		switch retType := actual.(type) {
+		switch retType := actualTypeName.(type) {
```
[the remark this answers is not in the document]
MrArnoldPalmer: Yeah these are funky because as far as I can tell its the only way to do a [type matching](https://tour.golang.org/methods/16) without reflection. It seems to be the preferred way of handling this.

packages/@jsii/go-runtime/jsii-calc-test/main_test.go
```diff
@@ -14,64 +18,137 @@ func TestMain(m *testing.M) {
 	os.Exit(status)
 }
 
+func initCalculator() calc.CalculatorIface {
+	calculatorProps := calc.CalculatorProps{InitialValue: float64(10), MaximumValue: math.MaxFloat64}
+	return calc.NewCalculator(&calculatorProps)
+}
+
 func TestCalculator(t *testing.T) {
```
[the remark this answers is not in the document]
MrArnoldPalmer: ooh I like that

Will work on this separately.

packages/@jsii/go-runtime/jsii-experimental/runtime.go
```diff
@@ -8,6 +8,8 @@ import (
 	"regexp"
 )
 
+type implementationMap = map[reflect.Type]reflect.Type
```
RomainMuller, line 11: That could have used a documentation string explaining what the key versus value is supposed to be.

packages/jsii-pacmak/lib/targets/go/runtime/method-call.ts
```diff
@@ -62,6 +70,13 @@ export class MethodCall {
     );
   }
 
+  private get implMapVar(): string {
+    return slugify(
+      'implMap',
+      this.parent.parameters.map((p) => p.name),
+    );
+  }
```
RomainMuller, line 78: Looks like this would be invoked a bunch of times and could be kinda slow & thrash's... Maybe store the result in a private and re-use?

packages/jsii-pacmak/lib/targets/go/runtime/method-call.ts
```diff
@@ -17,6 +21,7 @@ export class MethodCall {
 
   private emitDynamic(code: CodeMaker) {
     code.line(`var ${this.returnVarName} ${this.concreteReturnType}`);
+    code.line(`${this.implMapVar} := make(${JSII_IMPL_MAP_TYPE})`);
```
RomainMuller, line 24: You don't seem to ever add things to this map (same below on L48)... And it really just gets passed to a method...

This makes me wonder:
- Do we need the variable, in fact?
- Can we not use an "empty map literal" instead of using `make` here?

packages/jsii-pacmak/lib/targets/go/runtime/property-access.ts
```diff
@@ -14,12 +15,24 @@ export class GetProperty {
...
+    code.line(`${implMapVar} := make(${JSII_IMPL_MAP_TYPE})`);
+
+    const implMap =
+      this.parent.reference?.scopedImplMap(this.parent.parent.pkg) ?? [];
+    if (implMap.length) {
+      const [interfaceName, structName] = implMap;
+      code.line(
+        `${implMapVar}[reflect.TypeOf((*${interfaceName})(nil)).Elem()] = reflect.TypeOf((*${structName})(nil)).Elem()`,
+      );
+    }
```
RomainMuller, line 29: Same as before - is there any reason we cannot use a literal here?

And since this map is fed with respects to `this.parent.reference?`, can we not have this be computed *once* and stored somewhere for re-use?

packages/jsii-pacmak/lib/targets/go/runtime/property-access.ts
```diff
@@ -14,12 +15,24 @@ export class GetProperty {
...
     code.line(`var ${resultVar} ${this.parent.returnType}`);
+    code.line(`${implMapVar} := make(${JSII_IMPL_MAP_TYPE})`);
+
+    const implMap =
+      this.parent.reference?.scopedImplMap(this.parent.parent.pkg) ?? [];
+    if (implMap.length) {
+      const [interfaceName, structName] = implMap;
+      code.line(
+        `${implMapVar}[reflect.TypeOf((*${interfaceName})(nil)).Elem()] = reflect.TypeOf((*${structName})(nil)).Elem()`,
+      );
```
RomainMuller, line 28: (Also below on line 68)

It would be nice to have a comment on this bad boy showing what a line it produces could look like... The amount of braces there is nauseating and it's kinda hard to picture what this produces.

packages/jsii-pacmak/lib/targets/go/types/class.ts
```diff
@@ -241,6 +248,10 @@ export class ClassMethod extends GoMethod {
   public get instanceArg(): string {
     return this.parent.name.substring(0, 1).toLowerCase();
   }
+
+  public get usesReflectionPackage() {
+    return Boolean(this.reference?.scopedImplMap(this.parent.pkg).length);
```
RomainMuller, line 253: Just double-not things to coerce them into a boolean context...

```suggestion
    return !!this.reference?.scopedImplMap(this.parent.pkg).length;
```

packages/jsii-pacmak/lib/targets/go/types/class.ts
```diff
@@ -241,6 +248,10 @@ export class ClassMethod extends GoMethod {
   public get instanceArg(): string {
     return this.parent.name.substring(0, 1).toLowerCase();
   }
+
+  public get usesReflectionPackage() {
```
RomainMuller, line 252:
```suggestion
  public get usesReflectionPackage(): boolean {
```

packages/jsii-pacmak/lib/targets/go/types/go-type-reference.ts
```diff
@@ -74,66 +75,109 @@ export class GoTypeRef {
...
 
-  private scopedTypeName(
-    scope: Package,
-    asInterface: boolean,
-    asRef = false,
-  ): string {
-    if (this.primitiveType) {
-      return this.primitiveType;
+  public get typeMap(): TypeMap {
+    return this.buildTypeMap(this);
```
RomainMuller, line 90: That looks like something that perhaps should be stored (it's used in 3 distinct locations...).

packages/jsii-pacmak/lib/targets/go/types/go-type-reference.ts
```diff
@@ -74,66 +75,109 @@ export class GoTypeRef {
...
-          asRef,
-        ) ?? 'interface{}';
+    return { interface: ref };
+  }
+
+  /**
+   * Builds a map of interface to concrete types. This is passed to the runtime
+   * so that return types of interfaces can correctly be constructucted.
+   */
+  public scopedImplMap(scope: Package) {
```
RomainMuller, line 117:
```suggestion
  public scopedImplMap(scope: Package): [string, string] {
```

Also - this returns a pair of strings, which isn't extremely expressive. Perhaps returning some object `{ scopedInterface: string, scopedImplementingType: string }` would be more readable?

packages/jsii-pacmak/lib/targets/go/types/go-type-reference.ts
```diff
@@ -74,66 +75,109 @@ export class GoTypeRef {
...
+    asRef = false,
+  ): string {
+    if (typeMap.primitive) {
+      return typeMap.primitive;
+    } else if (typeMap.array || typeMap.map) {
+      const prefix = typeMap.array ? '[]' : 'map[string]';
       const innerName =
-        new GoTypeRef(this.root, this.reference.mapOfType).scopedTypeName(
+        this.scopedTypeName(
+          (typeMap.array ?? typeMap.map!).typeMap,
```
RomainMuller, line 140: Optional (I'm even starting to doubt myself here) - I kinda find moving the bang after the parentheses makes it better signal "one of these is defined"... I don't know...

```suggestion
          (typeMap.array ?? typeMap.map)!.typeMap,
```

packages/jsii-pacmak/lib/targets/go/types/go-type-reference.ts
```diff
@@ -74,66 +75,109 @@ export class GoTypeRef {
...
     }
 
     // type isn't handled
-    log.debug(
-      `Type ${baseName} does not resolve to a known Go type. It is being mapped to "interface{}".`,
+    log.warn(
+      `Type ${typeMap.interface?.name} does not resolve to a known Go type. It is being mapped to "interface{}".`,
     );
+
     return 'interface{}';
```
RomainMuller, line 172: This this even possible/legal? Should we throw here?

packages/jsii-pacmak/lib/targets/go/types/go-type-reference.ts
```diff
@@ -74,66 +75,109 @@ export class GoTypeRef {
...
 }
+
+interface TypeMap {
+  primitive?: string;
+  array?: GoTypeRef;
+  map?: GoTypeRef;
+  union?: GoTypeRef[];
+  void?: boolean;
+  interface?: GoTypeRef;
+}
```
RomainMuller, line 183: Since `scopedTypeName` will actually only care about ONE of them at a time, it appears to me this type is excessively permissive (it'll allow all the fields to be set, which appears to be illegal).

I'd also make this as immutable as possible to make sure nothing does anything stupid with that.

```suggestion
type TypeMap =
     { readonly primitive?: string; }
  || { readonly array?: GoTypeRef; }
  || { readonly map?: GoTypeRef; }
  || { readonly union?: readonly GoTypeRef[]; }
  || { readonly void?: boolean; }
  || { readonly interface?: GoTypeRef; }
```

RomainMuller (commented):
I don't think I've seen an `implMap` value have more than 1 entry in it (they're either empty, or have exactly one key-value pair). Are we over-engineering here? Do we really need a map? Is the fact we never get >1 item just because our tests don't get into the use-case that'd cause > 1 item to be generated (and then - can we fix that?)

SoManyHs (approved):

PR #2298, opened 2020-12-02
fix(runtime): excessive latency introduced by sleep

### LINKED ISSUE #2284: release 1.15.0 makes cdk synth radically slow
### Affected Languages

- [x] `Python`
- [x] `Java`
- [ ] .NET (`C#`, `F#`, ...)

### General Information
* **JSII Version:** 1.15.1
* **Platform:**     Python

```
sh-4.2# uname -a
Linux 454d0ce8e2e4 5.4.39-linuxkit #1 SMP Fri May 8 23:03:06 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
sh-4.2# python3 -V -V
Python 3.7.9 (default, Aug 27 2020, 21:59:41)
[GCC 7.3.1 20180712 (Red Hat 7.3.1-9)]
```

### What is the problem?

When upgrading to `pip3 install --upgrade jsii==1.15.0` the synthesis with CDK up to 10 times slower. Some of our pipeline jobs take now 80 minutes instead of 8.

We have a complex app.py with loads of stacks and also loads of lookups to the AWS account during synthesis. Normally it takes "only" 10 seconds to synthesize the CFN scripts. However as often as I switch between jsii==1.14.1 and jsii==1.15.0 I get the same result, that the synthesis takes much much longer. This is independent from the AWS version I use (client and library).

```
sh-4.2# date && python3 app.py && date
Fri Nov 27 09:29:22 UTC 2020
Fri Nov 27 09:29:32 UTC 2020
sh-4.2# pip3 install --upgrade jsii
...
      Successfully uninstalled jsii-1.14.1
Successfully installed jsii-1.15.0
sh-4.2# date && python3 app.py && date
Fri Nov 27 09:29:52 UTC 2020
Fri Nov 27 09:31:54 UTC 2020
```

### Verbose Log

Because of privacy reasons I dont want to provide verbose logs. However I made a diff of both verbose logs before and after upgrading to jsii==1.15.1 and there were just a couple and no spectacular differences:

* modules workdir
  * before: /tmp/jsii-kernel-7tSAAx
  * after: /tmp/jsii-kernel-WXbuZ8
* the cdk.out*-directory

Both log files are of same length and same content, except the module wordir and the cdk out dir.

RomainMuller:
When recovering from `EAGAIN`, the `@jsii/runtime` recently added a
synchronous sleep for 50 milliseconds. However in large and complex
applications, this can add up to a lot of time, as worst case scenario,
the sleep can occur before and after each call (once on read, once on
write).

This change removes the `sleep` until we come up with a better solution.

Fixes #2284

---


packages/@jsii/runtime/lib/sync-stdio.ts
```diff
@@ -53,7 +51,8 @@ export class SyncStdio {
         if (e.code !== 'EAGAIN') {
           throw e;
         }
-        sleep(50 /*ms*/);
+        // We'll retry immediately, and possibly thrash the CPU until the buffer was drained. We
```
NetaNir, line 54: How many time will we retry? "thrash the cpu" sounds scary, should we even retry?

RomainMuller: We most Rettel until it succeeds... if we don't we crash. And apparently we retry for less than 5ms apiece 🙃

MrArnoldPalmer:
Added `do-not-merge` so users affected by this have a chance to test the fix as detailed [here](https://github.com/aws/jsii/issues/2284#issuecomment-737771027). @RomainMuller obviously feel free to merge if you don't see the need to wait.

RomainMuller:
@MrArnoldPalmer - yeah I'd rather have at least one user confirm the fix is effective before merging. This kind of active polling makes me feel dirty and I'd rather not be doing it if it's no use 😬

PR #2324, opened 2020-12-08
feat: go runtime auto push

MrArnoldPalmer:
Adds action to auto push to the `jsii-runtime-go` repository whenever
master us updated. This will allow for testing against specific builds
of the go runtime between releases. This is especially useful during the
pre-stable phase as iteration is relatively rapid.

Fixes the name of the go-runtime module constant inside `jsii-pacmak` to
correspond to the published repo name.

Adds the `build-repo-dir` build script to @jsii/go-runtime that will
build the contents of the `jsii-runtime-go` repository for publishing.

Adds linting and formatting to the `.ts` files in
`@jsii/go-runtime/build-scripts` since they are growing and becoming
more complex.

---


.github/workflows/push-go-runtime.yml
```diff
@@ -0,0 +1,72 @@
...
+          git add .
+          # Check for modifications in tracked files
+          changed=$(git diff --name-only origin/main)
+          if [-z "$changed"]; then
+            echo "::set-output name=result::false"
+          else
+            echo "::set-output name=result::true"
+          fi
+      - name: Commit and Push
+        if: steps.check-for-changes.outputs.result == 'true'
```
MrArnoldPalmer, line 67: @RomainMuller when testing locally using `act`, I can't get this step to run even when verifying that the output is correctly set in the previous step. Wondering if you have some insight on this?

.github/workflows/push-go-runtime.yml
```diff
@@ -0,0 +1,72 @@
...
+          JSII_RUNTIME_REPO_TOKEN: ${{ secrets.AUTO_APPROVE_GITHUB_TOKEN }}
+        run: |-
+          REPO_DIR=$(mktemp -d)
+          npx ts-node ./build-tools/build-repo-dir.ts $REPO_DIR
+          echo "::set-output name=repodir::$REPO_DIR"
+      - name: Check for changes
+        id: check-for-changes
+        run: |-
+          # Change directory manually as using `working-directory` doesn't seem to support using outputs from the previous step
+          cd ${{ steps.build-repo-dir.outputs.repodir }}
```
MrArnoldPalmer, line 56: Additionally, using the `working-directory` key on this step like
`working-direction: steps.build-repo-dir.outputs.repodir`
fails with `no such directory`. It looks like its trying to use the string `steps.build-repo-dir.outputs.repodir` as the directory so also tried `${{ steps.build-repo-dir.outputs.repodir }}` which results in `invalid substitution`.

Manually running `cd` isn't ideal but it works 🤷

.github/workflows/push-go-runtime.yml
RomainMuller: Could be a bug in `act`? I guess the question is does it behave "as intended" on GitHub Actions? If it does, then maybe report to `nektos/act`?

.github/workflows/push-go-runtime.yml
```diff
@@ -0,0 +1,72 @@
...
+        run: |-
+          yarn build \
+          --scope codemaker \
+          --scope @jsii/spec \
+          --scope @jsii/kernel \
+          --scope @jsii/runtime
+      - name: Build
+        run: |-
+          yarn build \
+          --scope @jsii/go-runtime
```
RomainMuller, line 40: Would perhaps be more effective (and safer in the long run) to:

```suggestion
      - name: Build
        run: |-
          yarn build \
          --scope @jsii/go-runtime \
          --include-dependencies
```

.github/workflows/push-go-runtime.yml
```diff
@@ -0,0 +1,72 @@
...
+      - name: Build Repo Directory
+        id: build-repo-dir
+        working-directory: $GITHUB_WORKSPACE/packages/@jsii/go-runtime
+        env:
+          JSII_RUNTIME_REPO_NAME: ${{ secrets.JSII_RUNTIME_REPO_NAME }}
+          JSII_RUNTIME_REPO_USERNAME: ${{ secrets.JSII_RUNTIME_REPO_USERNAME }}
+          JSII_RUNTIME_REPO_TOKEN: ${{ secrets.AUTO_APPROVE_GITHUB_TOKEN }}
+        run: |-
+          REPO_DIR=$(mktemp -d)
+          npx ts-node ./build-tools/build-repo-dir.ts $REPO_DIR
```
RomainMuller, line 50:
```suggestion
          yarn ts-node ./build-tools/build-repo-dir.ts $REPO_DIR
```

.github/workflows/push-go-runtime.yml
```diff
@@ -0,0 +1,72 @@
...
+            echo "::set-output name=result::false"
+          else
+            echo "::set-output name=result::true"
+          fi
+      - name: Commit and Push
+        if: steps.check-for-changes.outputs.result == 'true'
+        run: |-
+          message=$(git log -1 --format=%s)
+          cd ${{ steps.build-repo-dir.outputs.repodir }}
+          git commit -m $message
```
RomainMuller, line 71:
```suggestion
          git commit -m "$message"
```

.github/workflows/push-go-runtime.yml
RomainMuller: Additionally... You should be able to make the "Push" unconditional (worst case, result in "already up to date" idempotent success).

Basically, in the "detect" step you'd check for changes and create the commit as appropriate... in the second step you simply (always) push.

MrArnoldPalmer: I kept this for now and assumed its a bug with act. If running in actions doesn't work I'll remove and just always push.

.github/workflows/push-go-runtime.yml
MrArnoldPalmer: Since `@jsii/go-runtime` devdepends on `jsii-pacmak` which devdepends on `@jsii/(dotnet|java)-runtime`, using `--include-dependencies` requires setting up those toolchains which I'd prefer to avoid. Removing `jsii-pacmak` from devDependencies also doesn't feel right, so I kept this for now.

As detailed in the [lerna docs](https://github.com/lerna/lerna/tree/main/core/filter-options#--include-dependencies), using `--include-dependencies` overrides `--scope` and `--ignore` flags.

RomainMuller (approved):

MrArnoldPalmer:
@Mergifyio update

PR #2348, opened 2020-12-14
feat: add runtime type info (fqn and version) to compiled sources

njlynch:
Adds a static member to each exported class with the fully-qualified name and
current version. This is the first, preparatory step for a richer runtime type
information module for jsii, as well as enhanced metadata for CDKv2.

See [RFC#0253][rfc253] for RFC and specification I'm following here.

[rfc253]: https://github.com/aws/aws-cdk-rfcs/blob/master/text/0253-cdk-metadata-v2.md#appendix-1-jsii-run-time-type-information

---


packages/jsii/lib/runtime-info.ts
```diff
@@ -0,0 +1,100 @@
+import * as spec from '@jsii/spec';
+import * as ts from 'typescript';
+
+/**
+ * A limited subset of the Map<ts.ClassDeclaration, spec.TypeBase> to ease testing.
+ */
+export interface RuntimeClassInfo {
+  get(key: ts.ClassDeclaration): spec.TypeBase | undefined;
```
rix0rrr, line 8: Feels a little unnecessary to take complete `spec.TypeBase` objects if you only need the FQN.

packages/jsii/lib/assembler.ts
```diff
@@ -45,6 +45,13 @@ export class Assembler implements Emitter {
   private readonly _submoduleMap = new Map<ts.Symbol, ts.Symbol>();
   private readonly _submodules = new Map<ts.Symbol, SubmoduleSpec>();
 
+  private readonly projectVersion: string;
+  // Map of nodes to types for exported class types. Used by the RuntimeTypeInfoInjector
+  private readonly classTypesMap = new Map<
```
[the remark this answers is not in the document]
njlynch: So after getting this working, I started a minor refactor to do just that. I ran into some issues and ended up dropping it. Most notably, just taking in and/or updating a Map with a signature like below didn't work because I wasn't able to get a reference to a `ts.ClassDeclaration` prior to running the transformation that would be `===` to the same node during the transformation. I tried a few things with the signatures and tests to hack around it, but wasn't super happy with the result. I'll give it another go and see if I can come up with something that looks prettier.

packages/jsii/lib/runtime-info.ts
```diff
@@ -0,0 +1,126 @@
...
+    };
+  }
+
+  public runtimeTypeTransformer(): ts.TransformerFactory<ts.SourceFile> {
+    // eslint-disable-next-line @typescript-eslint/no-this-alias
+    const self = this;
+
+    return (context) => {
+      return (sourceFile) => {
+        const rttiSymbolIdentifier = ts.createIdentifier('vfqnSym');
```
RomainMuller, line 46: I don't really understand what `vfqnSym` means... "Version and FQN Symbol"? I find this is over-specific.

And I think we should use `createUniqueName` to actually ensure we don't mistakenly get a variable name that conflicts with other declarations in the scope...

```suggestion
        const rttiSymbolIdentifier = ts.createUniqueName('jsiiRttiSymbol');
```

packages/jsii/lib/runtime-info.ts
```diff
@@ -0,0 +1,126 @@
...
+        const rttiSymbolDeclaration = ts.createVariableDeclaration(
+          rttiSymbolIdentifier,
+          undefined,
+          rttiSymbol,
+        );
+
+        sourceFile = ts.updateSourceFileNode(sourceFile, [
+          ts.createVariableStatement([], [rttiSymbolDeclaration]),
+          ...sourceFile.statements,
+        ]);
```
RomainMuller, line 64: Is there a chance this will cause "unused identifier" errors in any source file that does not declare an exported type?

packages/jsii/lib/runtime-info.ts
```diff
@@ -0,0 +1,126 @@
...
+    const runtimeProperty = ts.createProperty(
+      undefined,
+      ts.createModifiersFromModifierFlags(
+        ts.ModifierFlags.Private | ts.ModifierFlags.Static,
+      ),
+      ts.createComputedPropertyName(rttiSymbol),
+      undefined,
+      undefined,
+      runtimeInfo,
+    );
```
RomainMuller, line 115: This should also be `readonly`, I think.

packages/jsii/lib/runtime-info.ts
njlynch: Possibly. I didn't run into it in testing, but it's a reasonable call-out. I can adjust to only emit the symbol when necessary.

packages/jsii/lib/runtime-info.ts
njlynch: Totally fine with both changes. I originally was using `createUniqueName`, but it resulted in a slightly less friendly name by default, even without conflicts. But I suppose better safe than sorry in this regard.

eladb (approved):
Pleasure to read :-D

Minor comments.

RomainMuller (approved):

PR #2368, opened 2020-12-20
fix(dotnet): Use nested classes for proxies to avoid name collision

### LINKED ISSUE #2367: .net proxy classes can conflict with classes whose name ends with "proxy"
### Affected Languages

- [ ] `Python`
- [ ] `Java`
- [X] .NET (`C#`, `F#`, ...)

### General Information
* **JSII Version:** 1.15

### What is the problem?
If you have a class/interface named `XProxy` and also an interface named `X`, the generated proxy class for `X` will also have name `XProxy`. They actually can be written at the same time which results in a corrupt file that contains parts of each definition.

In cdktf, the azure provider results in both `ApiManagementHostnameConfiguration` and `ApiManagementHostnameConfigurationProxy`.

### Verbose Log
https://github.com/hashicorp/terraform-cdk/pull/462/checks?check_run_id=1580597921

### Possible Solution
Rather than generating these proxies in their own top level classes and files, generated as nested classes similar to how the Java generation works. While they can't be named `Jsii$Proxy`, `Jsii_Proxy` seems equally likely to not conflict.

jsteinich:
Fixes #2367 

C# proxy classes now nested in associated type using the name `Jsii_Proxy`. Using non-standard `_` in naming to further reduce chance of collision.

I would like to do some more testing since I'm not familiar with the details of how the proxies work.

---


RomainMuller:
Oh this is neat! I wonder if we shouldn't simply name the proxy type `_Proxy`... The `_` alone should guarantee there is no collision (this prefix - depending on the case - either signifies declarations are hidden from jsii, or is illegal).

Besides this, we'll need to dutifully make sure this is not a breaking change to existing users...

jsteinich:
> I wonder if we shouldn't simply name the proxy type `_Proxy`

I went ahead and updated the name.



> Besides this, we'll need to dutifully make sure this is not a breaking change to existing users...

Since the proxy class is an internal type, it would only be directly usable via reflection, which should reduce the chance of breakage.
Since users could be committing generated code, should definitely call out the change though.

I saw some references to the aws cdk in tests; does it run a full build of that? Or is there a good way to test things out against the cdk?

RomainMuller:
@all-contributors add @jsteinich for code

RomainMuller (dismissed):

RomainMuller:
(I like this, marked it as `pr/do-not-merge` so I have a chance to inspect the generated CDK artefacts before shipping this for realz)

You may want to make the PR "ready for review" (as far as I'm concerned, it is...).
I could do this on your behalf but I tend to think of it as overstepping a little (you're the one who should decide if you think it's ready or not).

jsteinich:
> You may want to make the PR "ready for review" (as far as I'm concerned, it is...).

I've taken some more time to understand what tests exist and how the proxies are used, so I'll go ahead and update.

The one not quite ideal piece is some warnings because of the `new` keyword the proxy class. It's there because otherwise subtypes of types with proxies will warn the other way. It seems possible to loop through base classes / implemented interfaces to determine if it is necessary, but that doesn't feel very clean or reliable.

RomainMuller:
> The one not quite ideal piece is some warnings because of the `new` keyword the proxy class. It's there because otherwise subtypes of types with proxies will warn the other way. It seems possible to loop through base classes / implemented interfaces to determine if it is necessary, but that doesn't feel very clean or reliable.

I think we should be able to get a relatively clean signal - if any type in the inheritance chain has a proxy, then we add the `new` modifier, otherwise not. We could also disable that warning (but I feel this is slightly worse overall).

RomainMuller:
Alright I have added some logic to only emit the `new` modifier when it's actually necessary, and I've asked the compiler to keep me honest on that front (turning the `new` warnings into errors).

RomainMuller (approved):

PR #2373, opened 2020-12-21
fix(pacmak): illegal static overrides in java & c#

### LINKED ISSUE #2358: Identical static methods in parent/child classes are mistakenly marked as overrides in java resulting in a compilation error
### Affected Languages

- [ ] `Python`
- [X] `Java`
- [X] .NET (`C#`, `F#`, ...) *[Probably]*

### General Information
* **JSII Version:** 1.15.0 (build 585166b), typescript 3.9.7
* **Platform:**     MacOS, CodeBuild (jsii/superchain)

### What is the problem?

We encountered a scenario in `aws-cdk` that generates illegal java code, annotating static methods with the `@Override` annotation. 

This happens when a child class has the same static method as its parent.

```ts
export class Parent {
  public static hello() {}
}

export class Child extends Parent {
  public static hello() {}
}
```

This code will generate a JSII assembly that marks the `Child.hello` function both as static and as overrides of `Parent.hello`.
The generated java code will look like:

```java
public class Child extends Parent {

  @Override
  public static hello() {}

}
```

Resulting in a compilation error: `method does not override or implement a method from a supertype`

See https://github.com/aws/aws-cdk/pull/12091 for the full scenario.

RomainMuller:
Stop emitting "overrides" in the respecting language idiomatic way for
static methods and properties (while ES6 supports those, this is not
true of Java and C#).

In Java, this is mostly getting rid of `@Overrides` on static
declarations. C# *allows* (but does not require) explicitly hiding the
parent declaration using the `new` keyword. This was introduced as it
provides additional safeguards against our generating incorrect code.

Fixes #2358

---


packages/jsii-pacmak/lib/targets/dotnet/dotnetgenerator.ts
```diff
@@ -833,7 +832,7 @@ export class DotNetGenerator extends Generator {
       );
       if (implementedInBase || datatype || proxy) {
         // Override if the property is in a datatype or proxy class or declared in a parent class
-        isOverrideKeyWord = 'override ';
+        isOverrideKeyWord = prop.static ? 'new ' : 'override ';
```
NetaNir, line 835: Oh god, it took me a moment to understand what the `new` keyword is doing here. Can we add a link to Microsoft documentation? https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/new-modifier
Added as a suggestion to the doc line.

packages/jsii-pacmak/lib/targets/dotnet/dotnetgenerator.ts
```diff
@@ -833,7 +832,7 @@ export class DotNetGenerator extends Generator {
       );
       if (implementedInBase || datatype || proxy) {
         // Override if the property is in a datatype or proxy class or declared in a parent class
```
NetaNir, line 834:
```suggestion
        // Override if the property is in a datatype or proxy class or declared in a parent class, for static methods use the `new` modifier: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/new-modifier 
```

NetaNir (approved):
Looks good to me, mostly "take it or leave it" comments, approved and added `do not merge` (we really need an emoji for this type of comments).

I initially thought we should implements this fix by emitting the `override` annotation for the assembly file, but Im guessing it will be less ideal since we will be losing important information?

packages/jsii-pacmak/lib/targets/dotnet/dotnetgenerator.ts
```diff
@@ -496,12 +496,11 @@ export class DotNetGenerator extends Generator {
...
     }
-    if (method.static) {
-      staticKeyWord = 'static ';
-    } else if (overrides) {
+    if (overrides) {
       // Add the override key word if the method is emitted for a proxy or data type or is defined on an ancestor
-      overrideKeyWord = 'override ';
+      overrideKeyWord = method.static ? 'new ' : 'override ';
     } else if (
+      !method.static &&
```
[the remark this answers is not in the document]
RomainMuller: I have been internally debating this a little bit... Initially I wanted to simply change the compiler so that statics are **never** considered to be overrides; but this caused problems, because in JavaScript (and hence, TypeScript), they actually are overrides... This manifests when you try the following:

```ts
export class Parent { public static foo: number; }

//! Child.foo incorrectly implements Parent.foo (types don't match)
export class Child extends Parent { public static foo: string; }
```

This made me decide that it was a more coherent story from the author's standpoint if we reproduced the ES6 semantics in the jsii assembly (where statics are inherited with the rest of the prototype). Carrying them as overrides has advantages - in languages where static are inherited (although JS is the only one that comes to mind now), this allows overrides to use `super`.

We *could* distinguish static overrides from non-static overrides, however this distinction is very much a code generation concern, and not so much a type model concern (after all, all we care about is whether this is an override or not, period).

PR #2377, opened 2020-12-22
feat(jsii): allow user-owned tsconfig.json files

RomainMuller:
Makes it possible for users to own more of the TypeScript configuration
by actually owing the `tsconfig.json` file instead of having it
generated by `jsii`.

The configuration is moved from the `jsii` key in `package.json` to the
`x-jsii` key in `tsconfig.json` (which the user must then check in
with their source code, as with any other TypeScript project).

The new configuration point is entirely optional (existing projects will
continue to work the exact same way they currently do), however if there
is configuration in `tsconfig.json`, it is an error to also have
configuration in `package.json`.

A new `@jsii/config` library was introduced to centralise the reading of
the configuration from either source. Several tools (including `jsii-pacmak`
make use of this for various purposes, including configuration that
has no business being recorded in the jsii assembly file, such as `outdir`).

---


packages/@scope/jsii-calc-base-of-base/tsconfig.json
```diff
@@ -0,0 +1,58 @@
...
+    "strictNullChecks": true,
+    "strictPropertyInitialization": true,
+    "stripInternal": false,
+    "composite": true,
+    "incremental": true
+  },
+  "include": ["**/*.ts"],
+  "exclude": ["node_modules"],
+  // The JSII configuration
+  "x-aws-jsii": {
```
eladb, line 27: I wouldn't use the "aws" name here. Let's call this "x-jsii" or just "jsii" (no need to be too pure about this).

RomainMuller: Yeah I guess `x-jsii` is good... (The `x-` prefix signifies "extension")

Neurostep:
@RomainMuller Any plans on going further with this one? Need any help? Currently have a use-case in a personal project where having managed `tsconfig.json` would allow me to go further

RomainMuller:
> @RomainMuller Any plans on going further with this one? Need any help? Currently have a use-case in a personal project where having managed `tsconfig.json` would allow me to go further

I still want to do this but did not have time and that PR is now very very stale I'm sorry.

PR #2411, opened 2021-01-11
feat(runtime): use dedicated file descriptors for sync IPC

### LINKED ISSUE #2406: Output to STDOUT corrupts the jsii runtime protocol
### Affected Languages

- [x] `Python`
- [x] `Java`
- [x] .NET (`C#`, `F#`, ...)

### General Information
* **JSII Version:** 1.16.0

### What is the problem?

The protocol between jsii languages and the javascript runtime is based on STDIN/STDOUT communication. If the javascript code (i.e. projen), writes data to STDOUT, this data intermingles with protocol commands and the jsii runtime crashes.

### Solutions

Perhaps there is a way to redirect the global `process.stdout` to `process.stderr` so that user code will be unaware and open stdout separately for communication. This approach has a disadvantage that now JS/TS will behave differently from other languages (i.e. I could write jsii code that writes to STDOUT, but it won't work in other languages).

Another option is to use a different communication pipe between the host language and the runtime, but as far as I know we dismissed most alternatives for various reasons (portability, security).

RomainMuller:
This replaces the `@jsii/runtime` package entry point so that it creates
a child `node` process with two new pipe file descriptors (at FD3 and
FD4). This means the `@jsii/runtime` now has two processes instead of
one:

1. A (new) "wrapper" process
2. The "core" process

The `@jsii/runtime` core process now uses `FD#3` for synchronous reading
of requests sent by the host to the `@jsii/kernel`, and `FD#4` to
synchronously write responses back to the host process. This alleviates
the need to resort to hot-polling around `EAGAIN` errors that are
symptomatic of attempting blocking operations on a non-blocking file
descriptor (aws/aws-cdk#12069). This works as the non-standard (`>2`)
file descriptors are not managed by `libuv`, and hence are not made
non-blocking at the start of `node`, allowing blocking operations to be
reliably used with those descriptors.

The `@jsii/runtime` core process was updated to use those two new file
descriptors instead of directly using `STDIN` and `STDOUT`, allowing
the **JavaScript** code to naturally use `process.stdout` and
`process.stderr` (as well as all other APIs that eventually use the same
file descriptors, such as `console.log` and `console.error`) without
causing corruption of the host-kernel communication channel as was
previously the case (#2406).

Data written directly to the `@jsii/runtime` core process' `FD#1`
(`STDOUT`) and `FD#2` (`STDERR`) is intercepted by the wrapper process,
and encoded in JSON objects before being sent to the host application
via the wrapper process' `STDERR` pipe. The host app is now reponsible
to maintain a thread that reads from the wrapper's `STDERR` pipe, decode
the JSON objects, then write the data to the appropriate host process'
STDIO stream to emulate direct operations as closely as possible.

Fixes aws/aws-cdk#12069
Fixes #2406

---


RomainMuller:
This PR would supersede #2365 as it appears to be a superior solution (more reliable, less complex, and with more beneficial side-effects).

rix0rrr (approved):
My brain explodes on this. I trust you?

PR #2417, opened 2021-01-13
feat(go): runtime release tagging

MrArnoldPalmer:
Adds github actions workflows for automatically tagging jsii-runtime-go
when jsii release branch is updated. This essentially publishes new
runtime version. This may eventually be served better in delivlib but is
much simpler in github actions for the alpha phase.

---


.github/workflows/main.yml
```diff
@@ -466,3 +466,84 @@ jobs:
           path: |-
             ${{ github.workspace }}/aws-cdk/dist/
             ${{ github.workspace }}/aws-cdk/**/dist/
+  push-go-runtime:
```
RomainMuller, line 469: Nitpicking, but can we add a blank line above this one?

```suggestion

  push-go-runtime:
```

.github/workflows/main.yml
```diff
@@ -466,3 +466,84 @@ jobs:
...
+      - name: Get Last Commit Message
+        id: get-last-commit-msg
+        working-directory: ${{ github.workspace }}/jsii-built-tree
+        run: |-
+          message=$(git log -1 --format=%s)
+          echo "::set-output name=result::$message"
+      - uses: actions/cache@v2
+        with:
+          path: '**/node_modules'
+          key: ${{ runner.os }}-modules-${{ hashFiles('**/yarn.lock') }}
```
RomainMuller, line 500: This cache is... okay, but it's not doing as much work as it could... You'd probably benefit from also caching Yarn's cache, like the other jobs do.

Also - not sure how this deals with symlinks... I'd be tempted to be somewhat afraid it caches the referents of the symlinks to packages within the monorepo... Could this have consequences?

.github/workflows/main.yml
```diff
@@ -466,3 +466,84 @@ jobs:
...
+          if [-z "$changed"]; then
+            echo "::set-output name=result::false"
+          else
+            echo "::set-output name=result::true"
+          fi
+      - name: Commit and Push
+        if: steps.check-for-changes.outputs.result == 'true'
+        working-directory: ${{ steps.build-repo-dir.outputs.repodir }}
+        run: |-
+          git config user.name ${{ secrets.JSII_RUNTIME_REPO_USERNAME }}
```
RomainMuller, line 539: Is that necessary? Could we make this be `AWS CDK Automation`, since the email address is hard-coded?

MrArnoldPalmer: Not necessary, relic from manual git authentication stuff.

.github/workflows/main.yml
MrArnoldPalmer: Lemme know if this is more what you were thinking. Since this only needs yarn deps tried to exclude all the other languages package caches.

RomainMuller (approved):

PR #2421, opened 2021-01-13
fix(python): process.terminate() got an unexpected keyword argument 'timeout'

### LINKED ISSUE #2398: process.terminate() got an unexpected keyword argument 'timeout'
### Affected Languages

- [X] `Python`

### General Information
* **JSII Version:** 1.16.0
* **Platform:** Debian

### What is the problem?
Hey, guys. While running **cdk diff** on a **python 3.7.6** app I'm getting the next error:

```
Error in atexit._run_exitfuncs:
  File "/usr/local/lib/python3.7/site-packages/jsii/_kernel/providers/process.py", line 293, in stop
    self._process.wait(timeout=5)
  File "/usr/local/lib/python3.7/subprocess.py", line 1019, in wait
    return self._wait(timeout=timeout)
  File "/usr/local/lib/python3.7/subprocess.py", line 1645, in _wait
    raise TimeoutExpired(self.args, timeout)
subprocess.TimeoutExpired: Command '['node', '/usr/local/lib/python3.7/site-packages/jsii/_embedded/jsii/jsii-runtime.js']' timed out after 5 seconds

During handling of the above exception, another exception occurred:

  File "/usr/local/lib/python3.7/site-packages/jsii/_kernel/providers/process.py", line 296, in stop
    self._process.terminate(timeout=5)
TypeError: terminate() got an unexpected keyword argument 'timeout'
```

The problem seems to be caused by [this change](https://github.com/aws/jsii/commit/c11999499e6f9679da0272d96a1422025bbc0502#diff-93240363d8ef76db5d41bb9017a4be449538459dff102c009f87f0996bbb55f6R296)  introduced in PR #2100:

image

Indeed, _timeout_ is not a valid parameter for that method, perhaps removing it should be enough to fix the issue. Thanks a lot!

@RomainMuller

RomainMuller:
Corrected usage of `process.terminate` to remove unexpected `timeout`
argument. Not sure how this did not surface in integration tests (will
investigate later).

Fixes #2398

---

### DIFF
--- a/packages/@jsii/python-runtime/src/jsii/_kernel/providers/process.py
+++ b/packages/@jsii/python-runtime/src/jsii/_kernel/providers/process.py
@@ -298,10 +298,7 @@ class _NodeProcess:
         try:
             self._process.wait(timeout=5)
         except subprocess.TimeoutExpired:
-            try:
-                self._process.terminate(timeout=5)
-            except subprocess.TimeoutExpired:
-                self._process.kill()
+            self._process.terminate()
 
         self._ctx_stack.close()
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## [1.17.1](https://github.com/aws/jsii/compare/v1.17.0...v1.17.1) (2021-01-13)
+
+### Bug Fixes
+
+* **python:** process.terminate() got an unexpected keyword argument 'timeout' ([#2421](https://github.com/aws/jsii/issues/2421)), closes [#2398](https://github.com/aws/jsii/issues/2398)
+
+
 ## [1.17.0](https://github.com/aws/jsii/compare/v1.16.0...v1.17.0) (2021-01-13)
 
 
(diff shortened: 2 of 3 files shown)


packages/@jsii/python-runtime/src/jsii/_kernel/providers/process.py
```diff
@@ -298,10 +298,7 @@ def stop(self):
         try:
             self._process.wait(timeout=5)
         except subprocess.TimeoutExpired:
-            try:
-                self._process.terminate(timeout=5)
-            except subprocess.TimeoutExpired:
-                self._process.kill()
+            self._process.terminate()
```
RomainMuller, line 301: Note: `terminate` does the same work as `kill`, except on POSIX platforms, where the signal sent to the subprocess is different (`SIGTERM` for `terminate`, `SIGKILL` for `kill`). The default handling for both of these signals is the same (terminates the process).

`SIGKILL` cannot be trapped (always kills the process immediately, not giving it a chance to clean up after itself).

`SIGTERM` can be trapped (and rescued to not terminate the process), and gives the process a modest slice of time to clean up after itself. Since we are not trapping `SIGTERM`, both signals achieve the same - terminating the `node` VM, except `SIGTERM` would allow VM exit handlers to run, which is desirable.

iliapolo (approved):

PR #2437, opened 2021-01-19
feat(jsii): experimental --strip-deprecated feature

RomainMuller:
This feature will cause `jsii` to erase all declarations marked with
`@deprecated` from the `.jsii` assembly and `.d.ts` files.

When a `@deprecated` type is extended (or implemented) by a type that
is not `@deprecated`, the inheritance chain will be corrected, erasing
the `@deprecated` type and replacing it with it's non `@deprecated`
parents, if any.

Remaining uses of `@deprecated` members will be reported through
diagnostic messages (with the `ERROR` severity).

---

A new module `bindings` was implemented to allow relating TypeScript AST
nodes to "arbitrary" points in the jsii assembly, allowing to recover
that information once the tree was fully processed. This makes
multi-pass transformations not need to re-inspect the whole AST again.

---

⚠️  This is currently unable to identify interface implementations that become
invalid (in the type definitions) as a result of erasing a base class.

---


packages/jsii/lib/transforms/deprecated-remover.ts
```diff
@@ -0,0 +1,891 @@
+import {
```
nija-at, line 1: I would recommend tests for such a large and complex code. I suppose it's coming soon.

nija-at (commented):
Architectural qn (non blocking): we have a couple of transformers here now  - this one and the one for experimental API docs. How far away do you think we are to making these PnP to the jsii core compiler (not necessarily to the CLI or configuration)?

packages/jsii/lib/transforms/deprecated-remover.ts
RomainMuller: ...granted I am not allocated to another project...

RomainMuller:
> Architectural qn (non blocking): we have a couple of transformers here now - this one and the one for experimental API docs. How far away do you think we are to making these PnP to the jsii core compiler (not necessarily to the CLI or configuration)?

To be honest - I very much dislike the current state of this; and it'd likely deserve a major overhaul. The main `jsii` business needs to happen as a transform, too, so we don't have to actually jump though the kind of hoops we are currently... The performance right now would be hit by the fact we actually cause the TypeScript compiler itself to emit (i.e: write output files, like .js & .d.ts files) at least twice, maybe thrice... This is...... unnecessary.

packages/jsii/bin/jsii.ts
```diff
@@ -54,6 +54,12 @@ const warningTypes = Object.keys(enabledWarnings);
             desc: `List of warnings to silence (warnings: ${warningTypes.join(
               ',',
             )})`,
+          })
+          .option('strip-deprecated', {
```
nija-at, line 58: Expand and document this in the jsii docs. Also explain any pre-requisites, such as deprecated types must not be used in such-and-such places, etc.

packages/jsii/test/deprecated-remover.test.ts
```diff
@@ -0,0 +1,161 @@
+import { compileJsiiForTest } from '../lib';
+
+const DEPRECATED = '/** @deprecated stripped */';
```
nija-at, line 3: Why is this additional `stripped` needed? Why not just `/** @deprecated */`

packages/jsii/test/deprecated-remover.test.ts
```diff
@@ -0,0 +1,161 @@
...
+
+test('produces correct output', async () => {
```
nija-at, line 5: Would've been nicer to organize these fixtures and expectations as actual folders with files. 
Would be much easier to add new cases as we fix or add features to the remove-deprecated.

WDYT?

packages/jsii/bin/jsii.ts
nija-at, line 58: Is there a way to ensure that when 'new' jsii features are supported (say, union types), they are correctly handled by this flag? 

Not expecting it in this PR, but wondering if you've thought about this. 

One way that comes to mind is - "some" automated mechanism (hand wave) of taking the jsii fixture (calc) and marking most or all of them as deprecated and stripping them out.

Another option - if there's some place where we've centralized the jsii specification and supported features, it might be easier walk through them and dynamically generate a fixture and run them through strip-deprecated.

nija-at, line 58: (Not necessary to do it in this PR)

What do you think about adding this option to the `jsii` component of `package.json` as well?

nija-at (approved):
Provisional approval

packages/jsii/bin/jsii.ts
RomainMuller: The whole point of it being `EXPERIMENTAL` here is I actually don't want to publicise it too much just yet...

packages/jsii/test/deprecated-remover.test.ts
RomainMuller: `/** @deprecated */` is in jsii required to have a deprecation note, or it'll generate an error... So I juts put something here...

packages/jsii/bin/jsii.ts
nija-at: Maybe add an experimental section to the docs 🤷‍♂️
Fine with me if you feel strongly

packages/jsii/test/deprecated-remover.test.ts
RomainMuller: Expectations as actual folders with files means we have to write our own way to update snapshots (`jest -u` will only update `.ts.snap` files, and the structure there is not very flexible). Also - how is more files to know about and modify, better than the whole GIVEN/WHEN/THEN is in a single place?

packages/jsii/bin/jsii.ts
RomainMuller: (Hand wave) Contributions are welcome 🤓

RomainMuller: I'm favorable to this but did not have time to go about it... Should be easy enough to do.

PR #2462, opened 2021-01-21
fix(go): invalid output for multi-line @return and @deprecated comments

### LINKED ISSUE #2457: Go: multi-line comments result in invalid go code
### Affected Languages

- [x] `golang`

### General Information
* **JSII Version:** 1.17.1

### What is the problem?

When generating go code from a module that includes multi-line comments (e.g. aws/constructs), the generated code is invalid.

For example, this typescript code:

```ts
  /**
   * All parent scopes of this construct.
   *
   * @returns a list of parent scopes. The last element in the list will always
   * be the current construct and the first element will be the root of the
   * tree.
   */
  public get scopes(): IConstruct[] { ... }
```

Results in this go code:

```go
	// All parent scopes of this construct.
	//
	// Returns: a list of parent scopes. The last element in the list will always
be the current construct and the first element will be the root of the
tree.
	Scopes []IConstructIface `json:"scopes"`
```

As you can see, we need to add `//` prefix to the multiline `@returns` docstring.

### Verbose Log

```
./constructs.go:457:8: syntax error: unexpected current, expecting semicolon or newline or }
./constructs.go:459:10: syntax error: unexpected ] after top level declaration
```

eladb:
Split multi-line comments and prefix with `//` so they are rendered correctly.

Fixes #2457

---


packages/jsii-calc/lib/documented.ts
```diff
@@ -12,11 +12,11 @@ export class DocumentedClass {
   /**
    * Greet the indicated person.
    *
-   * This will print out a friendly greeting intended for
-   * the indicated person.
+   * This will print out a friendly greeting intended for the indicated person.
```
RomainMuller, line 15: Is this intentional change?

packages/jsii-pacmak/lib/targets/go/documentation.ts
```diff
@@ -49,13 +46,19 @@ export class Documentation {
...
       } else {
-        this.code.line(`// ${this.code.toPascalCase(stability)}.`);
+        this.emitComment(`${this.code.toPascalCase(stability)}.`);
       }
     }
   }
 
+  private emitComment(text = '') {
+    for (const line of text.split('\n')) {
+      this.code.line(`// ${line}`);
```
RomainMuller, line 58:
```suggestion
      this.code.line(`// ${line}`.trim());
```

RomainMuller (commented):
I am somehow irked by the trailing whitespace, which is so easy to get rid of...

packages/jsii-calc/lib/documented.ts
eladb: Yes, just wrapped to 100 chars.

packages/jsii-pacmak/lib/targets/go/documentation.ts
eladb: I've added this but then the snapshots actually changed much more dramatically so I decided to revert. I'll add.

RomainMuller (approved):

PR #2476, opened 2021-01-26
fix(python): pin mypy to the exact version

### LINKED ISSUE #2464: MyPy Type Failures
### Affected Languages

- [x] `Python`
- [ ] `Java`
- [ ] .NET (`C#`, `F#`, ...)

### General Information
* **JSII Version:** v1.17.1
* **Platform:**  Darwin f8ffc25f2862.ant.amazon.com 19.6.0 Darwin Kernel Version 19.6.0: Tue Nov 10 00:10:30 PST 2020; root:xnu-6153.141.10~1/RELEASE_X86_64 x86_64

### What is the problem?
Python code generated by jsii-pacmak fails mypy type checks. This doesn't seem to have been triggered by code changes in JSII so is likely the result of an update to mypy or another related lib.

Adding `# type: ignore` comments can resolve some of these errors easily, additional logic may be required to only add these comments where needed so we aren't suppressing valid type errors. Not sure what the exact cause is though.

### Verbose Log
```bash
    #STDOUT> /tmp/harness.jsOMHEA9/python/src/jsii_calc/__init__.py:7010: error: Call to
    #STDOUT> untyped function "member" in typed context  [no-untyped-call]
    #STDOUT>         @jsii.member(jsii_name="baz")
    #STDOUT>          ^
    #STDOUT> /tmp/harness.jsOMHEA9/python/src/jsii_calc/__init__.py:7010: error: Untyped
    #STDOUT> decorator makes function "baz" untyped  [misc]
    #STDOUT>         @jsii.member(jsii_name="baz")
    #STDOUT>          ^
    #STDOUT> /tmp/harness.jsOMHEA9/python/src/jsii_calc/__init__.py:7012: error: Returning
    #STDOUT> Any from function declared to return "None"  [no-any-return]
    #STDOUT>             return jsii.invoke(self, "baz", [])
    #STDOUT>             ^
```

RomainMuller:
It's been a couple of times `mypy` breaks the pipeline by releasing new
versions that are more strict or start checking things it used to
silently ignored.

In order to avoid this happening in the future (and putting us back in
control), this pins the `mypy` version to the exact release (the last
we successfully built with, in this case), and moves everything we
install though `pip install` into `requirements-dev.txt` files so that
@dependabot can help us stay up-to-date going forward.

Fixes #2464

---


.github/dependabot.yml
```diff
@@ -5,7 +5,6 @@ updates:
     directory: '/'
     schedule:
       interval: daily
-    open-pull-requests-limit: 10
```
RomainMuller, line 8: Removed those limits because they're no longer necessary (they're mostly useful when you "just enabled" dependant on a codebase with a lot of outdated dependencies). This would improve our ability to remain up-to-date...

eladb (approved):
Looks like this PR includes more then just pinning mypy, but ok...

PR #2482, opened 2021-01-26
feat: submodules expose readmes and targets via `jsii-reflect`

rix0rrr:
For the documentation for monocdk/v2, individual submodules need to be
able to have `README`s attached to them, as well as the `targets`
configuration of submodules needs to be exposed so that we can use this
to determine the FQNs of types in other languages (Java/C#/...).

---


packages/jsii-calc/lib/submodule/README.md
```diff
@@ -0,0 +1,4 @@
+Read you, read me
```
RomainMuller, line 1: Some real Lionel Richie energy right there

RomainMuller (approved):

PR #2483, opened 2021-01-26
docs(decision): go callback formalism

RomainMuller:
Documents the decision for the userland formalism for go overrides. It
explains the way dynamic dispatch can be achieved in go, and how
overrides are to be declared.

The runtime mechanism itself derives directly from the userland syntax,
and is relaitvely straight-forward (as it does not differ significantly
from the mechanisms found in other language runtime libraries).

---

### DIFF
--- a/gh-pages/content/adr/2021-01-21-go-override-syntax.md
+++ b/gh-pages/content/adr/2021-01-21-go-override-syntax.md
@@ -1,0 +1,308 @@
+# Go Overrides Syntax
+
+* Status: proposed
+* Deciders: @RomainMuller, @MrArnoldPalmer, @eladb, @iliapolo
+* Date: 2021-01-21
+
+Technical Story: #2048
+
+## Context and Problem Statement
+
+**Go** is not an object-oriented programming language in the classical sense. It
+does not have classes, and consequently does not support inheritance. This poses
+an interesting challenge with respects to handling [callbacks and overrides].
+
+Generated bindings for **go** must allow developers to leverage classical object
+oriented inheritance semantics (including anonymously implementing interfaces),
+and have the overridden and implemented members correctly registered with the
+*jsii kernel*.
+
+!!! note
+    In the context of @aws/aws-cdk usage (more so in the `v2` release line),
+    abstract base classes are not exposed, and users are generally not expected
+    to sub-class types and override members.
+
+    Constructor overriding is not necessary in **go**, as the embedding model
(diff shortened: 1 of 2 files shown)


MrArnoldPalmer:
So I'm mostly onboard with the design here, I'm just wondering about how this relates to the current strategy of outputting interface method implementations for all parent classes/interfaces instead of embedding. Currently, when code generating a class, we don't anonymously embed interfaces we just implement them all the way down the chain. This is to prevent users from needing to know about long embed chains ie `x.parentiface.anotherparentiface.methodthing()`.

It seems like we could just change to interface embedding, and in code generating classes with parents, we would just generate delegating to parent methods for interface implementions, but then that would require the constructors to also call parent constructors and embed (also doable as shown in the proposal). Or is the idea to only have anonymous interface embedding within structs for the purpose of overriding? Without switching to embedding, would that create a conflict where two methods of the same signature are created on the same struct (since they are all top level) when implementing overrides on a non-abstract class?

gh-pages/content/adr/2021-01-21-go-override-syntax.md
```diff
@@ -0,0 +1,308 @@
...
+
+??? example
+    ```go
+    type DefaultGreeter struct {
+      *GreeterBase `overrides:"GetGreetee"`
+      greetee      string
+    }
+
+    type ExcitedGreeter struct {
+      *DefaultGreeter `overrides:"GetGreetee,Greet"`
```
eladb, line 272: Is it possible to put the tag on the method and not on the embedding?

eladb (changes requested):
I still struggle to fully understand how will this look to Go users. Maybe we can add a "working backwards from the README" section which describes the developer experience for:

1. Passing an anonymous implementation of a jsii interface from Go to a jsii method
2. Passing a Go object that implements a jsii interface
3. Passing a Go object derived from a jsii class _and_ implements a jsii interface
4. "Extending" an abstract jsii base class in Go


(we should have a set of compliance tests which cover this, perhaps we can use those to make sure we cover all the cases)

gh-pages/content/adr/2021-01-21-go-override-syntax.md
RomainMuller: Methods cannot carry tags... Only struct fields.

RomainMuller:
This was superseded by documentation introduced in #2717

PR #2485, opened 2021-01-26
feat(go): run "go build" on generated code

### LINKED ISSUE #2463: Go: run `go build` after code generation
### Affected Languages

- [x] `golang`

### General Information
* **JSII Version:** 1.17.1

To ensure that the generated go output is valid, I would recommend running `go build` on the resulting artifact so that errors will be discovered during build and not when the module is consumed.

### Description

The issue https://github.com/aws/jsii/issues/2457 uncovered a weakness in how go code is generated. The result was invalid but `jsii-pacmak` successfully finished. 

### Proposed Solution

By executing `go build` after code generation, we can ensure that the output compiles. Otherwise, this would have only be discovered when the module is consumed, which is definitely not a great experience.

eladb:
To ensure the generated go code is formatted and compiles by executing `go fmt`
and `go build` on the output directory.

Similar to other languages, in order to be able to build the go module within mono-repos (and in the jsii repo itself), we need to be able to resolve local dependencies. This is done in the following way:

1. Create a list of `dist/go` candidates (based on the `jsii.outdir` relative to dependency package dirs)
2. Add to this list the current `outDir`, in case `--outdir` is used
3. Iterate over the module's dependencies (_recursively_) and for each candidate `dist` directory, check if we can find a `<goModule>/go.mod` file that contains a `module <moduleName>` directive.
4. If we find one, we append a `replace` directive with a relative path to our own `go.mod` which tells `go fmt` and `go build` to find the dependency locally.
5. At the end, we restore the original `go.mod` file so that the published one will not include the `replace` directives.

TESTING: this is primarily tested using the `test:build` target of `jsii-pacmak` which generates & builds the go code for the "calc" fixture (in all pacmak execution variants).

NOTE: running `go fmt` on the generated sources turned out to be problematic because sources are generated in parallel by default, and therefore it was not possible to resolve local dependencies at that time (which is required since `go fmt` processes `go.mod`).

Resolves #2463

---


MrArnoldPalmer (approved):
Code looks good. Wondering if we should exclude 'go build' if the code-only option is set? Formatting would still be wanted so resolving dependencies is still necessary.

eladb:
> Code looks good. Wondering if we should exclude 'go build' if the code-only option is set? Formatting would still be wanted so resolving dependencies is still necessary.

Currently both `fmt` and `build` are executed only when we are actually building. I can look into resolving deps in the codegen case but I am currently leaning towards not formatting at all for now.

packages/jsii-pacmak/package.json
```diff
@@ -53,6 +53,7 @@
   "devDependencies": {
     "@jsii/dotnet-runtime": "^0.0.0",
     "@jsii/java-runtime": "^0.0.0",
+    "@jsii/go-runtime": "^0.0.0",
```
RomainMuller, line 56: The tests for the `go` runtime are similar to those of the `python` runtime... Does this not create a dependency cycle? (`@jsii/go-runtime` -> `jsii-pacmak` -> `@jsii/go-runtime`)

eladb: Didn't get any error to that affect. Do we have protection against cycles?

packages/@scope/jsii-calc-base/lib/index.ts
```diff
@@ -1,21 +1,34 @@
...
-    bar(): void;
-}
\ No newline at end of file
+  bar(): void;
+}
+
+/**
+ * Hides the transitive dependency of base-of-base
+ */
+export class StaticConsumer {
```
eladb, line 30: NOTE: this is required to keep `jsii-calc-base-of-base` a transitive dep of `jsii-calc`

packages/jsii-pacmak/lib/targets/go.ts
```diff
@@ -27,16 +33,91 @@ export class Golang extends Target {
...
 
 class GoGenerator implements IGenerator {
   private assembly!: Assembly;
+  public rootPackage!: RootPackage;
+
   private readonly code = new CodeMaker({
     indentCharacter: '\t',
     indentationLevel: 1,
   });
+
```
eladb, line 120:
```suggestion
```

packages/jsii-pacmak/package.json
MrArnoldPalmer: I assume this would break if there were circular deps with actual runtime `require()` statements. Since these are dev deps that are used in various non-parallel build/test phases I think its fine?

MrArnoldPalmer (approved):

packages/jsii-pacmak/lib/targets/go.ts
```diff
@@ -27,16 +33,91 @@ export class Golang extends Target {
...
+    ];
+
+    // try to resolve @jsii/go-runtime (only exists as a devDependency)
+    const jsiiRuntimeDir = tryFindLocalRuntime();
+    if (jsiiRuntimeDir) {
+      replace[JSII_RT_MODULE_NAME] = jsiiRuntimeDir;
+    }
+
+    // iterate (recursively) on all package dependencies and check if we have a
+    // local build directory (under ~/.jsii/go-build-cache) for this module. if
```
eladb, line 76:
```suggestion
    // local build directory for this module. if
```

packages/jsii-pacmak/lib/targets/go.ts
```diff
@@ -27,16 +33,91 @@ export class Golang extends Target {
...
+    const visit = (pkg: RootPackage) => {
+      for (const dep of pkg.packageDependencies) {
+        for (const baseDir of dirs) {
+          const moduleDir = tryFindLocalModule(baseDir, dep);
+          if (moduleDir) {
+            replace[dep.goModuleName] = moduleDir;
+          }
+        }
+
+        // recurse to transitive deps
```
eladb, line 88:
```suggestion
        // recurse to transitive deps ("replace" is only considered at the top level go.mod)
```

packages/jsii-pacmak/lib/targets/go.ts
```diff
@@ -94,3 +177,59 @@ class GoGenerator implements IGenerator {
...
+    const dir = path.dirname(p);
+    logging.debug(`Using @jsii/go-runtime from ${dir}`);
+    return dir;
+  } catch {
+    return undefined;
+  }
+}
+
+/**
+ * Executes a go CLI command against a `local.go.mod` file.
```
eladb, line 222:
```suggestion
 * Executes a go CLI command.
```

packages/jsii-pacmak/lib/targets/go.ts
```diff
@@ -94,3 +177,59 @@ class GoGenerator implements IGenerator {
...
+ *
+ * Since not all `go` commands support specifying an alternative `go.mod` file,
+ * we will copy `local.go.mod` over it and then restore the original version
+ * after execution.
```
eladb, line 226:
```suggestion
```

packages/jsii-calc/package.json
```diff
@@ -37,18 +37,16 @@
     "watch": "jsii --project-references -w",
     "lint": "eslint . --ext .js,.ts --ignore-path=.gitignore",
     "lint:fix": "yarn lint --fix",
-    "test": "node test/test.calc.js && diff-test test/assembly.jsii .jsii",
+    "test": "(ls test/test.*.js | xargs -n1 node) && diff-test test/assembly.jsii .jsii",
```
RomainMuller, line 40: Can we not just use `jest` like the grown ups do? This `ls | xargs` contraption is not very Windows-friendly.

packages/jsii-pacmak/lib/targets/go.ts
```diff
@@ -27,12 +33,86 @@ export class Golang extends Target {
...
+
+    // write `local.go.mod`
+
+    // read existing content
+    const goMod = path.join(pkgDir, GOMOD_FILENAME);
+    const lines = (await fs.readFile(goMod, 'utf-8')).split('\n');
+
+    for (const [from, to] of Object.entries(replace)) {
+      logging.info(`Local replace: ${from} => ${to}`);
+      lines.push(`replace ${from} => ${to}`);
```
RomainMuller, line 103: Somehow I thought you needed to have the version in the `replace` directive (`replace pkg version => localPath`), and I'm kinda glad I was wrong :)

packages/jsii-pacmak/lib/targets/go.ts
```diff
@@ -94,3 +176,56 @@ class GoGenerator implements IGenerator {
...
+  return path.resolve(path.dirname(gomodPath));
+}
+
+/**
+ * Check if we are running from inside the jsii repository, and then we want to
+ * use the local runtime instead of download from a released version.
+ */
+function tryFindLocalRuntime() {
+  try {
+    const p = require.resolve('@jsii/go-runtime/jsii-runtime-go/go.mod');
```
RomainMuller, line 211: The 'idiom' has been to have an actual `js` entry point for the private runtime package, and have this provide the path the relevant objects...

`require('@jsii/go-runtime').runtimePath` or something.

The "indirect" coupling with the location of specific filenames in other packages has been biting us several times in the past.

packages/jsii-pacmak/package.json
RomainMuller: I mean I'm literally just curious how `learn` manages to build a topological sort in order to build things in dependency orders. As long as "it works", this is fine by me!

RomainMuller (changes requested):

packages/jsii-pacmak/lib/targets/go.ts
eladb: No need for local "replace"

packages/jsii-pacmak/lib/targets/go.ts
eladb: Sure I can change that

packages/jsii-calc/package.json
eladb: Do we have a goal to be able to build jsii on Windows?

packages/jsii-pacmak/lib/targets/go.ts
eladb: Done

packages/jsii-calc/package.json
eladb: Deferring to a future commit

RomainMuller (approved):

PR #2492, opened 2021-01-28
feat(go): version file in the generated module directory

iliapolo:
Since golang version isn't available anywhere inside the module, we embed a version file so that consumers can access the version. 

This is needed for publishing [go modules](https://github.com/aws/jsii-release/pull/23) using `jsii-release` without the user having to supply the version externally.

---


eladb (approved):
Looks solid

RomainMuller (approved):

Issue #2506, opened 2021-02-02
go: module names must include major version for v2 and above


eladb:
## :bug: Bug Report

When trying to take a dependency on a go module with a major version other then v0 or v1, we get the following error when trying to `go get` the module:

```
require github.com/aws/constructs-go/constructs: version "v3.3.999" invalid: module contains a go.mod file, so major version must be compatible: should be v0 or v1, not v3
```

### Affected Languages

- [x] `golang`

### General Information
* **JSII Version:** 1.18.0
* **Platform:**     any

### What is the problem?

According to the [go modules spec](https://golang.org/ref/mod#major-version-suffixes), the module name must implicitly include the major version of the module. For example, instead of `github.com/aws/constructs-go/constructs` the module name should be `github.com/aws/constructs-go/constructs/v3` (everywhere where the module is referenced).

> Starting with major version 2, module paths must have a major version suffix like /v2 that matches the major version. For example, if a module has the path example.com/mod at v1.0.0, it must have the path example.com/mod/v2 at version v2.0.0.
> Major version suffixes are not allowed at major versions v0 or v1. There is no need to change the module path between v0 and v1 because v0 versions are unstable and have no compatibility guarantee. Additionally, for most modules, v1 is backwards compatible with the last v0 version; a v1 version acts as a commitment to compatibility, rather than an indication of incompatible changes compared with v0.

Some more context on this [here](https://research.swtch.com/vgo-import)

PR #2507, opened 2021-02-02
fix(go): major version suffix is missing in module names for >=v2

### LINKED ISSUE #2509: go: local.go.sum is accidentally published
### Affected Languages

- [x] `golang`

### General Information
* **JSII Version:** 1.18.0
* **Platform:**     

### What is the problem?

The file `local.go.sum` should not be included in the published directory.

eladb:
According to the [go modules spec], starting from 2.0.0, module names must include
A “major version suffix” (e.g. `/v2`).

[go modules spec]: https://golang.org/ref/mod#major-version-suffixes

Fixes https://github.com/aws/jsii/issues/2506

This commit also fixes #2509 by deleting the `local.go.sum` file from the output.

---

### DIFF
--- a/packages/jsii-pacmak/lib/targets/go.ts
+++ b/packages/jsii-pacmak/lib/targets/go.ts
@@ -44,8 +44,10 @@ export class Golang extends Target {
     // run `go build` with local.go.mod
     await go('build', ['-modfile', localGoMod], { cwd: pkgDir });
 
-    // delete local.go.mod from the output directory so it doesn't get published
-    await fs.unlink(path.join(pkgDir, localGoMod));
+    // delete local.go.mod and local.go.sum from the output directory so it doesn't get published
+    const localGoSum = `${path.basename(localGoMod, '.mod')}.sum`;
+    await tryUnlink(path.join(pkgDir, localGoMod));
+    await tryUnlink(path.join(pkgDir, localGoSum));
   }
 
   /**
@@ -228,4 +230,12 @@ function tryFindLocalRuntime() {
 async function go(command: string, args: string[], options: { cwd: string }) {
   const { cwd } = options;
   return shell('go', [command, ...args], { cwd });
+}
+
+async function tryUnlink(filePath: string) {
+  if (!(await fs.pathExists(filePath))) {
+    return;
+  }
+
+  await fs.unlink(filePath);
 }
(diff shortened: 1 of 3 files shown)


packages/jsii-pacmak/lib/targets/go.ts
```diff
@@ -229,3 +231,11 @@ async function go(command: string, args: string[], options: { cwd: string }) {
...
   return shell('go', [command, ...args], { cwd });
 }
+
+async function tryUnlink(filePath: string) {
+  if (!(await fs.pathExists(filePath))) {
+    return;
+  }
+
+  await fs.unlink(filePath);
+}
```
RomainMuller, line 241: If we have `fs-extra` available, you could use `remove` from it instead of implementing this...

packages/jsii-pacmak/lib/targets/go/package.ts
```diff
@@ -336,9 +340,26 @@ export class InternalPackage extends Package {
...
+      `Unable to parse version "${version}" as a semantic version`,
+    );
+  }
+
+  // suffix is only needed for 2.0 and above
+  if (sv.major <= 1) {
+    return '';
+  }
+
+  return `/v${sv.major.toString()}`;
```
RomainMuller, line 364: I reckon the `.toString()` is not necessary here, since this is in a template expression (this will happen by default).

packages/jsii-pacmak/lib/targets/go/package.ts
```diff
@@ -336,9 +340,26 @@ export class InternalPackage extends Package {
...
+
+function determineMajorVersionSuffix(version: string) {
+  const sv = semver.parse(version);
+  if (!sv) {
+    throw new Error(
+      `Unable to parse version "${version}" as a semantic version`,
+    );
+  }
+
+  // suffix is only needed for 2.0 and above
```
RomainMuller, line 359: Can we get a link to the go documentation about this?

RomainMuller (commented):
Mostly good, minor suggestions...

packages/jsii-pacmak/lib/targets/go.ts
eladb: Done.

packages/jsii-pacmak/lib/targets/go/package.ts
eladb: Done.

packages/jsii-pacmak/lib/targets/go/package.ts
eladb: Done

RomainMuller (approved):

PR #2519, opened 2021-02-03
fix(go): invalid major version suffix for submodules and jsii init

eladb:
The major version suffix introduced in #2507 was added at the end of the full module name, but for submodules, the suffix is needed only after the root package name.

Additionally, the `_init_` import failed to include the major version suffix as well.

In order to cover these cases in tests, change the version of `jsii-calc` to `3.20.120` and `@scope/jsii-calc-base-of-base` to `2.1.1` so that it will be a MV larger then 2.0 and we can verify this works both from top-level modules, submodules and transitive local dependencies.

Fixes #2507

---

### DIFF
--- a/packages/jsii-calc/package.json
+++ b/packages/jsii-calc/package.json
@@ -1,6 +1,6 @@
 {
   "name": "jsii-calc",
-  "version": "0.0.0",
+  "version": "3.20.120",
   "private": true,
   "description": "A simple calcuator built on JSII.",
   "stability": "stable",
--- a/packages/jsii-pacmak/test/generated-code/__snapshots__/target-go.test.ts.snap
+++ b/packages/jsii-pacmak/test/generated-code/__snapshots__/target-go.test.ts.snap
@@ -19,7 +19,7 @@ go 1.15
 
 require (
 	github.com/aws/jsii-runtime-go v0.0.0
-	github.com/aws/jsii/jsii-calc/go/scopejsiicalcbaseofbase v0.0.0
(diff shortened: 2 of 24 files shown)


packages/jsii-pacmak/test/generated-code/__snapshots__/target-go.test.ts.snap
```diff
@@ -1583,10 +1583,10 @@ package jsiicalc
 
 import (
 	_jsii_ "github.com/aws/jsii-runtime-go"
-	_init_ "github.com/aws/jsii/jsii-calc/go/jsiicalc/jsii"
+	_init_ "github.com/aws/jsii/jsii-calc/go/jsiicalc/v3/jsii"
 	"reflect"
 	"github.com/aws/jsii/jsii-calc/go/scopejsiicalclib"
-	"github.com/aws/jsii/jsii-calc/go/jsiicalc/composition"
+	"github.com/aws/jsii/jsii-calc/go/jsiicalc/v3/composition"
```
eladb, line 1589: submodule

packages/jsii-calc/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "jsii-calc",
-  "version": "0.0.0",
+  "version": "3.20.120",
```
RomainMuller, line 3: You could have chosen `3.2.2021` (today's date 🙃) but I guess this is as good as any other pick :)

RomainMuller (approved):

packages/jsii-calc/package.json
eladb: Lol didn't even realize

PR #2535, opened 2021-02-04
fix(go): enums are not encoded/decoded correctly

### LINKED ISSUE #2534: go: enums not working - Expected enum value, got "<string>"
### Affected Languages

- [x] `Golang`
- [ ] `Python`
- [ ] `Java`
- [ ] .NET (`C#`, `F#`, ...)

### General Information
* **JSII Version:** 1.20.0
* **Platform:**     *any*

### What is the problem?

Enums are not working.

Consider the following jsii code packaged as `github.com/eladb/go-enum-repro/enumrepro`:

```ts
export enum MyEnum {
  MY_VALUE = 'my',
  YOUR_VALUE = 'your'
}

export class Hello {
  public sayHello(x: MyEnum) {
    return `hello, ${x}!`;
  }
}
```

And the consuming go code:

```go
package main

import (
	"github.com/eladb/go-enum-repro/enumrepro"
)

func main() {
	var hello = enumrepro.NewHello()
	hello.SayHello(enumrepro.MyEnumYourValue)
}
```

When running this program, we get the following error:

```
panic: "Expected enum value, got \"YOUR_VALUE\""

goroutine 1 [running]:
github.com/aws/jsii-runtime-go.Invoke(0x1124220, 0x1376d60, 0x11461c8, 0x8, 0xc00015bdf0, 0x1, 0x1, 0x1376d01, 0x111c400, 0xc000104720, ...)
        /Users/benisrae/go/pkg/mod/github.com/aws/jsii-runtime-go@v1.20.0/runtime.go:87 +0x305
github.com/eladb/go-enum-repro/enumrepro.(*Hello).SayHello(0x1376d60, 0x11466ca, 0xa, 0x111e4a0, 0xc000102058)
        /Users/benisrae/playground/play-0204133257/enum-repro/dist/go/enumrepro/enumrepro.go:34 +0x165
main.main()
        /Users/benisrae/playground/play-0204133257/enum-repro/dist/go/test/main.go:9 +0x4d
```
- [ ] blocked by https://github.com/aws/jsii/issues/2561

eladb:
Enum values are represented in the jsii kernel as `{ "$jsii.enum": "fqn/Member" }`. To that end, generate custom marshaling and unmarshalling code for enum types.

Marshaling is mostly straightforward: use the value of the Go constant to render the `$jsii.enum` object.

In order to for the custom unmarshling logic to be triggered, we need to call `json.Unmarshal` with a pointer to the specific enum type. Luckily, we know it since jsii APIs are strongly-typed. To that end, change the code of `castAndSetToPtr` to round-trip to JSON and back with the specific return value pointer, which encodes the enum type. This will allow custom unmarshlers in Go to be used.

Added a bunch of tests to verify.

Fixes #2534

---

Additional tests:

- [ ] Nested enum inside of struct
- [ ] Optional enum

---

### DIFF
--- a/packages/@jsii/go-runtime/jsii-runtime-go/api.go
+++ b/packages/@jsii/go-runtime/jsii-runtime-go/api.go
@@ -89,6 +89,11 @@ type objref struct {
 	InstanceID string `json:"$jsii.byref"`
 }
 
+// EnumRef is a reference to a jsii enum member
+type EnumRef struct {
+	MemberFqn string `json:"$jsii.enum"`
+}
+
 type loadRequest struct {
 	kernelRequester
 
--- a/packages/@jsii/go-runtime/jsii-runtime-go/runtime.go
+++ b/packages/@jsii/go-runtime/jsii-runtime-go/runtime.go
@@ -1,6 +1,7 @@
 package jsii
 
 import (
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"os"
@@ -286,8 +287,18 @@ func castAndSetToPtr(ptr interface{}, data interface{}, implMap implementationMa
 		client := getClient()
 		client.objects[ptrVal.Interface()] = ref.InstanceID
 	} else {
-		val := reflect.ValueOf(data)
-		ptrVal.Set(val)
+		// Attempt to unmarshal (again) with the concrete return type. This gives
+		// custom unmarshal logic (e.g. for $jsii.enums) an opportunity to be discovered
+		// based on the actual return type.
+		bytes, err := json.Marshal(data)
+		if err != nil {
+			panic(err)
+		}
+
+		err = json.Unmarshal(bytes, ptr)
+		if err != nil {
+			panic(err)
+		}
 	}
 }
 
--- a/packages/jsii-pacmak/lib/targets/go/types/enum.ts
+++ b/packages/jsii-pacmak/lib/targets/go/types/enum.ts
@@ -1,4 +1,5 @@
-import { EnumType } from 'jsii-reflect';
+import { CodeMaker, toPascalCase } from 'codemaker';
+import { EnumMember, EnumType } from 'jsii-reflect';
 
 import { EmitContext } from '../emit-context';
 import { Package } from '../package';
@@ -11,6 +12,10 @@ export class Enum extends GoType {
 
   public constructor(pkg: Package, public type: EnumType) {
     super(pkg, type);
+
+    this.stdImports.add('encoding/json');
+    this.stdImports.add('fmt');
+    this.stdImports.add('errors');
   }
 
   public emit(context: EmitContext) {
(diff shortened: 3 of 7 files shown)


packages/@jsii/go-runtime/jsii-runtime-go/runtime.go
```diff
@@ -286,8 +287,18 @@ func castAndSetToPtr(ptr interface{}, data interface{}, implMap implementationMa
...
+		// based on the actual return type.
+		bytes, err := json.Marshal(data)
+		if err != nil {
+			panic(err)
+		}
+
+		err = json.Unmarshal(bytes, ptr)
+		if err != nil {
+			panic(err)
+		}
```
MrArnoldPalmer, line 301: I'm not sure why enum types wouldn't be caught by first pass deserialization unless we weren't declaring the type signature correctly in certain places. Are they being seen as target type `interface{}` on the first pass or "string" (in the case of string enums?). If the latter then this makes sense to me.

eladb: The `Result` field in the "Invoke" response structure is `interface{}` (which makes sense), and that's the type we unmarshal from the wire.

packages/@jsii/go-runtime/jsii-runtime-go/api.go
```diff
@@ -89,6 +89,11 @@ type objref struct {
 	InstanceID string `json:"$jsii.byref"`
 }
 
+// EnumRef is a reference to a jsii enum member
+type EnumRef struct {
+	MemberFqn string `json:"$jsii.enum"`
```
RomainMuller, line 94: Let's try to be consistent with go local naming style, which has acronyms spelled out ALL UPPERCASE.

```suggestion
	MemberFQN string `json:"$jsii.enum"`
```

packages/@jsii/go-runtime/jsii-runtime-go/runtime.go
RomainMuller: That looks a little late to me and I kinda wish we'd have a way that did not involve re-marshaling, then re-unmarshaling here. This is also not gonna work (guessing) for enum entries returned by an `any` method/property?

packages/jsii-pacmak/lib/targets/go/types/enum.ts
```diff
@@ -11,6 +12,10 @@ export class Enum extends GoType {
 
   public constructor(pkg: Package, public type: EnumType) {
     super(pkg, type);
+
+    this.stdImports.add('encoding/json');
+    this.stdImports.add('fmt');
+    this.stdImports.add('errors');
```
RomainMuller, line 18: I'd rather we follow the Law of Demeter here... Actually, since we're in the constructor, and this is the only use of `this.stdImports` (from `GoType`) being accessed yet, it would make sense that the values are provided to the `super` constructor.

This would allow `GoType` to (publicly) expose only a `ReadonlySet`.

packages/jsii-pacmak/lib/targets/go/types/enum.ts
```diff
@@ -25,16 +30,68 @@ export class Enum extends GoType {
...
+      code.close();
+    }
+
+    code.line('}');
+    code.line(
+      `return errors.New(fmt.Sprintf("Invalid enum member FQN '%s' for enum '${this.fqn}'", ref.MemberFqn))`,
+    );
+    code.closeBlock();
+    code.line();
+  }
```
RomainMuller, line 96: Not sure this is necessary (here)... Arguably, all enums are unmarshaled from the same representation (`jsii$enum`), so a "standard" un-marshaler could detect this pattern, and then:

- identify the correct enum type
- coerce the value in that enum type
- profit!

This might however require enums be registered into the runtime library, so it can get the correct `reflect.Type` from a `FQN`. This can be achieved within the `func init() {}` block of the libraries that declare the enums.

packages/@jsii/go-runtime/jsii-runtime-go/api.go
eladb: Done

packages/@jsii/go-runtime/jsii-runtime-go/runtime.go
eladb: Indeed, this won't work if the return type if `any`. Any suggestion on how to tackle?

packages/jsii-pacmak/lib/targets/go/types/enum.ts
eladb: Sure, can do

eladb: Done

eladb:
Superseded by  https://github.com/aws/jsii/pull/2585

PR #2550, opened 2021-02-08
fix(pacmak): TypeError when operating with Worker threads

RomainMuller:
This works around isaacs/node-graceful-fs#204, where `graceful-fs`
cannot be loaded from within `Worker` threads. Since the language
generators are loaded in worker threads whenever they are available,
they cannot use `fs-extra` which itself uses `graceful-fs`.
    
This is hopefully a temporary measure, until isaacs/node-graceful-fs#205
is merged & released.

---


packages/jsii-pacmak/lib/targets/go.ts
```diff
@@ -46,8 +46,16 @@ export class Golang extends Target {
 
     // delete local.go.mod and local.go.sum from the output directory so it doesn't get published
     const localGoSum = `${path.basename(localGoMod, '.mod')}.sum`;
-    await fs.remove(path.join(pkgDir, localGoMod));
-    await fs.remove(path.join(pkgDir, localGoSum));
+    await fs.unlink(path.join(pkgDir, localGoMod)).catch((err) =>
```
eladb, line 49: Please extract to a helper function

packages/jsii-pacmak/lib/targets/go.ts
```diff
@@ -78,27 +86,34 @@ export class Golang extends Target {
...
+                  replace[dep.goModuleName] = moduleDir;
+                }
+              }),
+            ),
+          ).then(() =>
+            // recurse to transitive deps ("replace" is only considered at the top level go.mod)
+            visit(dep),
+          ),
+        ),
+      ).then(() => void undefined);
```
eladb, line 105: Why not use await?

packages/jsii-pacmak/lib/targets/go.ts
```diff
@@ -78,27 +86,34 @@ export class Golang extends Target {
...
+    await visit(this.goGenerator.rootPackage);
 
     // write `local.go.mod`
 
     // read existing content
     const goMod = path.join(pkgDir, GOMOD_FILENAME);
-    const lines = (await fs.readFile(goMod, 'utf-8')).split('\n');
+    const lines = await fs
+      .readFile(goMod, 'utf8')
+      .then((text) => text.split('\n'));
```
eladb, line 116: Does this handle errors properly?

eladb (changes requested):

packages/jsii-pacmak/lib/targets/go.ts
RomainMuller: [`no-await-in-loop`](https://eslint.org/docs/rules/no-await-in-loop) is why I transformed this block. Now I used this particular form because it does not require the outer function closure to be `async`, which itself reduces the cost of it.

This is a bit of a micro-optimisation (I got a reflex of writing those these way now... cuz in some cases it matters - not su much here though). I'm open to re-formulate here if you feel this isn't legible enough.

packages/jsii-pacmak/lib/targets/go.ts
RomainMuller: Yep - `.then` has no `catch` handler, so it'd simply use the original rejection.

eladb (approved):

PR #2554, opened 2021-02-09
fix(rosetta,pacmak): TypeError in node 10 with --experimental-worker

RomainMuller:
isaacs/node-graceful-fs#204 prevents `graceful-fs` (a dependency of
`fs-extra`) from loading in a worker thread when using node < 12. The
issue affects only version `4.2.5`, so pinning both `jsii-rosetta` and
`jsii-pacmak` (both of which may use worker threads when available) to
the last known good version: `4.2.4`, until a fixed release of
`graceful-fs` is released.

---

Note - this replaces the solution introduced in 5822e48 (#2550), which
turned out to not be complete enough (I mistakenly through the issue was
introduced by a recent change in our code, while it was triggered in other
places, actually due to a recent release of `graceful-fs`).

---

### DIFF
--- a/packages/jsii-pacmak/lib/targets/go.ts
+++ b/packages/jsii-pacmak/lib/targets/go.ts
@@ -1,5 +1,5 @@
 import { CodeMaker } from 'codemaker';
-import { promises as fs } from 'fs';
+import * as fs from 'fs-extra';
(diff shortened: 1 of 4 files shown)


eladb (approved):
Much nicer!

PR #2567, opened 2021-02-12
chore(go): runtime-wide type registration

### LINKED ISSUE #2561: Go runtime type registration
## :rocket: Feature Request

We need to re-work the type handling, moving way from the `impMap` passed to the runtimes `Invoke`, `StaticInvoke`, `Get` and `StaticGet` calls, and instead have a client-wide type registry.

This is necessary to enable returning the correct dynamic type in several occasions:
- Declared type is `ParentClass`, dynamic value is `ChildClass extends ParentClass`
- Returning values that implement a collection of interfaces
- ...

This may also be necessary to be able to correctly handle enum values; and to be able to re-hydrate structs correctly.

RomainMuller:
The previous convention of passing of `implementationMap` values at the
call sites for value-returning calls (`Get`, `Invoke`, and their static
counterparts) did not provide sufficient information to properly convert
certain values (JavaScript returning a descendant of the declared return
type, decoding enum members, ability to represent values that implement
multiple interfaces in JavaScript, ...).

In order to address this shortcomings, this mechanism was replaced with
a new `func init(){}` block emitted in each package, which is
responsible for registering the generated types with the runtime
library. This allows the `client` to determine the correct go type to
use when converting arbitrary values returned from JavaScript, as it has
visibility on the complete set of available types, by fully qualified
name.

This minimally changes the conversion of JavaScript values into go
values, so as to minimize the footprit of this change. Of course,
subsequent effort will be made to fully leverage the potential of the
extended information (i.e: correctly processing enum members, ...).

Fixes #2561

---


packages/@jsii/go-runtime/jsii-runtime-go/type-registry.go
```diff
@@ -0,0 +1,117 @@
...
+		return fmt.Errorf("another struct was already registered with %v: %v", iface, existing)
+	}
+
+	t.fqnToType[fqn] = strct
+	t.ifaceToStruct[iface] = strct
+
+	return nil
+}
+
+func (t *typeRegistry) concreteTypeFor(typ reflect.Type) (structType reflect.Type, ok bool) {
```
eladb, line 107: Should this return an `Error` instead of a boolean?

packages/jsii-pacmak/lib/targets/go/package.ts
```diff
@@ -107,6 +107,25 @@ export abstract class Package {
     this.emitTypes(context);
     code.closeFile(this.file);
 
+    if (this.types.length > 0) {
+      const initFile = this.file.replace(/(\.go)$/, '.init$1');
```
eladb, line 111: Maybe use `path.basename(this.file, '.go')` instead?

packages/jsii-pacmak/lib/targets/go/package.ts
```diff
@@ -107,6 +107,25 @@ export abstract class Package {
     this.emitTypes(context);
     code.closeFile(this.file);
 
+    if (this.types.length > 0) {
```
eladb, line 110: Add some commentary on what is this function (and consider extracting to a separate method)

packages/jsii-pacmak/lib/targets/go/package.ts
```diff
@@ -107,6 +107,25 @@ export abstract class Package {
...
+      const initFile = this.file.replace(/(\.go)$/, '.init$1');
+      code.openFile(initFile);
+      code.line(`package ${this.packageName}`);
+      code.line();
+      code.open('import (');
+      code.line('"reflect"');
+      code.line();
+      code.line(`${JSII_RT_ALIAS} "${JSII_RT_MODULE_NAME}"`);
+      code.close(')');
```
eladb, line 119: This seems like a reporting pattern in the Go code gen, maybe extract to a helper method on `Package`?

```ts
this.emitImports('reflect', `${JSII_RT_ALIAS} "${JSII_RT_MODULE_NAME}"`);
```

Or:

```ts
this.emitGoClause('imports', [ ... ]);
```

packages/jsii-pacmak/test/generated-code/__snapshots__/target-go.test.ts.snap
```diff
@@ -198,11 +191,44 @@ func StaticConsumer_Consume(args interface{}) {
...
+
+import (
+	"reflect"
+
+	_jsii_ "github.com/aws/jsii-runtime-go"
+)
+
+func init() {
+	_jsii_.RegisterClass(
+		"@scope/jsii-calc-base.Base",
```
eladb, line 211: Is that the idiomatic way in Go to obtain a reflection type?

eladb (commented):
LGTM, I'll leave it to @MrArnoldPalmer to approve

packages/@jsii/go-runtime/build-tools/gen-calc.ts
```diff
@@ -15,6 +15,7 @@ runCommand(
     '-t',
     'go',
     '-v',
+    '-c',
```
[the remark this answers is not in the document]
RomainMuller: Could not get it to build locally without specifying this... The `go build` would always fail to resolve local dependency overrides, and it fried my brain... This was the easiest way for me to not be blocked by that anymore.

packages/jsii-pacmak/lib/targets/go/package.ts
RomainMuller: Duh I always forget this is a thing... -_-

packages/jsii-pacmak/test/generated-code/__snapshots__/target-go.test.ts.snap
RomainMuller: Yes, it is. There is no way to get a type by name, because a type that isn't actually used (as in - being the type of some declared memory location) is not even present in the compiled program.

packages/@jsii/go-runtime/build-tools/gen-calc.ts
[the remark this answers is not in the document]
eladb: Local deps should work but I don't think it's a huge problem to use code-only here. Maybe use the fully qualified long name

packages/@jsii/go-runtime/jsii-runtime-go/type-registry.go
RomainMuller: That's a good idea.

MrArnoldPalmer (approved):
👍🏻 love it @RomainMuller

PR #2571, opened 2021-02-14
fix(pacmak): examples with multi-line strings produce invalid python docstrings

### LINKED ISSUE #2569: python: newlines in literal strings example code results in invalid syntax
### Affected Languages

- [x] `Python`
- [ ] `Java`
- [ ] .NET (`C#`, `F#`, ...)

### General Information
* **JSII Version:** 1.20.1

### What is the problem?

```ts
/**
 * Hello world, this is cool.
 *
 * @example
 * const s1 = "\nFooBar";
 * const s2 = `Hello
 *   world!`;
 */
export class Hello {
  public sayHello() {
    return 'hello, world!';
  }
}
```

The generated python output looks like this:

```py
class Hello(metaclass=jsii.JSIIMeta, jsii_type="python-repro.Hello"):
    """Hello world, this is cool.

    Example::

        # Example automatically generated without compilation. See https://github.com/aws/jsii/issues/826
        s1 = """
        FooBar"""
        s2 = """Hello
          world!"""
    """

    def __init__(self) -> None:
        jsii.create(Hello, self, [])

    @jsii.member(jsii_name="sayHello")
    def say_hello(self) -> builtins.str:
        return typing.cast(builtins.str, jsii.invoke(self, "sayHello", []))
```

Notice that the `\n` in the example code was converted to `"""` pairs which should be escaped or just left as is since they are part of a string literal.

This was introduced in https://github.com/aws/jsii/pull/2419

eladb:
Since docstrings are now rendered inside single-quoted multiline strings to allow the docstring itself to include example code with double-quoted multiline string literals. Those are the ones produced by rosetta for python.

Fixes #2569

---

### DIFF
--- a/packages/jsii-pacmak/lib/targets/python.ts
+++ b/packages/jsii-pacmak/lib/targets/python.ts
@@ -42,6 +42,10 @@ const requirementsFile = path.resolve(
   'requirements-dev.txt',
 );
 
+// we use single-quotes for multi-line strings to allow examples within the
+// docstrings themselves to include double-quotes (see https://github.com/aws/jsii/issues/2569)
+const DOCSTRING_QUOTES = "'''";
+
 export default class Python extends Target {
   protected readonly generator: PythonGenerator;
 
@@ -1571,9 +1575,9 @@ class PythonModule implements PythonType {
       this.fqn === this.assembly.name &&
       this.package.convertedReadme.trim().length > 0
     ) {
-      code.line('"""');
+      code.line(DOCSTRING_QUOTES);
       code.line(this.package.convertedReadme);
-      code.line('"""');
+      code.line(DOCSTRING_QUOTES);
     }
   }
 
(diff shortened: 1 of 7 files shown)


packages/jsii-pacmak/lib/targets/python.ts
```diff
@@ -42,6 +42,10 @@ const requirementsFile = path.resolve(
   'requirements-dev.txt',
 );
 
+// we use single-quotes for multi-line strings to allow examples within the
+// docstrings themselves to include double-quotes (see https://github.com/aws/jsii/issues/2569)
```
RomainMuller, line 46: That sounds very brittle to me... but okay...

RomainMuller (approved):

packages/jsii-pacmak/lib/targets/python.ts
eladb: I believe this is sufficient to fix this p0.

PR #2577, opened 2021-02-15
chore: increase test timeout to 5 minutes

eladb:
We are seeing sporadic timeouts on Windows

---

### DIFF
--- a/packages/@jsii/runtime/test/playback.test.ts
+++ b/packages/@jsii/runtime/test/playback.test.ts
@@ -54,7 +54,7 @@ function createRecords(): string {
       ),
       env: { ...process.env, JSII_RECORD: records, JSII_NOSTACK: '1' },
       stdio: ['inherit', 'pipe', 'pipe'],
-      timeout: 5 * 60_000, // 5 minutes
+      timeout: 300_000, // 5 minutes
     },
   );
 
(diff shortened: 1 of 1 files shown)


packages/@jsii/runtime/test/playback.test.ts
```diff
@@ -54,7 +54,7 @@ function createRecords(): string {
       ),
       env: { ...process.env, JSII_RECORD: records, JSII_NOSTACK: '1' },
       stdio: ['inherit', 'pipe', 'pipe'],
-      timeout: 60_000, // 1 minute
+      timeout: 5 * 60_000, // 5 minutes
```
RomainMuller, line 57:
```suggestion
      timeout: 300_000, // 5 minutes
```

RomainMuller (changes requested):

packages/@jsii/runtime/test/playback.test.ts
eladb: Why? `5 * 60` feels more readable no?

eladb: (and don't tell me performance)

RomainMuller: `// 5 minutes` is jolly well enough to explain this is 5 times 60 times one thousand milliseonds...

eladb: I won't insist, but from a maintenance perspective, if the future me would need to update this from 5 to 7 minutes, it would be easier if I could just change this number from 5 to 7 instead of having to calculate the total value again.

RomainMuller (approved):

PR #2578, opened 2021-02-15
chore: remove npm deps from dependabot + auto-approve

eladb:
We have a custom workflow which updates deps.

Additionally, add a GitHub workflow to auto-approve PRs with the `pr/auto-approve` label.

---


.github/workflows/auto-approve.yml
```diff
@@ -0,0 +1,13 @@
...
+on: pull_request
+
+jobs:
+  auto-approve:
+    if: contains(github.event.pull_request.labels.*.name, 'pr/auto-approve')
+    runs-on: ubuntu-latest
+    steps:
+    - uses: hmarr/auto-approve-action@v2.0.0
+      with:
+        github-token: "${{ secrets.AUTO_APPROVE_GITHUB_TOKEN }}"
```
RomainMuller, line 13: We already have such a workflow, which does not use `pr/auto-approve`, and instead triggers on the author...

https://github.com/aws/jsii/blob/716a8994dc46cc44d38ae61289a2193b3b16ebec/.github/workflows/dependabot.yml#L1-L28

Why create yet another one?

RomainMuller (changes requested):

.github/workflows/auto-approve.yml
eladb: The `pr/auto-approve` is more generic and can be applied to additional automation tasks. This is what we are normalizing on for the CDK repo.

eladb: Deleted the old one

RomainMuller: Well sure but you are giving away one property of the old workflow with __YOU__ had insisted be added: the ability to only auto-approve changes that have an impact of "minor" or less (i.e: the dependency is verified not to be a major version bump).

eladb: Not sure I understand. Isn't this taken care of by the yarn-update workflow?

RomainMuller (approved):

.github/workflows/auto-approve.yml
eladb: okay, got it. this workflow only auto-approve dependabot PRs for minor versions and not major versions.

PR #2588, opened 2021-02-17
refactor(go): re-organize the go runtime

RomainMuller:
Split the go runtime into multiple submodules, in an attempt to ensure
we do not horizontally couple components that should not be coupled to
each other, have a cleaner API for each component (since unexported
members are simply not visible). This also improves the quality of code
completion is improved as a result, too.

Concretely, this produces the following layout:
- `github.com/aws/jsii-runtime-go` is the client entry point
- `github.com/aws/jsii-runtime-go/api` holds shared type definitions
- `github.com/aws/jsii-runtime-go/kernel` is the kernel's main engine (interface to the `node` child process)
- `github.com/aws/jsii-runtime-go/typeregistry` has the code used to manage type registration

One could image we would add:
- `github.com/aws/jsii-runtime-go/objectregistry` to hold the code that manages the object registry
- `github.com/aws/jsii-runtime-go/nodeprocess` to isolate management of the node process 
- ...

I've also broken down "large" files into smaller units, in particular there is now one file per individual kernel API message in the `github.com/aws/jsii-runtime-go/kernel` package, as this makes it much easier to navigate.

---


eladb (approved):
Blindly approving assuming this is basically just code moves and we want to merge that ASAP so to allow the team to continue to work on this code base.

RomainMuller:
> Blindly approving assuming this is basically just code moves and we want to merge that ASAP so to allow the team to continue to work on this code base.

This is correct!

PR #2591, opened 2021-02-19
fix(java): invalid code when multi-inheriting optional properties

RomainMuller:
The generated code for interfaces includes default implementations for
optional properties (since adding a new optional property is a non
breaking change in TypeScript, this allows us to preserve this property
in Java, too).

However, if two distinct super-interfaces declare the same property, the
child inherits two unrelated default implementations for the same
method, which results in dispatch ambiguity that is illegal.

In order to remove the ambiguity, a new default implementation must be
provided on the child interface, so that the dispatch is unambiguous
again.

Fixes #22556

---


packages/jsii-pacmak/test/generated-code/__snapshots__/target-java.test.ts.snap
```diff
@@ -6642,6 +6979,217 @@ public interface DerivedStruct extends software.amazon.jsii.JsiiSerializable, so
...
+        return null;
+    }
+
+    /**
+     */
+    @software.amazon.jsii.Stability(software.amazon.jsii.Stability.Level.Deprecated)
+    @Deprecated
+    default @org.jetbrains.annotations.Nullable java.lang.String getHoistedTop() {
+        return null;
+    }
```
RomainMuller, line 7006: This right here is what used to not be emitted - leading to `DiamondBottom` inheriting the default implementation from both `DiamondLeft` and `DiamondRight` (both of which hoisted that from their internal parent `InternalDiamondTop`) -- causing illegal dispatch ambiguity.

New default declared here replaces the parent's, and solves the issue.

packages/jsii-pacmak/lib/targets/java.ts
```diff
@@ -911,6 +919,60 @@ class JavaGenerator extends Generator {
...
+        continue;
+      }
+      this.onInterfaceProperty(ifc, spec);
+    }
+
+    function allOptionalProps(this: JavaGenerator, fqn: string) {
+      const type = this.findType(fqn) as spec.InterfaceType;
+      const result: Record<string, spec.Property> = {};
+      for (const prop of type.properties ?? []) {
+        // Adding artifical "overrides" here for code-gen quality's sake.
```
nija-at, line 960: why artificial?

packages/jsii-pacmak/lib/targets/java.ts
```diff
@@ -911,6 +919,60 @@ class JavaGenerator extends Generator {
...
+        result[prop.name] = { ...prop, overrides: type.fqn };
+      }
+      for (const base of type.interfaces ?? []) {
```
nija-at, line 963: Might be useful to add a comment like this

```suggestion
      // Include optional properties of all super interfaces in the result
      for (const base of type.interfaces ?? []) {
```

packages/jsii-pacmak/lib/targets/java.ts
```diff
@@ -911,6 +919,60 @@ class JavaGenerator extends Generator {
...
+        for (const [name, spec] of Object.entries(entry)) {
+          histogram[name] = histogram[name] ?? { spec, count: 0 };
+          histogram[name].count += 1;
+        }
+        return histogram;
+      }, {} as Record<string, { readonly spec: spec.Property; count: number }>);
+
+    const localProps = new Set(ifc.properties?.map((prop) => prop.name) ?? []);
+    for (const { spec, count } of Object.values(inheritedOptionalProps)) {
+      if (count < 2 || localProps.has(spec.name)) {
```
nija-at, line 950: What's the latter half of this check doing?

nija-at (approved):
Looks good!

skinny85 (approved):
LGTM!

I think Niranjan's comments make a lot of sense, no additional ones from me 🙂.

packages/jsii-pacmak/lib/targets/java.ts
RomainMuller: Technically, the property as we fetch it is *not* overridden... We actually generate an override, but per the jsii type model, it's only inherited.

packages/jsii-pacmak/lib/targets/java.ts
RomainMuller: Avoids re-declaring a property that was already overridden locally (that would be illegal).

PR #2596, opened 2021-02-23
feat(go): emit indirect dependencies in go.mod

RomainMuller:
This helps IDEs (eg: GoLand from JetBrains, etc..) be more comfortable
with the codebase, as indirect dependencies are correctly mapped. This
generates code more in line with what `go mod` would have generated if
one had tried to manually generate those packages.

This was manually confirmed to have a fully working experience in GoLand,
where even generated code modules are completely browsable, compiling,
and IntelliSense-able in this configuration; while previously it would be
unable to build and check the generated code, resulting in an extremely
poor experience (with UIs jumping around as the IDE failed to resolve modules).

---

### DIFF
--- a/packages/jsii-pacmak/lib/targets/go/package.ts
+++ b/packages/jsii-pacmak/lib/targets/go/package.ts
@@ -3,6 +3,7 @@ import { Assembly, Type, Submodule as JsiiSubmodule } from 'jsii-reflect';
 import { basename, dirname, join } from 'path';
 import * as semver from 'semver';
 
+import { VERSION } from '../../version';
 import { EmitContext } from './emit-context';
 import { ReadmeFile } from './readme-file';
 import {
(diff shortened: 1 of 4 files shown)


packages/jsii-pacmak/lib/targets/go/package.ts
```diff
@@ -253,17 +254,38 @@ export class RootPackage extends Package {
...
     code.open('require (');
     // Strip " (build abcdef)" from the jsii version
-    code.line(
-      `${JSII_RT_MODULE_NAME} v${this.assembly.jsiiVersion.replace(
-        / .*$/,
-        '',
-      )}`,
-    );
-    for (const dep of this.packageDependencies) {
+    code.line(`${JSII_RT_MODULE_NAME} v${VERSION}`);
```
RomainMuller, line 257: The previous code was actually incorrect - the version of the jsii runtime to link against must correspond to the version of `jsii-pacmak` that generated the code, NOT the version of `jsii` that created the assembly.

eladb (approved):

PR #2600, opened 2021-02-23
feat(go): represent jsii structs as go structs (only)

RomainMuller:
As proposed in aws/aws-cdk-rfcs#292 (this is *Approach 4*), stop
rendering go interfaces to represent jsii structs, and instead only emit
a plain go struct with the flattened list of fields (own + all super
interfaces).

Made the necessary code changes to de-serialize structs returned
by-reference by the `@jsii/kernel` by eagerly fecthing all properties.

Also, implemented the option to offer convenience conversion functions
to easily create a parent type from a child type.

---


packages/jsii-pacmak/test/generated-code/__snapshots__/target-go.test.ts.snap
```diff
@@ -1142,80 +1102,26 @@ import (
...
 // Deprecated.
-type DiamondLeftIface interface {
-	GetHoistedTop() string
-	GetLeft() float64
-}
-
-// Deprecated.
-// Struct proxy
 type DiamondLeft struct {
+	// Deprecated.
```
MrArnoldPalmer, line 1107: Not quite sure how we are getting this double stability banner here, but yeah we shouldn't =)

packages/jsii-pacmak/lib/targets/go/types/type-member.ts
```diff
@@ -99,14 +99,15 @@ export class GoProperty implements GoTypeMember {
     const { code } = context;
     const receiver = this.parent.name;
     const instanceArg = receiver.substring(0, 1).toLowerCase();
+    const resultVar = slugify('r', [instanceArg]);
 
     code.openBlock(
       `func (${instanceArg} *${receiver}) ${
         this.getter
-      }()${` ${this.returnType}`}`,
+      }()${` (${resultVar} ${this.returnType})`}`,
```
MrArnoldPalmer, line 107: Whats the value of using a named return value here instead of letting the function body decide what the return var name is? I don't see the need for the parent data structure to care. I find named return values to generally be confusing. When someone sees `return`, it looks a lot like returning nil unless you go back up. Not that anyone is looking at our generated code but still.

I'm not super opinionated here if you want to keep them, though if you do lets also convert the `StaticGet` implementation to do the same thing. In the future I'd say all of the generated functions with return values should do it if we choose to go that way.

MrArnoldPalmer (changes requested):
Only immediate concern is removing the duplicate `deprecated` docs.

packages/jsii-pacmak/lib/targets/go/types/type-member.ts
eladb: Yeah if we can avoid this quirky idiom I'd be in favor.

RomainMuller: I guess the go style guide advises to avoid naming return values unless there's > 1 of them, so I guess fine...

The idea here was to actually avoid duplicating the return type signature - which I thought made it nicely apparent that the jsii runtime library call was actually putting it's result there (and that was __the__ result). But two of you agreeing that this isn't quite readable in this way (at least not without explanation) is enough to convince me otherwise 😅

packages/jsii-pacmak/test/generated-code/__snapshots__/target-go.test.ts.snap
RomainMuller: Well - the "member" classes were inconsistent - some emit docs, other don't. An easy fix :)

MrArnoldPalmer (approved):

PR #2603, opened 2021-02-24
feat(go): require go 1.16, use native embed

RomainMuller:
Instead of generating code to build byte slices with the embedded
assets, upgraded the baseline go requirement to go 1.16, and use the new
native embed feature released as part of this go release.

This includes the update to `jsii/superhain` to the correct go release,
updates to the `.github/workflows/main.yml` workflow to ensure the
correct go version is installed in PR validation builds, and the changes
in `jsii-pacmak` and `@jsii/go-runtime` to leverage the feature.

This makes the code easier to deal with by IDEs (the byte slices were
otherwise very large and caused IDEs to occasionally choke on them),
and removes the risk we make a mistake in generating the slices. It
also makes the generated code easier to troubleshoot, as the tarball
is present as-is, and can be inspected without additional hoop jumping.

See: https://pkg.go.dev/embed

---


RomainMuller:
Note - this has a bootstrap paradox for the CodeBuild PR validation, as the `jsii/superchain` it uses would ship with a pre-1.16 go release. I'd manually solve this if/when we get consensus that this is something we want to move forward with.

packages/jsii-pacmak/lib/targets/go/package.ts
```diff
@@ -344,7 +348,9 @@ export class RootPackage extends Package {
     importGoModules(code, toImport);
 
     code.line();
-    code.line('var once sync.Once');
+    code.line(`//go:embed ${tarballName(this.assembly)}`);
```
MrArnoldPalmer, line 351: wow, go:embed is weird

MrArnoldPalmer (approved):

RomainMuller:
I have manually resolved the bootstrap paradox at this stage (I think)

PR #2604, opened 2021-02-24
feat(pacmak): emit LICENSE file with SPDX license text, NOTICE file

RomainMuller:
In order to properly attribute a license to the generated code, so it
shows up on the godoc.org website, we must produce a `LICENSE` file with
the appropriate content.

We now copy the `LICENSE` file from the source package's root if one
exists, otherwise we attempt to generate one using the SPDX standard
license text.

We also copy the `NOTICE` file from the source package's root if one
exists (but there is no fallback here, as we cannot infer the contents
of it).

See: https://pkg.go.dev/license-policy

This was applied uniformly across all existing bindings.

---


packages/jsii-pacmak/test/generated-code/__snapshots__/target-go.test.ts.snap
```diff
@@ -246,11 +324,89 @@ exports[`Generated code for "@scope/jsii-calc-base-of-base": <outDir>/ 1`] = `
...
+
+END OF TERMS AND CONDITIONS
+
+APPENDIX: How to apply the Apache License to your work.
+
+To apply the Apache License to your work, attach the following boilerplate notice, with the fields enclosed by brackets "[]" replaced with your own identifying information. (Don't include the brackets!)  The text should be enclosed in the appropriate comment syntax for the file format. We also recommend that a file or class name and description of purpose be included on the same "printed page" as the copyright notice for easier identification within third-party archives.
+
+Copyright [yyyy] [name of copyright owner]
```
eladb, line 394: Is it required to replace this with year and owner?

RomainMuller: No - the license file should NOT be modified, ever. This is actually the template for the NOTICE file that should accompany the source code (either in a NOTICE file, or in source file headers, or both)...

I guess we should copy any NOTICE file present in the source package though... this would be a nice touch.

eladb (approved):

PR #2606, opened 2021-02-24
fix(python): accept Sequence[T] for array parameters

RomainMuller:
Instead of typing array (list) parameters as `typing.List[T]`, using the
more flexible `typing.Sequence[T]` type, as recommended by the `typing`
package documentation for this use-case.

Using `List[T]` is inconvenient as it is invariant ([ref]), and requires
very intentional typing, which elads to boilerplate. `Sequence[T]` is
variant, which removes the need for boilerplate, as long as all items
in the sequence match the expected type.

[ref]: https://mypy.readthedocs.io/en/latest/common_issues.html#variance

This was reported in aws/aws-cdk#13203

---


packages/jsii-pacmak/test/generated-code/harness.ts
```diff
@@ -67,7 +67,7 @@ export function verifyGeneratedCodeFor(
 
       expect({ [TREE]: checkTree(outDir) }).toMatchSnapshot('<outDir>/');
 
-      if (targetName !== TargetName.PYTHON) {
+      if (targetName !== TargetName.PYTHON || process.env.SKIP_MYPY_CHECK) {
```
RomainMuller, line 70: The Mypy check adds between 40 and 100 seconds to the execution... Disabling it temporarily can help iterating much faster...

MrArnoldPalmer (approved):

PR #2612, opened 2021-02-25
refactor(go): stop emitting unused fields in go structs

RomainMuller:
Go struct fields used to be emitted for all properties of any type,
including for go structs merely there to act as a proxy to a JS object.
Those properties were never accessed (neither set not read), and could
have been cause for confusion. They additionally needlessly increased
the memory footprint of such data types.

This changes the code generation to no longer emit fields on go structs
that are intended as proxies to JS objects, instead embedding the base
type(s) (super-class and/or super-interfaces) to properly inherit their
methods without duplicating every declaration; and uses a `byte`
placeholder when needed to ensure no 0-width struct is ever generated
(go would have those take 0 bytes of memory, and as a consequence, any
such vlaue is deemed equal to any other, which is not what we want).

In cases when a type descends from more than one direct parent, and
any of their (transitive) bases belongs to a different assembly, all members,
including inherited, are re-implemented to avoid the risk of a conflict
occurring if two parents bring in the same declaration (then automated
promotion cannot happen anymore).

This implied making a change to the method registration functions, such
that each type (class and interface) registers a proxy `maker`
function that is used by the runtime library to properly initialize jsii
proxy instances  of objects without requiring user intervention. This
makes it safe to operate even when embedding data structures from other
libraries. Additionally, it enables hiding the JS proxy structs from the
exported APIs, which further reduces the risk of mis-use.

These changes were proposed in aws/aws-cdk-rfcs#292.

---

### DIFF
--- a/packages/@jsii/go-runtime/jsii-calc-test/main_test.go
+++ b/packages/@jsii/go-runtime/jsii-calc-test/main_test.go
@@ -24,7 +24,7 @@ func TestMain(m *testing.M) {
 
 // Only uses first argument as initial value. This is just a convenience for
 // tests that want to assert against the initialValue
-func initCalculator(initialValue float64) calc.CalculatorIface {
+func initCalculator(initialValue float64) calc.Calculator {
 	return calc.NewCalculator(calc.CalculatorProps{
 		InitialValue: initialValue,
 		MaximumValue: math.MaxFloat64,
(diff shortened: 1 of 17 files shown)


MrArnoldPalmer (approved):
This is a meaty one 🥩 . Can't find anything to complain about. Snapshots look good and codegen feels simpler without the `GoStruct` type from `go-type` and/or the need to deal with name vs interfaceName in type-ref. 👍🏻

Issue #2048, opened 2020-09-26
Go Runtime Handle Callbacks


MrArnoldPalmer:
## :rocket: Feature Request

Runtime methods currently don't handle callback responses from the API. Basically, they should 😄. This is a non-trivial task that kernel request methods need to handle.

The custom `UnmarshalJSON` implementation for runtime response apis may be an appropriate place to distinguish between a callback response and everything else, as this is where error and ok responses are currently deciphered. Decoding the callback should trigger additional runtime request round trips until a complete response is received.

- [x] blocked by https://github.com/aws/jsii/issues/2561

iliapolo:
Emm..not sure what happening here with all these linked issues, I definitely did not intend to close this out.

PR #2614, opened 2021-02-26
feat(go): support direct implementation of jsii interfaces

RomainMuller:
Pure go implementations of jsii interfaces are detected upon being
passed as a parameter to a JavaScript call. When this happens, a
`create` call is sent to the `@jsii/kernel` process with the correct
list of implemented interface FQNs, and all necessary overrides records.
The object is then registered into the new `ObjectStore` for later
reference.

When a callback request interrupts the normal request/response flow, the
`kernel.Client` will handle the callback by getting the appropriate
receiver object, invoking the designated go implementation, and sending
the result to the `@jsii/kernel` process before resuming normal response
handling.

Related to #2048 (partial support)

---

Minor additional content:

- `yarn doc` will start `godoc` (serving the go documentation - including
  for the runtime modules - at `localhost:6060`)
- `yarn lint` now runs `golint` on top of `go vet`, for improved code quality

---


packages/@jsii/go-runtime/jsii-runtime-go/api/api.go
```diff
@@ -64,28 +64,3 @@ type EnumRef struct {
...
-type GetCallback struct {
-	Property string    `json:"property"`
-	ObjRef   ObjectRef `json:"objref"`
-}
-
-type SetCallback struct {
-	Property string      `json:"property"`
-	Value    interface{} `json:"value"`
-	ObjRef   ObjectRef   `json:"objref"`
-}
```
RomainMuller, line 91: Those were all moved & hidden as implementation details of the kernel response unmarshaling.

packages/@jsii/go-runtime/jsii-runtime-go/api/api.go
```diff
@@ -22,17 +22,17 @@ func (o override) isOverride() {
 type MethodOverride struct {
 	override
 
-	Method *string `json:"method"`
-	Cookie *string `json:"cookie"`
+	JsiiMethod string `json:"method"`
+	GoMethod   string `json:"cookie"`
```
RomainMuller, line 26: Renamed for better go-local semantics.

packages/@jsii/go-runtime/jsii-runtime-go/api/api.go
```diff
@@ -22,17 +22,17 @@ func (o override) isOverride() {
...
 
 // PropertyOverride is used to register a "go-native" implementation to be
 // substituted to the default javascript implementation on the created object.
 type PropertyOverride struct {
 	override
 
-	Property *string `json:"property"`
-	Cookie   *string `json:"cookie"`
+	JsiiProperty string `json:"property"`
+	GoGetter     string `json:"cookie"`
```
RomainMuller, line 35: Renamed for better go-local semantics.

packages/@jsii/go-runtime/jsii-runtime-go/kernel/callbacks.go
```diff
@@ -1,20 +1,146 @@
 package kernel
```
RomainMuller, line 1: This file was changed to no longer actually expose a user-accessible API... The reason being there is no use for users to call `Callbacks` directly, as this feature is currently unused, and in any case should be hidden from end-user code, and entirely managed by the runtime library.

This now handles "inline callback requests" received as part of the standard request-response flow, and does not actually handle asynchronous methods anymore (this is unused - and would be re-introduced in the future if needed).

packages/@jsii/go-runtime/jsii-runtime-go/kernel/client.go
```diff
@@ -37,8 +39,8 @@ type client struct {
 	stdin  io.WriteCloser
 	tmpdir string
 
-	types   typeregistry.TypeRegistry
-	objects map[reflect.Value]string
+	types   *typeregistry.TypeRegistry
+	objects *objectstore.ObjectStore
```
RomainMuller, line 43: Re-factored into separate module, as keying on `reflect.Value` resulted in bugs, because they compare `==` if their values are `==`, regardless of whether they are distinct object instances... But this map actually should track individual instances (i.e: pointers).

packages/@jsii/go-runtime/jsii-runtime-go/typeregistry/discover-implemented.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+func (t *TypeRegistry) DiscoverImplementation(vt reflect.Type) (interfaces []api.FQN, overrides []api.Override) {
+	registeredOverrides := make(map[string]bool)
+
+	for fqn, members := range t.typeMembers {
+		iface := t.fqnToType[fqn]
+		if iface.Kind == classType || !vt.AssignableTo(iface.Type) {
+			continue
+		}
```
RomainMuller, line 23: This might be expensive (I don't know). I decided to KISS for now, but we should be able to add a cache layer around this to mitigate the cost.

packages/jsii-pacmak/lib/targets/go/runtime/class-constructor.ts
```diff
@@ -38,10 +40,11 @@ export class ClassConstructor {
...
 
-  public get argsString(): string {
-    const argsList = this.parent.parameters
-      .map((param) => param.name)
-      .join(', ');
-    return `[]interface{}{${argsList}}`;
+  private get argsString(): string {
+    const argsList = this.parent.parameters.map((param) => param.name);
+    if (argsList.length === 0) {
+      return 'nil /* no parameters */';
```
MrArnoldPalmer, line 46: teeny tiny nit:
All the others (no parameters, no overrides etc) emit `nil, // no x`, line comment instead of block, following the `,`.

MrArnoldPalmer (approved):
👍🏻

packages/jsii-pacmak/lib/targets/go/runtime/class-constructor.ts
RomainMuller: Yeah this is different because the comment is emitted before the comma due to how the code is structured...

PR #2636, opened 2021-03-02
fix(go): enums inside structs are not properly serialized

eladb:
Serializes struct values according to the jsii protocol, wrapping
the data in a `$jsii.struct` object recording both the struct's
fully qualified type name, and key-value pairs.

This includes recursively transforming field values, so the
behavior is now correct.

---


RomainMuller (approved):

eladb:
LGTM

PR #2645, opened 2021-03-03
chore(go): make Load call idempotent

RomainMuller:
Making the `Load` runtime library call idempotent makes it so that the
`Client` can be fully disposed (via the `Close` call), while
subsequently getting a new client starte up if needed.

This is in particular useful for implementing test suites that operate
on a clean kernel instance each time.

This entails making sure the registered type information (as in the
`typeregistry.TypeRegistry` instance) survives `*Client.Close()`, as
the `init` functions will not re-trigger.

---


iliapolo (commented):
Looks good. What about a test that makes sure the type registration survives the `Close` method though? We'll have it covered eventually with the compliance suite but I think its worth an independent test since it will have a big affect. 

Maybe add a `runtime_test.go` in `go-runtime-test`?

iliapolo (approved):

PR #2650, opened 2021-03-04
fix(go): nested types are not namespaced

### LINKED ISSUE #2649: go: namespaces with the same name as a class are broken
### Affected Languages

- [x] `GoLang`
- [ ] `Python`
- [ ] `Java`
- [ ] .NET (`C#`, `F#`, ...)

### General Information
* **JSII Version:** 1.24.0
* **Platform:**     *any*

### What is the problem?

```ts
export class Namespace1 {
  public foo() {
    return;
  }
}

export class Namespace2 {
  public foo() {
    return;
  }
}

export namespace Namespace1 {
  export interface Foo {
    readonly bar: string;
  }
}

export namespace Namespace2 {
  export interface Foo {
    readonly baz: string;
    readonly bug: number;
  }
}
```

The go output does not treat the namespaces as submodules and results in duplicate `Foo` type:

```
./duplicatenames.go:79:6: Foo redeclared in this block
        previous declaration at ./duplicatenames.go:42:6
```

The go output looks like this:

```go
// duplicate-names
package duplicatenames

import (
	_jsii_ "github.com/aws/jsii-runtime-go"
	_init_ "github.com/foo/bar/duplicatenames/jsii"
)

type Namespace1 interface {
	Foo()
}

// The jsii proxy struct for Namespace1
type namespace1_jsiiProxy struct {
	_ byte // padding
}

func NewNamespace1() Namespace1 {
	_init_.Initialize()

	n := namespace1_jsiiProxy{}

	_jsii_.Create(
		"duplicate-names.Namespace1",
		nil /* no parameters */,
		[]_jsii_.FQN{},
		nil, // no overrides
		&n,
	)

	return &n
}

func (n *namespace1_jsiiProxy) Foo() {
	_jsii_.InvokeVoid(
		n,
		"foo",
		nil /* no parameters */,
	)
}

type Foo struct {
	Bar string `json:"bar"`
}

type Namespace2 interface {
	Foo()
}

// The jsii proxy struct for Namespace2
type namespace2_jsiiProxy struct {
	_ byte // padding
}

func NewNamespace2() Namespace2 {
	_init_.Initialize()

	n := namespace2_jsiiProxy{}

	_jsii_.Create(
		"duplicate-names.Namespace2",
		nil /* no parameters */,
		[]_jsii_.FQN{},
		nil, // no overrides
		&n,
	)

	return &n
}

func (n *namespace2_jsiiProxy) Foo() {
	_jsii_.InvokeVoid(
		n,
		"foo",
		nil /* no parameters */,
	)
}

type Foo struct {
	Baz string `json:"baz"`
	Bug float64 `json:"bug"`
}
```

RomainMuller:
Nested types were generated in the same package as they parent, without
any namespacing additions, meaning that if two types in the same package
have nested types with the same name, the generated code would be
invalid.

This namespaces the nested types in go by prefixing their names with
their nesting type's name, using an `_` delimiter, which is the same as
what is done for static methods.

Also added a validation in the `jsii` compiler that prohibits that a
nested type and a static method share the same PascalCase
transformation, as this would result in conflicts in Go, but also in C#.

Fixes #2649

---


packages/jsii-pacmak/lib/targets/go/types/go-type.ts
```diff
@@ -14,9 +14,20 @@ export abstract class GoType {
...
+      let parent = type.nestingParent;
+      parent != null;
+      parent = parent.nestingParent
+    ) {
+      this.name = `${toPascalCase(parent.name)}_${this.name}`;
+    }
+
+    // Add "jsiiProxy_" suffix to private struct name to avoid keyword conflicts
+    // such as "default". See https://github.com/aws/jsii/issues/2637
+    this.proxyName = `jsiiProxy_${this.name}`;
```
eladb, line 29: was this intentionally changed to a prefix. if so, update the comment

packages/jsii-pacmak/test/generated-code/examples.test.ts
```diff
@@ -0,0 +1,67 @@
...
+import * as pacmak from '../../lib';
+import { checkTree, TREE } from './harness';
+
+const EXAMPLES_ROOT = path.resolve(__dirname, 'examples');
+
+for (const name of fs.readdirSync(EXAMPLES_ROOT)) {
+  const file = path.join(EXAMPLES_ROOT, name);
+  test(name, async () => {
+    const source = await fs.readFile(file, 'utf8');
+    const compiled = await jsii.compileJsiiForTest(source);
```
eladb, line 16: yap this is what I had in mind

packages/jsii-pacmak/test/generated-code/examples.test.ts
```diff
@@ -0,0 +1,67 @@
...
+        spaces: 2,
+      });
+      await fs.writeJson(path.join(tmpdir, 'package.json'), {
+        name: compiled.assembly.name,
+        version: compiled.assembly.version,
+        jsii: {
+          outdir: 'dist',
+          targets: {
+            go: { moduleName: 'example.test/demo' },
+          },
```
eladb, line 32: why not include all the target types already?

packages/jsii-pacmak/test/generated-code/examples/nested-types.ts
```diff
@@ -0,0 +1,31 @@
+// See: https://github.com/aws/jsii/issues/2649
```
eladb, line 1: wondering if perhaps we should afford for multiple files so basically each example will be in a separate directory?

packages/jsii-pacmak/test/generated-code/examples/nested-types.ts
```diff
@@ -0,0 +1,31 @@
...
+    readonly bar: string;
+    method(): void;
+  }
+}
+
+export namespace Namespace2 {
+  export enum Foo {
+    BAR,
+    BAZ,
+  }
```
eladb, line 30: Add a case of more than one nesting level (your code accounts for this)

eladb (commented):
This looks great! Let's merge that in so I can use for my bugs :-)

packages/jsii-pacmak/lib/targets/go/types/go-type.ts
RomainMuller: Yes it was (it looks better for nested type names). And I changed the comment (from `postfix` to `suffix`, because my brain farted).

packages/jsii-pacmak/test/generated-code/examples.test.ts
RomainMuller: Speed of iteration 🙃

packages/jsii-pacmak/test/generated-code/examples/nested-types.ts
RomainMuller: We could update it so that - if it's a file, it's a single-file example... If it's a directory, it's a multiple-files example. I went for just what I needed at this point.

packages/jsii-pacmak/test/generated-code/examples/nested-types.ts
RomainMuller: Done!

packages/jsii-pacmak/lib/targets/go/types/go-type.ts
```diff
@@ -14,9 +14,20 @@ export abstract class GoType {
 
   public constructor(public pkg: Package, public type: Type) {
     this.name = toPascalCase(type.name);
-    // add "_jsiiProxy" postfix to private struct name to avoid keyword
-    // conflicts such as "default". see https://github.com/aws/jsii/issues/2637
-    this.proxyName = `${toCamelCase(type.name)}_jsiiProxy`;
+
+    // Prefix witht he nesting parent name(s), using an _ delimiter.
```
eladb, line 18:
```suggestion
    // Prefix with the nesting parent name(s), using an _ delimiter.
```

packages/jsii-pacmak/test/generated-code/examples/nested-types.ts
eladb: Okay, we can iterate on this

eladb (approved):

PR #2654, opened 2021-03-05
fix(python): duplicated kwargs when field is multi-inherited

### LINKED ISSUE #2653: Struct multi-inheritance generates wrong Python code
### Affected Languages

- [x] `Python`
- [ ] `Java`
- [ ] .NET (`C#`, `F#`, ...)

### General Information
* **JSII Version:** `1.23.0`
* **Platform:**     all of them

### What is the problem?

After merging https://github.com/aws/aws-cdk/pull/13356 in the CDK repo, JSII generates the following code for the [`forAsset` static method in the `@aws-cdk/aws-lambda` module](https://github.com/aws/aws-cdk/blob/664133a35da2bd096a237971ce662f3dd38b297f/packages/%40aws-cdk/aws-lambda/lib/code.ts#L51-L58):

```python
    @jsii.member(jsii_name="fromAsset") # type: ignore[misc]
    @builtins.classmethod
    def from_asset(
        cls,
        path: builtins.str,
        *,
        readers: typing.Optional[typing.List[aws_cdk.aws_iam.IGrantable]] = None,
        source_hash: typing.Optional[builtins.str] = None,
        exclude: typing.Optional[typing.List[builtins.str]] = None,
        follow: typing.Optional[aws_cdk.assets.FollowMode] = None,
        ignore_mode: typing.Optional[aws_cdk.core.IgnoreMode] = None,
        exclude: typing.Optional[typing.List[builtins.str]] = None,
        follow_symlinks: typing.Optional[aws_cdk.core.SymlinkFollowMode] = None,
        ignore_mode: typing.Optional[aws_cdk.core.IgnoreMode] = None,
        asset_hash: typing.Optional[builtins.str] = None,
        asset_hash_type: typing.Optional[aws_cdk.core.AssetHashType] = None,
        bundling: typing.Optional[aws_cdk.core.BundlingOptions] = None,
    ) -> "AssetCode":
        '''Loads the function code from a local disk path.

        :param path: Either a directory with the Lambda code bundle or a .zip file.
        :param readers: (experimental) A list of principals that should be able to read this asset from S3. You can use ``asset.grantRead(principal)`` to grant read permissions later. Default: - No principals that can read file asset.
        :param source_hash: (deprecated) Custom hash to use when identifying the specific version of the asset. For consistency, this custom hash will be SHA256 hashed and encoded as hex. The resulting hash will be the asset hash. NOTE: the source hash is used in order to identify a specific revision of the asset, and used for optimizing and caching deployment activities related to this asset such as packaging, uploading to Amazon S3, etc. If you chose to customize the source hash, you will need to make sure it is updated every time the source changes, or otherwise it is possible that some deployments will not be invalidated. Default: - automatically calculate source hash based on the contents of the source file or directory.
        :param exclude: (deprecated) Glob patterns to exclude from the copy. Default: nothing is excluded
        :param follow: (deprecated) A strategy for how to handle symlinks. Default: Never
        :param ignore_mode: (deprecated) The ignore behavior to use for exclude patterns. Default: - GLOB for file assets, DOCKER or GLOB for docker assets depending on whether the '
        :param exclude: Glob patterns to exclude from the copy. Default: - nothing is excluded
        :param follow_symlinks: A strategy for how to handle symlinks. Default: SymlinkFollowMode.NEVER
        :param ignore_mode: The ignore behavior to use for exclude patterns. Default: IgnoreMode.GLOB
        :param asset_hash: Specify a custom hash for this asset. If ``assetHashType`` is set it must be set to ``AssetHashType.CUSTOM``. For consistency, this custom hash will be SHA256 hashed and encoded as hex. The resulting hash will be the asset hash. NOTE: the hash is used in order to identify a specific revision of the asset, and used for optimizing and caching deployment activities related to this asset such as packaging, uploading to Amazon S3, etc. If you chose to customize the hash, you will need to make sure it is updated every time the asset changes, or otherwise it is possible that some deployments will not be invalidated. Default: - based on ``assetHashType``
        :param asset_hash_type: Specifies the type of hash to calculate for this asset. If ``assetHash`` is configured, this option must be ``undefined`` or ``AssetHashType.CUSTOM``. Default: - the default is ``AssetHashType.SOURCE``, but if ``assetHash`` is explicitly specified this value defaults to ``AssetHashType.CUSTOM``.
        :param bundling: (experimental) Bundle the asset by executing a command in a Docker container. The asset path will be mounted at ``/asset-input``. The Docker container is responsible for putting content at ``/asset-output``. The content at ``/asset-output`` will be zipped and used as the final asset. Default: - uploaded as-is to S3 if the asset is a regular file or a .zip file, archived into a .zip file and uploaded to S3 otherwise
        '''
        options = aws_cdk.aws_s3_assets.AssetOptions(
            readers=readers,
            source_hash=source_hash,
            exclude=exclude,
            follow=follow,
            ignore_mode=ignore_mode,
            exclude=exclude,
            follow_symlinks=follow_symlinks,
            ignore_mode=ignore_mode,
            asset_hash=asset_hash,
            asset_hash_type=asset_hash_type,
            bundling=bundling,
        )

        return typing.cast("AssetCode", jsii.sinvoke(cls, "fromAsset", [path, options]))
```

This is clearly wrong - the `exclude`  and `ignore_mode` keyword args are duplicated.

I suspect that's because the argument to `fromAsset()`, `AssetOptions` from the `@aws-cdk/aws-s3-assets` module, is currently defined as:

```ts
export interface AssetOptions extends assets.CopyOptions, cdk.FileCopyOptions, cdk.AssetOptions {
  // ...
```

Where both `assets.CopyOptions` and `cdk.FileCopyOptions` define the properties `exclude` and `ignoreMode` (of the same types, `string[]` and `cdk.IgnoreMode`, respectively).

RomainMuller:
When a struct field is inherited from more than one parent (this could
be twice from the same type - in the case of a diamond shape; or from
two distinct parents that declare the same field), the lifted keyword
arguments in python would be duplicated for this field.

This is because the method that performs the keyword argument lifting
did not perform name-based de-duplication, and operates directly on the
"raw" assembly (whereby it must traverse the inheritance tree itself,
as opposed to the Go generator which uses `jsii-reflect` and has the
field collection done by that library).

Added the necessary de-duplication logic and confirmed the produced code
now looks correct using the exact same test harness as I had introduced
in #2650.

Fixes #2653

---


packages/jsii-pacmak/lib/targets/python.ts
```diff
@@ -693,24 +693,36 @@ abstract class BaseMethod implements PythonBase {
     const liftedProperties: spec.Property[] = [];
 
     const stack = [this.liftedProp];
-    let current = stack.shift();
-    while (current !== undefined) {
+    const knownIfaces = new Set<string>();
+    const knownProps = new Set<string>();
+    for (
+      let current = stack.shift();
+      current != null;
```
nija-at, line 700: Why change from `undefined` to `null`. Per [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/shift#description), this will return `undefined` when empty?

Maybe just check for truthiness instead?

packages/jsii-pacmak/package.json
```diff
@@ -64,6 +64,7 @@
     "@types/yargs": "^16.0.0",
     "eslint": "^7.21.0",
     "jest": "^26.6.3",
+    "jsii": "^0.0.0",
```
nija-at, line 67: Intentional? pacmak depends on jsii?

Not familiar with the jsii layering but smells fishy.

packages/jsii-pacmak/test/generated-code/__snapshots__/examples.test.ts.snap
```diff
@@ -0,0 +1,1437 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
nija-at, line 1: Does this require another new snapshot test? 

Did this not manifest in the previous snapshot test for diamond dependencies - https://github.com/aws/jsii/pull/2591/files#diff-305aec203005012473bdfe3d0c5d9c734c7ee318a0e813759c82aed7067323e3 ?

Snapshot tests make it really hard to evaluate if a problem was actually fixed. It's useful to detect unintended regressions but perhaps it shouldn't be the sole testing mechanism.

nija-at (approved):

packages/jsii-pacmak/lib/targets/python.ts
RomainMuller: `x != null` is the same as `x !== nul && x !== undefined`, but more compact.

And I don't like truthiness checks, because this is ripe for bugs.

packages/jsii-pacmak/package.json
RomainMuller: New tests actually use jsii to compile some tiny projects to validate code-gen output. This is very practical.

packages/jsii-pacmak/test/generated-code/__snapshots__/examples.test.ts.snap
RomainMuller: The alternative is to build (I mean compile) the generated code, for possibly many small projects, to ensure they "work" -- this is bound to be very slow/expensive. Also - next up is "but the fact the code compiles is not enough - we have to run it, too", and then it becomes a full time job?

RomainMuller:
@mergifyio refresh

PR #2685, opened 2021-03-14
fix(go): missing imports needed by base members

### LINKED ISSUE #2647: go: missing imports needed by base ctors and reimplemented methods/properties
### Affected Languages

- [x] `GoLang`
- [ ] `Python`
- [ ] `Java`
- [ ] .NET (`C#`, `F#`, ...)

### General Information
* **JSII Version:** 1.24.0
* **Platform:**     *any*

### What is the problem?

`module1.ts`:

```ts
import { Construct } from 'constructs';

export class Base {
  public foo(c: Construct) { }
}
```

`module2.ts`:

```ts
import { Base } from 'module1';

export class Derived extends Base {
}
```

The go code generated for `Derived` will implement all the derived methods from `Base` but it won't take into account the fact that `constructs` needs to be imported.

The output is:

```
undefined: constructs
```

eladb:
Go pacmak will reimplement methods and properties in case the class has more than a single "base" (interface/class). The Go code generator failed to include these members when determining which imports to generate.

Additionally, if a constructor of a base class requires an import, it was also missing.

Add a calc test fixture to verify (failed without this change).

Fixes #2647

---


iliapolo (approved):

packages/jsii-pacmak/lib/targets/go/types/class.ts
```diff
@@ -326,18 +327,18 @@ export class GoClass extends GoType {
   }
 }
 
-export class GoClassConstructor {
+export class GoClassConstructor extends GoMethod {
+  public readonly usesInitPackage = false;
+  public readonly usesRuntimePackage = false;
```
iliapolo, line 332: Why `false`? The constructor actually does use `_init_` and `_jsii_`.

eladb: Good catch. Fixing.

eladb: I believe this was implicitly okay because these are enabled at the class level (for this reason).

PR #2687, opened 2021-03-14
feat(go): packageName and versionSuffix

### LINKED ISSUE #2632: Impossible to control go module name - add 'preview' version module name
Consider revising the go configuration block, and [documenting it](https://aws.github.io/jsii/user-guides/lib-author/configuration/targets/go/).

In particular:
- [ ] The "module name root" should probably be called `repositoryName` instead of `moduleName`
- [ ] The go package name should be user-configurable (with sane defaults), as `packageName`
- [ ] Whether the `packageName` is appended to the `repositoryName` to make the fully qualified module name or not should be user-configurable, too (it makes sense for mono-repos, but otherwise is likely undesirable)

eladb:
Add two features to the `go` targets config in package.json:

- `packageName` can be used to customize the name of the go package (previously this was always derived from the javascript name).
- `versionSuffix` can be used to append a suffix to the version number (e.g. `-devpreview`).

Resolves #2632

This is needed in order to control the name of the AWS CDK go bindings so they are the same between v1 and v2. Currently the v1 name would have been `monocdk` and v2 `awscdklib` and we want both to actually be `awscdk` per the decision in [this RFC](https://github.com/aws/aws-cdk-rfcs/blob/master/text/0079-cdk-2.0.md#releasing-to-package-managers).

---

### DIFF
--- a/gh-pages/content/user-guides/lib-author/configuration/targets/go.md
+++ b/gh-pages/content/user-guides/lib-author/configuration/targets/go.md
@@ -5,10 +5,11 @@
 
 To enable go package generation, add the `go` key to the jsii targets configuration:
 
-- `packageName` (optional) - The name of the Go package name. If this is not
-  specified, the package name will be derived from the npm module name by
-  filtering non-alphanumeric characters (e.g. `aws-cdk-lib` will be
-  `awscdklib`). You can use this field to set a specific package name.
+- `packageName` (optional) - The name of the Go package name. If not specified,
+  package name will be derived from the JavaScript module name by removing
+  non-alphanumeric characters (e.g. `@aws-cdk/aws-s3` will be `awscdkawss3`). If
+  this is set on a submodule config file, it refers to the submodule package
+  name.
 - `moduleName` (required) - The name of the **target repository** in which this
   module will be published (e.g. `github.com/foo/bar`). The module itself will
   *always* be published under a subdirectory named according to the Go package
--- a/gh-pages/content/user-guides/lib-author/configuration/targets/go.md
+++ b/gh-pages/content/user-guides/lib-author/configuration/targets/go.md
@@ -3,15 +3,31 @@
 !!! danger
     The **go** target is currently unstable and not suitable for production use.
 
-To enable go package generation, add the **go** key with an empty object to the jsii targets configuration.
+To enable go package generation, add the `go` key to the jsii targets configuration:
 
-This will add generated go package code to your specified `outDir` for testing and experimentation.
+- `packageName` (optional) - The name of the Go package name. If this is not
+  specified, the package name will be derived from the npm module name by
+  filtering non-alphanumeric characters (e.g. `aws-cdk-lib` will be
+  `awscdklib`). You can use this field to set a specific package name.
+- `moduleName` (required) - The name of the **target repository** in which this
+  module will be published (e.g. `github.com/foo/bar`). The module itself will
+  *always* be published under a subdirectory named according to the Go package
+  name of the module (e.g. `github.com/foo/bar/awscdk`).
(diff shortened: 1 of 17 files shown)


gh-pages/content/user-guides/lib-author/configuration/targets/go.md
```diff
@@ -3,15 +3,31 @@
 !!! danger
     The **go** target is currently unstable and not suitable for production use.
 
-To enable go package generation, add the **go** key with an empty object to the jsii targets configuration.
+To enable go package generation, add the `go` key to the jsii targets configuration:
 
-This will add generated go package code to your specified `outDir` for testing and experimentation.
+- `packageName` (optional) - The name of the Go package name. If this is not
```
iliapolo, line 8: This is confusing to me as go modules can have multiple packages. I would rename to `rootPackageName` and explain what exactly it contains (i.e types that are directly exported from the entry point?)

gh-pages/content/user-guides/lib-author/configuration/targets/go.md
```diff
@@ -3,15 +3,31 @@
...
+  specified, the package name will be derived from the npm module name by
+  filtering non-alphanumeric characters (e.g. `aws-cdk-lib` will be
+  `awscdklib`). You can use this field to set a specific package name.
+- `moduleName` (required) - The name of the **target repository** in which this
```
iliapolo, line 12: Maybe just rename to `repositoryName`?

packages/jsii-pacmak/lib/targets/go/package.ts
```diff
@@ -206,21 +211,26 @@ export class RootPackage extends Package {
...
-    const packageName = goPackageName(assembly.name);
+    const goConfig = assembly.targets?.go ?? {};
+    const packageName = goPackageNameForAssembly(assembly);
     const filePath = '';
-    const moduleName = assembly.targets?.go?.moduleName ?? '';
+    const moduleName = goConfig.moduleName ?? '';
+    let version = assembly.version;
+    if (goConfig.versionSuffix) {
+      version += goConfig.versionSuffix;
+    }
```
iliapolo, line 221:
```suggestion
    const version = `${assembly.version}${goConfig.versionSuffix ?? ''}`;
```

iliapolo (approved):

gh-pages/content/user-guides/lib-author/configuration/targets/go.md
eladb: We wanted to keep this change to a minimum since this is consumed by downstream tools like jsii-release, so decided against changing the current semantics. We can evolve this at a later stage. I understand it's not ideal.

gh-pages/content/user-guides/lib-author/configuration/targets/go.md
eladb: Are you referring to submodules? Because I believe it should be possible to specify the package name for each submodule by placing a jsii config file near the module root. @RomainMuller will just JUST WORK for Go or do we need extra support for this?

gh-pages/content/user-guides/lib-author/configuration/targets/go.md
iliapolo: Fair enough

gh-pages/content/user-guides/lib-author/configuration/targets/go.md
iliapolo: I actually blanked out and thought every subdirectory directory will translate to its own go package, but of course it needs to be explicitly exported from `index.ts`. 

 > Added a test which verifies that this works for submodules, in which case, packageName is a better name because it can be used recursively.

Agree. I didn't know we can specify a nested jsii config, TIL 👍

PR #2699, opened 2021-03-15
fix(go,python/java): bad code for members with same name with different casing

### LINKED ISSUE #2508: go/python: members with same name but different capitalization result in duplicate declarations
### Affected Languages

- [x] Python (second method overrides first)
- [x] `Golang`

### General Information
* **JSII Version:** 1.18.0
* **Platform:**    _all_

### What is the problem?

If a class has two methods with the same name but slightly different capitalization, the go code generator may end up with the same method name and eventually result in a duplicate definition error.

For example, in CDK8s, we had the following:

```ts
  public toIsoString(): string { ... }

  /** @deprecated */
  public toISOString(): string { ... }
```

This resulted in the following error during jsii-pacmak:

```
# github.com/awslabs/cdk8s-go/cdk8s/cdk8s
./cdk8s.go:1074:2: duplicate method ToIsoString
./cdk8s.go:1237:20: (*Duration).ToIsoString redeclared in this block
 	previous declaration at ./cdk8s.go:1224:6
```

eladb:
In TypeScript, it is possible to give two members (methods/properties) the same name but with different capitalization. This results in duplicate members in Go and Python.

In Go, where the member is expected to be PascalCase, use the same conversion used in .NET, in which we only capitalize the first letter (assuming TypeScript uses camelCase). This offers more tolerance.

In Python, where members use snake_case, we implemented a different heuristic in which we only render the non-deprecated member and fail if there is more than one deprecated member.

In Java, this issue existed only for property names.

Fixes #2508

BREAKING CHANGE: if multiple members have the same name with different capitalization, only one is allowed to be non-deprecated. This will currently only manifest when producing python bindings, but will be added as a jsii compiler error in the future.

---


packages/jsii-pacmak/lib/targets/python.ts
```diff
@@ -184,6 +184,61 @@ const setDifference = <T>(setA: Set<T>, setB: Set<T>): Set<T> => {
...
+
+  // now return all the members
+  const ret = new Array<PythonBase>();
+  for (const v of Object.values(map)) {
+    if (v.length !== 1) {
+      throw new Error('assertion failed');
+    }
+
+    ret.push(v[0]);
+  }
```
iliapolo, line 237: Seems like this whole loop can be part of the preceding loop, and offer a clearer error message.

packages/jsii-pacmak/lib/targets/python.ts
```diff
@@ -2911,3 +2967,5 @@ function nestedContext(
         : context.surroundingTypeFqns,
   };
 }
+
+const isDeprecated = (x: PythonBase) => x.docs?.deprecated !== undefined;
```
iliapolo, line 2971: I'd just add `isDeprecated` on `PythonBase` or something.

RomainMuller (approved):

packages/jsii-pacmak/lib/targets/go/types/go-type.ts
```diff
@@ -13,15 +13,15 @@ export abstract class GoType {
   public readonly proxyName: string;
 
   public constructor(public pkg: Package, public type: Type) {
-    this.name = toPascalCase(type.name);
+    this.name = type.name;
```
iliapolo, line 16: How come?

packages/jsii-pacmak/lib/targets/python.ts
```diff
@@ -184,6 +184,61 @@ const setDifference = <T>(setA: Set<T>, setB: Set<T>): Set<T> => {
...
+    }
+
+    // we found more than one member with the same python name, filter all
+    // deprecated versions and check that we are left with exactly one.
+    // otherwise, they will overwrite each other
+    // see https://github.com/aws/jsii/issues/2508
+    const nonDeprecated = list.filter((x) => !isDeprecated(x));
+    if (nonDeprecated.length > 1) {
+      throw new Error(
+        `Multiple non-deprecated members which map to the Python name "${name}"`,
```
iliapolo, line 221: Mention as a breaking change?

packages/jsii-pacmak/lib/targets/go/types/go-type.ts
eladb: jsii type names (enums/classes/interfaces) are already pascal cased, so there is not reason to screw up with them at this point.

packages/jsii-pacmak/lib/targets/python.ts
eladb: Sure

packages/jsii-pacmak/lib/targets/python.ts
eladb: done.

packages/jsii-pacmak/lib/targets/python.ts
eladb: `PythonBase` is an interface 😭

packages/jsii-pacmak/lib/targets/python.ts
```diff
@@ -184,6 +184,57 @@ const setDifference = <T>(setA: Set<T>, setB: Set<T>): Set<T> => {
...
+      // otherwise, they will overwrite each other
+      // see https://github.com/aws/jsii/issues/2508
+      const nonDeprecated = list.filter((x) => !isDeprecated(x));
+      if (nonDeprecated.length > 1) {
+        throw new Error(
+          `Multiple non-deprecated members which map to the Python name "${name}"`,
+        );
+      }
+
+      member = nonDeprecated[0];
```
iliapolo, line 229: I'd still add an `if nonDeprecated.length == 0` check here like you did before.

eladb: good catch

iliapolo (approved):

PR #2703, opened 2021-03-15
 fix(go): embedded bases may conflict with member names

### LINKED ISSUE #2702: go: invalid code generated when a method is named like a base class
### Affected Languages

- [ ] `Python`
- [ ] `Java`
- [ ] .NET (`C#`, `F#`, ...)
- [x] `Go`

### General Information
* **JSII Version:** 1.42.0
* **Platform:**     *any*

### What is the problem?

```ts
// member has the same name as a base class
// @see https://github.com/aws/jsii/issues/2702

import { Base } from '@scope/jsii-calc-base';

export class Class1 extends Base {
  public base() {
    return;
  }
}

export class Class2 extends Base {
  public readonly base = 'hello';
}
```

Errors:

```
#STDERR> module2702/module2702.go:34:9: jsiiProxy_Class1.Base is a field, not a method
#STDERR> module2702/module2702.go:34:9: cannot use &j (type *jsiiProxy_Class1) as type Class1 in return argument:
#STDERR> 	*jsiiProxy_Class1 does not implement Class1 (missing Base method)
#STDERR> module2702/module2702.go:37:6: type jsiiProxy_Class1 has both field and method named Base
#STDERR> module2702/module2702.go:70:6: type jsiiProxy_Class2 has both field and method named Base
#STDERR> module2702/module2702.go:94:9: jsiiProxy_Class2.Base is a field, not a method
#STDERR> module2702/module2702.go:94:9: cannot use &j (type *jsiiProxy_Class2) as type Class2 in return argument:
#STDERR> 	*jsiiProxy_Class2 does not implement Class2 (missing Base method)
```

eladb:
When we generate proxy structs in Go (for classes or interfaces), we embed the base types (base class or interfaces) in the struct. If the base type has a name that is the same as a member of the hosting type, there will be a conflict.

To solve this, we create a unique type alias for each base interface we want to embed.

Fixes #2702

---


packages/jsii-pacmak/lib/targets/go/types/go-type.ts
```diff
@@ -60,7 +61,7 @@ export abstract class GoType {
       const instanceVar = this.proxyName[0];
       code.line(`${instanceVar} := ${this.proxyName}{}`);
       for (const base of bases) {
-        const baseEmbed = base.pkg === this.pkg ? base.proxyName : base.name;
+        const baseEmbed = embedForBase(this.pkg, this.proxyName, base).embed;
         code.line(
           `${JSII_RT_ALIAS}.InitJsiiProxy(&${instanceVar}.${baseEmbed})`,
```
MrArnoldPalmer, line 66: looks like we need to remove the `${instanceVar}.` prefix here?

MrArnoldPalmer (changes requested):
Just fixing the broken references to the new type alias, Otherwise LGTM 👍🏻

packages/jsii-pacmak/lib/targets/go/types/go-type.ts
eladb: Possibly, checking...

eladb: I believe it is required. 

Sample output:

```go
	_jsii_.RegisterClass(
		"jsii-calc.module2702.Class1",
		reflect.TypeOf((*Class1)(nil)).Elem(),
		[]_jsii_.Member{
			_jsii_.MemberMethod{JsiiMethod: "base", GoMethod: "Base"},
			_jsii_.MemberMethod{JsiiMethod: "typeName", GoMethod: "TypeName"},
		},
		func() interface{} {
			j := jsiiProxy_Class1{}
			_jsii_.InitJsiiProxy(&j.jsiiProxy_Class1_jcbBase)
			return &j
		},
	)
```

eladb:
@MrArnoldPalmer I am seeing some runtime errors, so this needs a bit more work.

packages/jsii-pacmak/lib/targets/go/types/go-type.ts
MrArnoldPalmer: You're correct this is required. I thought we were referencing the type here for some reason and not the embedded interface field.

packages/jsii-pacmak/lib/targets/go/package.ts
```diff
@@ -114,16 +122,63 @@ export abstract class Package {
...
+    const exists = this.embeddedTypes[type.fqn];
+    if (exists) {
+      return exists;
+    }
+
+    const typeref = new GoTypeRef(this.root, type.type.reference);
+    const original = typeref.scopedInterfaceName(this);
+
+    const slug = original.replace(/[^A-Za-z0-9]/g, '');
+    const aliasName = `Type_${slug}`;
```
RomainMuller, line 170: Suggest using TWO `_` delimiter, so it will _NEVER_ conflict with a nested type (those would have only 1).

eladb: Fixed

RomainMuller (approved):

Issue #2671, opened 2021-03-10
go: optional values are treated as empty values


eladb:
### Affected Languages

- [x] `GoLang`
- [ ] `TypeScript` or `Javascript`
- [ ] `Python`
- [ ] `Java`
- [ ] .NET (`C#`, `F#`, ...)

### General Information
* **JSII Version:** 1.42.0
* **Platform:**     *any*

### What is the problem?

Consider:

```ts
interface MyStruct {
  readonly value1?: string;
  readonly value2?: string;
}

class Foo {
  public static isValue2Undefined(my: MyStruct) {
    return my.value2 === undefined;
  }
}
```

In go:

```go
my := MyStruct{Value1: "Hello"}
assert(Foo_IsValue2Undefined(my), true); // <== value2 is an empty string and therefore this assertion fails
```

See compliance test `eraseUnsetDataValues`.

Issue #2442, opened 2021-01-20
Go Optional Function Argument Support


MrArnoldPalmer:
The Go runtime and code generation don't currently support TS functions with optional arguments. In JSII libraries currently, most methods that accept complex data structures use interfaces with optional properties. These are currently handled "correctly" implicitly because when these are "zero values" in go, they are serialized as undefined and therefore not passed over the wire.

A more explicit approach will likely have to be taken here. A couple of options that have been discussed previously.

1. Wrapper types. This would cause us to wrap all optional types in some wrapper type. This however has the huge downside of when a type goes from required to optional in TS, a breaking change would occur in the generated Go code. This is likely untenable.

2. Pointers everywhere. All types in all public APIs would be converted from `T` to `*T`. This would allow users to pass null pointers for optional values. This also has some huge downsides. The first being, users can also pass null pointers to non-optional types and the go compiler wouldn't complain, these would manifest as a runtime crash. Additionally, increased verbosity is an issue. In Go you cannot pass a pointer to a primitive type anonymously. ie `callThing(&"MyString")`. Instead the user would have to allocate a variable with `"MyString"` and specify a pointer to that. This however could be mitigated by creating wrapper types from "pointer to a thing".

Issue #2686, opened 2021-03-14
go: unable to set an array of interfaces/enums


iliapolo:
## :bug: Bug Report

Its impossible to set a TS property typed as an array of classes.

### Affected Languages

- [ ] `TypeScript` or `Javascript`
- [ ] `Python`
- [ ] `Java`
- [ ] .NET (`C#`, `F#`, ...)
- [X] Golang

### General Information
* **JSII Version: ALL**
* **Platform: ALL**

### What is the problem?

Consider the following TS code:

https://github.com/aws/jsii/blob/45b527c0b2f35a09b715c1a6c5940ec0578007fb/packages/jsii-calc/lib/calculator.ts#L205-L209

Which generates the following Golang code:

```go
type Sum interface {
  SetParts(val []scopejsiicalclib.NumericValue)
}
```

Trying to set this property from Go produces an error:

```go
sum.SetParts([]calclib.NumericValue{calclib.NewNumber(5), calclib.NewNumber(10), calc.NewMultiply(calclib.NewNumber(2), calclib.NewNumber(3))});
```

```console
test panicked: "Value did not match any type in union: Expected a number, got {\"Base\":{}}, Expected object reference, got {\"Base\":{}}"
```

eladb:
@MrArnoldPalmer I suspect the same bug also applies to an array of enums. When setting an array of enums it seems like the enums are marshaled as strings instead of as jsii enums. Updating title to reflect.

RomainMuller:
@eladb  - this would have been correct - the _contents_ of the map was not recursively transformed on the way in from JavaScript, so only the top level map was correctly processed.

There was also an additional bug whereby you could not receive a list as an `interface{}` in certain conditions (including when that was part of a map)

PR #2705, opened 2021-03-15
fix(go): optional values

MrArnoldPalmer:
Adds support for passing `nil` to jsii nullable types in generated go code.

Changes go code generation to always use "pointer types" for all struct fields, method arguments, and method return values. Update runtime type conversion to correctly handle serializing and deserializing to these pointer types where necessary.

Removes unnecessary functionality for resolving scoped interface names for generated go types as each generated type now has only one public name. Adds support for resolving `scopedReferenceName` to easily generate "pointer types" for complex data structures from a type reference.

[RFC](https://github.com/aws/aws-cdk-rfcs/pull/297)
fix: #2442
fix: #2671

---


eladb (approved):
LGTM; added "do-not-merge" to let others chime in (@RomainMuller, @iliapolo)

packages/@jsii/go-runtime/jsii-runtime-go/runtime.go
```diff
@@ -359,3 +359,8 @@ func convertArguments(args []interface{}) []interface{} {
 func Close() {
 	kernel.CloseClient()
 }
+
+// Helpers to store primitives and return pointers to them
+func Bool(v bool) *bool         { return &v }
+func Number(v float64) *float64 { return &v }
+func String(v string) *string   { return &v }
```
RomainMuller, line 366: We need to either move those to another package, or move the runtime API (except for `Close`) to a non-internal sub-package, and use it from there...

Right now - this likely will make people itch to use toys we don't want them to play with.

packages/jsii-pacmak/lib/targets/go/types/class.ts
```diff
@@ -357,12 +353,11 @@ export class ClassMethod extends GoMethod {
...
-    return (
-      this.reference?.scopedInterfaceName(this.parent.pkg) ??
-      this.method.toString()
-    );
-  }
+  // public get returnType(): string {
+  //   return (
+  //     this.reference?.scopedName(this.parent.pkg) ?? this.method.toString()
+  //   );
+  // }
```
RomainMuller, line 361:
```suggestion
```

packages/jsii-pacmak/lib/targets/go/types/go-type-reference.ts
```diff
@@ -72,6 +72,11 @@ export class GoTypeRef {
     return this.type?.name;
   }
 
+  public get datatype() {
+    const reflectType = this.type?.type;
+    return reflectType?.isInterfaceType() && reflectType?.datatype;
```
RomainMuller, line 77: I'd assume the compiler can infer the following is correct... Can it not?

```suggestion
    return reflectType?.isInterfaceType() && reflectType.datatype;
```

packages/@jsii/go-runtime/jsii-runtime-go/runtime.go
MrArnoldPalmer: Wrote a ticket to extract user land functions into a separate package. https://github.com/aws/jsii/issues/2721

packages/@jsii/go-runtime-test/project/callbacks_test.go
```diff
@@ -7,20 +7,21 @@ import (
 )
 
 func TestPureInterfacesCanBeUsedTransparently(t *testing.T) {
-	expected := calc.StructB{RequiredString: "It's Britney b**ch!"}
-	delegate := &StructReturningDelegate{expected: expected}
-	consumer := calc.NewConsumePureInterface(delegate)
+	requiredstring := "It's Britney b**ch!"
+	expected := calc.StructB{RequiredString: &requiredstring}
+	delegate := StructReturningDelegate{expected: &expected}
+	consumer := calc.NewConsumePureInterface(&delegate)
```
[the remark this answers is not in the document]
MrArnoldPalmer: There appears to be a bug related to registering managed objects when this is passed by value instead of by reference. The only relevant change here is to the logic within the CastPtrToRef.

This also feels related to #2686 where structs passed by value that are fields of structs, and/or items within a slice, aren't being initialized correctly. Began hunting this down but need some more time.

RomainMuller (approved):

PR #2712, opened 2021-03-16
fix(rosetta): 'extract' does not translate samples in submodule READMEs

rix0rrr:
---


RomainMuller (approved):

rix0rrr:
I don't understand why the integ test is failing

RomainMuller:
There is an *unhandled* promise rejection in `jsii-rosetta` (it exists successfully, but really failed):

```
UnhandledPromiseRejectionWarning: Error: Sample uses fixture with-lambda-trigger, but not found: /home/runner/work/jsii/jsii/aws-cdk/packages/aws-cdk-lib/rosetta/with-lambda-trigger.ts-fixture
```

It appears some CDK documentation block uses a non-existent fixture file!

rix0rrr:
Thanks for digging that one out for me!

Guess there's no getting around the fact that the strategies Rosetta uses to get at fixtures and types aren't going to work for a monocdk that has been stitched together from the individual source packages.

Here's what Rosetta sees:

* Here we are extracting all the samples from `.../packages/aws-cdk-lib`.
* Some submodule has a README that uses a fixture named `with-lambda-trigger`.

Here's what Rosetta needs to conclude:

* Oooooh OF COURSE I know that file lives in `.../packages/aws-cdk-lib/lib/aws-lambda/rosetta/with-lambda-trigger.ts`.

....but how? I don't think it can.

Bottom line: this can't ever work without a lot more investment.

rix0rrr:
Supposedly we can move all fixtures to the right location... but they'll still have the wrong import names.

PR #2713, opened 2021-03-16
fix(pacmak): xmldom error when generating packages

RomainMuller:
In some cases, an `xmldom` error would occur when trying to render C#
XML-style documentation blocks, where inline html within a Markdown
document is not valid as far as the `xmldom` parser is considered. This
introduces a safer fall-back strategy.

---

### DIFF
--- a/packages/jsii-rosetta/lib/markdown/xml-comment-renderer.ts
+++ b/packages/jsii-rosetta/lib/markdown/xml-comment-renderer.ts
@@ -79,8 +79,24 @@ export class CSharpXmlCommentRenderer extends MarkdownRenderer {
    */
   public html_inline(node: cm.Node, _context: RendererContext) {
     const html = node.literal ?? '';
-    const doc = new DOMParser().parseFromString(html, 'text/html');
-    return new XMLSerializer().serializeToString(doc);
+    try {
+      const doc = new DOMParser().parseFromString(html, 'text/html');
+      return new XMLSerializer().serializeToString(doc);
+    } catch {
+      // Could not parse - we'll escape unsafe XML entities here...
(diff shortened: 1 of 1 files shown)


packages/jsii-rosetta/lib/markdown/xml-comment-renderer.ts
```diff
@@ -79,8 +79,24 @@ export class CSharpXmlCommentRenderer extends MarkdownRenderer {
...
   public html_inline(node: cm.Node, _context: RendererContext) {
     const html = node.literal ?? '';
-    const doc = new DOMParser().parseFromString(html, 'text/html');
-    return new XMLSerializer().serializeToString(doc);
+    try {
+      const doc = new DOMParser().parseFromString(html, 'text/html');
+      return new XMLSerializer().serializeToString(doc);
+    } catch {
+      // Could not parse - we'll escape unsafe XML entities here...
+      return html.replace(/[<>&]/g, (char: string) => {
```
iliapolo, line 87: So in case of success this method returns the result of `serializeToString`, but in case of failure it just replaces some chars in the html? Shouldn't this go back through the parser?

RomainMuller: There is no XML/HTML to parse anymore once you've stripped out all the `<`, and `>`, is there?

I've tried it out of curiosity, and it results in double-XML-escaping the entities (which is weird, but heh).

iliapolo: Its just dawned on me that the thing we are parsing is an XML comment 👍

iliapolo (approved):

PR #2726, opened 2021-03-19
feat(superchain): `$NVM_USE_VERSION` selects Node.js version

rix0rrr:
We added multiple Node versions into the Superchain image, and made it
possible to switch between them using NVM.

Doing so would require putting `nvm use 14` at the top of every script
that wanted to run against a newer Node version, and so would require
modification to scripts in our branches.

Another, more effective, way to configure the Node version is to be able
to do it without modifying the scripts themselves, but by specifying it
via the runtime environment. In this case, via environment variable that
can be set on the Docker environment.

The `$NVM_USE_VERSION` environment variable will lead to an implicit
execution of `nvm use $NVM_USE_VERSION` at container startup time, if
present.

---


nija-at (approved):
Noice!

PR #2730, opened 2021-03-22
fix(go): dates are mistreated as strings

### LINKED ISSUE #2659: go: dates are mistreated as strings
## :bug: Bug Report

Golang generated code results in `Date` arguments treated as plain strings.

### Affected Languages

- [ ] `Python`
- [ ] `Java`
- [ ] .NET (`C#`, `F#`, ...)
- [X] Golang

### General Information
* **JSII Version: ALL**
* **Platform: ALL**

### What is the problem?

Consider the following TS code:

https://github.com/aws/jsii/blob/188336275e5d48bfc47c6d43295ede1ee7c7a14d/packages/jsii-calc/lib/compliance.ts#L118-L124

The resulting Golang code would be:

```go
func (a *allTypes_jsiiProxy) SetDateProperty(val string) {
	_jsii_.Set(
		a,
		"dateProperty",
		val,
	)
}
```

Which causes a deserialization error:

```console
test panicked: "Expected Date, got \"12345\""
```

Stemmed from here:

https://github.com/aws/jsii/blob/188336275e5d48bfc47c6d43295ede1ee7c7a14d/packages/%40jsii/kernel/lib/serialization.ts#L153-L155

This seems to be a known issue, as evident here:

https://github.com/aws/jsii/blob/188336275e5d48bfc47c6d43295ede1ee7c7a14d/packages/jsii-pacmak/lib/targets/go/types/go-type-reference.ts#L15-L16

RomainMuller:
Implements the correct code generation flow & runtime logic to correctly
handle jsii date values as *time.Time instances.

Fixes #2659

---


MrArnoldPalmer (approved):
LGTM, special dependencies stuff is 👍🏻 much easier to reason about.

PR #2736, opened 2021-03-23
fix(superchain): use entrypoint to set up nvm

### LINKED ISSUE #2734: No access to NodeJS in jsii/superchain when overriding the default container image command
### Affected Languages
- [X] `TypeScript` or `Javascript`
- [ ] `Python`
- [ ] `Java`
- [ ] .NET (`C#`, `F#`, ...)
- [ ] `Go`

### General Information
* **JSII Version:** latest `jsii/superchain` image (Image ID: `53fcfbbc6bda`)
* **Platform:**     N/A

### What is the problem?
In #2718, NVM was added instead of directly installing NodeJS 10.x in the `jsii/superchain` image. The default Docker `CMD` was changed to use a bash login shell:

https://github.com/aws/jsii/blob/c79a346ada8789efe63db2a1f798f55f31c56178/superchain/Dockerfile#L129

This works when you use the container without specifying an override command. When a command is specified, for example:

```sh
docker run --rm -it jsii/superchain node
```

this code is bypassed and NodeJS is no longer available:

```sh
docker: Error response from daemon: OCI runtime create failed: container_linux.go:370: starting container process caused: exec: "node": executable file not found in $PATH: unknown.
```

RomainMuller:
This ensures node is always in the PATH at the start of a container (unless
the user replaced the entrypoint). This attempts to minimize overhead of the
overall setup by using exec in the run.sh script.

Fixes #2734

---

### DIFF
--- a/superchain/run.sh
+++ b/superchain/run.sh
@@ -3,4 +3,4 @@
 source "$NVM_DIR/nvm.sh"
 [[ -z "$NVM_USE_VERSION" ]] || nvm use "$NVM_USE_VERSION"
 
-exec $@
+exec "$@"
--- a/superchain/run.sh
+++ b/superchain/run.sh
@@ -1,0 +1,6 @@
+#!/bin/sh
+
+source "$NVM_DIR/nvm.sh"
+[[ -z "$NVM_USE_VERSION" ]] || nvm use "$NVM_USE_VERSION"
+
+exec $@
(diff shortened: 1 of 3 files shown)


superchain/run.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+
+source "$NVM_DIR/nvm.sh"
+[[ -z "$NVM_USE_VERSION" ]] || nvm use "$NVM_USE_VERSION"
+
+exec $@
```
rix0rrr, line 6:
```suggestion
exec "$@"
```

rix0rrr (changes requested):

rix0rrr (approved):

PR #2741, opened 2021-03-25
feat(superchain): also build a superchain image with Node 14

rix0rrr:
For CDKv2, we want to require (and test) against a newer version of
Node. We've attempted to build multiple Node versions into a single
Docker image, but because we cannot control the entrypoint there is no
reliable way of switching between the Node versions at execution time.

Therefore, we will now build two versions of the "superchain" image.
The Node version built into each is controlled by the `NODE_VERSION`
build argument.

Workflow has been changed to build and push the following images:

* Nightlies:
    * `jsii/superchain:node10-nightly`
    * `jsii/superchain:node14-nightly`
    * `jsii/superchain:nightly` will remain an alias for `:node10-nightly`
* Releases:
    * `jsii/superchain:node10`
    * `jsii/superchain:node14`
    * `jsii/superchain:latest` will remain an alias for `:node10`.

---


superchain/README.md
```diff
@@ -13,7 +13,8 @@ SDK             | Version
 `OpenJDK 8`     | Amazon Corretto `>= 8.242.08.1`
 `.NET SDK`      | `>= 3.1.101`
 `mono`          | `>= 6.8.0.105`
-`Javascript`    | `node >= 10.19.0` with `npm >= 6.13.4`
+`Javascript`    | `node >= 10.19.0` with `npm >= 6.13.4` (default)
+                | OR `node >= 14.16.0` with `npm >= 6.14.11` (`--build-arg NODE_VERSION=14`)
```
RomainMuller, line 17: This won't render properly... Seems like the rendering engine does not like when the first column is empty.

Also - you don't want to make this relative to `--build-arg`, instead refer to the tag names.

Perhaps we should actually extract the `node` and `npm` info to a sub-section - and in the table mention this is tag-dependent, and link to it?

.github/workflows/docker-images.yml
```diff
@@ -53,31 +59,45 @@ jobs:
...
-            -v${{ github.workspace }}:${{ github.workspace }}                   \
-            -w${{ github.workspace }}                                           \
-            'jsii/superchain:nightly'                                           \
+          docker run                                                          \
+            --rm                                                              \
+            --tty                                                             \
+            --network=host                                                    \
+            -v${{ github.workspace }}:${{ github.workspace }}                 \
+            -w${{ github.workspace }}                                         \
+            "jsii/superchain:node${{ matrix.node }}-nightly"                  \
```
RomainMuller, line 78: You could have left `'` here - the `${{}}` stuff gets interpolated *before* the job runs.

RomainMuller (approved):

PR #2763, opened 2021-03-31
fix(dotnet): unable to pass 2d-array of interfaces

RomainMuller:
When a 2-dimensional array of interfaces was passed, the `InferType`
function failed to account for interfaces, and failed instead of
correctly returning the relevant type reference.

Fixes aws/aws-cdk#12587

---


RomainMuller:
@all-contributors add @ChristopheVico for bug

packages/@jsii/dotnet-runtime-test/test/Amazon.JSII.Runtime.IntegrationTests/ComplianceTests.cs
```diff
@@ -1512,5 +1512,18 @@ public override String Repeat(String word)
                 return word;
             }
         }
+
+        [Fact(DisplayName = Prefix + nameof(ArrayOfInterfaces))]
```
eladb, line 1516: Do we need to implement this in all other languages?

What about arrays of enums?

eladb (approved):

Issue #2768, opened 2021-04-01
(jsii-go) Could we resolve the set of overriden methods by reflect not annotation.


yglcode:
### Affected Languages

- [ ] `TypeScript` or `Javascript`
- [ ] `Python`
- [ ] `Java`
- [ ] .NET (`C#`, `F#`, ...)
- [X] `Go`

### General Information
* **JSII Version:** 1.26.0 (build 7d76e02), typescript 3.9.9
* **Platform:**     linux

* [x] I may be able to implement this feature request

* [X] This feature might incur a breaking change

### Description
In current jsii go package, a struct will embed the "overriden" interface and annotate it with the implemented "overriding" methods [doc](https://aws.github.io/jsii/user-guides/lib-user/language-specific/go/), such as:
type childClass struct {
   // We implement AbstractMethod, AbstractReadonlyProperty, and override the setter for ConcreteProperty
   jsiimodule.AbstractBaseClass `overrides:"AbstractMethod,AbstractReadonlyProperty,ConcreteProperty"`
   ...
}.

### Proposed Solution
My understanding is that all overriding methods will have pointer receiver. If this is correct, we may be able to resolve the set of overriding methods thru reflect automatically, for the following two reasons:
   1> If S contains an embedded field *T, the method sets of S and *S both include promoted methods with receiver T or *T.
   2> if a method is defined/overriden with pointer receiver *T, in reflect, it is removed from method-set of T, only available in method-set of *T.

So thru reflect, we can find the difference between method-sets of T and *T, which are the methods overriden. Here is some [sample code](https://gist.github.com/yglcode/910552910223e76a0e2eed179a873215).

MrArnoldPalmer:
@yglcode this is interesting, I was unaware of this
>if a method is defined/overriden with pointer receiver *T, in reflect, it is removed from method-set of T, only available in method-set of *T.

So we would still need to delegate to the "overrides constructor" of a class in order to register the overrides with the runtime but this could remove the need for our custom "overrides" annotation as we can find this list via the difference between methods on *T and T.

@RomainMuller does this seem feasible to you? Adding this to the GA project board for now.

RomainMuller:
I'm favorable to this. I wonder if we should "keep" the `overrides` tag, but make it optional. It could be needed if for some reason someone was to implement a method using a non-pointer receiver (which right now we say they should not do - so maybe this is enough for us to not care?)

yglcode:
[...]

PR #2780, opened 2021-04-08
feat(jsii-go): use reflect to resolve overriden methods

yglcode:
Use reflect to resolve overridden methods, per issue #2768, leveraging
how such methods will no longer be visible on the non-pointer receiver,
while methods promoted from anonymous embeds are.

---


packages/@jsii/go-runtime/jsii-runtime-go/runtime.go
```diff
@@ -159,6 +143,36 @@ func Create(fqn FQN, args []interface{}, inst interface{}) {
 		}
 	}
 
+	//find method overrides thru reflection
+	mOverrides := getMethodOverrides(inst, "jsiiProxy_")
+	//if overriding struct has no overriding methods, could happen if
+	//overriding methods are not defined with pointer receiver.
+	if len(mOverrides) == 0 && !strings.HasPrefix(instType.Name(), "jsiiProxy_") {
+		panic(fmt.Errorf("%s has no overriding methods. Please verify overriding methods are defined with pointer receiver", instType.Name()))
```
RomainMuller, line 151:
```suggestion
		panic(fmt.Errorf("%s has no overriding methods. Overriding methods must be defined with a pointer receiver", instType.Name()))
```

RomainMuller:
Made some (vey) minor style changes in your code (mostly adding white spaces and capitalisation in comments)...

I wonder if the error message you added could be improved a little (but I'm not too concerned about it though).

That looks pretty good to me at this stage; it's a very clever idea and I like it.

---

Would you be able to update the documentation under `gh-pages`? I can take that part up otherwise (no problem!)

yglcode:
> Made some (vey) minor style changes in your code (mostly adding white spaces and capitalisation in comments)...
> 
> I wonder if the error message you added could be improved a little (but I'm not too concerned about it though).
> 
> That looks pretty good to me at this stage; it's a very clever idea and I like it.
> 
> Would you be able to update the documentation under `gh-pages`? I can take that part up otherwise (no problem!)

Thanks for your comments. I will update the code. It will be great if you could update the docs. Thanks.

RomainMuller:
Updated the docs page now to remove references to the `overrides` tag and insist on the fact overridden methods must have a pointer receiver... I think we're good to ship now...

Thank you again for the clever idea & implementation!

RomainMuller (approved):

RomainMuller (dismissed):

RomainMuller:
@yglcode - don't mind me I'm trying to diagnose the transient windows PR validation failure as it appears to be happening on this particular PR (no idea why or what's wrong 🤷🏻‍♂️).

I'll get your code merged soon (even if I don't manage to understand the failure).

RomainMuller (approved):

PR #2783, opened 2021-04-09
fix: deprecated symbol stripping does not strip out heritage clause with imported type

nija-at:
When a non-deprecated class or interface inherits a deprecated class or
interface, the `--strip-deprecated` option does not strip out the
heritage clause.

This particularly affects cases where the inherited class or interface
is non-local, i.e., imported from another file.

---


eladb (commented):
Missing test

nija-at:
### Testing

- [X] Ran `run-against-dist` with `test-typescript-app.sh` and verified that the tests succeed. Before this change, the run would fail with errors that classes are missing in the declaration file.

eladb (approved):

RomainMuller (approved):

PR #2786, opened 2021-04-12
chore(go): retract v1.27.0

### LINKED ISSUE #2775: go: pacmak fails to add a local "replace" directive when using an unreleased version of jsii
### Affected Languages

- [ ] `Python`
- [ ] `Java`
- [ ] .NET (`C#`, `F#`, ...)
- [x] `Go`

### General Information
* **JSII Version:** 1.26.0
* **Platform:**     *any*

### What is the problem?

When trying to build Go bindings using an _unreleased_ version of jsii-pacmak (e.g. in the jsii integ tests), the generated `local.go.mod` points to a non-existent version of `github.com/aws/jsii-runtime-go`.

This results in the following error (requires `-vvv`):

```
go: github.com/aws/jsii-runtime-go@v1.27.0: reading github.com/aws/jsii-runtime-go/go.mod at revision v1.27.0: unknown revision v1.27.0
```

There is an additional error but it seems like a red herring:

```
(node:57300) UnhandledPromiseRejectionWarning: Error: ENOENT: no such file or directory, open 'local.go.mod'
```

We need the logic that generates the `local.go.mod` file to add a `replace` directive which points to the non-released version.

RomainMuller:
It was manually created, then deleted, then re-created and now does not
work properly as the proxy.golang.org has incorrect artifacts in, and
sum.golang.org has the incorrect checksum, and both of these are
designed to be immutable (as a security measure).

Fixing this requires retracting 1.27.0 and releasing a new version (e.g:
1.27.1).

Related: https://github.com/golang/go/issues/45517

The original issue was found & reported by @Cristim

Fixes #2775

---


RomainMuller:
@all-contributors add @Cristim for bug

eladb (approved):

eladb (approved):

packages/@jsii/go-runtime/jsii-runtime-go/go.mod
```diff
@@ -6,3 +6,5 @@ require (
 	github.com/Masterminds/semver/v3 v3.1.1
 	github.com/stretchr/testify v1.7.0
 )
+
+retract v1.27.0
```
eladb, line 10: How the hell did you know this is a thing...

cristim: I tried to run CDK synth as per the preview instructions and noticed signature mismatch errors when installing this dependency.

PR #2809, opened 2021-04-22
fix(python): classes do not correctly implement interfaces

rectalogic:
Fixes  aws/aws-cdk#13474

---


RomainMuller:
Whee thanks for that! I've re-titled this a bit, as this will actually be the fix for the problem once we figure out a solution....

I will take it from there and devise a fix.

@all-contributors add @rectalogic for bug, code

PR #2816, opened 2021-04-27
feat(rosetta): JSII_ROSETTA_MAX_WORKER_COUNT limits max workers

njlynch:
The current behavior of rosetta (extract) is to use a number of node worker
threads equal to half the number of CPU cores. On large CI/CD build hardware,
this may mean creating a huge number of worker threads which causes more
contention than benefit.

Introduce `JSII_ROSETTA_MAX_WORKER_COUNT` to limit the maximum number of workers
(defaults to `16`).

See https://github.com/aws/aws-cdk/pull/14389 for more motivation from the AWS CDK.

---

### DIFF
--- a/packages/jsii-rosetta/README.md
+++ b/packages/jsii-rosetta/README.md
@@ -195,6 +195,6 @@ Worker threads are enabled by default on NodeJS 12.x, and can be enabled on Node
 $ node --experimental-worker /path/to/jsii-rosetta extract ...
 ```
 
-If worker thread support is enabled, `jsii-rosetta` will use a number of workers equal to half the number of CPU cores,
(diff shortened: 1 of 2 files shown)


packages/jsii-rosetta/README.md
```diff
@@ -194,3 +194,7 @@ Worker threads are enabled by default on NodeJS 12.x, and can be enabled on Node
 ```
 $ node --experimental-worker /path/to/jsii-rosetta extract ...
 ```
+
+If worker thread support is enabled, `jsii-rosetta` will use a number of workers equal to half the number of CPU cores,
```
RomainMuller, line 198:
```suggestion
If worker thread support is available, `jsii-rosetta` will use a number of workers equal to half the number of CPU cores,
```

RomainMuller (approved):
Minor tweak in the README.

PR #2837, opened 2021-05-13
fix: update lineMap after altering source.text

### LINKED ISSUE #2680: Bug: inline sourcemaps are broken
The AWS CDK inline sourcemaps are broken. This is because sourcemaps are based on the original typescript source files, while the inlined typescript source code has been stripped down in some way so they're shorter and don't align with sourcemap locations.

This mismatch prevents debuggers from stepping through source correctly. It also causes wrong code excerpts to appear in CDK CLI error output. Both of these problems can make it much harder to debug a CDK app than it needs to be. That's too bad, because the CDK is a wonderful tool, and simpler debugging would make it even better!

### Reproduction Steps

There are different ways to show the problem for a given CDK version. I've checked 1.91.0 (currently the latest 1.x) as well as the earlier 1.85.0 release. I'll use the vscode debugger below with screenshots, which seems simplest here.

(Note that I created the annotations in the screenshots before writing this text, and some imprecisely refer to the original source as 'correct' and the inline source as 'wrong' in terms of sourcemap locations. Of course, what really matters is that sourcemap locations align with the inlined code, not that they necessarily match the original files.)

#### Repro 1

1. Find the source for the version we're using as a reference. For example, use the github branch picker online, or clone locally:
```bash
git clone https://github.com/aws/aws-cdk.git
git checkout tags/v1.91.0
# No need to waste time building here, we just want to see source files
```

2. Initialize and build a minimal CDK app to debug, then open in vscode:
```bash
cdk init app --language typescript
npm run build
code .
```

3. Switch to the debugger tab in vscode, and create a new nodejs launch config.
3_cdkbug_launch_config

4. Set a breakpoint at the `new cdk.App` call:
4_cdkbug_breakpoint_at_new_app_call

5. Step into the cdk source for the `App` constructor. This is really the inline typescript source at the end of the js file. Notice that the debugger has placed us at the end of the file. That's because the 101:5 location it tries to navigate to is based on the original source that has 147 lines, but the debugger uses the inline source that only has 86 lines:
5_cdkbug_after_stepping_in_stack_correct_but_wrong_file_location

6. Here's a screenshot from the original v1.91.0 `app.ts` source file, confirming that the sourcemap line:column indicator in the stack (101:5) matches it:
6_cdkbug_correct_location_in_original_ts_source

#### Repro 2

As a further example, I added this code to force a `cdk synth` CLI error. The fact that this causes an error is surprising to me, since it seems I should only need to create a single `HostedZone` instance once using `fromLookup` that I can pass to multiple stacks via `props`. Anyway, I'll probably create a separate issue for it. Debugging it is what prompted reporting this issue.

1. Here's the main source change:
1_cdk_synth_error_app_update

2. Here's the `cdk synth` CLI output stack trace with incorrect source code excerpt. In the screenshot, notice all stack locations match the original typescript source but not the inline source:
2_cdk_synth_error_stack_and_bad_src_excerpt

3. Here's the reported line 191 from the original source, which is the correct source of the thrown error:
3_cdk_synth_error_line_191_matches_orig_src

4.a And here's where the code excerpt text is from in the original source at line 369:
4_cdk_synth_error_bad_src_excerpt_from_orig_line_369

### What did you expect to happen?

I expect the sourcemap locations to match the inline source code exactly.

### What actually happened?

The sourcemaps do not match the inline source code, which causes two problems:
1. The debugger cannot step through the source correctly, causing a debugging nightmare lol.
2. Source code excerpts in stack trace errors during CDK CLI operations are wrong and misleading.

### Environment

  - **CDK CLI Version  :** 1.91.0 (doesn't matter but also 1.85.0)
  - **Framework Version:** 1.91.0 (also earlier in at least 1.85.0)
  - **Node.js Version:** 14.15.4 (doesn't matter but also 12.18.3)
  - **OS               :** Windows 10 (doesn't matter but also macOS)
  - **Language (Version):** 

It appears that there's a code reduction step in the build process that's stripping down the source that ultimately gets inserted inline in the js files. Unfortunately, it seems the sourcemaps are created before this happens, so they match the original source instead of the inlined source. Hopefully, there may be a simple fix by just reordering some build steps, but maybe that's just wishful thinking :-)

--- 

This is :bug: Bug Report

lzhoucs:
Fixes #2680.

I debugged through the compilation process and found that `source.lineMap` never gets recomputed after  `source.text` is updated. This causes wrong source lines getting mapped into 3rd field of each segment(a.k.a the line in the original source, see [spec](https://sourcemaps.info/spec.html) for more details) when the sourcemap is generated against out dated lineMap.

Here is an example before the fix:
image

And here is after the fix:
image

---


RomainMuller:
@all-contributors add lzhoucs for bugs, code

RomainMuller (approved):

PR #2843, opened 2021-05-17
chore(deps): reject auto upgrade of prettier

nija-at:
prettier@2.3.0 is now enforcing Allman brace styles, while the code
base uses 1tbs (one true brace style).

In fact, the latter has been explicitly configured in eslint.
There does not seem to be a way to configure prettier to switch styles.

---

### DIFF
--- a/.github/workflows/yarn-upgrade.yml
+++ b/.github/workflows/yarn-upgrade.yml
@@ -78,11 +78,12 @@ jobs:
       - name: Run "ncu -u"
         # We special-case @types/fs-extra because the current major (9.x) is broken with @types/node@10
         # We special-case typescript because it's not semantically versionned, and major.minor is the API contract
+        # We special-case prettier because 2.3.0 introduces Allman brace style, while we use one true brace style
         run: |-
           # Upgrade devDependencies at repository root
           ncu --upgrade --target=minor --filter=@types/node,@types/fs-extra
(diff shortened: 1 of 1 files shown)


otaviomacedo (approved):

eladb:
Isn't it possible to configure prettier accordingly?

MrArnoldPalmer:
It looks related to this change which only applies to classes with multiline `extends/inherits` https://prettier.io/blog/2021/05/09/2.3.0.html

We likely could make eslint ignore this and just let prettier handle it (thats what eslint-config-prettier is for but maybe it hasn't caught up?).

At the top of that same release notes page it recommends not using a ^ version of prettier though so this seems good anyway.

nija-at:
> Isn't it possible to configure prettier accordingly?

@eladb - I could not find such an option. I believe this is because of prettier's [philosophy](https://prettier.io/docs/en/option-philosophy.html)

> We likely could make eslint ignore this and just let prettier handle it

@MrArnoldPalmer - Here's how that would look - https://github.com/aws/jsii/pull/2835/. Is that a better change?

MrArnoldPalmer:
@nija-at yeah lets take that. I think just letting Prettier handle everything it does and using Eslint only for non-formatting related stuff is our preferred experience. TY for that!

PR #2856, opened 2021-05-26
chore: browser compatibility for `jsii-reflect` and `codemaker`

iliapolo:
We have started using `jsii-reflect` and `codemaker` in the [construct-hub-webapp](https://github.com/cdklabs/construct-hub-webapp) to generate API references on the client side.

These are the modifications needed in order for the modules to load in a browser environment. 

---


iliapolo:
@RomainMuller Can you think of a quick test for this scenario? As far as regressions go i'm leaning heavily on the existing test suites...

RomainMuller (commented):
I mean we could have jest operate with browser environments... That should be good enough.... I guess? I'm somewhat wondering if we should be looking at making consumers of `codemaker` require it lazily, so it is *not* loaded on browsers...

If you tried to require `fs` on the browser, I guess it would crash, right? That makes me a little uneasy (i.e: we allow it to be loaded there, but it effectively cannot be used there... and the `engines` setting in `package.json` does NOT mention browser compatibility, either...

iliapolo:
I actually explicitly load `codemaker` for the casing function. I considered copying it over but it comes with its own dependencies and I believe we'll be using more of its functionality later on.

Regarding requiring `fs` - yes that makes the browser crash. There is a [browser shim](https://github.com/jvilk/BrowserFS) for `fs`, but I couldn't get it to work quickly so I dropped it. 

I think that if consumers target browser environments, they won't be using any file system API's the module has to offer (because why would they) -  and then webpack tree shaking effectively removes those imports from the bundle. 

Are you good with merging this though and coming up with a more robust plan (maybe separating to different modules..) later on?

PR #2863, opened 2021-05-28
fix(rosetta): extract does not respect strict metadata entry

### LINKED ISSUE #2861: rosetta extract command doesn't respect strict metadata entry
### Affected Languages

- [ ] `Python`
- [ ] `Java`
- [ ] .NET (`C#`, `F#`, ...)
- [ ] `Go`

### General Information
* **JSII Version:** 1.29.0
* **Platform:** macOS 10.15.7

### What is the problem?

The [README for jsii-rosetta](https://github.com/aws/jsii/tree/main/packages/jsii-rosetta#enforcing-compilability) states:
> Package owners can enable enforcement of compiling code sample by setting the `jsii.rosetta.strict` assembly metadata entry to `true`

However, this does not seem to be the case. For example, adding the corresponding entry to the [`@aws-cdk/aws-lambda` module](https://github.com/aws/aws-cdk/blob/061fd558a3327b805bb5fe0abc72de7c21bbbdb9/packages/%40aws-cdk/aws-lambda/package.json#L7) and running `jsii-rosetta extract` produces no error messages and an exit code of 0, while running the same command with the `--strict` flag produces a whole bunch of diagnostics (77 at the time of writing) and produces an exit code of 1. This can easily be reproduced with a jsii projen project by adding the following to the .projenrc.js:
```
Object.assign(project.manifest['jsii'], {
  metadata: {
    jsii: {
      rosetta: {
        strict: true
      }
    }
  }
});
```
Then, any non-compiling snippet in the README or otherwise will not fail when running `jsii-rosetta extract`.

There seem to be two separate issues that are causing this problem, one that is preventing compilation and one that is preventing failure when compilation fails.

1. During the `fixturize` command that is used to combine snippets with fixtures, when [returning the fixturized snippet](https://github.com/aws/jsii/blob/0fbba6bc1af09cbab19acdabaea14e5eb1681ae3/packages/jsii-rosetta/lib/fixtures.ts#L41-L46), does not copy [the `strict` flag](https://github.com/aws/jsii/blob/0fbba6bc1af09cbab19acdabaea14e5eb1681ae3/packages/jsii-rosetta/lib/snippet.ts#L25-L30) that marks whether the snippet should be compiled and cause a failure. This can be fixed by instead returning
```
{                                                                                                                                                                      
  ...snippet,                                                                                                                                                                      
  completeSource: source,                                                                                                                                                          
  parameters,                                                                                                                                                                      
};
```
2. During translation, [the `worker_threads` module](https://nodejs.org/api/worker_threads.html#worker_threads_worker_threads) is used to parallelize translation of snippets. Each worker passes the compiler "diagnostics" (which contain any failures that may have arose during compilation of a snippet) in its return value. These diagnostics should be [annotated by the translator](https://github.com/aws/jsii/blob/main/packages/jsii-rosetta/lib/translate.ts#L156-L159) to contain [a special `Symbol` property](https://github.com/aws/jsii/blob/0fbba6bc1af09cbab19acdabaea14e5eb1681ae3/packages/jsii-rosetta/lib/util.ts#L41-L53) that marks the diagnostic as "strict" if the extraction run was configured as such. Unfortunately, the [message passing protocol](https://nodejs.org/api/worker_threads.html#worker_threads_port_postmessage_value_transferlist) that `worker_threads` uses [does not support object properties that are `Symbol`s](https://nodejs.org/api/worker_threads.html#worker_threads_considerations_when_cloning_objects_with_prototypes_classes_and_accessors). This means that the diagnostics that are returned do not satisfy [the criteria for being an error](https://github.com/aws/jsii/blob/0fbba6bc1af09cbab19acdabaea14e5eb1681ae3/packages/jsii-rosetta/lib/util.ts#L61) and so extraction does not fail. This can potentially be fixed by defining a diagnostic "schema" in the same vein as [`TranslatedSnippetSchema`](https://github.com/aws/jsii/blob/main/packages/jsii-rosetta/lib/tablets/schema.ts#L33) which seems to exist for serialization purposes only.

### Verbose Log

N/A – should be getting an error, but passing silently.

BenChaimberg:
The `jsii.rosetta.strict` assembly metadata entry was not being respected by the `extract` command due to two bugs:
1. Fixturizing (the process by which snippets are enriched by a separate fixture) did not transfer the "strict" flag from the original snippet to the fixturized snippet. Fix: transfer the strict flag to the fixturized snippet.
2. Translation of snippets to other languages using the `worker_threads` node module involves sending compiler diagnostics across the wire. These diagnostics should be annotated with a "strict brand" that marks them as failing compilation and allows further steps to fail if such a diagnostic exist. Because the brand was a [Symbol](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol), it is not correctly serialized. Fix: make the brand a string so it can be properly serialized.

fixes #2861

---


BenChaimberg:
I opted for a simpler fix than described in the issue: just changed the "strict brand" from a symbol to a string with a "jsii" prefix

packages/jsii-rosetta/lib/util.ts
```diff
@@ -38,15 +38,15 @@ export function printDiagnostic(
   stream.write(message);
 }
 
-const StrictBrand = Symbol('strict');
+const StrictBrand = 'jsiiStrict';
```
RomainMuller, line 41:
```suggestion
const StrictBrand = 'jsii.strict';
```

packages/jsii-rosetta/test/util.test.ts
```diff
@@ -0,0 +1,63 @@
+import * as ts from 'typescript';
+
+import { annotateStrictDiagnostic, isErrorDiagnostic } from '../lib/util';
+
+const strictBrand = 'jsiiStrict';
```
RomainMuller, line 5: Shouldn't be copied here... instead export the brand in the library, and import it there.

nija-at (approved):
Could you summarize the issue and the fix in the PR description?
Would be nice to get that into the git commit message with the ability to read directly on the IDE/console, rather than have to click into a Github issue.

There's a ton of material you'll find on the internet on the value of good commit messages. Here's one that's standard token - https://medium.com/swlh/why-should-i-write-good-commit-messages-e15d37bf45cb.


Change look good to me.

Added 'do-not-merge' in case @RomainMuller wants to take another look.

BenChaimberg:
@Mergifyio refresh

BenChaimberg:
@Mergifyio refresh

BenChaimberg:
@RomainMuller for some reason, the jsii-pacmak test:build is failing with error:
> jsii-pacmak: #STDERR> /home/runner/work/jsii/jsii/packages/@jsii/go-runtime/jsii-runtime-go/internal/kernel/process/handshake.go:7:2: missing go.sum entry for module providing package github.com/Masterminds/semver/v3 (imported by github.com/aws/jsii-runtime-go/internal/kernel/process); to add:
jsii-pacmak: #STDERR> 	go get github.com/aws/jsii-runtime-go/internal/kernel/process@v0.0.0

I don't understand why this would suddenly be a problem and why other builds are succeeding. The go.sum in question does in fact have entries for [Masterminds/semver/v3].

[Masterminds/semver/v3]: https://github.com/aws/jsii/blob/d9028c8cf91297a44cd4b1f128f769716bf0e74e/packages/%40jsii/go-runtime/jsii-runtime-go/go.sum#L1-L2

RomainMuller (approved):

RomainMuller:
The build problem is apparently caused by a bug in `go mod download` in the 1.16 versions...

https://github.com/aws/jsii/pull/2893

PR #2869, opened 2021-06-02
feat(jsii-rosetta transliterate): transliterate a jsii assembly

RomainMuller:
The new `jsii-rosetta transliterate` command can be used to
transliterate one or more jsii assemblies (the `.jsii` file) to one or
more target languages. The output assembly has all code examples
transliterated into the correct target language.

The current feature is experimental. It does not include renaming API
elements within the assembly (only code examples are touched) - this
feature may be added in the future.

---


packages/jsii-rosetta/bin/jsii-rosetta.ts
```diff
@@ -166,6 +168,55 @@ function main() {
...
+            type: 'string',
+            string: true,
+            default: new Array<string>(),
+            required: true,
+            describe: 'Assembly to transliterate',
+          })
+          .positional('LANGUAGE', {
+            type: 'string',
+            string: true,
+            default: new Array<string>(),
```
eladb, line 186: Is this supposed to also be an array?

packages/jsii-rosetta/bin/jsii-rosetta.ts
```diff
@@ -166,6 +168,55 @@ function main() {
...
+            describe: 'Language ID to transliterate to',
+          })
+          .option('tablet', {
```
eladb, line 189: What's that? Not sure I understand this option (and `strict` as well). Maybe add some `.example()`s?

packages/jsii-rosetta/lib/commands/transliterate.ts
```diff
@@ -0,0 +1,172 @@
...
+  assemblyLocations: readonly string[],
+  targetLanguages: readonly TargetLanguage[],
+  tabletLocation?: string,
+  options: TransliterateAssemblyOptions = {},
+): Promise<void> {
+  const rosetta = new Rosetta({ liveConversion: true, targetLanguages });
+  if (tabletLocation) {
+    await rosetta.loadTabletFromFile(tabletLocation);
+  }
+  const assemblies = await Promise.all(
```
eladb, line 36: Style: the functional style is very neat but TBH makes the logic pretty hard to read (at least for me). I think imperative style will be more readable and easier to maintain long-term.

packages/jsii-rosetta/lib/commands/transliterate.ts
```diff
@@ -0,0 +1,172 @@
...
+}
+
+type Mutable<T> = T extends ReadonlyArray<infer E>
+  ? Array<Mutable<E>>
+  : T extends Array<infer E>
+  ? Array<Mutable<E>>
+  : T extends string
+  ? T
+  : { -readonly [K in keyof T]: Mutable<T[K]> };
+function mutableCopy<T>(val: T): Mutable<T> {
```
eladb, line 91: I'd just read the assembly again in every iteration...

packages/jsii-rosetta/lib/commands/transliterate.ts
```diff
@@ -0,0 +1,172 @@
...
+      case 'csharp':
+      case 'java':
+      case 'go':
+      default:
+        return '//';
+    }
+  }
+}
+
+function transliterateType(
```
eladb, line 126: I am surprised this code doesn't exist already somewhere...

eladb (changes requested):
Missing test

packages/jsii-rosetta/bin/jsii-rosetta.ts
RomainMuller: Yeah, you can submit several languages if you want to. Interestingly, I don't know how it would work (since it's a sequence of string[] positional arguments, so how'd it tell when it's one and when it's the other? I gotta fix that).

packages/jsii-rosetta/lib/commands/transliterate.ts
RomainMuller: Rosetta just didn't get in the business of final result substitutions previously... This particular feature is scattered in consumers of Rosetta tablet files... Which is parts of what I am trying to solve here.

packages/jsii-rosetta/test/commands/transliterate.test.ts
```diff
@@ -0,0 +1,917 @@
...
+            "name": "ClassNameProps",
+            "properties": Array [
+              Object {
+                "abstract": true,
+                "immutable": true,
+                "locationInModule": Object {
+                  "filename": "index.ts",
+                  "line": 37,
+                },
+                "name": "foo",
```
eladb, line 281: Are these inline snapshots actually update? I never managed to get that to work..

Also not sure that's better. It really pollutes the source

eladb (approved):

packages/jsii-rosetta/test/commands/transliterate.test.ts
```diff
@@ -0,0 +1,917 @@
...
+import * as jsii from 'jsii';
+import * as os from 'os';
+import * as path from 'path';
+
+import { transliterateAssembly } from '../../lib/commands/transliterate';
+import { TargetLanguage } from '../../lib/languages/target-language';
+
+jest.setTimeout(60_000);
+
+test('single assembly, all languages', () =>
```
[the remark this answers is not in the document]
RomainMuller: It is :) The function returns a promise, which is what jest uses to identify tests.

packages/jsii-rosetta/test/commands/transliterate.test.ts
RomainMuller: They do update, yes (I have iterated over those a bit).

Best practice is to treat the snapshots as code... I agree that they take up quite a lot of space, though.... But I find they are more difficult to inspect when they're in a different file... I don't feel super strongly about this though... If you'd rather I use "regular" snapshots, it's easy to flip to that.

PR #2892, opened 2021-06-29
feat(rosetta): transliterate loose mode

RomainMuller:
When transliterating code samples from a published package tarball,
supporting files such as fixtures and literate source files may be
missing (typically because they are part of the `.npmignore` file).

The new `--loose` option to `jsii transliterate` allows ignoring those
problems by:

- Ignoring missing fixtures (the behavior is then the same as if no
  fixture was requested)
- Falling back to the `.lit.js` file when the `.lit.ts` file cannot be
  found, or replacing the source with a placeholder if neither file
  could be found

This addresses the issue outlined in https://github.com/cdklabs/construct-hub/issues/127

---


packages/jsii-rosetta/lib/fixtures.ts
```diff
@@ -46,10 +49,22 @@ export function fixturize(snippet: TypeScriptSnippet): TypeScriptSnippet {
...
+  loose = false,
+) {
   const fullPath = path.join(directory, literateFileName);
   const exists = fs.existsSync(fullPath);
   if (!exists) {
+    if (loose) {
+      // In loose mode, we'll fall back to the `.js` file if it exists...
+      const jsFile = fullPath.replace(/\.ts(x?)$/, '.js$1');
+      if (fs.existsSync(jsFile)) {
+        return fs.readFileSync(jsFile, { encoding: 'utf-8' });
```
iliapolo, line 64: Maybe add a header here saying we are showing `js` code because the expected file is missing?

iliapolo: Oh well, we tried

iliapolo (approved):

PR #2893, opened 2021-06-29
fix(pacmak/go): missing go.sum entry

RomainMuller:
Switch to using `go mod tidy` instead of `go mod download` to work
around https://github.com/golang/go/issues/44129. This caused tests to
fail due to using `go1.16.x`, which is susceptible to this bug (`go1.17`
will have the fix).

---


BenChaimberg (approved):
Should we be tracking 1.17 somewhere to remove this change when we eventually upgrade?

RomainMuller:
> Should we be tracking 1.17 somewhere to remove this change when we eventually upgrade?

Well _maybe_. This isn't clear, because our generated code targets `1.16` and so we should test against that release...

PR #2894, opened 2021-06-29
fix(dotnet): strong-name sign assemblies

RomainMuller:
In order for generated assemblies to be strong-name signed, their
dependencies need to also be strong-name signed, but we were not
strong-name signing the jsii runtime for .NET.

Additionally, the `AssemblyOriginatorKeyFile` was configured, but it was
not actually copied into the generated codebase, leading to the
signature process being skipped entirely (for we only sign if they key
is present).

---


rix0rrr (approved):

rix0rrr:
I've been trying to find documentation on whether or not we can reuse key pairs, because a key pair for each and every assembly looks sucky to manage (and a nightmare for CDK).

I believe it's not necessary to have individual keys... shouldn't we simplify the jsii repo as well in that case?


Also, for future reference in case someone complains about committing keys:

>  Important
>
> When the identity of the publisher of the code is desired, Authenticode and NuGet Package Signing are recommended. Code Access Security (CAS) should not be used as a security mitigation.

https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/strong-naming

RomainMuller:
> I believe it's not necessary to have individual keys... shouldn't we simplify the jsii repo as well in that case?

Correct - it is a common thing to re-use the same SNK for multiple assemblies of the same project. In the case of the `jsii-calc` et al. packages, I've intentionally used a separate key for each of them, in order to "model" that they are from different authors, and ensure that there is no interference. We could also add a check that the assemblies did _indeed_ get strong-name signed with the correct public key (right now not doing that, but actually the more I think about it...).

Our "public" packages (there are 3 of them, plus their unit test packages, and the integration test package - although those are not published) from this repository are actually all strong-name signed with the same keypair.

PR #2897, opened 2021-07-01
feat(rosetta): hoist declare statements to top-level of document

BenChaimberg:
`declare` statements allow snippet writers to create a mock variable of a certain type that can be used within a snippet without having to provide a concrete value of that type. This is useful when referencing a type from another module without needing to create an instance of that type, which may be complex. Example:
```
declare const lambdaFunction: lambda.Function;
const myResource = ...
myResource.grantRead(lambdaFunction);
```
`declare` statements can only exist at the top level of a document, meaning snippets with fixtures may not be able to define a `declare` statement, since the snippet may be embedded into the fixture within a class, function, etc. This feature "hoists" `declare` statements to the top level of the document, similarly to how `import` statements are hoisted currently.

---


eladb (commented):
Cool stuff

RomainMuller (approved):
👌🏻 Thank you so much!

PR #2902, opened 2021-07-08
revert: "fix(dotnet): strong-name sign assemblies (#2894)"

RomainMuller:
This reverts commit 7b90f27c7819209f6485c28eabefb328a0304ab5, and removes all support for .NET Strong-Name Signing, as this would be impossible to use as long as the `Amazon.Jsii.Runtime` assembly is not strong-name signed itself (which would be a breaking change to do!), since strong-name signed assemblies can only depend on other strong-name signed assemblies.

---


packages/@jsii/dotnet-runtime/src/Amazon.JSII.Runtime/AssemblyInfo.cs
```diff
@@ -1,9 +1,8 @@
...
 
 [assembly:CLSCompliant(false)] // among other things, array parameters are passed to attributes
 
-[assembly:InternalsVisibleTo("Amazon.JSII.Runtime.IntegrationTests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100d7206fa09e081f23fa91938ceede7b25c0dfb8010fa55038a859a3e0150f652e655b8f0d9c17c41bb1d2b2c5ee2eb6af43a930cd8e80c2d7fe7703b24e66a1c674f6328e44bda8000b5df4686d77be2aeef092da587672a7ca9b9a9502ad960bcb7f35154d9a855bf9db546daf79a34c9fb5ecb23c6291a92587d8ece0354e9c")]
-[assembly:InternalsVisibleTo("Amazon.JSII.Runtime.UnitTests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100d7206fa09e081f23fa91938ceede7b25c0dfb8010fa55038a859a3e0150f652e655b8f0d9c17c41bb1d2b2c5ee2eb6af43a930cd8e80c2d7fe7703b24e66a1c674f6328e44bda8000b5df4686d77be2aeef092da587672a7ca9b9a9502ad960bcb7f35154d9a855bf9db546daf79a34c9fb5ecb23c6291a92587d8ece0354e9c")]
-[assembly:InternalsVisibleTo(InternalsVisible.ToDynamicProxyGenAssembly2)]
+[assembly:InternalsVisibleTo("Amazon.JSII.Runtime.IntegrationTests")]
+[assembly:InternalsVisibleTo("Amazon.JSII.Runtime.UnitTests")]
+[assembly:InternalsVisibleTo("DynamicProxyGenAssembly2")]
```
eladb, line 8: Is this intentionally different?

packages/@jsii/spec/lib/configuration.ts
```diff
@@ -77,16 +77,6 @@ export interface Config {
        * version field
        */
       versionSuffix?: string;
-
-      /**
-       * whether the assembly should be strong-name signed
-       */
```
eladb, line 83: Is there a chance this is going to somehow break existing customers that pass these values?

eladb (commented):
I am in favor of getting rid of this redundant feature. A few questions

packages/@jsii/dotnet-runtime/src/Amazon.JSII.Runtime/AssemblyInfo.cs
RomainMuller: Yes - you gotta refer to it "by the string" if you're not-strong-name-signed, and using the constant if you are. This is awkward, I agree.

packages/@jsii/spec/lib/configuration.ts
RomainMuller: Impossible - this feature **could not possibly** have been used (it never worked!!!)

PR #2917, opened 2021-07-21
fix(rosetta): fails on "Debug Failure"

RomainMuller:
In cases where a literate source file was missing, the substitution
value was not valid TypeScript, which could cause the comipler to fail
on an opaque error (`Debug Failure`).

This falls back to the visible code for the snippet instead of inserting a
placeholder, and also makes sure to rescue the `Debug Failure` where
it is emitted from, that is in the call to
`program.getDeclarationDiagnostics`.

Both of this changes result in greater chances of transliteration
success.

Related to cdklabs/jsii-docgen#369

---

### DIFF
--- a/packages/jsii-rosetta/lib/fixtures.ts
+++ b/packages/jsii-rosetta/lib/fixtures.ts
@@ -62,7 +62,7 @@ function loadLiterateSource(
       if (fs.existsSync(jsFile)) {
         return fs.readFileSync(jsFile, { encoding: 'utf-8' });
       }
-      return `Missing literate source file ${literateFileName}`;
+      return `throw new Error("Could not render example: missing literate source file ${literateFileName}");`;
     }
     // This couldn't really happen in practice, but do the check anyway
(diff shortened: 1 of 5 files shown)


packages/jsii-rosetta/lib/fixtures.ts
```diff
@@ -62,7 +62,7 @@ function loadLiterateSource(
       if (fs.existsSync(jsFile)) {
         return fs.readFileSync(jsFile, { encoding: 'utf-8' });
       }
-      return `Missing literate source file ${literateFileName}`;
+      return `throw new Error("Could not render example: missing literate source file ${literateFileName}");`;
```
iliapolo, line 65: This is kind of funky - why have `throw new Error`? I think `Could not render example...` is clearer no?

iliapolo (approved):
Added `do-not-merge` in case you want to address the comment

iliapolo (approved):

Issue #2927, opened 2021-07-26
Python: function overrides with different parameter names are incompatible


gshpychka:
### Affected Languages

- [ ] `TypeScript` or `Javascript`
- [X] `Python`
- [ ] `Java`
- [ ] .NET (`C#`, `F#`, ...)
- [ ] `Go`

### What is the problem?
All generated Python code uses named parameters in function declarations. Python <3.8 doesn't have positional-only (unnamed) parameters at all.

As a result, any function overrides that change the parameter name make the subclass incompabitle with the parent. 

Also, each change in a parameter name is a breaking change on the Python side.

Related issue on the AWS CDK side:

https://github.com/aws/jsii/issues/4541

### Proposed solution

Treating all parameter name changes as breaking changes. Enforce the same parameter names in interface implementations.

NGL321:
Hey @gshpychka,

It looks like [a commit](https://github.com/aws/jsii/commit/c665b7e73a6c76ed6b342e7de2ad9b9ca76821e1) was made that might resolve this issue. Could you confirm this? If it did not resolve the issue we will continue to track this, otherwise we can close-out the related issues.

😸

gshpychka:
@NGL321 I don't think I have enough experience to test this, but from my understanding, that commit is still in a draft PR and hasn't been merged.

sirrus233:
Is there any progress on this? It seems like the fix in question is still in a draft state and needs to be updated.

rafrafek:
> Enforce the same parameter names in interface implementations.

Please go with this proposed solution other than the positional-only draft PR. We use Python because it is capable of programmer friendly things like keyword arguments.

Consider the example below where you want to feed cats:

```python
program_the_robot(False, True)
```

What went wrong?

If you use keyword arguments it is much easier to spot the mistake:

```python
program_the_robot(feed_cats=False, destroy_world=True)
```

As you can see keyword parameters can save the world and it is much harder to make a mistake when using them.

The positional-only draft PR is not a fix, it is a workaround trying to make Python less capable to match other languages so the error is ignored rather than fixed. It would be better to level other languages up rather than level Python down.

PR #2937, opened 2021-08-04
fix(jsii-diff): renaming a positional argument is a breaking change in Python

RomainMuller:
Related: #2927

---


gshpychka:
@RomainMuller any progress on this?

gshpychka:
@RomainMuller do you think this has any chance of getting merged? I'm assuming the lack of progress is due to the high number of deprecations required on the CDK side.

RomainMuller:
> @RomainMuller do you think this has any chance of getting merged? I'm assuming the lack of progress is due to the high number of deprecations required on the CDK side.

We are also consider alternate solutions to this...

Quite frankly... it might be "easier" to wait for Python 3.7 EOL next June, as Python 3.8+ supports "positional-only" argument notation (using the `/` modifier), which we could then adopt and forego this problem entirely.

gshpychka:
> > @RomainMuller do you think this has any chance of getting merged? I'm assuming the lack of progress is due to the high number of deprecations required on the CDK side.
> 
> 
> 
> We are also consider alternate solutions to this...
> 
> 
> 
> Quite frankly... it might be "easier" to wait for Python 3.7 EOL next June, as Python 3.8+ supports "positional-only" argument notation (using the `/` modifier), which we could then adopt and forego this problem entirely.

Got it. But that would still be a breaking change, some folks may be using them as keyword arguments.

PR #2946, opened 2021-08-10
chore: upgrade baseline node requirement to >=12.7.0

RomainMuller:
Node 10.x has been end-of-lifed on `2021-04-30`, and it is more than
time to drop support for it entirely. Per our support policy, only
current releases of runtimes are supported, and hence we are at liberty
to make this change without incurring a major version bump.

BREAKING CHANGE: the minimum supported version of `node` changed from
`10.3.0` to `12.7.0`. We recommend using `16.x` if possible.

---


RomainMuller:
This should not ship before the deadline set in https://github.com/aws/jsii/pull/2948 has expired.

tools/jsii-compliance/package.json
```diff
@@ -16,7 +16,7 @@
     "tablemark": "^2.0.0"
   },
   "devDependencies": {
-    "@types/node": "^10.17.60",
+    "@types/node": "^12.20.19",
```
Chriscbr, line 19: Slightly curious why @types/node is set to 12.20.0 even though we only require 12.7.0 - maybe worth adding a note why in the PR description.

Chriscbr (approved):

tools/jsii-compliance/package.json
RomainMuller: The `@types/node` versions are __NOT__ related to the `node` versions in any way. Only the major version number is relevant (`12.x` is for `node12`, `14.x` is for `node14`, etc...).

The `@types/node` packages are generally speaking maintained as if the API for a given node major release are always the same (i.e: once it hit GA, no new APIs get added, modified or removed).

See: https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/54465

Chriscbr: Ah! I was misled by this stackoverflow post [here](https://stackoverflow.com/questions/42035263/relationship-between-the-version-of-node-js-and-the-version-of-types-node)... but after diving deeper I see you are totally correct. Thanks for clarifying!

PR #2948, opened 2021-08-10
feat: node runtime version check

RomainMuller:
Checks the current version of `node` against a list of SemVer rangers
that are configured in `@jsii/node-check`, and emit a BIG FAT warning
to STDERR (with color if possible) in case the version is found to be
end-of-life, deprecated, or untested, so that users know that they are
subjecting themselves to random problems.

When marking a range as deprecated, a deadline is also provided in the
message, informing users about when the releases will move to the
end-of-life support level.

---

The check is performed at the entry point of all CLI utilities, including the
`@jsii/runtime` entry point.

---


packages/@jsii/node-check/lib/constants.ts
```diff
@@ -0,0 +1,57 @@
...
+  DEPRECATED = 'end-of-life',
+
+  /**
+   * Supported releases, nothing to see. No message will be output.
+   */
+  SUPPORTED = 'supported',
+  /**
+   * EOL releases are unsupported, so this will emit a SCARY RED WARNING.
+   */
+  END_OF_LIFE = 'unsupported',
```
Chriscbr, line 16: Any reason for naming enum choices different from their string values? I feel like this could cause confusion perhaps

packages/@jsii/node-check/test/intex.test.ts
```diff
@@ -0,0 +1,20 @@
+import {
```
MrArnoldPalmer, line 1: I assume this file name should be `index.test.ts`?

packages/jsii-diff/bin/jsii-diff.ts
```diff
@@ -4,6 +4,8 @@ import * as reflect from 'jsii-reflect';
 import * as log4js from 'log4js';
 import * as yargs from 'yargs';
 
+import '@jsii/node-check';
```
MrArnoldPalmer, line 7: I wonder if having `@jsii/node-check` be a script that just runs on import vs a function that needs to be called explicitly is the way to go? Since this is a separate module that is consumed by various CLIs and the runtime it seems like we should let those consumers decide where exactly this runs in their lifecycle?

packages/@jsii/node-check/lib/constants.ts
RomainMuller: Oh oops! This is the result of me being indecisive about naming... 🤓

packages/jsii-diff/bin/jsii-diff.ts
RomainMuller: I was trying to go for a boilerplate-free approach as much as possible. Import & forget is a pretty simple way to go... but I can actually offer both...

Have it export a check function... And have an import-and-run name like `@jsii/node-check/run` (although I don't love that particular name).

packages/@jsii/check-node/README.md
```diff
@@ -0,0 +1,27 @@
+# `@jsii/check-node`
+
+This package can be used to check whether the node runtime used to run an
+application is known to be compatible with the `jsii` constellation of packages,
```
Chriscbr, line 4: I really like the visual I have of jsii packages as a constellation! ✨

Chriscbr (approved):
Leaving pr/do-not-merge in case @MrArnoldPalmer wants to take another look.

RomainMuller:
> Leaving pr/do-not-merge in case @MrArnoldPalmer wants to take another look.

Well I do need this to ship sooner rather than later, so I removed the label... That doesn't mean @MrArnoldPalmer doesn't get to have another look (but feedback will be addressed in a follow-up)

PR #2965, opened 2021-08-20
feat: @struct type system hint

RomainMuller:
Using the hint allows any interface to always be treated as a struct (as
long as all it's properties are readonly and it has no methods) even if
it has an `I`-prefixed name.

---

### DIFF
--- a/packages/@jsii/spec/lib/assembly.ts
+++ b/packages/@jsii/spec/lib/assembly.ts
@@ -358,32 +358,11 @@ export interface Docs {
   default?: string;
 
   /**
-   * Hints provided to the compiler via TSDoc annotations.
-   */
-  hints?: TypeSystemHints;
-
-  /**
    * Custom tags that are not any of the default ones
    *
    * @default none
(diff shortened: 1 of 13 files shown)


packages/@jsii/spec/lib/assembly.ts
```diff
@@ -357,6 +357,11 @@ export interface Docs {
    */
   default?: string;
 
+  /**
+   * Hints provided to the compiler via TSDoc annotations.
+   */
+  hints?: TypeSystemHints;
```
rix0rrr, line 363: Does this type hint even need to persist in the assembly?

It does not add information, and the only thing its presence here will do is allow downstream consumers to make mistakes, because they can now accidentally look at `docs.hints.struct` instead of `datatype`, which they should hav ebeen.

eladb: +1 feels like a leak

rix0rrr (approved):

PR #2989, opened 2021-09-16
feat: added warnings for usage of deprecated elements

otaviomacedo:
## Contained in this draft
* Handling of deprecated types in the inheritance graph
* Handling of deprecated methods and functions
* Code generation (this may need a different approach. See the TODOs)
* Feature flag

## Coming up
* Handling of parameters
* Resolution of TODOs
* Unit tests
* README
* Other things that I probably missed

---

### DIFF
--- a/packages/jsii/lib/transforms/deprecated-warnings-transformer.ts
+++ b/packages/jsii/lib/transforms/deprecated-warnings-transformer.ts
@@ -1,0 +1,258 @@
+import {
+  Assembly,
+  ClassType,
+  InterfaceType,
+  isClassOrInterfaceType,
+  Stability,
+} from '@jsii/spec';
+import * as ts from 'typescript';
+import { Block } from 'typescript';
+
+import * as bindings from '../node-bindings';
+import { fullyQualifiedName, isDeprecated } from './utils';
+
+interface Warning {
+  readonly elementName: string;
+  readonly message: string;
+}
+
+export class DeprecatedWarningInjector {
+  private readonly index = new Map<
+    string,
+    ts.ClassDeclaration | ts.InterfaceDeclaration | undefined
+  >();
+  private deprecatedTypes: ts.Type[] = [];
+
+  public constructor(private readonly typeChecker: ts.TypeChecker) {}
+
+  public get customTransformers(): ts.CustomTransformers {
+    return {
+      before: [
+        (context) => {
+          const transformer = new DeprecatedWarningsTransformer(
+            this.typeChecker,
+            context,
+            this.deprecatedTypes,
+            this.index,
+          );
+          return transformer.transform.bind(transformer);
+        },
+      ],
+    };
+  }
+
+  public process(assembly: Assembly) {
+    if (assembly.types != null) {
+      const types: Array<InterfaceType | ClassType> = Object.values(
+        assembly.types,
+      ).filter(isClassOrInterfaceType);
+
+      for (const type of types) {
+        const node = bindings.getClassOrInterfaceRelatedNode(type)!;
+        this.index.set(fullyQualifiedName(this.typeChecker, node)!, node);
+      }
+
+      this.deprecatedTypes = Object.values(assembly.types)
+        .filter((typeInfo) => typeInfo.docs?.stability === Stability.Deprecated)
+        .filter(isClassOrInterfaceType)
+        .map((typeInfo) => bindings.getClassOrInterfaceRelatedNode(typeInfo)!)
+        .map((node) => this.typeChecker.getTypeAtLocation(node));
+    }
+  }
+}
+
+class DeprecatedWarningsTransformer {
+  private addWarning = false;
+  private warning?: Warning;
+
+  public constructor(
+    private readonly typeChecker: ts.TypeChecker,
+    private readonly context: ts.TransformationContext,
+    private readonly deprecatedTypes: ts.Type[],
+    private readonly index: Map<
+      string,
+      ts.ClassDeclaration | ts.InterfaceDeclaration | undefined
+    >,
+  ) {}
+
+  public transform<T extends ts.Node>(node: T): T {
+    return this.visitEachChild(node);
+  }
+
+  private visitEachChild<T extends ts.Node>(node: T): T {
+    return ts.visitEachChild(node, this.visitor.bind(this), this.context);
+  }
+
+  private visitor<T extends ts.Node>(node: T): ts.VisitResult<T> {
+    if (ts.isClassDeclaration(node)) {
+      const deprecatedType = this.findDeprecated([
+        this.typeChecker.getTypeAtLocation(node),
+      ]);
+      this.addWarning = deprecatedType != null;
+      if (this.addWarning) {
+        const deprecatedNode = this.index.get(
+          this.typeChecker.getFullyQualifiedName(deprecatedType!.symbol),
+        );
+        this.warning = getWarning(this.typeChecker, deprecatedNode!);
+      }
+    }
+
+    if (ts.isFunctionDeclaration(node)) {
+      this.addWarning = isDeprecated(node);
+      if (this.addWarning) {
+        this.warning = getWarning(this.typeChecker, node);
+      }
+    }
+
+    if (isMethodLikeDeclaration(node)) {
+      this.addWarning = this.addWarning || isDeprecated(node);
+      if (isDeprecated(node)) {
+        this.warning = getWarning(this.typeChecker, node);
+      }
+    }
+
+    if (ts.isBlock(node) && this.addWarning) {
+      const block = node as Block;
+      const warningStatement = this.createWarningStatement();
+      return ts.updateBlock(
+        node,
+        ts.createNodeArray([...warningStatement, ...block.statements]),
+      ) as any;
+    }
+
+    return this.visitEachChild(node);
+  }
+
+  private createWarningStatement() {
+    const warning = this.warning;
+    const message = `[WARNING] ${warning?.elementName} is deprecated.\n  ${warning?.message}\n  This API will be removed in the net major release.`;
+
+    // TODO There should be a simpler way...
+    return [
+      // TODO Instead of declaring the function everywhere it's used, should we create a single declaration somewhere and reuse it?
+      //  If so, where should this declaration go?
+      ts.createFunctionDeclaration(
+        undefined,
+        undefined,
+        undefined,
+        'printWarning',
+        [],
+        [],
+        undefined,
+        ts.createBlock(
+          [
+            ts.createExpressionStatement(
+              ts.createAssignment(
+                ts.createIdentifier('const deprecated'),
+                ts.createIdentifier('process.env.DEPRECATED'),
+              ),
+            ),
+            ts.createExpressionStatement(
+              ts.createAssignment(
+                ts.createIdentifier('const deprecationMode'),
+                ts.createIdentifier(
+                  "['warn', 'error', 'quiet'].includes(deprecated) ? deprecated : \"warn\"",
+                ),
+              ),
+            ),
+            ts.createExpressionStatement(
+              ts.createAssignment(
+                ts.createIdentifier('const message'),
+                ts.createLiteral(
+                  `${warning?.elementName} is deprecated.\n  ${warning?.message}\n  This API will be removed in the net major release.`,
+                ),
+              ),
+            ),
+            // ts.createExpressionStatement(
+            ts.createSwitch(
+              ts.createIdentifier('deprecationMode'),
+              ts.createCaseBlock([
+                ts.createCaseClause(ts.createLiteral('error'), [
+                  ts.createThrow(
+                    ts.createNew(ts.createIdentifier('Error'), [], []),
+                  ),
+                  ts.createBreak(),
+                ]),
+                ts.createCaseClause(ts.createLiteral('warn'), [
+                  ts.createExpressionStatement(
+                    ts.createCall(
+                      ts.createIdentifier('console.warn'),
+                      [],
+                      [ts.createLiteral(message)],
+                    ),
+                  ),
+                  ts.createBreak(),
+                ]),
+              ]),
+            ),
+          ],
+          true,
+        ),
+      ),
+      ts.createExpressionStatement(
+        ts.createCall(ts.createIdentifier('printWarning'), [], []),
+      ),
+    ];
+  }
+
+  private findDeprecated(toProcess: ts.Type[]): ts.Type | undefined {
+    if (toProcess.length === 0) {
+      return undefined;
+    }
+
+    if (this.deprecatedTypes.includes(toProcess[0])) {
+      return toProcess[0];
+    }
+
+    const node = this.index.get(
+      this.typeChecker.getFullyQualifiedName(toProcess[0].symbol),
+    )!;
+    if (!ts.isClassDeclaration(node) && !ts.isInterfaceDeclaration(node)) {
+      throw new Error('Node should be a class or interface declaration.');
+    }
+
+    const newBatch = toProcess.slice(1);
+    const heritageClauses = node.heritageClauses ?? [];
+    for (const clause of heritageClauses) {
+      for (const expression of clause.types) {
+        const type = this.typeChecker.getTypeAtLocation(expression);
+        if (!toProcess.includes(type)) {
+          newBatch.push(type);
+        }
+      }
+    }
+
+    return this.findDeprecated(newBatch);
+  }
+}
+
+function isMethodLikeDeclaration(node: ts.Node): boolean {
+  return (
+    ts.isMethodDeclaration(node) ||
+    ts.isGetAccessorDeclaration(node) ||
+    ts.isSetAccessorDeclaration(node) ||
+    ts.isConstructorDeclaration(node)
+  );
+}
+
+function getWarning(
+  typeChecker: ts.TypeChecker,
+  node: ts.Node,
+): Warning | undefined {
+  const original = ts.getOriginalNode(node);
+  const warnings = ts
+    .getJSDocTags(original)
+    .filter(
+      (tag) => (tag.tagName.text ?? tag.tagName.escapedText) === 'deprecated',
+    )
+    .map((tag: ts.JSDocTag) => {
+      const type = typeChecker.getTypeAtLocation(node);
+      // TODO We need a better fully qualified name
+      const fqn = typeChecker.getFullyQualifiedName(type.symbol);
+      return {
+        elementName: fqn,
+        message: tag.comment,
+      } as Warning;
+    });
+  return warnings[0];
+}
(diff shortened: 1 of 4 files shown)


packages/jsii/lib/transforms/deprecated-warnings-transformer.ts
```diff
@@ -0,0 +1,258 @@
...
+      const types: Array<InterfaceType | ClassType> = Object.values(
+        assembly.types,
+      ).filter(isClassOrInterfaceType);
+
+      for (const type of types) {
+        const node = bindings.getClassOrInterfaceRelatedNode(type)!;
+        this.index.set(fullyQualifiedName(this.typeChecker, node)!, node);
+      }
+
+      this.deprecatedTypes = Object.values(assembly.types)
```
otaviomacedo, line 55: There is some repetition here that we can probably optimize away.

packages/jsii/lib/transforms/deprecated-warnings-transformer.ts
```diff
@@ -0,0 +1,258 @@
...
+    }
+
+    return this.visitEachChild(node);
+  }
+
+  private createWarningStatement() {
+    const warning = this.warning;
+    const message = `[WARNING] ${warning?.elementName} is deprecated.\n  ${warning?.message}\n  This API will be removed in the net major release.`;
+
+    // TODO There should be a simpler way...
```
otaviomacedo, line 130: I tried using `ts.createSourceFile()` with the whole code, but it makes a mess in the generated file.

packages/jsii/lib/transforms/deprecated-warnings-transformer.ts
```diff
@@ -0,0 +1,258 @@
...
+    return [
+      // TODO Instead of declaring the function everywhere it's used, should we create a single declaration somewhere and reuse it?
+      //  If so, where should this declaration go?
```
otaviomacedo, line 133: Any ideas here are welcome.

packages/jsii/lib/transforms/deprecated-warnings-transformer.ts
```diff
@@ -0,0 +1,258 @@
...
+): Warning | undefined {
+  const original = ts.getOriginalNode(node);
+  const warnings = ts
+    .getJSDocTags(original)
+    .filter(
+      (tag) => (tag.tagName.text ?? tag.tagName.escapedText) === 'deprecated',
+    )
+    .map((tag: ts.JSDocTag) => {
+      const type = typeChecker.getTypeAtLocation(node);
+      // TODO We need a better fully qualified name
```
otaviomacedo, line 250: The name that gets generated here is something like `"/Users/otaviom/projects/experiments/jsii-experiment/index".MyClass`. But we probably want the JSII style: `jsii-experiment.MyClass`.

packages/jsii/lib/transforms/deprecated-warnings-transformer.ts
```diff
@@ -0,0 +1,354 @@
...
+
+function createWarningStatement(warning: Warning): ts.Statement[] {
+  const message = `${warning?.elementName} is deprecated.\n  ${warning?.message}\n  This API will be removed in the net major release.`;
+  // TODO Are random numbers a good idea?
+  const functionName = `printWarning${getRandomInt(
+    0,
+    Number.MAX_SAFE_INTEGER,
+  )}`;
+
+  // TODO There must be a simpler way...
```
otaviomacedo, line 247: I tried using `ts.createSourceFile()` with the whole code, but it makes a mess in the generated file.

packages/jsii/lib/transforms/deprecated-warnings-transformer.ts
```diff
@@ -0,0 +1,413 @@
...
+          newBatch.push(type);
+        }
+      }
+    }
+
+    return this.findDeprecatedInTheTypeHierarchy(newBatch);
+  }
+
+  private createWarningStatement(warning: Warning): ts.Statement[] {
+    const message = `${warning?.elementName} is deprecated.\n  ${warning?.message}\n  This API will be removed in the net major release.`;
```
RomainMuller, line 290: Why all the `?.`? `warning` is a required parameter here...

```suggestion
    const message = `${warning.elementName} is deprecated.\n  ${warning.message}\n  This API will be removed in the next major release.`;
```

packages/jsii/lib/transforms/deprecated-warnings-transformer.ts
```diff
@@ -0,0 +1,413 @@
...
+    // TODO Are random numbers a good idea?
+    const functionName = `printWarning${getRandomInt(
+      0,
+      Number.MAX_SAFE_INTEGER,
+    )}`;
```
RomainMuller, line 295: There is an TS API that allows you to get a new identifier for a given scope... It'll make sure there are no conflicts.

packages/jsii/lib/transforms/deprecated-warnings-transformer.ts
```diff
@@ -0,0 +1,413 @@
...
+        functionName,
+        [],
+        [],
+        undefined,
+        ts.createBlock(
+          [
+            ts.createExpressionStatement(
+              ts.createAssignment(
+                ts.createIdentifier('const deprecated'),
+                ts.createIdentifier('process.env.DEPRECATED'),
```
RomainMuller, line 314: That environment variable name is way too generic IMHO. Should likely be `JSII_DEPRECATIONS` or something along those lines.

packages/jsii/lib/transforms/deprecated-warnings-transformer.ts
```diff
@@ -0,0 +1,413 @@
...
+            ts.createExpressionStatement(
+              ts.createAssignment(
+                ts.createIdentifier('const message'),
+                ts.createLiteral(message),
+              ),
+            ),
+            ts.createSwitch(
+              ts.createIdentifier('deprecationMode'),
+              ts.createCaseBlock([
+                ts.createCaseClause(ts.createLiteral('error'), [
```
RomainMuller, line 334: I'd call this mode `throw` or `fail` instead of `error`, because otherwise I'd have expected this outputs a message to STDERR.

packages/jsii/lib/transforms/deprecated-warnings-transformer.ts
```diff
@@ -0,0 +1,413 @@
...
+    if (ts.isFunctionDeclaration(node)) {
+      this.handleFunctionDeclaration(node);
+    }
+
+    if (isMethodLikeDeclaration(node)) {
+      this.handleMethodLikeDeclaration(node);
+    }
+
+    if (ts.isBlock(node)) {
+      if (this.warnings.length > 0) {
```
rix0rrr, line 122: I would not do it like this.

As far as I know (correct me if wrong, but this is what I believe) a "block" is any pair of braces `{ ... }`, so also a pair of braces that could occur within something other than a function declaration like `if (condition) { ... }`.

Right now, it looks like your code assumes that we will do something with warnings in the "first block" that appears after a function declaration? There is some reliance on ordering there that's making me uncomfortable, through mutating the shared state of `this.warnings`.

Can't we just get the block from the method declaration, which we KNOW is the main function body, and work on that directly?

Or am I just misunderstanding how this works?

rix0rrr (commented):
This does not yet check for deprecated struct fields, right?

It's also an interesting question of where we generate the functions necessary for doing those. In the source file of the interface, or in each consuming source file?

The former seems best, but I don't know how hard that is to achieve...

packages/jsii/lib/transforms/deprecated-warnings-transformer.ts
```diff
@@ -0,0 +1,413 @@
...
+  private visitor<T extends ts.Node>(node: T): ts.VisitResult<T> {
+    if (ts.isClassDeclaration(node)) {
+      this.handleClassDeclaration(node);
+    }
+
+    if (ts.isFunctionDeclaration(node)) {
+      this.handleFunctionDeclaration(node);
+    }
+
+    if (isMethodLikeDeclaration(node)) {
```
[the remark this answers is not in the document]
rix0rrr: Oh I see, you added it yourself. Disregard.

packages/jsii/lib/transforms/deprecated-warnings-transformer.ts
```diff
@@ -0,0 +1,413 @@
...
+import * as ts from 'typescript';
+import { Block } from 'typescript';
+
+import * as bindings from '../node-bindings';
+import { fullyQualifiedName, isDeprecated } from './utils';
+
+type MethodLikeDeclaration =
+  | ts.MethodDeclaration
+  | ts.ConstructorDeclaration
+  | ts.FunctionDeclaration
```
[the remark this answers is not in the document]
otaviomacedo: Are you talking specifically about function declarations?

packages/jsii/lib/transforms/deprecated-warnings-transformer.ts
```diff
@@ -0,0 +1,413 @@
...
+interface Warning {
+  readonly elementName: string;
+  readonly message: string;
+  readonly type: ElementType;
+}
+
+export class DeprecatedWarningInjector {
+  private readonly index = new Map<
+    string,
+    ts.ClassDeclaration | ts.InterfaceDeclaration | undefined
```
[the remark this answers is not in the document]
otaviomacedo: Yes, you're right.

packages/jsii/lib/transforms/deprecated-warnings-transformer.ts
```diff
@@ -0,0 +1,413 @@
...
+  >();
+  private deprecatedTypes: ts.Type[] = [];
```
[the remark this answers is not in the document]
otaviomacedo: I think a `Set` is perfectly appropriate here, thanks.

packages/jsii/lib/transforms/deprecated-warnings-transformer.ts
otaviomacedo: That's a good idea! I had missed the `body` property. However, some reliance on the stated of `this.warnings` is still necessary, I believe; when an entire class is deprecated, for example.

packages/jsii/lib/transforms/deprecated-warnings-transformer.ts
otaviomacedo: Yes, they shouldn't be there. Thanks for catching this.

packages/jsii/lib/transforms/deprecated-warnings-transformer.ts
otaviomacedo: Thanks, I'll look into that.

packages/jsii/lib/transforms/deprecated-warnings-transformer.ts
```diff
@@ -0,0 +1,413 @@
...
+        }
+      }
+    }
+
+    return this.findDeprecatedInTheTypeHierarchy(newBatch);
+  }
+
+  private createWarningStatement(warning: Warning): ts.Statement[] {
+    const message = `${warning?.elementName} is deprecated.\n  ${warning?.message}\n  This API will be removed in the net major release.`;
+    // TODO Are random numbers a good idea?
```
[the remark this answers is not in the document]
otaviomacedo: I'll follow up on Romain's idea to get rid of this.

packages/jsii/lib/transforms/deprecated-warnings-transformer.ts
otaviomacedo: Is there any argument against adding a new file? Otherwise, I'm onboard with this idea.

packages/jsii/lib/transforms/deprecated-warnings-transformer.ts
otaviomacedo: Makes sense. I'll rename it.

packages/jsii/lib/transforms/deprecated-warnings-transformer.ts
otaviomacedo: Makes sense, too.

packages/jsii/lib/transforms/deprecated-warnings-transformer.ts
```diff
@@ -0,0 +1,413 @@
...
+  | ts.FunctionDeclaration
+  | ts.GetAccessorDeclaration
+  | ts.SetAccessorDeclaration;
+
+enum ElementType {
+  METHOD = 0,
+  FUNCTION = 1,
+  CLASS = 2,
+  PARAMETER = 3,
+}
```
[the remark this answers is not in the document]
otaviomacedo: To be fair, I'm not happy with this enum, either. But I think if I follow Rico's suggestion of processing the body of the method as soon as I see it, I won't need this enum anymore.

otaviomacedo:
> This does not yet check for deprecated struct fields, right?

It does, in `getParameterWarnings` (I guess I should add some comments to guide the reading of that method). But it's not yet checking whether the field has been set. So it's too broad at the moment. I'm working on that right now.
 
> It's also an interesting question of where we generate the functions necessary for doing those. In the source file of the interface, or in each consuming source file?
> 
> The former seems best, but I don't know how hard that is to achieve...

By the source file of the interface, do you mean the `.d.ts` file?

packages/jsii/lib/transforms/deprecated-warnings-transformer.ts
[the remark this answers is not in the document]
RomainMuller: Yes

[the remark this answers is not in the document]
otaviomacedo: Cool, I'll remove it, then.

packages/jsii/test/deprecation-warnings-transformer.test.ts
```diff
@@ -0,0 +1,425 @@
...
+    export class Foo {
+      public bar(policy: SecurityPolicy){return policy;}
+    }
+    `,
+      undefined /* callback */,
+      { addDeprecationWarnings: true },
+    );
+
+    jsFile(result);
+    fail('How should we handle this case?');
```
otaviomacedo, line 57: The problem here is this: the third argument to `printJsiiDeprecationWarnings` is the value of the deprecated element, as a de-referencing chain (`foo?.bar?.zee`), starting from the parameter, if applicable. This is used by the function to decide whether to show the warning. For example, let's say in the source code compiled by JSII we have:

```ts
interface FooProps {
  readonly fooName: string,
 
  /** @deprecated Blah blah blah */
  readonly bar: string,
}
```
 
And in the user's application code we have:
 
```ts
new Foo(this, 'foo', {
  fooName: 'myFoo',
});
```

Since they are only passing the `fooName` property, which is not deprecated, no warning should be printed. The solution in this PR is to inject a call like this in the constructor:

```ts
printJsiiDeprecationWarnings("my-package.FooProps.bar", "Blah blah blah", props.bar);
```

Because `props.bar` is `undefined`, the function will not print the warning.

So far, so good. Now if we have deprecated enum elements, like:

```ts
enum MyEnum {
  /** @deprecated */
  P1 = 0,
  P2 = 1
}

class Foo {
   bar(e: MyEnum){...}
}
```
and the user's code is:

```ts
foo.bar(MyEnum.P1); // <-- SHOULD print a warning
foo.bar(MyEnum.P2); // <-- SHOULD NOT print a warning
```
What is the best way to address this? The best solution I can come up with is to generate a call for each deprecated enum member like this:

```ts
printJsiiDeprecationWarnings("my-package.MyEnum.P1", message, e === MyEnum.P1 ? e : undefined);
```

This would work, but my concern is the use of `MyEnum`. Depending on how it was imported, we would need to use different namespaces (or none) to reference it. Any ideas?

packages/jsii/lib/transforms/deprecation-warnings-transformer.ts
```diff
@@ -0,0 +1,618 @@
...
+such as deprecated types and class and interface declarations. It also creates a
+new Javascript file with a function that prints warnings.
+
+At emit time, DeprecatedWarningsTransformer.transform is invoked for each source
+file and then recursively for each node in the AST. We are interested in two
+types of nodes: class declarations and method-like declarations (proper methods,
+constructors and accessors).
+
+For classes, we navigate up in the type hierarchy looking for deprecated types.
+If one is found, a warning for that will added to each method of the class. For
```
rix0rrr, line 50: And I'm also not sure we need to do that. We would already get the deprecation warning when the class is constructed; what would additional deprecation warnings add every time we then call a member method on the class?

On the other hand, for STATIC methods we probably DO need to do this.

packages/jsii/lib/transforms/deprecation-warnings-transformer.ts
```diff
@@ -0,0 +1,618 @@
...
+invoked so that it can compute and store general information about the module,
+such as deprecated types and class and interface declarations. It also creates a
+new Javascript file with a function that prints warnings.
+
+At emit time, DeprecatedWarningsTransformer.transform is invoked for each source
+file and then recursively for each node in the AST. We are interested in two
+types of nodes: class declarations and method-like declarations (proper methods,
+constructors and accessors).
+
+For classes, we navigate up in the type hierarchy looking for deprecated types.
```
rix0rrr, line 49: Wait. I'm not sure we need to do that. 

I don't think it's true that if a class' base class is deprecated, it automatically means the subclass is deprecated.

packages/jsii/lib/transforms/deprecation-warnings-transformer.ts
rix0rrr: I'm somewhat half-convinced the behavior of the assembler should already be to push deprecation markers down to subtypes, if applicable (so that if a method is deprecated in a superclass, it also counts as deprecated in a subclass). 

Not actually sure if that is true or not... but especially for consistency with the `deprecated-remover`, I would hope we don't do complex type analysis here and trust the analysis and markings done by the `assembler` already. We should centralize those kinds of decisions if we can.

packages/jsii/test/deprecation-warnings-transformer.test.ts
rix0rrr: I think for every type we need a function, and call them recursively:

```ts
enum MyEnum {
  /** @deprecated */
  P1 = 0,
  P2 = 1
}

interface FooProps {
  readonly okay: string;

  /** @deprecated */
  readonly ohno: string;

  readonly enum: MyEnum;
}

class Foo {
   bar(props: FooProps) {
  }
}
```

To my mind should synth to something like this:

```ts
class Foo {
  bar(props) { 
    $_checkDeprecations_FooProps(props, `Foo#bar: fooprops`);
  }
}

function $_checkDeprecations_FooProps(props, path) {
  if (props?.ohno) { warnDeprecation(`${path}.ohno`); }
  $_checkDeprecations_MyEnum(props?.enum, `${path}.enum`);
}

function $_checkDeprecations_MyEnum(enum, path) {
  if (enum === MyEnum.P1) { warnDeprecation(`${path}.P1`); }
}
```

Or something. Functions will save on generated code if the same interface type is referenced in multiple places, and is also the only way to deal with recursive types:

```ts
interface MyStruct {
  thing: string;
  substructs: MyStruct[];
}
```

PR #2997, opened 2021-09-22
fix(jsii): improve checks around peerDependencies

### LINKED ISSUE #2994: jsii should help users get devDependencies/peerDependencies right
## :rocket: Feature Request

- Right now, jsii adds `dependencies` to `peerDependencies`. It should not, that is wrong.
- jsii should make sure `peerDependencies` are in `devDependencies`, with a point range that is the bottom of the caret range in `peerDependencies`.
- jsii should make sure `peerDependencies` *are not* in `dependencies` <-- this will break all v1 packages, so needs to go behind a flag probably... ?? :/

rix0rrr:
There are two changes here:

- The check that forces package authors to list all `dependencies` in
  `peerDependencies` is superfluous, and has been removed. This
  effectively removes the `--fix-peer-dependencies` flag as well.
- Add a warning to tell people that when they add a `peerDependency`,
  they should have a dependency in `devDependencies` as well.

Fixes #2994.

---


packages/jsii/bin/jsii.ts
```diff
@@ -106,20 +104,20 @@ const warningTypes = Object.keys(enabledWarnings);
...
+  for (const diagnostic of allDiagnostics) {
+    utils.logDiagnostic(diagnostic, projectRoot);
+  }
+  if (emitResult.emitSkipped) {
+    process.exitCode = 1;
+  }
+})().catch((e) => {
+  console.error(`Error: ${e.stack}`);
+  process.exitCode = -1;
+});
```
RomainMuller, line 120:
```suggestion
   const result = await (argv.watch ? compiler.watch() : compiler.emit());

   const allDiagnostics = [...projectInfoDiagnostics, ... result.diagnostics];

   for (const diagnostic of allDiagnostics) {
     utils.logDiagnostic(diagnostic, projectRoot);
   }
   if (result.emitSkipped) {
     process.exitCode = 1;
   }
 })().catch((e) => {
   console.error(`Error: ${e.stack}`);
   process.exitCode = -1;
 });
```

packages/jsii/lib/helpers.ts
```diff
@@ -145,9 +145,8 @@ async function makeProjectInfo(
     spaces: 2,
   });
 
-  return loadProjectInfo(path.resolve(process.cwd(), '.'), {
-    fixPeerDependencies: true,
-  });
+  const result = await loadProjectInfo(path.resolve(process.cwd(), '.'));
+  return result.projectInfo;
```
RomainMuller, line 149:
```suggestion
  const { projectInfo } = await loadProjectInfo(path.resolve(process.cwd(), '.'));
  return projectInfo;
```

packages/jsii/lib/jsii-diagnostic.ts
```diff
@@ -238,6 +238,20 @@ export class JsiiDiagnostic implements ts.Diagnostic {
     name: 'metadata/missing-peer-dependency',
   });
 
+  public static readonly JSII_0006_MISSING_DEV_DEPENDENCY = Code.warning({
```
RomainMuller, line 241: Maybe drop a comment here stating that _for now_, it is not possible to increase the severity of this because it's generated _before_ the severity settings are applied?

packages/jsii/lib/project-info.ts
```diff
@@ -8,6 +8,8 @@ import * as ts from 'typescript';
 
 import { parsePerson, parseRepository } from './utils';
 
+import { JsiiDiagnostic } from '.';
```
RomainMuller, line 11:
```suggestion
import { JsiiDiagnostic } from './jsii-diagnostics';
```

packages/jsii/bin/jsii.ts
```diff
@@ -41,7 +40,7 @@ const warningTypes = Object.keys(enabledWarnings);
           .option('fix-peer-dependencies', {
             type: 'boolean',
             default: true,
-            desc: 'Automatically add missing entries in the peerDependencies section of package.json',
+            desc: 'Automatically add missing entries in the peerDependencies section of package.json (deprecated, defunct)',
```
RomainMuller, line 43: I'd say we hide this from the output of `jsii --help` and make it clear that it is now NOOP.

```suggestion
            desc: 'This option no longer has any effect.',
            hidden: true,
```

RomainMuller (approved):

PR #3018, opened 2021-10-04
chore(superchain): set UID and GID of superchain user

RomainMuller:
In order to facilitate use with GitHub Actions, explicitly set the UID and GID
of the `superchain` user to `1001`, which is the UID/GID set on the GitHub
working directory on the file system.

---

### DIFF
--- a/superchain/Dockerfile
+++ b/superchain/Dockerfile
@@ -31,7 +31,7 @@ RUN apt-get update && apt-get install -y curl gpg tar zsh
 SHELL ["/bin/zsh", "-c"]
 
 # Prepare maven binary distribution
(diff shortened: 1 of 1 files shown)


eladb (approved):

eladb:
@RomainMuller Should we move superchain to a separate repository so it can evolve independently of jsii?

RomainMuller:
> @RomainMuller Should we move superchain to a separate repository so it can evolve independently of jsii?

This would fit in the scope of https://github.com/aws/aws-cdk-rfcs/pull/373 -- I believe `jsii/superchain` should actually be tied to the compiler and runtime requirements... So it sort of makes sense to have it all in one place.

PR #3069, opened 2021-10-15
chore(superchain): add libffi-dev, explicitly install python3-dev

RomainMuller:
The `cryptography` Python package may require `libffi-dev` to be
available for installs on operating systems / architectures that are not
supported by `manylinux`. This appears to now be the case of Debian<11,
Python<3.8.

Having `libffi` on the image is going to alleviate this issue.

Also, moved the `rust` install to happen before the `python3` install,
as it is a logical dependency of `python` (it is only installed because
`cryptography` might need it).

---


RomainMuller:
@mergifyio refresh

PR #3101, opened 2021-10-27
feat(rosetta): reuse translation results from a cache

rix0rrr:
By fingerprinting types in an assembly and storing the fingerprints along with translations, we can reuse existing translations if we know nothing changed.

A snippet needs to be recompiled if:

- Its source text changes: hash key will be different
- Its fixture changes: fullSource will be different
- The referenced types have changed: fingerprint will be different
- The translator's capabilities have changed: translation version will be different

Since we have many duplicate snippets (especially `cluster.enableSingleUserRotation()` crops up a lot 🙂), I've had to encode the source location for the snippet into the key. Since that means that location encoding is now significant and needs to be correct, some refactoring around encoding snippet locations has become important.

Incidentally, this will also fix broken translations originating from `monocdk` from overwriting perfectly cromulent translations in the `@aws-cdk` packages.

Refactorings in this PR:

- Snippet source locations are no longer a semi-free form string -- they are now a structured data type that captures the location accurately (and incidentally gets rendered to a filename-like string for interaction with the TS compiler and human display).
- Snippet source locations now need to be passed in every place where some code is passed to Rosetta. These take the form of `{ api: 'member', fqn: '@ns/package.type', memberName: 'bucketName' }` (an `ApiLocation`), and have been added to many, many places, including `jsii-pacmak`.
- Translators now have a `VERSION`, which is used to invalidate cached translations if we change something about the implementation of the translator.
- Existing tablet files may now need to span multiple jsii release versions, so they are now validated based on their schema version instead of the version of the tool that produced them.
- `TranslatedSnippet` had a copy of all the data in a `TranslatedSnippetSchema`, and was copying back and forth. Change it to just hold on to a `TranslatedSnippet` internally. It now represents the operations that can be done to a `TranslatedSnippetSchema`. 
- The translation of `ts.Diagnostic` to `RosettaDiagnostic` is now done even earlier (closer to the code that deals with the TS compiler).
- Snippets are no longer in an `IterableIterator`, but just an plain old array.
- Snippets in Markdown are no longer identified by their snippet counter, but by their line number.
- Mutable fields turned into `readonly` fields on a number of structs.

---


packages/jsii-rosetta/lib/languages/csharp.ts
```diff
@@ -70,6 +70,14 @@ interface CSharpLanguageContext {
 type CSharpRenderer = AstRenderer<CSharpLanguageContext>;
 
 export class CSharpVisitor extends DefaultVisitor<CSharpLanguageContext> {
+  /**
+   * Translation version
+   *
+   * Bump this when you change something in the implementation to invalidate
+   * existing cached translations.
+   */
+  public static readonly VERSION = '1';
```
njlynch, line 79: Food for (future) thought -- this feels very much like a good intentions thing. Could we mechanize this a bit better somehow?

packages/jsii-rosetta/lib/tablets/tablets.ts
```diff
@@ -67,60 +79,43 @@ export class LanguageTablet {
...
-  private _didCompile?: boolean;
-  private _where = '';
-  private fullSource?: string;
-
-  private constructor() {}
-
-  public get didCompile() {
-    return this._didCompile;
-  }
+  public readonly snippet: TranslatedSnippetSchema;
```
njlynch, line 111: `~` The naming here is a bit odd. `TranslatedSnippet.snippet`, which returns a `TranslatedSnippetSchema`, which reads to me like the snippet itself. It leads to a lot of `snippet.snippet` in the code, which is never great...

It's _almost_ like `TranslatedSnippet` should implement `TranslatedSnippetSchema`. Am I off there?

njlynch (approved):
Honestly, my eyes did glaze over a bit here at one point, but I think I was paying attention for all the salient points.

🤞

packages/jsii-rosetta/lib/commands/extract.ts
```diff
@@ -70,21 +99,17 @@ interface TranslateAllResult {
...
 }
 
 /**
  * Translate all snippets
  *
  * We are now always using workers, as we are targeting Node 12+.
  */
-async function translateAll(
-  snippets: IterableIterator<TypeScriptSnippet>,
+export async function translateAll(
```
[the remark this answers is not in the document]
rix0rrr: It was originally, no longer needs to be. Good catch.

packages/jsii-rosetta/lib/languages/csharp.ts
rix0rrr: Maybe. Not sure!

packages/jsii-rosetta/lib/tablets/tablets.ts
rix0rrr: Maybe the names would be better to say `TranslatedSnippetOperations` and `TranslatedSnippet`. The class represents the operations on the pure-data object.

But I do need the ability to get the pure-data object out in a lightweight way.

PR #3127, opened 2021-11-05
feat(pacmak): fail on untranslated snippets

rix0rrr:
Because of backwards compatibility reasons, pacmak will try to live-translate
example code it encounters. However, this can add unpredictably to the compile
times, especially while we're making sure that rosetta and pacmak agree on API
locations for all snippets that are being translated (which are now part of the
snippet keys).

Add an option to pacmak to fail if any snippets are encountered that are not
pre-translated in any of the tablets. This most likely signifies an error.

This will be enabled in cdklabs/cdk-ops#1777.

---

### DIFF
--- a/packages/jsii-pacmak/lib/logging.ts
+++ b/packages/jsii-pacmak/lib/logging.ts
@@ -3,10 +3,12 @@ export enum Level {
   QUIET = 0,
   INFO = 1,
   VERBOSE = 2,
+  SILLY = 3,
 }
 
 export const LEVEL_INFO: number = Level.INFO;
 export const LEVEL_VERBOSE: number = Level.VERBOSE;
+export const LEVEL_SILLY: number = Level.SILLY;
 
 /** The minimal logging level for messages to be emitted. */
 export let level = Level.QUIET;
--- a/packages/jsii-pacmak/lib/packaging.ts
+++ b/packages/jsii-pacmak/lib/packaging.ts
@@ -54,7 +54,6 @@ export class JsiiModule {
    */
   public async npmPack() {
     this._tarball = await Scratch.make(async (tmpdir) => {
-      logging.debug(`Running "npm pack ${this.moduleDirectory}" in ${tmpdir}`);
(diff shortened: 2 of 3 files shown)


packages/jsii-rosetta/test/rosetta.test.ts
```diff
@@ -68,10 +75,23 @@ test('Rosetta object can do live translation', () => {
...
+    unknownSnippets: UnknownSnippetMode.FAIL,
+    targetLanguages: [TargetLanguage.PYTHON],
+  });
+
+  // WHEN
+  expect(() => {
+    rosetta.translateSnippet(SAMPLE_CODE, TargetLanguage.PYTHON);
+  }).toThrow(/snippet was not found/);
+});
+
```
kaizen3031593, line 90: Should we test `UnknownSnippetMode.VERBATIM` as well?

packages/jsii-rosetta/lib/rosetta.ts
```diff
@@ -147,19 +169,40 @@ export class Rosetta {
    *   will be based on the snippet key, which consists of a hash of the
    *   visible source and the API location.
    * - Otherwise, translate the snippet as-is (without fixture information).
+   *
+   * This will do and store a full conversion the given snippet, even if it only
```
kaizen3031593, line 173:
```suggestion
   * This will do and store a full conversion of the given snippet, even if it only
```

packages/jsii-rosetta/lib/rosetta.ts
```diff
@@ -147,19 +169,40 @@ export class Rosetta {
...
+   * returns one language. Subsequent retrievals for the same snippet but other
```
kaizen3031593, line 174:
```suggestion
   * returns one language. Subsequent retrievals for the same snippet in other
```

kaizen3031593 (approved):
Minor edits, and maybe an additional test necessary? There seems to be a failing test as well.

PR #3129, opened 2021-11-05
fix(rosetta): remember live-translated snippets without fixtures

rix0rrr:
We used to only remember snippets we live-translated using the rosetta
API if we were able to find a fixture for them.

There's no good reason for this. Also remember them if we live-translate
without a fixture.

---

### DIFF
--- a/packages/jsii-rosetta/test/rosetta.test.ts
+++ b/packages/jsii-rosetta/test/rosetta.test.ts
@@ -1,6 +1,13 @@
 import * as mockfs from 'mock-fs';
 
-import { Rosetta, LanguageTablet, TranslatedSnippet, TypeScriptSnippet, DEFAULT_TABLET_NAME } from '../lib';
+import {
+  Rosetta,
+  LanguageTablet,
+  TranslatedSnippet,
+  TypeScriptSnippet,
+  DEFAULT_TABLET_NAME,
+  Translation,
+} from '../lib';
 import { TargetLanguage } from '../lib/languages';
 import { fakeAssembly } from './jsii/fake-assembly';
 import { testSnippetLocation } from './testutil';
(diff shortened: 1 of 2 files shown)


packages/jsii-rosetta/test/rosetta.test.ts
```diff
@@ -10,20 +17,29 @@ const SAMPLE_CODE: TypeScriptSnippet = {
...
 
-test('Rosetta object can do live translation', () => {
-  // GIVEN
-  const rosetta = new Rosetta({
-    liveConversion: true,
-    targetLanguages: [TargetLanguage.PYTHON],
+describe('Rosetta object can do live translation', () => {
+  let rosetta: Rosetta;
+  let translated: Translation | undefined;
+  beforeEach(() => {
```
kaizen3031593, line 23: Wondering about using `beforeEach` vs `beforeAll`... It seems like `beforeAll` would work here since the tests are not modifying `rosetta` or `translated` (yay it saves 1 translation) however it might make more sense to default to `beforeEach` as it is safer.

My question is, did you mean `beforeAll`? Or do you like to default to `beforeEach` whenever possible?

kaizen3031593 (approved):
Approved, can merge whenever you want. Just want to give you a shot to change it to `beforeAll` if it was a typo.

packages/jsii-rosetta/test/rosetta.test.ts
rix0rrr: I did have `beforeAll` at first, but I felt `beforeEach` was safer.

I would say the rule would be:

* Default to using `forEach`, it is most likely going to save you trouble if it so happens that things are accidentally mutable, or become mutable after a while (even though you didn't intend them to be at the start).
* If the performance cost of `beforeEach` is horrible, fall back to `beforeAll` (and make *damn* sure your fixture is truly immutable)

PR #3131, opened 2021-11-05
feat(rosetta): find fixtures based on submodules

rix0rrr:
The `rosetta` fixtures directory is assembly-scoped. That means that in
the monocdk transform, all Rosetta fixtures from the individual
assemblies need to be tossed together into one `rosetta` directory for
all of them.

Inevitably this leads to conflicts, where the various individual
`default.ts-fixture` files compete with each other to be the one,
global `default.ts-fixture`. What ends up happening is they end up
overwriting each other and the last module to get copied in wins.
All other modules' examples fail to compile.

Ideally the `ubergen` tool that combines these fixtures would ensure
it doesn't clobber existing files, and renames the fixtures to unique
names (maybe `s3.default.ts-fixture`). But if it does that, *none* of them are
called `default.ts-fixture` and so all snippets require explicit
fixture references.

Unfortunately, we now need to also automatically add in those references
and it's nontrivial to do this automatically. `ubergen` would need to
parse and generate both Markdown and Typescript, to add these markers
in the right places.

Instead, this PR adds the concept of scoped fixtures. If a class
(or README) is defined in a submodule, say `aws_s3`, then we will
first try `rosetta/aws_s3/default.ts-fixture` before trying
`rosetta/default.ts-fixture`.

That way, `ubergen` and other tools like it will have an easier
time combining default fixtures.

Also renamed `AssemblyFixture` => `TestJsiiModule` and made a distinction
between the two directories exposed by the object for more clarity on
what it's doing.

---

### DIFF
--- a/packages/jsii-rosetta/lib/fixtures.ts
+++ b/packages/jsii-rosetta/lib/fixtures.ts
@@ -159,9 +159,6 @@ function submodules(location: ApiLocation): string[] {
   }
 
   function middle(xs: string[]) {
-    if (xs.length <= 2) {
-      return [];
-    }
     return xs.slice(1, xs.length - 1);
   }
 }
--- a/packages/jsii-rosetta/lib/fixtures.ts
+++ b/packages/jsii-rosetta/lib/fixtures.ts
@@ -2,7 +2,7 @@ import * as fs from 'fs-extra';
 import * as path from 'path';
 import { createSourceFile, ScriptKind, ScriptTarget, SyntaxKind } from 'typescript';
 
-import { TypeScriptSnippet, SnippetParameters } from './snippet';
+import { TypeScriptSnippet, SnippetParameters, ApiLocation } from './snippet';
 
 /**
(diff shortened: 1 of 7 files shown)


packages/jsii-rosetta/lib/fixtures.ts
```diff
@@ -99,6 +127,45 @@ function loadAndSubFixture(directory: string, fixtureName: string, source: strin
...
+    case 'member':
+    case 'type':
+    case 'parameter':
+      return middle(location.fqn.split('.'));
+    case 'moduleReadme':
+      return location.moduleFqn.split('.').slice(1);
+  }
+
+  function middle(xs: string[]) {
+    if (xs.length <= 2) {
```
otaviomacedo, line 162: [minor] Why is this a special case?

rix0rrr: You're right, it doesn't need to be.

otaviomacedo (approved):

PR #3133, opened 2021-11-05
fix(pacmak): Support more than 255 properties for interfaces in Java

### LINKED ISSUE #3132: Java: `too many parameters` error for interfaces with more than 255 properties (sorry!)
### Affected Languages

- [ ] `Python`
- [x] `Java`
- [ ] .NET (`C#`, `F#`, ...)
- [ ] `Go`

### General Information
* **JSII Version:** 1.42.0
* **Platform:** OS X

### What is the problem?

The CDK for Terraform generates interfaces for all attributes that can be set on resources of Terraform providers. Some of these (nested) types have more than 255 attributes. This already surfaced with the [AWS VPC Terraform module](https://github.com/hashicorp/terraform-cdk/issues/365) some time ago, but a workaround was possible for that specific case.

Recently a new version of the AWS Terraform provider was released, which added support for setting more AWS service endpoints which in turn caused the number of properties on the AwsProviderEndpoint class to rise above the number of 255. JSII then produces invalid Java code that fails at compile time. [Related issue in our repo](https://github.com/hashicorp/terraform-cdk/issues/1260)

For Java JSII creates such a Builder (examples from snapshot test)
https://github.com/aws/jsii/blob/0638345e7712638f01e809994f4b38b50523622e/packages/jsii-pacmak/test/generated-code/__snapshots__/examples.test.ts.snap#L733-L736

and a proxy object
https://github.com/aws/jsii/blob/0638345e7712638f01e809994f4b38b50523622e/packages/jsii-pacmak/test/generated-code/__snapshots__/examples.test.ts.snap#L783-L786

which has the following constructor
https://github.com/aws/jsii/blob/0638345e7712638f01e809994f4b38b50523622e/packages/jsii-pacmak/test/generated-code/__snapshots__/examples.test.ts.snap#L802

which in turn is used in `Builder.build()`
https://github.com/aws/jsii/blob/0638345e7712638f01e809994f4b38b50523622e/packages/jsii-pacmak/test/generated-code/__snapshots__/examples.test.ts.snap#L774-L776

In our case this constructor is generated with more than 255 attributes which is not allowed in Java and fails the compilation.

### Verbose Log
See https://github.com/hashicorp/terraform-cdk/issues/1260 for more info and for example [this Github Action Run](https://github.com/hashicorp/terraform-cdk/runs/4114745215?check_suite_focus=true).

ansgarm:
- fix(pacmak): support more than 255 properties in interfaces for Java
- chore(tests): update snapshots

Resolves #3132

This is a naive approach to fix the problem outlined in #3132 – it should not be user facing and should not break anything (famous last words before CI fails). Happy for any feedback!

---

### DIFF
--- a/packages/jsii-pacmak/lib/targets/java.ts
+++ b/packages/jsii-pacmak/lib/targets/java.ts
@@ -2009,7 +2009,7 @@ class JavaGenerator extends Generator {
     );
 
     props.forEach((prop) =>
-      this.code.line(`protected ${prop.fieldJavaType} ${prop.fieldName};`),
+      this.code.line(`${prop.fieldJavaType} ${prop.fieldName};`),
     );
     props.forEach((prop) =>
       this.emitBuilderSetter(prop, BUILDER_CLASS_NAME, classSpec),
--- a/packages/jsii-pacmak/lib/targets/java.ts
+++ b/packages/jsii-pacmak/lib/targets/java.ts
@@ -2009,7 +2009,7 @@ class JavaGenerator extends Generator {
     );
 
     props.forEach((prop) =>
-      this.code.line(`private ${prop.fieldJavaType} ${prop.fieldName};`),
+      this.code.line(`protected ${prop.fieldJavaType} ${prop.fieldName};`),
     );
     props.forEach((prop) =>
       this.emitBuilderSetter(prop, BUILDER_CLASS_NAME, classSpec),
@@ -2028,9 +2028,7 @@ class JavaGenerator extends Generator {
     this.code.line('@Override');
     this.code.openBlock(`public ${classSpec.name} build()`);
 
-    const propFields = props.map((prop) => prop.fieldName).join(', ');
-
-    this.code.line(`return new ${constructorName}(${propFields});`);
+    this.code.line(`return new ${constructorName}(this);`);
     this.code.closeBlock();
     // End build()
 
@@ -2104,7 +2102,7 @@ class JavaGenerator extends Generator {
     this.code.closeBlock();
     // End JSII reference constructor
 
-    // Start literal constructor
+    // Start builder constructor
     this.code.line();
     this.code.line('/**');
     this.code.line(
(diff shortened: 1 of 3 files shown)


packages/jsii-pacmak/lib/targets/java.ts
```diff
@@ -2114,11 +2112,8 @@ class JavaGenerator extends Generator {
     if (props.some((prop) => prop.fieldJavaType !== prop.paramJavaType)) {
       this.code.line('@SuppressWarnings("unchecked")');
     }
-    const constructorArgs = props
-      .map((prop) => `final ${prop.paramJavaType} ${prop.fieldName}`)
-      .join(', ');
     this.code.openBlock(
-      `protected ${INTERFACE_PROXY_CLASS_NAME}(${constructorArgs})`,
+      `protected ${INTERFACE_PROXY_CLASS_NAME}(final ${BUILDER_CLASS_NAME} builder)`,
```
jsteinich, line 2116: Certainly a bit non-traditional use of a builder, but the alternative would require the fields to be mutable. Does seem fine given that this is an internal class.

packages/jsii-pacmak/lib/targets/java.ts
```diff
@@ -2009,7 +2009,7 @@ class JavaGenerator extends Generator {
     );
 
     props.forEach((prop) =>
-      this.code.line(`private ${prop.fieldJavaType} ${prop.fieldName};`),
+      this.code.line(`protected ${prop.fieldJavaType} ${prop.fieldName};`),
```
jsteinich, line 2012: Package-private (no modifier) is probably a little better since there aren't subclasses involved.

jsteinich (changes requested):

packages/jsii-pacmak/lib/targets/java.ts
ansgarm: Sounds better – My Java is quite rusty 😁

RomainMuller (approved):
LGTM - Thank you!

RomainMuller:
@all-contributors add ansgarm  for code,bug

RomainMuller:
@all-contributors add @ansgarm for code and bug

PR #3134, opened 2021-11-06
feat(@jsii/spec): `replaceAssembly` and `_fingerprint` functions

kaizen3031593:
This PR moves two functions to `@jsii/spec` that are used in a few other locations in jsii.
`replaceAssembly()` updates an assembly file with a modified file (and updates the fingerprint).
It is utilized in `jsii-rosetta` and will be utilized in the `generate-examples` project.
`_fingerprint()` is also used in the `jsii` folder.
The motivation is to avoid copy/pasting similar functions around.  

---

### DIFF
--- a/packages/@jsii/spec/lib/replace-assembly.ts
+++ b/packages/@jsii/spec/lib/replace-assembly.ts
@@ -1,0 +1,48 @@
+import * as crypto from 'crypto';
+import * as fs from 'fs-extra';
+import * as path from 'path';
+
+import { Assembly } from './assembly';
+
+// eslint-disable-next-line @typescript-eslint/no-var-requires, @typescript-eslint/no-require-imports
+const sortJson = require('sort-json');
+
+/**
+ * Options for replacing the assembly
+ */
+interface ReplaceAssemblyOptions {
+  /**
+   * The assembly file extension that you want to overwrite.
+   */
+  readonly fileExtension: string;
+
+  /**
+   * The directory where the assembly is located.
+   */
+  readonly directory: string;
+}
+
+/**
+ * Replaces the file where the original assembly file *should* be found with a new assembly file.
+ * Recalculates the fingerprint of the assembly to avoid tampering detection.
+ */
+export async function replaceAssembly(
+  assembly: Assembly,
+  options: ReplaceAssemblyOptions,
+): Promise<void> {
+  const fileName = path.join(options.directory, options.fileExtension);
+  await fs.writeJson(fileName, _fingerprint(assembly), {
+    encoding: 'utf8',
+    spaces: 2,
+  });
+}
+
+function _fingerprint(assembly: Assembly): Assembly {
+  delete (assembly as any).fingerprint;
+  assembly = sortJson(assembly);
+  const fingerprint = crypto
+    .createHash('sha256')
+    .update(JSON.stringify(assembly))
+    .digest('base64');
+  return { ...assembly, fingerprint };
+}
--- a/packages/@jsii/spec/lib/replace-assembly.ts
+++ b/packages/@jsii/spec/lib/replace-assembly.ts
@@ -37,6 +37,9 @@ export async function replaceAssembly(
   });
 }
 
+/**
+ * Create a new fingerprint for the assembly.
+ */
 export function _fingerprint(assembly: Assembly): Assembly {
   delete (assembly as any).fingerprint;
   assembly = sortJson(assembly);
--- a/packages/@jsii/spec/lib/replace-assembly.ts
+++ b/packages/@jsii/spec/lib/replace-assembly.ts
@@ -37,7 +37,7 @@ export async function replaceAssembly(
   });
 }
 
-function _fingerprint(assembly: Assembly): Assembly {
+export function _fingerprint(assembly: Assembly): Assembly {
   delete (assembly as any).fingerprint;
   assembly = sortJson(assembly);
   const fingerprint = crypto
--- a/packages/@jsii/spec/package.json
+++ b/packages/@jsii/spec/package.json
@@ -31,11 +31,14 @@
     "package": "package-js"
   },
   "dependencies": {
-    "jsonschema": "^1.4.0"
+    "jsonschema": "^1.4.0",
+    "sort-json": "^2.0.0",
+    "fs-extra": "^9.1.0"
   },
   "devDependencies": {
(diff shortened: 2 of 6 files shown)


packages/@jsii/spec/test/testutil.ts
```diff
@@ -0,0 +1,51 @@
+import * as fs from 'fs-extra';
+import * as os from 'os';
+import * as path from 'path';
+
+import * as spec from '../lib/assembly';
+
+/**
+ * Construct a temp directory and write an assembly file inside it
+ */
+export class TestAssembly {
```
kaizen3031593, line 10: I initially wanted to add `TestJsiiModule` here but that introduces a circular dependency between `@jsii/spec` and `jsii`. Instead, I created something very similar but geared towards my testing needs, what do you think?

In particular, I didn't see the need to do the whole `tmpDir` and `modDir` thing as I don't really care where the assembly file is as long as I can access it.

packages/@jsii/spec/test/testutil.ts
```diff
@@ -0,0 +1,51 @@
...
+  public async syncAssembly() {
+    this._assembly = await fs.readJSON(path.join(this.directory, '.jsii'));
+  }
+
+  public async cleanup() {
+    await fs.remove(this.directory);
+  }
+}
+
+export function makeType(
```
kaizen3031593, line 44: I figure it is better to pull this function here rather than copy/paste it from `name-tree.test.ts`...

packages/@jsii/spec/lib/replace-assembly.ts
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Replaces the file where the original assembly file *should* be found with a new assembly file.
+ * Recalculates the fingerprint of the assembly to avoid tampering detection.
+ */
+export async function replaceAssembly(
+  assembly: Assembly,
+  options: ReplaceAssemblyOptions,
+): Promise<void> {
+  const fileName = path.join(options.directory, options.fileExtension);
```
kaizen3031593, line 33: This is an open question in my mind -- should we hardcode `.jsii` or just expose it as an option for those that use `replaceAssembly`? I figure that this is more versatile but maybe unnecessary.

kaizen3031593 (commented):
Leaving a couple comments on things I'm unsure of. In addition -

- Is it in scope of this PR to refactor `infuse.ts` to use this `replaceAssembly`?
- I haven't added `insertExample` into `spec`. The README for this module says that `@jsii/spec` contains jsii specifications, and `insertExample` doesn't seem to fall under that category. I'm not even sure `replaceAssembly` does...

packages/@jsii/spec/lib/replace-assembly.ts
```diff
@@ -0,0 +1,51 @@
...
+   * The directory where the assembly is located.
+   */
+  readonly directory: string;
+}
+
+/**
+ * Replaces the file where the original assembly file *should* be found with a new assembly file.
+ * Recalculates the fingerprint of the assembly to avoid tampering detection.
+ */
+export async function replaceAssembly(
```
rix0rrr, line 29: Feels like this ought to be called `saveAssembly`, wdyt?

packages/@jsii/spec/lib/replace-assembly.ts
rix0rrr: How about we just pass a `fileName` to `saveAssembly()`, that saves us having to decide anything, leaves the flexibility, and makes for a still pretty sane API from the PoV of the consumer?

packages/@jsii/spec/lib/replace-assembly.ts
```diff
@@ -0,0 +1,51 @@
...
+  await fs.writeJson(fileName, _fingerprint(assembly), {
+    encoding: 'utf8',
+    spaces: 2,
+  });
+}
+
+/**
+ * Create a new fingerprint for the assembly.
+ */
+export function _fingerprint(assembly: Assembly): Assembly {
```
rix0rrr, line 43: I don't think `_fingerprint` needs to be `export`ed. It's only used by `saveAssembly()`, right?

packages/@jsii/spec/package.json
```diff
@@ -31,11 +31,14 @@
     "package": "package-js"
   },
   "dependencies": {
-    "jsonschema": "^1.4.0"
+    "jsonschema": "^1.4.0",
+    "sort-json": "^2.0.0",
+    "fs-extra": "^9.1.0"
```
rix0rrr, line 36: Not a big fan of adding `fs-extra` as a dependency. I'd like to keep dependencies minimal if possible.

`fs-extra` doesn't really deliver anything that the regular `fs` module doesn't deliver (starting at Node 12 or so).

Instead, do the following:

```ts
import { promises as fs } from 'fs';

// ...

await fs.writeFile('filename', data, { encoding: 'utf-8' });
```

This uses the [fs.promises](https://nodejs.org/docs/latest-v12.x/api/fs.html#fs_fs_promises_api) API, which is also async (just like `fs-extra`).

packages/@jsii/spec/test/testutil.ts
rix0rrr: I don't think you need this helper at all.

The test could already be:

* You have an instance of a `spec.Assembly`. You can use `saveAssembly()` to save it to disk, and if you load it back `validateAssembly` doesn't complain.
* You have an instance of a `spec.Assembly`. After you save it, change it and save it again, the fingerprints are different between the two of them.

You could write all of that without this class, and I'm not sure the class makes it more readable.

rix0rrr (commented):
> Is it in scope of this PR to refactor infuse.ts to use this replaceAssembly?

Yes. And also the original use in `jsii`.

> I haven't added insertExample into spec. The README for this module says that @jsii/spec contains jsii specifications, and insertExample doesn't seem to fall under that category. 

Agreed.

> I'm not even sure replaceAssembly does...

That's fair. It does have "validateAssembly", so there is some operating on assemblies. Since `fingerprint` is defined here, it makes sense to me that the logic that determines the `fingerprint` would *also* live here. But it is sure debatable...

Alternatively, we just put `calculateFingerprint` here, but we never do `fs.writeFile()` if we feel that's a bridge too far.

wdyt?

kaizen3031593:
Closing this PR as this is something we decided not to do.

PR #3148, opened 2021-11-09
fix(pacmak): remove disclaimer on compiling examples

rix0rrr:
We're pretty confident in our rosetta translations by now. If we
successfully compiled the example, no need to prefix it with
a disclaimer anymore: the translation is most likely correct.

---


otaviomacedo (approved):
Do we have a way to measure this, by the way?

eladb (approved):
Thank you!

PR #3151, opened 2021-11-10
fix: dependency submodules may not be discovered

RomainMuller:
When a dependency exposes submodules, and the entry point of that
dependency is never imported by the program being compiled, symbols
from the dependency were not correctly mapped to their submodule because
the entry point was not part of the compiler input (and hence the
`ts.Program` did not hold a `ts.SourceFile` for it, and the
`ts.TypeChecker` did not have symbols for exports therein).

In order to address this problem, this always explicitly references
dependencies' entry point (the `.d.ts` file referenced by the `types`
key) to the compiler input path. This guarantees the compiler binds
symbols within those, and that the submodule mapper can appropriately
discover those.

------------------------------------------------------------------------

An alternative approach was considered (but rejected because it
introduced significant complexity): we could have created
`ts.SourceFile` objects using `ts.createSourceFile`, which allows us to
obtain a set of `ts.Statement` nodes, however those are not bound to
symbols and we then have to process the AST without the assistance of
the type checker (and would end up having to re-implement a sub-set of
the binder/type-checker's features).

---

### DIFF
--- a/packages/jsii/lib/compiler.ts
+++ b/packages/jsii/lib/compiler.ts
@@ -504,7 +504,7 @@ export class Compiler implements Emitter {
 
     if (files.length > 0) {
       for (const file of files) {
-        ret.add(path.resolve(this.options.projectInfo.projectRoot, file));
+        ret.add(file);
       }
     } else {
       const parseConfigHost = parseConfigHostFromCompilerHost(
--- a/packages/jsii/lib/compiler.ts
+++ b/packages/jsii/lib/compiler.ts
@@ -504,7 +504,7 @@ export class Compiler implements Emitter {
 
     if (files.length > 0) {
       for (const file of files) {
-        ret.add(file);
+        ret.add(path.resolve(this.options.projectInfo.projectRoot, file));
       }
     } else {
       const parseConfigHost = parseConfigHostFromCompilerHost(
--- a/packages/jsii/lib/compiler.ts
+++ b/packages/jsii/lib/compiler.ts
@@ -138,7 +138,10 @@ export class Compiler implements Emitter {
         ...pi.tsc,
         ...BASE_COMPILER_OPTIONS,
         noEmitOnError: false,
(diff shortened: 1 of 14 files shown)


packages/jsii/lib/compiler.ts
```diff
@@ -495,23 +497,43 @@ export class Compiler implements Emitter {
...
+      const { resolvedModule } = ts.resolveModuleName(
+        assm.name,
+        path.join(this.options.projectInfo.projectRoot, this.options.projectInfo.types),
+        this.typescriptConfig?.compilerOptions ?? {},
+        ts.sys,
+      );
+      if (!resolvedModule) {
+        continue;
+      }
+      ret.add(resolvedModule.resolvedFileName);
```
rix0rrr, line 533: Are the filenames from the config also resolved? If not, the deduplication probably won't work properly.

packages/jsii/lib/compiler.ts
```diff
@@ -495,23 +497,43 @@ export class Compiler implements Emitter {
...
+      for (const file of fileNames) {
+        ret.add(file);
+      }
     }
 
-    return ret;
+    // Bonus: ensure all dependencies' entry points are included in the compiler
+    // input path. This guarantees we have symbols for all types, from the
+    // module root, which is necessary in order to properly detect submodules.
+    for (const assm of this.options.projectInfo.dependencyClosure) {
```
rix0rrr, line 523: Do we really need the closure? Would the direct dependencies not be good enough?

rix0rrr (approved):
Provisional approval if you don't think the previous 2 comments are a concern.

packages/jsii/lib/compiler.ts
RomainMuller: Great question :) I had to check this, too... They are indeed resolved.

packages/jsii/lib/compiler.ts
RomainMuller: Yes it is. We need to do the submodule traversal for _all_ transitive dependencies, as types thereof might be referred to implicitly but the user code (i.e: via a forwarded return value), but explicitly in the type model.

The problem we have where the submodule traders could not work if the dependency's entry point is not in the compiler's input set also applies to transitive dependencies.

PR #3161, opened 2021-11-11
feat(rosetta): expose the 'extract' API

rix0rrr:
Expose parts of the `extract` API (cache reading and mass translating)
to consumers, so that downstream tools can build example manipulators
as well.

* The original exposed `Rosetta` class is renamed to `RosettaTabletReader`, and is intended to be used by `pacmak` (or tools that primarily read tablets).
* A new class is introduced, `RosettaTranslator`, which is intended to be used by tools that need to do mass translation. This new API exposes the worker-based caching and `translateAll` functionality.
* The `extract` command is rewritten to use `RosettaTranslator`
* The `translateAll` features are moved from `extract.ts` and `extract_worker.ts` to `translate_all.ts` and `translate_all_worker.ts`.
* Add a function to build a snippet from source with `/// !show|!hide` directives, make it clear when each is supposed to be used.
* Fix the `console.error()` suppression from the TypeScript compiler that was using an incorrect substring.

---


packages/jsii-rosetta/lib/commands/extract.ts
```diff
@@ -50,50 +40,44 @@ export async function extractSnippets(
...
 
-  const tablet = new LanguageTablet();
+  const translator = (options.translatorFactory ?? mk(RosettaTranslator))({
+    assemblies: assemblies.map((a) => a.assembly),
+  });
 
   if (options.cacheTabletFile) {
-    await reuseTranslationsFromCache(snippets, tablet, options.cacheTabletFile, fingerprinter);
+    await translator.loadCache(options.cacheTabletFile);
+    const reused = translator.readFromCache(snippets);
```
njlynch, line 59: Lets put a comment here noting that this removes/manipulates `snippets`. It's probably documented on the method itself, but reading this flow as-is it's certainly not clear.

packages/jsii-rosetta/lib/rosetta-translator.ts
```diff
@@ -0,0 +1,155 @@
...
+    }
+  }
+
+  /**
+   * For all the given snippets, try to read translations from the cache
+   *
+   * Will remove the cached snippets from the input array.
+   */
+  public readFromCache(snippets: TypeScriptSnippet[], addToTablet = true): TranslatedSnippet[] {
+    if (!this.cache) {
```
njlynch, line 69: this.cache is always instantiated, ya? Should this be something like `this.cache.isEmpty` or `!this.cache.loaded` instead, or just be removed?

packages/jsii-rosetta/lib/rosetta-translator.ts
```diff
@@ -0,0 +1,155 @@
...
+    return {
+      translatedSnippets: fingerprinted,
+      diagnostics: result.diagnostics,
+    };
+  }
+
+  /**
+   * Insert an example into the docs of a type, and insert it back into the tablet under a new key
+   */
+  public insertExample(example: TranslatedSnippet, type: spec.Type): void {
```
njlynch, line 112: This looks unused (part of the next draft PR)?

njlynch (changes requested):

njlynch (approved):

PR #3176, opened 2021-11-16
fix(rosetta): Rosetta is not submodule-aware

rix0rrr:
Rosetta was not properly aware of submodules and namespaces, at all. This manifested in two different ways:

- Imports of (Python) modules from submodules were not translated properly
- Synthetically generated struct constructors were not namespaced with a proper import.

The translation of a call involving a struct:

```ts
import { aws_ec2 as ec2 } from 'aws-cdk';

const construct = new ec2.MyConstruct('some', 'values', {
  arg: { ...  }
});
```

Used to be translated into this:

```py
from aws_cdk import aws_ec2 as ec2

construct = ec2.MyConstruct('some', 'values',
    arg=DeeperStruct(...)
    )
```

Notice that `DeeperStruct` is missing the module prefix that would
allow this code to work.

We trace what jsii FQNs every imported symbol refers to, and see
if we can reuse an import that already exists to retrieve our struct.
If not, we'll generate a new import at the top of the example.

To properly track what is being imported in an example (if it's a class or
a module, or potentially a submodule) we needed to be able to identify
submodules in TypeScript source, so we've also attached a `symbolId`
to `submodules` in the jsii assembly.

---


rix0rrr:
NOTE: not complete yet for submodules. Requires #3174 to finish.

packages/jsii-rosetta/lib/jsii/assemblies.ts
```diff
@@ -257,3 +265,19 @@ function findPackageJsonLocation(currentPath: string): string | undefined {
     currentPath = parentPath;
   }
 }
+
+/**
+ * Find the jsii [sub]module that contains the given FQN
+ *
+ * Return `undefined` if the type is a member of the assembly root.
```
RomainMuller, line 272:
```suggestion
 * @returns `undefined` if the type is a member of the assembly root.
```

packages/jsii-rosetta/lib/jsii/assemblies.ts
```diff
@@ -257,3 +265,19 @@ function findPackageJsonLocation(currentPath: string): string | undefined {
...
+ */
+export function findContainingSubmodule(assembly: spec.Assembly, fqn: string): string | undefined {
+  const submoduleNames = Object.keys(assembly.submodules ?? {});
+  sortBy(submoduleNames, (s) => [-s.length]); // Longest first
+  for (const s of submoduleNames) {
+    if (fqn.startsWith(`${s}.`)) {
+      return s;
+    }
+  }
+  return undefined;
```
RomainMuller, line 282: Very much optional, but this can be done in a purely functional style (I'm not sure either form is clearly more readable than the other).

```suggestion
  Object.keys(assembly.submodules ?? {})
    // Keep only those that are prefixes of our fqn
    .filter((s) => fqn.startsWith(`${s}.`))
    // Sort by length (longest to shortest)
    .sort((l, r) => r.length - l.length)
    // Longest prefix is the one we're looking for
    [0]
```

packages/jsii-rosetta/lib/util.ts
```diff
@@ -102,6 +102,7 @@ export function mkDict<A extends string, B>(xs: Array<readonly [A, B]>): Record<
 
 export function fmap<A, B>(value: NonNullable<A>, fn: (x: A) => B): B;
 export function fmap<A, B>(value: undefined, fn: (x: A) => B): undefined;
+export function fmap<A, B>(value: A | undefined, fn: (x: A) => B): B | undefined;
 export function fmap<A, B>(value: A, fn: (x: A) => B): B | undefined {
```
RomainMuller, line 106: All of these direly lack doc-strings.

I have no idea what `fmap` means.
The body looks like this is a pretty trivial wrapper for `value != null ? fn(value) : undefined`... And having realized this I still don't see what `fmap` stands for?

packages/jsii-rosetta/lib/util.ts
```diff
@@ -116,3 +117,47 @@ export function mapValues<A, B>(xs: Record<string, A>, fn: (x: A) => B): Record<
   }
   return ret;
 }
+
+export function sortBy<A>(xs: A[], keyFn: (x: A) => Array<string | number>) {
```
RomainMuller, line 121: This could use some doc-strings.

packages/jsii/lib/symbol-id.ts
```diff
@@ -6,6 +6,11 @@ export function symbolIdentifier(
   typeChecker: ts.TypeChecker,
   sym: ts.Symbol,
 ): string | undefined {
+  // If this symbol happens to be an alias, resolve it first
+  if ((sym.flags & ts.SymbolFlags.Alias) !== 0) {
+    sym = typeChecker.getAliasedSymbol(sym);
+  }
```
RomainMuller, line 12: Is it maybe possible that an aliased symbol refers to another symbol that is itself an alias? Or does the `typeChecker.getAliasedSymbol` resolve all the way out?

packages/@jsii/spec/lib/assembly.ts
```diff
@@ -192,7 +192,14 @@ export interface ReadMe {
...
  * that the submodule is annotated with its location in the repository.
  */
-export type Submodule = SourceLocatable & Targetable;
+export interface Submodule extends SourceLocatable, Targetable {
+  /**
+   * Identifier of the symbol that represents the root of this submodule
+   *
+   * Used to map from TypeScript code back into the assembly.
+   */
+  symbolId?: string;
```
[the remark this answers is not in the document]
rix0rrr: I considered that, but too many things have `SourceLocatable` though, also things that don't have a symbol id.

I'll make a new interface.

packages/jsii-rosetta/lib/util.ts
rix0rrr: It's very much a wrapper for that. Just leads to more readable code than the tangle of `: undefined` you'd otherwise get into. 

I'll add docs.

packages/jsii-rosetta/lib/jsii/assemblies.ts
rix0rrr: Nice.

packages/jsii/lib/symbol-id.ts
rix0rrr: Good question. I'll turn it into a `while` loop just to be safe.

RomainMuller (approved):

Issue #3078, opened 2021-10-19
Python submodule import not found


jsteinich:
### Affected Languages

- [ ] `TypeScript` or `Javascript`
- [X] `Python`
- [ ] `Java`
- [ ] .NET (`C#`, `F#`, ...)
- [ ] `Go`

### General Information
* **JSII Version:** 1.39
* **Platform:**     Windows 10. Python 3.8.5

### What is the problem?
I'm doing some testing with cdktf version 0.6.4-pre.375. This includes a recent change to generate submodules for the aws provider. This works fine until runtime when the following error is produced:
```
Traceback (most recent call last):
      File "main.py", line 4, in <module>
        from imports.aws import AwsProvider
      File "C:\project\imports\aws\__init__.py", line 3869, in <module>
        import aws.access_analyzer
    ModuleNotFoundError: No module named 'aws'
```

The `__init__.py` file contains at the end a list of imports for all of the submodules. i.e. `import aws.access_analyzer`. This matches the behavior introduced with #3049. However, the `aws` module is being defined within that same file and doesn't seem to be referenceable at that point.

I manually updated the file with the imports replaced with a slightly different form: `from ..aws import access_analyzer` and everything seems to be running just fine.

### Repro steps

1. Install latest cdktf (npm install -g cdktf-cli@next)
2. Clone https://github.com/hashicorp/terraform-cdk/tree/main/examples/python/aws
3. Populate `cdktf.json` `terraformProviders` with `"aws@~> 3"` (May also happen in v2)
4. Run `cdktf get`
5. Run `cdktf synth` and observe error

### Other details

We've been having some issues reproducing this and similar issues in different environments. I haven't tracked down the source of difference yet, but I'm wondering if various python / virtualenv differences could be at play here.

skorfmann:
I'll create a test repo to reproduce this on a small case. In a nutshell, this is causing issues for cdktf in scenarios where users are generating their own code bindings for a Terraform provider or module. The generated code ends up in user configured output directories, by default `imports/aws/...` for the AWS provider.

This is essentially a breaking change for existing Python users of the AWS provider (the only one with namespaces so far) in cdktf which are generating the code bindings themselves. For this reason, we locked jsii to `1.37.0` in our last release and can't upgrade to newer versions until we found a solution or at least a workaround. 

[...]

PR #3181, opened 2021-11-17
fix(pacmak): Generate Relative Module Imports in Python

skorfmann:
The goal of this PR is, to find a way for generated Python code to work with namespaced modules regardless of the way it is consumed. While the current implementation of submodule imports seem to work fine when distributed as packages, it's breaking apart when the generated code is imported relatively as part of a single code base. That's what we do in cdktf and the reason why we have locked jsii to `1.37.0` at the moment.

The general idea is, to generate relative imports rather than fqn imports in Python. This could be a potential way to address https://github.com/aws/jsii/issues/3078

```python
# Before    
import jsii_calc.submodule
import jsii_calc.submodule.nested_submodule
import jsii_calc.submodule.nested_submodule.deeply_nested

# After
from . import submodule
from .submodule import nested_submodule
from .submodule.nested_submodule import deeply_nested
```

Besides from a few snapshots being changed, it seems like the tests are still working with this change. However, I'm not sure  about the general implications. So, this entire PR is really meant to be a conversation starter. 

I have confirmed that an import like `import . from some_aws_namespace` would work for us in cdktf.

## References

- https://github.com/aws/jsii/issues/3078
- https://github.com/aws/jsii/pull/3049
- https://github.com/hashicorp/terraform-cdk/issues/1152
- https://github.com/hashicorp/terraform-cdk/pull/1288

---

### DIFF
--- a/packages/jsii-pacmak/lib/targets/python.ts
+++ b/packages/jsii-pacmak/lib/targets/python.ts
@@ -25,6 +25,7 @@ import {
   PythonImports,
   mergePythonImports,
   toPackageName,
+  toPythonFqn,
 } from './python/type-name';
 import { die, toPythonIdentifier } from './python/util';
 import { toPythonVersionRange, toReleaseVersion } from './version-utils';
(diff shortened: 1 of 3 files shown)


RomainMuller (approved):
LGTM - Thanks for fixing that!

PR #3183, opened 2021-11-18
fix: C# NamespaceDoc emitted to wrong location

RomainMuller:
The namespace determination for the `NamespaceDoc` class generated for
submodules was incorrect, as the `jsiiNs` passed to the type resovler
was absolute, instead of being the required assembly-relative form.

Additionally, removed the `readme` and `locationInSource` property from
submodule descriptors included in the jsii assembly under
`dependencyClosure`, as those are not necessary and needlessly bloat the
assembly document.

---

### DIFF
--- a/packages/jsii/lib/assembler.ts
+++ b/packages/jsii/lib/assembler.ts
@@ -3100,9 +3100,9 @@ function noEmptyDict<T>(
 }
 
 function toDependencyClosure(assemblies: readonly spec.Assembly[]): {
-  [name: string]: spec.AssemblyConfiguration;
+  [name: string]: spec.DependencyConfiguration;
 } {
-  const result: { [name: string]: spec.AssemblyConfiguration } = {};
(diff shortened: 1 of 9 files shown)


packages/jsii/lib/assembler.ts
```diff
@@ -3102,17 +3102,42 @@ function noEmptyDict<T>(
...
+   * assembly object.
+   *
+   * This removes the `readme` and `locationInModule` fields from the submodule
+   * descriptios if present.
+   *
+   * @param submodules the submodules list to clean up.
+   *
+   * @returns the cleaned up submodules list.
+   */
+  function cleanUp(
```
rix0rrr, line 3129: This should be accompanied by a different type signature in the spec, no? 

Those fields should not even be present in a dependency's submodules field.

rix0rrr (changes requested):

rix0rrr (approved):

PR #3184, opened 2021-11-18
fix(jsii): require statement for the warning file is generated when it's not used

otaviomacedo:
When `add-deprecation-warnings` is `true` but no calls were injected in a given file, the `require` statement for `.warnings.jsii.js` should not be generated.

---


nija-at (commented):
What would happen if a `@deprecated` tsdoc is added to a function in a non-jsii typescript file?

nija-at:
Isn't the correct fix here to not run this transformer on non-jsii files? 
Is it not possible to check this?

PR #3185, opened 2021-11-18
fix(jsii): require statement for the warning file is generated for non-exported files

otaviomacedo:
This change makes the transformer only run on module exports, which are the ones actually processed by the assembler.

---


nija-at (approved):
Adding 'do-not-merge' in case this needs more testing with real CDK modules.

PR #3200, opened 2021-11-23
feat(rosetta):  metadata tag for fixtures in docs

kaizen3031593:
Currently the best thing to do to specify metadata in `@example` docs is to supply it in the example like so:

```
/**
* @example
* /// metadata
* my example
*/
```

This works, but is cumbersome because the `/// metadata` line gets picked up in typescript doc gen and gets displayed as part of the example in docs. Because of backwards compatibility, this functionality stays, but this PR introduces a better way to supply example metadata via the `@exampleMetadata` tag. For example:

```
/**
 * @exampleMetadata nofixture
 * @example 
 * my example
 */
```

`@exampleMetadata` will be read by rosetta and the necessary fixtures supplied. While this PR focuses on adding functionality for fixture metadata, it theoretically should work on any kind of metadata we may want to add in the future.

---


packages/jsii-rosetta/lib/rosetta-reader.ts
```diff
@@ -291,7 +290,7 @@ export class RosettaTabletReader {
   }
 
   public get hasErrors() {
-    return this.diagnostics.some(isError);
+    return this.diagnostics.some((d) => d.isError);
```
kaizen3031593, line 293: This is a pesky little bug independent of this PR that came about when the type of `this.diagnostics` was changed. It means that assemblies in `strict` mode would never actually bail (in my case, tests in `strict` mode would succeed) since `rosetta.hasErrors()` would always return `false`.

packages/jsii-rosetta/lib/commands/transliterate.ts
```diff
@@ -228,3 +230,11 @@ function transliterateType(
     }
   }
 }
+
+export function parseMetadata(metadata: string) {
```
rix0rrr, line 234: Move this?

packages/jsii-rosetta/lib/jsii/assemblies.ts
```diff
@@ -74,7 +75,7 @@ async function loadAssemblyFromFile(filename: string, validate: boolean): Promis
 
 export type AssemblySnippetSource =
   | { type: 'markdown'; markdown: string; location: ApiLocation }
-  | { type: 'example'; source: string; location: ApiLocation };
+  | { type: 'example'; source: string; metadata: { [key: string]: any } | undefined; location: ApiLocation };
```
rix0rrr, line 78: Can be typed as `string -> string`.

packages/jsii-rosetta/lib/jsii/assemblies.ts
```diff
@@ -132,6 +133,7 @@ export function allSnippetSources(assembly: spec.Assembly): AssemblySnippetSourc
       ret.push({
         type: 'example',
         source: docs.example,
+        metadata: docs.custom?.exampleMetadata ? parseMetadataLine(docs.custom.exampleMetadata) : undefined,
```
rix0rrr, line 136:
```suggestion
        metadata: fmap(docs.custom?.exampleMetadata, parseMetadataLine),
```

rix0rrr (approved):

PR #3205, opened 2021-11-25
fix: improve compatibility with restricted-export modules

rix0rrr:
Modules can decide they don't export `package.json`, and our methods
of using `require.resolve('the-package/package.json')` to find the
package will fail.

Instead, resolve the main entry point and search upward for
`package.json`, bypassing `require()`.

We have been doing all of:

- `require('the-package/package.json')`
- `require.resolve('the-package/.jsii')`
- `require.resolve('the-package/tsconfig.json')`

Replace all these with a `require.resolve('the-package')` then `findUp`

Relates to https://github.com/aws/aws-cdk/pull/17707

---


packages/jsii-pacmak/lib/util.ts
```diff
@@ -6,19 +6,72 @@ import * as path from 'path';
...
-  const lookupPaths = [path.join(packageDir, 'node_modules')];
-  return path.dirname(
-    require.resolve(`${dependencyName}/package.json`, { paths: lookupPaths }),
+  searchStart: string,
+) {
+  // Explicitly do not use 'require("dep/package.json")' because that will fail if the
+  // package does not export that particular file.
+  const entryPoint = require.resolve(dependencyName, {
+    paths: [searchStart],
+  });
```
RomainMuller, line 19: We *could* try `require("dep/package.json")` in first intention (although this could return a different file than what you expect due to import/export maps.

packages/jsii-pacmak/lib/util.ts
```diff
@@ -6,19 +6,72 @@ import * as path from 'path';
...
+      (await fs.pathExists(pjFile)) &&
+      (await fs.readJson(pjFile)).name === packageName
+    );
+  });
+}
+
+/**
+ * Find a directory up the tree from a starting directory matching a condition
+ *
+ * Will return `undefined` if a no directory matches
```
RomainMuller, line 56:
```suggestion
 * Will return `undefined` if no directory matches
```

packages/jsii-reflect/lib/type-system.ts
```diff
@@ -300,7 +305,7 @@ export class TypeSystem {
    * @param validate Whether to validate the assembly or just assume it matches the schema
    */
   private async loadAssembly(file: string, validate = true) {
-    const spec = JSON.parse((await readFile(file)).toString());
+    const spec = JSON.parse(await fs.readFile(file, { encoding: 'utf-8' }));
```
RomainMuller, line 308: `fs-extra` has a `readJson` function, too ;)

packages/jsii-reflect/lib/util.ts
```diff
@@ -1,7 +1,79 @@
...
   for (const x of xs) {
     ret[f(x)] = x;
   }
   return ret;
 }
+
+/**
+ * Find the directory that contains a given dependency, identified by its 'package.json', from a starting search directory
+ */
+export async function findDependencyDirectory(
```
RomainMuller, line 15: Is this implementation a copy of the one in `jsii-pacmak`? I wouldn't mind a `@jsii/utils` package...

packages/jsii/lib/project-info.ts
```diff
@@ -67,7 +70,7 @@ export async function loadProjectInfo(
 ): Promise<ProjectInfoResult> {
   const packageJsonPath = path.join(projectRoot, 'package.json');
   // eslint-disable-next-line @typescript-eslint/no-var-requires,@typescript-eslint/no-require-imports
-  const pkg = require(packageJsonPath);
+  const pkg = await fs.readJson(packageJsonPath);
```
RomainMuller, line 73: Nice!

packages/jsii-reflect/lib/util.ts
rix0rrr: Yeeeeaaa -- hhh... I don't disagree, but I didn't want to make too many changes at once

packages/jsii-pacmak/lib/util.ts
rix0rrr: Yeah, doesn't seem to be much of a point.

RomainMuller (approved):
Approved granted you add notes in the duplicated code that it is... duplicated, and create an issue to refactor in a follow-up PR (I don't expect you'll implement that PR right away, though).

rix0rrr:
https://github.com/aws/jsii/issues/3236

PR #3206, opened 2021-11-25
fix(jsii): deprecation message is not displayed for deprecated classes

otaviomacedo:
---


rix0rrr:
@nija-at is this going to fuck inherited deprecated classes? Is that still something we need to think about?

nija-at:
Not sure. We only twiddled around with interface inheritance.
We should write a test for it.

rix0rrr (approved):

PR #3210, opened 2021-11-26
feat(rosetta): reuse output file as additional cache and introduce `--infuse` option for `extract`

kaizen3031593:
Introduces two new features for `rosetta:extract`:

- `rosetta:extract --infuse`: combines a common workflow of calling `infuse` right after `extract`.
- extract now uses the output file as an additional cache. This allows us to reuse existing snippets that have already been translated.

---


packages/jsii-rosetta/lib/commands/extract.ts
```diff
@@ -49,16 +63,19 @@ export async function extractSnippets(
...
   };
 
   const translator = options.translatorFactory
     ? options.translatorFactory(translatorOptions)
     : new RosettaTranslator(translatorOptions);
 
-  if (options.cacheTabletFile) {
-    await translator.loadCache(options.cacheTabletFile);
+  // Choose the cache if provided, or the output file if we plan to append to it.
+  const cacheTabletFile = options.cacheTabletFile ?? (options.append ? options.outputFile : undefined);
```
rix0rrr, line 74: We need to support BOTH the append-file and the cache file as sources of existing snippets.

rix0rrr (changes requested):

packages/jsii-rosetta/lib/rosetta-translator.ts
```diff
@@ -60,6 +69,15 @@ export class RosettaTranslator {
     }
   }
 
+  public async addToCache(filename: string) {
```
kaizen3031593, line 72: It almost feels like `addToCache` should be the way we load caches rather than `loadCache`, since it does the same thing but allows for us to add multiple tablets. Any chance you'd let me deprecate `loadCache`?

kaizen3031593 (commented):
Some comments on the new lines i added

packages/jsii-rosetta/lib/rosetta-translator.ts
rix0rrr: Yep, that's fine.

packages/jsii-rosetta/lib/commands/extract.ts
```diff
@@ -49,16 +62,20 @@ export async function extractSnippets(
...
   const translator = options.translatorFactory
     ? options.translatorFactory(translatorOptions)
     : new RosettaTranslator(translatorOptions);
 
   if (options.cacheTabletFile) {
-    await translator.loadCache(options.cacheTabletFile);
+    await translator.addToCache(options.cacheTabletFile);
+  }
+  await translator.addToCache(options.outputFile);
+  if (translator.hasCache()) {
```
rix0rrr, line 76: Instead of doing this we can also always try it, and only print if the number of successful reads is > 0.

packages/jsii-rosetta/test/commands/extract.test.ts
```diff
@@ -57,25 +62,39 @@ test('extract samples from test assembly', async () => {
...
+        cacheTabletFile,
+        translatorFactory: (o) => new MockTranslator(o, translationFunction),
+        ...defaultExtractOptions,
+      });
+
+      expect(translationFunction).not.toHaveBeenCalled();
     });
 
-    expect(translationFunction).not.toHaveBeenCalled();
+    test('because append to file acts as cache', async () => {
```
rix0rrr, line 87: This test description isn't really accurate anymore

rix0rrr (approved):
Conditional approve, some small things around describing the changes that could be improved.

As does the title: are we still appending? Or do we just overwrite the output tablet with everything from the assembly, but do we reuse existing snippets from the output assembly? That would be a better PR description.

packages/jsii-rosetta/lib/commands/extract.ts
kaizen3031593: `readFromCache` will loop through all the snippets and check to see if it exists in the cache. This seems like redundant work if there are no elements in the cache. That is why I added `hasCache`, not because of the printing.

PR #3223, opened 2021-12-02
feat(rosetta): generate rosetta tablets next to each assembly

rix0rrr:
Rather than write our Rosetta translations to a single tablet file which
we throw away at the end of the build, `rosetta extract` will now write
translations to individual `.jsii.tabl.json` files, located next to
the assemblies themselves.

Construct library authors can publish these to NPM, so that downstream
tools that process JSII modules for documentation have access to the
translations and don't need to redo the work (especially relevant
for large libraries with lots of examples, where otherwise a lot
of CPU time would be wasted).

The "single output tablet" can still be used, but is now intended to be
used as a cache to speed up repeated runs of `rosetta extract` (to skip
translating unchanged snippets). Add options to trim the cache files
down so they won't grow without bounds.

This PR also contains a bugfix: `infuse` did not re-inject a copied
snippet's metadata into the assembly's `exampleMetadata` field. This
is strictly speaking not a problem as long as the tablet always stays
with the assembly, but better to fix anyway.

---

### DIFF
--- a/packages/jsii-rosetta/bin/jsii-rosetta.ts
+++ b/packages/jsii-rosetta/bin/jsii-rosetta.ts
@@ -4,12 +4,13 @@ import * as fs from 'fs-extra';
 import * as path from 'path';
 import * as yargs from 'yargs';
 
-import { TranslateResult, DEFAULT_TABLET_NAME, translateTypeScript, RosettaDiagnostic } from '../lib';
+import { TranslateResult, translateTypeScript, RosettaDiagnostic } from '../lib';
 import { translateMarkdown } from '../lib/commands/convert';
-import { extractAndInfuse, ExtractResult, extractSnippets } from '../lib/commands/extract';
+import { extractAndInfuse, extractSnippets, ExtractOptions } from '../lib/commands/extract';
 import { infuse, DEFAULT_INFUSION_RESULTS_NAME } from '../lib/commands/infuse';
 import { readTablet } from '../lib/commands/read';
 import { transliterateAssembly } from '../lib/commands/transliterate';
+import { trimCache } from '../lib/commands/trim-cache';
 import { TargetLanguage } from '../lib/languages';
 import { PythonVisitor } from '../lib/languages/python';
 import { VisualizeAstVisitor } from '../lib/languages/visualize';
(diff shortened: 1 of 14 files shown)


packages/jsii-rosetta/bin/jsii-rosetta.ts
```diff
@@ -197,30 +215,26 @@ function main() {
         // compilerhost. Have to make all file references absolute before we chdir
         // though.
         const absAssemblies = (args.ASSEMBLY.length > 0 ? args.ASSEMBLY : ['.']).map((x) => path.resolve(x));
-        const absOutput = path.resolve(args.output);
-        const absCache = fmap(args['cache-from'], path.resolve);
+
+        const absCacheFrom = fmap(args.cache ?? args['cache-from'], path.resolve);
+        const absCacheTo = fmap(args.cache ?? args['cache-to'] ?? args.output, path.resolve);
```
kaizen3031593, line 220: Did you consider adding `.conflicts('output', ['cache-to', 'cache'])` instead of ignoring `output` if it is specified along with the cache arguments?

packages/jsii-rosetta/lib/tablets/tablets.ts
```diff
@@ -35,16 +36,22 @@ export class LanguageTablet {
...
     }
     return ret;
   }
 
   private readonly snippets: Record<string, TranslatedSnippet> = {};
 
-  public addSnippet(snippet: TranslatedSnippet) {
-    const existingSnippet = this.snippets[snippet.key];
-    this.snippets[snippet.key] = existingSnippet ? existingSnippet.mergeTranslations(snippet) : snippet;
+  public addSnippet(...snippets: TranslatedSnippet[]) {
```
kaizen3031593, line 50: `addSnippets`?

packages/jsii-rosetta/lib/tablets/tablets.ts
```diff
@@ -54,9 +61,11 @@ export class LanguageTablet {
   /**
    * Add all snippets from the given tablet into this one
    */
-  public addTablet(tablet: LanguageTablet) {
-    for (const snippet of Object.values(tablet.snippets)) {
-      this.addSnippet(snippet);
+  public addTablet(...tablets: LanguageTablet[]) {
```
kaizen3031593, line 64: `addTablets`? And maybe the documentation should be pluralized too.

kaizen3031593 (approved):
Super minor comments that don't have bearing on the functionality.

packages/jsii-rosetta/bin/jsii-rosetta.ts
```diff
@@ -69,37 +70,31 @@ function main() {
...
-          })
-          .option('output', {
-            alias: 'o',
             type: 'string',
             describe: 'Output file to store logging results. Ignored if -log is not true',
             default: DEFAULT_INFUSION_RESULTS_NAME,
           })
-          .demandOption('TABLET'),
+          .option('cache-to', {
+            alias: 'o',
```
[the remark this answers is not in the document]
rix0rrr: I don't think we can have two-letter aliases. Doesn't fit GNU commandline flag style.

Since this is a replacement for the old `--output` flag, `-o` seemed okay.

packages/jsii-rosetta/lib/tablets/tablets.ts
rix0rrr: Yeh, but that breaks API. This seemed simpler and just as effective.

packages/jsii-rosetta/bin/jsii-rosetta.ts
rix0rrr: We have `.conflicts('cache-to', 'output')`. Is that not good enough?

PR #3225, opened 2021-12-03
fix(rosetta): classes are not correctly identified if package uses an outDir

kaizen3031593:
`rosetta extract` breaks down when a project structure includes `outDir` or `rootDir` in the `tsconfig` file. Given such a project structure, `extract` may find symbolIds that look like `outDir/index:ClassA` or `rootDir/index:ClassA`. Since the jsii assembly is created from the source files in the `rootDir`, `rootDir/index:ClassA` is the symbolId we want. So if the `outDir` is specified, we swap it with the specified `rootDir` (and if `rootDir` is not specified, we guess at it the same way node does). This ensures that `rosetta extract` is structure agnostic.

---

### DIFF
--- a/packages/jsii/lib/symbol-id.ts
+++ b/packages/jsii/lib/symbol-id.ts
@@ -150,11 +150,18 @@ export function normalizePath(
     sourcePath =
       rootDir === '.' ? paths.join('/') : `${rootDir}/${paths.join('/')}`;
   }
-  return sourcePath;
+  return unixize(sourcePath);
 
   function removeEndSlash(filePath: string) {
     return filePath.endsWith(path.sep)
       ? filePath.slice(0, filePath.length - 1)
       : filePath;
   }
+}
+
+/**
+ * Turn backslashes in a path into forward slashes
+ */
+function unixize(p: string) {
+  return p.replace(/\\/g, '/');
 }
--- a/packages/jsii/lib/symbol-id.ts
+++ b/packages/jsii/lib/symbol-id.ts
@@ -120,6 +120,13 @@ function assemblyRelativeSourceFile(
   }
 }
 
+/**
+ * Ensures that the sourcePath is pointing to the source code
+ * and not compiled code. This can happen if the root directory
+ * and/or out directory is set for the project. We check to see
+ * if the out directory is present in the sourcePath, and if so,
+ * we replace it with the root directory.
+ */
 export function normalizePath(
   sourcePath: string,
   rootDir?: string,
--- a/packages/jsii/lib/symbol-id.ts
+++ b/packages/jsii/lib/symbol-id.ts
@@ -3,7 +3,17 @@ import * as fs from 'fs';
 import * as path from 'path';
 import * as ts from 'typescript';
 
+/**
+ * Additional options that may be provided to the symbolIdentifier.
+ */
 interface SymbolIdOptions {
+  /**
+   * The assembly that the symbol is found in.
+   * This is used to provide the correct root directory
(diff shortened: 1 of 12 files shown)


packages/jsii/lib/symbol-id.ts
```diff
@@ -5,6 +5,8 @@ import * as ts from 'typescript';
 export function symbolIdentifier(
   typeChecker: ts.TypeChecker,
   sym: ts.Symbol,
+  valueAsIs = true,
```
rix0rrr, line 8: Can we think of a better name for this argument?

Perhaps it should go into an `options` struct as well...

packages/jsii/lib/symbol-id.ts
```diff
@@ -61,11 +71,27 @@ function assemblyRelativeSourceFile(sourceFileName: string) {
...
+  let sourcePath = removePrefix(
     packageJson.jsii?.outdir ?? '',
     path.relative(path.dirname(packageJsonLocation), sourceFileName),
   );
 
+  // Modify the namespace if we allow changes and if the outDir exists.
+  // We should only allow changes if we call the function from rosetta.
+  if (!valueAsIs && packageJson.jsii?.tsc?.outDir) {
+    const paths = path.normalize(sourcePath).split(path.sep);
+    const pathDir = paths.shift();
```
rix0rrr, line 83: Looks like we're only comparing the very first path segment. What if `outDir` or `rootDir` consists of 2 path segments? 

```
{
  "rootDir": "src/public",
  "outDir": "dist/minified",
}
```
?

packages/jsii/test/compiler.test.ts
```diff
@@ -103,6 +111,43 @@ describe(Compiler, () => {
...
+          },
+        },
+        failOnWarnings: true,
+        projectReferences: false,
+      });
+
+      await compiler.emit();
+
+      const assembly = await readJson(join(sourceDir, '.jsii'), 'utf-8');
+      expect(assembly.metadata).toEqual({
```
rix0rrr, line 144: For robustness against future additions to metadata:

```suggestion
      expect(assembly.metadata).toEqual(expect.objectContaining({
```

rix0rrr (changes requested):

packages/jsii/test/symbol-identifiers.test.ts
```diff
@@ -92,3 +92,85 @@ test('Submodules also have symbol identifiers', async () => {
     'index:cookie',
   );
 });
+
+describe(normalizePath, () => {
```
rix0rrr, line 96: Nice! That is hot damn thorough! 😅

rix0rrr (approved):

rix0rrr:
Oh no, issue on Windows:

```
FAIL test/symbol-identifiers.test.ts
jsii:   ● normalizePath › specify multiple directories
jsii:     expect(received).toEqual(expected) // deep equality
jsii:     Expected: "root/extra/lib/filename.ts"
jsii:     Received: "root\\extra/lib/filename.ts"
jsii:       164 |       'root/filename.ts',
jsii:       165 |     );
jsii:     > 166 |     expect(normalizePath('out/lib/filename.ts', 'root/extra', 'out')).toEqual(
jsii:           |                                                                       ^
jsii:       167 |       'root/extra/lib/filename.ts',
jsii:       168 |     );
jsii:       169 |     expect(normalizePath('out/lib/filename.ts', '.', 'out/lib')).toEqual(
jsii:       at Object.<anonymous> (test/symbol-identifiers.test.ts:166:71)
```

PR #3233, opened 2021-12-06
fix(jsii): correctly identify types regardless of import method

rix0rrr:
Types imported from dependency's submodules are only correctly
identified if they are imported via a top-level namespace import.

Otherwise, they are incorrectly identified as coming from the top-level
scope and compilation will fail.

Compare:

```ts
// Works
import { aws_ec2 as ec2 } from 'aws-cdk-lib';

// Does not work
import * as ec2 from 'aws-cdk-lib/aws-ec2';
```

Fortunately we already have symbol ids to solve this exact issue
for Rosetta, so we can resort to looking up FQN by symbol id.

Fixes aws/aws-cdk#17860.

---

### DIFF
--- a/packages/@jsii/spec/lib/configuration.ts
+++ b/packages/@jsii/spec/lib/configuration.ts
@@ -87,6 +87,16 @@ export interface Config {
   metadata?: {
     [key: string]: any;
   };
+
+  /**
+   * TypeScript compiler options
+   */
+  tsc?: {
+    outDir?: string;
+    rootDir?: string;
+
+    [key: string]: any;
+  };
 }
 
 /**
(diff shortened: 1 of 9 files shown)


rix0rrr:
This PR is currently blocked on https://github.com/aws/jsii/pull/3225, as the build of `jsii-calc-lib` uses an `outDir` and our symbol Id implementation is currently not yet robust against outdir-imported declarations.

njlynch (commented):
Any chance of some tests to prove out this works (and prevent future regressions)?

rix0rrr:
> Any chance of some tests to prove out this works (and prevent future regressions)?

Yessir! Coming up. I got sidetracked here... it's a little annoying because I had the test already, but apparently I didn't commit it so it must have gotten lost in a `git reset --hard` somewhere 😢

njlynch (approved):

PR #3234, opened 2021-12-06
fix(jsii): symbolid for single-valued enums is incorrect

rix0rrr:
We used to be doing our own `symbolId` calculations, but this was silly:
TS has a perfectly cromulent function built-in to do the same.

Implement ours in terms of theirs.

This uncovered a bug in calculating symbolIds for single-valued enums, which is rectified as well.

---

### DIFF
--- a/packages/jsii/lib/symbol-id.ts
+++ b/packages/jsii/lib/symbol-id.ts
@@ -1,7 +1,9 @@
 import { Assembly } from '@jsii/spec';
-import * as fs from 'fs';
+import * as fs from 'fs-extra';
 import * as path from 'path';
 import * as ts from 'typescript';
+
+import { findUp } from './utils';
 
 /**
  * Additional options that may be provided to the symbolIdentifier.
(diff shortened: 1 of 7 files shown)


packages/jsii/lib/symbol-id.ts
```diff
@@ -11,41 +39,21 @@ export function symbolIdentifier(
...
-  if (!namespace) {
+  const [, fileName, typeName] = groups; // typeName may be absent
+
+  const relFile = assemblyRelativeSourceFile(fileName);
+  if (!relFile) {
     return undefined;
   }
 
-  return `${namespace}:${inFileNameParts.join('.')}`;
+  return `${relFile}:${typeName ?? ''}`;
```
otaviomacedo, line 56: Is it intentional that we leave the ':' in, even when there is no type name?

packages/jsii/lib/symbol-id.ts
```diff
@@ -11,41 +39,21 @@ export function symbolIdentifier(
...
-        inFileNameParts.unshift(declSym.name);
-      }
-    }
-    decl = decl.parent;
-  }
-  if (!decl) {
+  const tsName = typeChecker.getFullyQualifiedName(sym);
+  // TypeScript fqn looks like "/path/to/file"[.name.in.file]
+  const groups = /^"([^"]+)"(?:\.(.*))?$/.exec(tsName);
+  if (!groups) {
```
otaviomacedo, line 45: What does it mean if the name doesn't match the regex? Should we log a warning?

rix0rrr: I don't know that we have the mechanisms to log a warning. I am fine ignoring this for now, we have end-to-end tests that will ensure everything works correctly. [Alarm on symptoms, not causes](https://paulbellamy.com/2017/08/symptoms-not-causes)

otaviomacedo: I believe we do (`log4js`). But anyway, this is just a minor point.

otaviomacedo (approved):

PR #3238, opened 2021-12-07
fix(jsii): move sourceMaps into separate files

deftomat:
Importing the `aws-cdk-lib` in NodeJS may be quite slow as it is a massive library with a lot of files.

Yes, it is a good idea to import only a part which you need, like `aws-cdk-lib/aws-lambda`.

Unfortunately, developers ofter forget about it and they import the whole CDK instead.

As NodeJs needs to parse all imported JS files. It is a good idea to reduce the number of characters which needs to be parsed/loaded to minimum.

For example, moving inline source maps from each JS file into separate MAP file may reduce the import time by 30%.

On my machine (Macbook Pro 2017 15inch), just a plain import could took a 1200ms. So, every CDK command is slowed down by importing the whole CDK by 1200ms. 

It turns out, I was able to reduce it to 850ms by removing the inlined source maps from JS files and moving them into separate `*.js.map` files.

To really use the source maps, user needs to use `--enable-source-maps` flag to enable them. It works both with inlined source maps and with external source maps, so I think there is no reason to keep them inlined.

---

### DIFF
--- a/packages/jsii/lib/compiler.ts
+++ b/packages/jsii/lib/compiler.ts
@@ -18,7 +18,7 @@ const BASE_COMPILER_OPTIONS: ts.CompilerOptions = {
   declaration: true,
   experimentalDecorators: true,
   incremental: true,
-  inlineSourceMap: true,
+  sourceMap: true,
   inlineSources: true,
   lib: ['lib.es2019.d.ts'],
   module: ts.ModuleKind.CommonJS,


rix0rrr:
I agree with the goal.

Here's another thing we could be doing: bundle and minify each submodule. In my tests that also increases the load speed a bunch. Something we could/should be doing in the CDK repo, not in jsii though.

And I think the same for this change: this flag could go into `package.json#jsii.tsc` (I believe) where compileroptions go.

deftomat:
@rix0rrr Yes, bundling the submodules could help a lot.

I think flag should be in `jsii` as a default. It is a welcome optimisation for everyone which is using `jsii`.

rix0rrr:
For most libraries you wouldn't be able to tell the difference, it's just that `aws-cdk-lib` is huge.

I'm going to close this for now, we're addressing the issue in the CDK build. Thanks for bringing it up!

PR #3243, opened 2021-12-09
fix(rosetta): infuse drops first assembly

kaizen3031593:
As part of a refactor, `rosetta infuse` no longer utilizes a single tablet file in favor of tablets at the assembly level. However the `infuse` command still takes in a positional argument `TABLET` that goes unused. This means that the first argument sent into the command gets dropped.

---


kaizen3031593:
btw, it looks like `yargs` would not "eat" an option as the first positional argument.

```bash
rosetta:infuse --cache-from='blah' assembly1 assembly2
```
result is `TABLET=assembly1`.

PR #3248, opened 2021-12-10
fix(rosetta): `rosetta extract` fails if run after `rosetta infuse`

kaizen3031593:
There is a valid use case to run `rosetta extract` after `rosetta infuse`.
However, we run into problems because we `fixturize` snippets
before we check the cache, and sometimes `fixturizing` will fail for
infused snippets (specifically, literate sources like `integ.blah.lit.ts`
will fail).

We do not actually need to successfully `fixturize` infused examples.
We rely on these examples to be in the cached tablet file and thus
do not need to be compiled and translated again. So we introduce
a new metadata tag, `infuse=true` for all infused examples. This tag, 
when consumed by `extract`, will skip any errors associated with
that snippet. 

ALSO: increased test timeout from `30000` to `50000` to allow enough
time for mac PR test to succeed.

---


packages/jsii-rosetta/lib/commands/infuse.ts
```diff
@@ -207,17 +207,20 @@ function insertExample(
   type: spec.Type,
   tablets: LanguageTablet[],
 ): void {
-  const exampleMetadata = fmap(original?.parameters, renderMetadataline);
+  const parameters = {
+    ...original?.parameters,
+    infused: 'true',
```
rix0rrr, line 212: Empty string?

rix0rrr (changes requested):

rix0rrr (approved):

PR #3266, opened 2021-12-17
chore: centralize disclaimers to extract

kaizen3031593:
Draft PR; no tests because I am not sure if we want to continue down this road.

Currently we have two disclaimers -- one for non-compiling examples and the 
other for generated examples. The former is appended in `pacmak`, while the
latter is appended to the actual source code.

The idea proposed in this PR is to centralize disclaimers to `extract` as a property
of `TranslatedSnippet`. They then get rendered when the target language code
gets rendered.

Since the TypeScript doc workflow does not include `pacmak` or `extract`, so 
the non-compiling disclaimer does not show. If we add the generated disclaimer
to the central `extract` location as well, it will not show in the TypeScript docs
either.

---

### DIFF
--- a/packages/jsii-pacmak/lib/targets/dotnet/dotnetdocgenerator.ts
+++ b/packages/jsii-pacmak/lib/targets/dotnet/dotnetdocgenerator.ts
@@ -185,8 +185,8 @@ export class DotNetDocGenerator {
   }
 
   private prefixDisclaimer(translated: Translation) {
-    if (!translated.didCompile && INCOMPLETE_DISCLAIMER_NONCOMPILING) {
-      return `// ${INCOMPLETE_DISCLAIMER_NONCOMPILING}\n${translated.source}`;
+    if (translated.disclaimer) {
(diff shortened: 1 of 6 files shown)


packages/jsii-pacmak/lib/targets/dotnet/dotnetdocgenerator.ts
```diff
@@ -185,8 +185,8 @@ export class DotNetDocGenerator {
   }
 
   private prefixDisclaimer(translated: Translation) {
-    if (!translated.didCompile && INCOMPLETE_DISCLAIMER_NONCOMPILING) {
-      return `// ${INCOMPLETE_DISCLAIMER_NONCOMPILING}\n${translated.source}`;
+    if (translated.disclaimer) {
+      return `// ${translated.disclaimer}\n${translated.source}`;
```
rix0rrr, line 189: I don't think there's a big downside to just having the disclaimer be a part of `translated.source` here, do you?

PR #3269, opened 2021-12-17
feat(rosetta): Rosetta manages dependencies automatically

rix0rrr:
It's quite an annoying bother and pretty brittle to have to set up a full dependency directory to compile examples in.

In CDK, we use the directory of the `decdk` package to do this, because we have external knowledge that all packages
will be symlinked into this directory.

That mechanism is brittle and revolves around having outside knowledge. Instead, make Rosetta manage the dependency closure itself, when user declare `exampleDependencies` in `package.json`:

```
{
  "jsiiRosetta": {
    "exampleDependencies": {
      "@some-other/package": "^1.2.3",
      "@yet-another/package": "*",
    }
  }
}
```

It is not necessary to redeclare any of the `dependencies` and `peerDependencies` already in `package.json`: those are automatically included.

---


packages/jsii-rosetta/lib/find-utils.ts
```diff
@@ -0,0 +1,86 @@
...
+ * Whether the given dependency is a built-in
+ *
+ * Some dependencies that occur in `package.json` are also built-ins in modern Node
+ * versions (most egregious example: 'punycode'). Detect those and filter them out.
+ */
+export function isBuiltinModule(depName: string) {
+  // eslint-disable-next-line @typescript-eslint/no-require-imports,@typescript-eslint/no-var-requires
+  const { builtinModules } = require('module');
+  return (builtinModules ?? []).includes(depName);
+}
```
RomainMuller, line 86: AFAIK, the behavior of `node` module resolution is that, unless the module specifier (in `require(moduleSpecifier)` or `import from moduleSpecifier`) starts with `node:` (at which point it will ALWAYS resolve to a built-in module), it'll start by trying to resolve from external library search paths, and ONLY if not found will it resort to built-ins.

This makes me feel like this behavior is, in fact, incorrect in this respects.

packages/jsii-rosetta/lib/jsii/assemblies.ts
```diff
@@ -171,36 +186,50 @@ export function allSnippetSources(assembly: spec.Assembly): AssemblySnippetSourc
...
-    }
-  }
-  return ret;
+export async function allTypeScriptSnippets(
+  assemblies: readonly LoadedAssembly[],
+  loose = false,
+): Promise<TypeScriptSnippet[]> {
+  return Promise.all(
+    assemblies
+      .flatMap((loaded) => allSnippetSources(loaded.assembly).map((source) => ({ source, loaded })))
```
RomainMuller, line 195: Oh jeez, I though `flatMap` was a node 14+ and realizing that it is, in fact, node 11+ kind of made my day 🙃

packages/jsii-rosetta/lib/jsii/assemblies.ts
```diff
@@ -171,36 +186,50 @@ export function allSnippetSources(assembly: spec.Assembly): AssemblySnippetSourc
...
+              },
+            ];
+          case 'markdown':
+            return extractTypescriptSnippetsFromMarkdown(source.markdown, source.location, isStrict(loaded)).map(
+              (snippet) => ({ snippet, loaded }),
+            );
+        }
+      })
+      .map(async ({ snippet, loaded }) => {
+        const isInfused = snippet.parameters?.infused !== undefined;
```
RomainMuller, line 219: Nit - I prefer `!= null`....

```suggestion
        const isInfused = snippet.parameters?.infused != null;
```

packages/jsii-rosetta/lib/rosetta-translator.ts
```diff
@@ -164,3 +201,12 @@ export interface ReadFromCacheResults {
   readonly translations: TranslatedSnippet[];
   readonly remaining: TypeScriptSnippet[];
 }
+
+export interface TranslateAllOptions {
+  readonly compilationDirectory?: string;
```
RomainMuller, line 206: I guess `@default - a new temporary directory`?

packages/jsii-rosetta/lib/snippet-dependencies.ts
```diff
@@ -0,0 +1,245 @@
...
+import { TypeScriptSnippet, CompilationDependency } from './snippet';
+import { mkDict, formatList } from './util';
+
+/**
+ * Collect the dependencies of a bunch of snippets together in one declaration
+ *
+ * We assume here the dependencies will not conflict.
+ */
+export function collectDependencies(snippets: TypeScriptSnippet[]) {
+  const ret: Record<string, CompilationDependency> = {};
```
RomainMuller, line 20: FYI (don't go refactoring this!), I kind of pivoted to using `Map<String, T>` in such cases (i.e: where the key space isn't known upfront).

packages/jsii-rosetta/lib/snippet-dependencies.ts
```diff
@@ -0,0 +1,245 @@
...
+
+    return a;
+  }
+
+  if (a.type === 'symbolic' && b.type === 'concrete') {
+    // Reverse roles so we fall into the previous case
+    return resolveConflict(name, b, a);
+  }
+
+  throw new Error('Cases should have been exhaustive');
```
RomainMuller, line 70: I kind of like using assertion errors for "this should have been impossible".

```suggestion
  throw new assert.AssertionError('Cases should have been exhaustive');
```

packages/jsii-rosetta/lib/snippet-dependencies.ts
```diff
@@ -0,0 +1,245 @@
...
+    isSymbolic(dep) ? [`${name}@${dep.versionRange}`] : [],
+  );
+  const linkedInstalls = mkDict(
+    Object.entries(resolvedDeps).flatMap(([name, dep]) =>
+      isConcrete(dep) ? [[name, dep.resolvedDirectory] as const] : [],
+    ),
+  );
+
+  // Run 'npm install' on it
+  await fs.mkdirp(tmpDir);
```
RomainMuller, line 139: Doesn't `tmpDir` come from `fs.mkdtemp` here? In this case... should already exist?

packages/jsii-rosetta/lib/snippet-dependencies.ts
```diff
@@ -0,0 +1,245 @@
...
+
+  if (symbolicInstalls.length > 0) {
+    logging.debug(`Installing example dependencies: ${symbolicInstalls.join(' ')}`);
+    cp.execSync(`npm install ${symbolicInstalls.join(' ')}`, { cwd: tmpDir, encoding: 'utf-8' });
```
RomainMuller, line 143: Perhaps also turn on "global style" install? This will prevent the package's own dependencies from being installed/de-duped at the same level, too, so you guarantee the list of directories under `node_modules` after this is done is exactly the list of package names you passed.

That might however have unintended consequences (i.e: multiple different versions of transitive dependencies), which might be problematic (I'm not sure).

packages/jsii-rosetta/lib/snippet-dependencies.ts
```diff
@@ -0,0 +1,245 @@
...
+  }
+
+  // Symlink the rest
+  if (Object.keys(linkedInstalls).length > 0) {
+    logging.debug(`Symlinking example dependencies: ${Object.values(linkedInstalls).join(' ')}`);
+    const modDir = path.join(tmpDir, 'node_modules');
+    await Promise.all(
+      Object.entries(linkedInstalls).map(async ([name, source]) => {
+        const target = path.join(modDir, name);
+        await fs.mkdirp(path.dirname(target));
```
RomainMuller, line 153: I wondered why you did this for a moment, and then I remembered...

```suggestion
        // Package could be namespaced, so ensure the namespace dir exists
        await fs.mkdirp(path.dirname(target));
```

packages/jsii-rosetta/lib/snippet-dependencies.ts
```diff
@@ -0,0 +1,245 @@
...
+        await fs.symlink(source, target, 'dir');
```
RomainMuller, line 154: So... because of how you might have run `npm install` before, isn't it possible that some package/directory already exists (if it was a transitive dependency of a symbolic package)?

This collision *might* be solved by moving the symlink creation before the symbolic package installs, as those would then be tempted to re-use existing packages instead of installing fresh?

packages/jsii-rosetta/lib/snippet-dependencies.ts
```diff
@@ -0,0 +1,245 @@
...
+    );
+  }
+
+  return tmpDir;
+}
+
+/**
+ * Map package name to directory
+ */
+async function scanMonoRepos(startingDirs: string[]): Promise<Record<string, string>> {
```
RomainMuller, line 165:
```suggestion
async function scanMonoRepos(startingDirs: readonly string[]): Promise<Record<string, string>> {
```

packages/jsii-rosetta/lib/snippet.ts
```diff
@@ -30,8 +30,22 @@ export interface TypeScriptSnippet {
    * @default false
    */
   readonly strict?: boolean;
+
+  /**
+   * Dependencies necessary to compile this snippet
+   *
+   * Value is a regular { name -> semver } map like NPM's `dependencies`,
```
RomainMuller, line 37: Is that true, though? CompilationDependency is not `semver` :)

packages/jsii-rosetta/lib/util.ts
```diff
@@ -24,6 +24,12 @@ export function printDiagnostics(diags: readonly RosettaDiagnostic[], stream: No
   }
 }
 
+export function formatList(xs: string[], n = 5) {
+  const tooMany = xs.length - n;
+
+  return tooMany > 0 ? `${xs.join(', ')} (and ${tooMany} more)` : xs.join(', ');
+}
```
RomainMuller, line 31: Given the usage I see in `snippet-dependencies.ts`, it looks like this function's implementation is not correct.

```suggestion
export function formatList(xs: readonly string[], n = 5) {
  const tooMany = xs.length - n;
  // Limit xs to the desired visible item count
  xs = XS.slice(0, n);

  const suffix = tooMany > 0
    ? ` (and ${suffix} more)`
    : '';

  return `${xs.join(', ')}${suffix}`;
}
```

packages/jsii-rosetta/lib/find-utils.ts
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  return result ? directory : recurse();
+
+  function recurse() {
+    const parent = path.dirname(directory);
+    if (parent === directory) {
+      return undefined;
+    }
+    return findUp(parent, pred as any);
```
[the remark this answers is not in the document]
rix0rrr: I felt the same way, but nevertheless

```
No overload matches this call.
  Overload 1 of 2, '(directory: string, pred: (dir: string) => Promise<boolean>): Promise<string | undefined>', gave the following error.
    Argument of type '(dir: string) => boolean | Promise<boolean>' is not assignable to parameter of type '(dir: string) => Promise<boolean>'.
      Type 'boolean | Promise<boolean>' is not assignable to type 'Promise<boolean>'.
        Type 'false' is not assignable to type 'Promise<boolean>'.
  Overload 2 of 2, '(directory: string, pred: (dir: string) => boolean): string | undefined', gave the following error.
    Argument of type '(dir: string) => boolean | Promise<boolean>' is not assignable to parameter of type '(dir: string) => boolean'.
      Type 'boolean | Promise<boolean>' is not assignable to type 'boolean'.
        Type 'Promise<boolean>' is not assignable to type 'boolean'.
```

packages/jsii-rosetta/lib/find-utils.ts
rix0rrr: It doesn't, though.

```
$ npm init -y
$ npm install punycode

added 1 package, and audited 2 packages in 601ms

found 0 vulnerabilities

$ node -p "require.resolve('punycode')"
punycode                <--- indicating a built-in module
```

packages/jsii-rosetta/lib/snippet-dependencies.ts
rix0rrr: I'm always going back and forth on this.

What's your motivation for picking `Map` over an object?

packages/jsii-rosetta/lib/snippet-dependencies.ts
rix0rrr: The arguments to this thing are making me uncomfortable though. It looks pretty narrowly targeted towards unit test assertions:
image

https://nodejs.org/api/assert.html#new-assertassertionerroroptions

packages/jsii-rosetta/lib/snippet-dependencies.ts
rix0rrr: I don't think this is an issue. The packages should all be able to be installed into a single flat closure.

packages/jsii-rosetta/lib/snippet-dependencies.ts
rix0rrr: Do you reckon? I'd say NPM just tramples over everything that might already exist, if I know it well enough :).

It's a good point that the `symlink` operation might fail though, so I need to check for that.

packages/jsii-rosetta/lib/snippet.ts
rix0rrr: Well. The string that's there is supposed to be :).

packages/jsii-rosetta/lib/commands/coverage.ts
```diff
@@ -7,7 +7,7 @@ export async function checkCoverage(assemblyLocations: readonly string[]): Promi
   logging.info(`Loading ${assemblyLocations.length} assemblies`);
   const assemblies = await loadAssemblies(assemblyLocations, false);
 
-  const snippets = Array.from(allTypeScriptSnippets(assemblies, true));
+  const snippets = Array.from(await allTypeScriptSnippets(assemblies, true));
```
RomainMuller, line 10: `allTypeScriptSnippets` returns a promise to an array already...

```suggestion
  const snippets = await allTypeScriptSnippets(assemblies, true);
```

RomainMuller (approved):

packages/jsii-rosetta/lib/find-utils.ts
RomainMuller: Ah you're correct, the module resolution STARTS with "if this is a built-in, return the built-in". My bad, sorry.

packages/jsii-rosetta/lib/snippet-dependencies.ts
RomainMuller: There are a couple of reasons:
- `Map` typically won't lead you to prototype corruption routes, because you'll probably never be tempted to `Object.assign` on them.
  - Alternatively, you'd initialize your object with `Object.create(null)` to remove risk for future prototype pollution.
- `Map` won't create a new hidden class each time you assign a new key/property on them. This reduces memory churn, and perhaps also footprint (although hidden classes now get GC'd -- in earlier versions of V8, I reckon they'd land in "permanent gen" and stick around forever).
- `Map` can be used in `for [key, value] of map`, while for Objects you have to `Object.entries(obj)` there (which is also not `null`-safe)
- `Map` does not have some of the "surprising" behavior of objects with their keys, e.g:
  ```js
  const obj = Object.create(null);
  const map = new Map();

  obj[123] = 'FOO';
  assert(obj['123'] === 'FOO');

  map.set(123, 'FOO');
  assert(map.get('123') !== 'FOO');
  ```
- `Object` properties need to be valid property names. `Map` keys can be valid JS value.

PR #3277, opened 2021-12-21
fix(pacmak): fails on bundled dependency without entry point

### LINKED ISSUE #3275: Cannot package project with non-importable dependencies
### Affected Languages

- [x] `Python`
- [x] `Java`
- [x] .NET (`C#`, `F#`, ...)
- [x] `Go`

### General Information
* **JSII Version:** jsii 1.49.0
* **Platform:** Unix

### What is the problem?

If a JSII project takes a dependency on a package that is not importable, `jsii-pacmak` fails. A package may not be importable if it only provides command-line scripts, for example the [shx](https://www.npmjs.com/package/shx) package. shx does not specify the "main" field in package.json, and it does not have a default "index.js" file that serves as an entrypoint -- its main utility is through the "bin" field which directly runs JS code.

This previously worked, but stopped working when we upgraded our project from JSII v1.47.0 to v1.49.0.

Our use case for taking "non-importable" libraries as dependencies is to bundle them along with our project, so that the command-line scripts in "shx" can be used for functionality in our construct library.

Related issue: https://github.com/projen/projen/pull/1336

### Verbose Log

Reproduction repo: https://github.com/Chriscbr/jsii-pacmak-repro

```
$ npx jsii-pacmak
[jsii-pacmak] [WARN] Exception occurred, not cleaning up
[jsii-pacmak] [WARN] java failed
Error: Cannot find module 'shx'
Require stack:
- /Volumes/workplace/jsii-pacmak-repro/node_modules/jsii-pacmak/lib/util.js
- /Volumes/workplace/jsii-pacmak-repro/node_modules/jsii-pacmak/lib/packaging.js
- /Volumes/workplace/jsii-pacmak-repro/node_modules/jsii-pacmak/lib/npm-modules.js
- /Volumes/workplace/jsii-pacmak-repro/node_modules/jsii-pacmak/lib/index.js
- /Volumes/workplace/jsii-pacmak-repro/node_modules/jsii-pacmak/bin/jsii-pacmak.js
- /Volumes/workplace/jsii-pacmak-repro/node_modules/jsii-pacmak/bin/jsii-pacmak
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:933:15)
    at Function.resolve (node:internal/modules/cjs/helpers:108:19)
    at Object.findDependencyDirectory (/Volumes/workplace/jsii-pacmak-repro/node_modules/jsii-pacmak/lib/util.js:18:32)
    at /Volumes/workplace/jsii-pacmak-repro/node_modules/jsii-pacmak/lib/dependency-graph.js:46:42
    at Array.map (<anonymous>)
    at real$traverseDependencyGraph (/Volumes/workplace/jsii-pacmak-repro/node_modules/jsii-pacmak/lib/dependency-graph.js:45:10)
    at async Object.findLocalBuildDirs (/Volumes/workplace/jsii-pacmak-repro/node_modules/jsii-pacmak/lib/target.js:76:5)
    at async /Volumes/workplace/jsii-pacmak-repro/node_modules/jsii-pacmak/lib/targets/java.js:171:29
    at async JavaBuilder.generateMavenSettingsForLocalDeps (/Volumes/workplace/jsii-pacmak-repro/node_modules/jsii-pacmak/lib/targets/java.js:173:51)
    at async /Volumes/workplace/jsii-pacmak-repro/node_modules/jsii-pacmak/lib/targets/java.js:103:1
 ```

### PR DESCRIPTION
**Pull request by rix0rrr**
`projen` uses a bundled dependency called `shx`, and this library does
not have an entry point. It exposes some command-line tools and nothing
else.

This conflicts with our dependency finding mechanism, which performs
a `require()` on the library to find it.

Ignore bundled dependencies in `pacmak`, it never needs to pack them
individually anyway.

Fixes #3275.

---

### DIFF
--- a/packages/jsii-pacmak/lib/dependency-graph.ts
+++ b/packages/jsii-pacmak/lib/dependency-graph.ts
@@ -58,6 +58,7 @@ export interface TraverseDependencyGraphHost {
 export interface PackageJson {
   readonly dependencies?: { readonly [name: string]: string };
   readonly peerDependencies?: { readonly [name: string]: string };
+  readonly bundleDependencies?: string[];
   readonly bundledDependencies?: string[];
 
   readonly [key: string]: unknown;
@@ -93,7 +94,7 @@ async function real$traverseDependencyGraph(
       .filter(
         (m) =>
           !util.isBuiltinModule(m) &&
-          !(meta.bundledDependencies ?? []).includes(m),
+          (meta.bundledDependencies ?? meta.bundleDependencies)?.includes(m),
       )
       .map(async (dep) => {
         const dependencyDir = await host.findDependencyDirectory(
--- a/packages/jsii-pacmak/lib/dependency-graph.ts
+++ b/packages/jsii-pacmak/lib/dependency-graph.ts
@@ -94,7 +94,7 @@ async function real$traverseDependencyGraph(
       .filter(
         (m) =>
           !util.isBuiltinModule(m) &&
-          (meta.bundledDependencies ?? meta.bundleDependencies)?.includes(m),
+          !(meta.bundledDependencies ?? meta.bundleDependencies)?.includes(m),
       )
       .map(async (dep) => {
         const dependencyDir = await host.findDependencyDirectory(
--- a/packages/jsii-pacmak/lib/dependency-graph.ts
+++ b/packages/jsii-pacmak/lib/dependency-graph.ts
@@ -58,6 +58,7 @@ export interface TraverseDependencyGraphHost {
 export interface PackageJson {
   readonly dependencies?: { readonly [name: string]: string };
   readonly peerDependencies?: { readonly [name: string]: string };
+  readonly bundledDependencies?: string[];
 
   readonly [key: string]: unknown;
 }
@@ -88,7 +89,12 @@ async function real$traverseDependencyGraph(
   ]);
   return Promise.all(
     Array.from(deps)
-      .filter((m) => !util.isBuiltinModule(m))
+      // No need to pacmak the dependency if it's built-in, or if it's bundled
+      .filter(
+        (m) =>
+          !util.isBuiltinModule(m) &&
+          !(meta.bundledDependencies ?? []).includes(m),
+      )
       .map(async (dep) => {
         const dependencyDir = await host.findDependencyDirectory(
           dep,
--- a/packages/jsii-pacmak/lib/dependency-graph.ts
+++ b/packages/jsii-pacmak/lib/dependency-graph.ts
@@ -94,7 +94,8 @@ async function real$traverseDependencyGraph(
       .filter(
         (m) =>
           !util.isBuiltinModule(m) &&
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/jsii-pacmak/lib/dependency-graph.ts [hunk 1]
```diff
@@ -58,6 +58,7 @@ export interface TraverseDependencyGraphHost {
 export interface PackageJson {
   readonly dependencies?: { readonly [name: string]: string };
   readonly peerDependencies?: { readonly [name: string]: string };
+  readonly bundledDependencies?: string[];
```
RomainMuller, line 61: This can be named `bundleDependencies` or `bundledDependencies` (they are synonyms). Probably should account for that... As we don't mandate people use one or the other.

packages/jsii-pacmak/lib/dependency-graph.ts
```diff
@@ -88,7 +89,12 @@ async function real$traverseDependencyGraph(
   ]);
   return Promise.all(
     Array.from(deps)
-      .filter((m) => !util.isBuiltinModule(m))
+      // No need to pacmak the dependency if it's built-in, or if it's bundled
+      .filter(
+        (m) =>
+          !util.isBuiltinModule(m) &&
+          !(meta.bundledDependencies ?? []).includes(m),
```
RomainMuller, line 96:
```suggestion
          !meta.bundledDependencies?.includes(m),
```

packages/jsii-pacmak/lib/dependency-graph.ts
```diff
@@ -88,7 +90,12 @@ async function real$traverseDependencyGraph(
   ]);
   return Promise.all(
     Array.from(deps)
-      .filter((m) => !util.isBuiltinModule(m))
+      // No need to pacmak the dependency if it's built-in, or if it's bundled
+      .filter(
+        (m) =>
+          !util.isBuiltinModule(m) &&
+          !(meta.bundledDependencies ?? meta.bundleDependencies)?.includes(m),
```
RomainMuller, line 97: Unfortunately, the spec could be interpreted as implying that if both are provided, both will be honored (sigh).

```suggestion
          !meta.bundledDependencies?.includes(m) &&
          !meta.bundleDependencies?.includes(m),
```

packages/jsii-pacmak/test/dependency-graph.test.ts
```diff
@@ -113,3 +113,41 @@ test('stops traversing when callback returns false', async () => {
   expect(mockHost.readJson).toHaveBeenCalledTimes(2);
   expect(mockHost.findDependencyDirectory).toHaveBeenCalledTimes(1);
 });
+
+test('dont call findDependencyDirectory for bundledDependencies', async () => {
```
RomainMuller, line 117: Can you add flavors of this test for:
- the alternate spelling
- when both are used

RomainMuller (approved):

PR #3282, opened 2021-12-21
fix(rosetta): diagnostics from infused snippets were not ignored

kaizen3031593:
Recently we implemented the `infused` metadata that turns `strict=false`
and `loose=true` for the specific snippet. However we still output the
diagnostics along with other snippets (with `isError=true`). One thing we
could do is make these diagnostics `isError=false`, but I argue that we
want to ignore these diagnostics entirely. They aren't actually diagnostics;
we will find them in the cache and they do compile.

---

### DIFF
--- a/packages/jsii-rosetta/lib/translate.ts
+++ b/packages/jsii-rosetta/lib/translate.ts
@@ -57,7 +57,7 @@ export class Translator {
       }),
     );
 
-    if (snip.strict) {
+    if (snip.parameters?.infused === undefined) {
       this.#diagnostics.push(...translator.diagnostics);
     }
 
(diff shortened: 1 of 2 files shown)


rix0rrr:
I agree with the conclusion, but I'm not sure the premise is right. There is a bug, but I think it lies somewhere else:

> Recently we implemented the infused metadata that turns strict=false and loose=true for the specific snippet. However we still output the diagnostics along with other snippets (with isError=true).

AFAIK setting `strict=false` should prevent `isError` from turning `true`, no? Why is that not working?

> I argue that we want to ignore these diagnostics entirely. They aren't actually diagnostics; we will find them in the cache and they do compile.

The diagnostics must originate from trying to compile the snippet. So my counter-question to this observation would be: if we find them in the cache, then why are we trying to compile them in the first place?

rix0rrr (approved):

PR #3291, opened 2021-12-23
fix(rosetta): infused snippets not returned from cache

kaizen3031593:
Infused snippets do not have the right full source since we are adding the local fixture to the snippet
rather than the fixture from where the snippet came from. Since there is no hope in actually translating
an infused snippet (unless it is in the same directory), we might as well take any cache result as is.
So this PR adds logic that returns the cache result for infused snippets always.

This was tested on my local machine and dropped all errors from infused snippets. There are still errors
from directories that are not yet strict, which is understandable.

Also, the infuse command was missing `cache-from` argument that is present in `infuseOptions`.
We were using those options as part of `rosetta extract --infuse` but should be available via `rosetta infuse`
as well.

Also fixed up minor mistake in tests where I was testing for `infused=true` instead of `infused=''`.

---


packages/jsii-rosetta/lib/rosetta-translator.ts
```diff
@@ -149,6 +149,13 @@ export class RosettaTranslator {
 function tryReadFromCache(sourceSnippet: TypeScriptSnippet, cache: LanguageTablet, fingerprinter: TypeFingerprinter) {
   const fromCache = cache.tryGetSnippet(snippetKey(sourceSnippet));
 
+  // infused snippets won't pass the full source check or the fingerprinter
```
kaizen3031593, line 152: Why do they not pass the fingerprinter? I tried to investigate this, and I can't really make sense of it.
Verified that the fqns are the same between `fromCache.fqnsReferenced()` and `fromCache.snippet.fqnsReferenced`. The only difference is that locally I removed the `monocdk` assembly and the fingerprinter takes in all assemblies at initialization. But I fail to see how that impacts anything here.

A bit of a cop out, but I think its unnecessary. We _cant_ translate infused examples, so we might as well take what we've got. Thoughts?

rix0rrr: I'm surprised as well, but I agree with the solution. It's fine like this.

rix0rrr (approved):

PR #3298, opened 2021-12-25
chore: merge dependabot with mergify

iliapolo:
Currently, we are (trying) to merge dependabot PR's by automatically approving them with a comment saying `@dependabot squash and merge`.

The problem is that the entity approving the PR, which is the default github token, doesn't have write permissions to the repo, resulting in:

Screen Shot 2021-12-25 at 2 42 38 PM

This might be fixable with granting the token in this specific job with `content: write` permissions. 
However, dependabot doing the merge also has other downsides:

- No observability. Its impossible to understand if dependabot is actually doing something, or what PR's are scheduled for merge. This is as opposed to mergify, where we can always look at the merge queue.
- I've noticed that sometimes dependabot will either ignore or take a long time rebasing the PR. Since there is no observability, its hard to understand what is happening. 
- Fairly complex workflow setup. To make this work properly, we need several workflows to be explicitly aware of dependabot. As opposed to the proposed solution, where only the labeler needs to know about dependabot. 

Instead, I propose to simply let mergify handle these PR's as all other PR's. 

The only potential problem with this is that once mergify rebases the PR, dependabot will no longer take any action on it, including fixing merge conflicts that might arise. However, I believe these occurrences were more prolific when dependabot did our npm upgrades, but will be rare in the current form where it only updates specific non npm dependencies that should cause conflicts. 

Let's give this a try, if we see it doesn't workout we'll consider alternatives. 

---

### DIFF
--- a/.github/workflows/auto-approve.yml
+++ b/.github/workflows/auto-approve.yml
@@ -12,44 +12,11 @@ on:
 
 jobs:
   auto-approve:
-    if: contains(github.event.pull_request.labels.*.name, 'pr/auto-approve') || github.actor == 'dependabot[bot]' || github.actor == 'dependabot-preview[bot]'
-
+    if: contains(github.event.pull_request.labels.*.name, 'pr/auto-approve')
     permissions:
       pull-requests: write
     runs-on: ubuntu-latest
-
     steps:
-      # Check whether the PR needs an automated approval or not. We will only add an automated
-      # approval if the PR currently has no reviews on it. Dismissed reviews are ignored in the
-      # context of this check. This way, the automated workflow will not do anything if a manual
-      # review has already been done for the PR (implicit opt-out).
-      - uses: actions/github-script@v5.0.0
-        id: needs-approving
-        with:
-          script: |-
-            const { issue: { number: pull_number }, repo: { owner, repo }  } = context;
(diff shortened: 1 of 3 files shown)


.github/workflows/auto-approve.yml
```diff
@@ -12,44 +12,11 @@ on:
 
 jobs:
   auto-approve:
-    if: contains(github.event.pull_request.labels.*.name, 'pr/auto-approve') || github.actor == 'dependabot[bot]' || github.actor == 'dependabot-preview[bot]'
-
+    if: contains(github.event.pull_request.labels.*.name, 'pr/auto-approve')
```
iliapolo, line 15: I just reverted this workflow to its simple revision, before we even added special handling for dependabot

.github/workflows/pr-labeler.yml
```diff
@@ -0,0 +1,18 @@
+# Apply various labels on PRs
```
iliapolo, line 1: This labeler was mistakenly removed [here](https://github.com/aws/jsii/pull/3274). The label is needed regardless of the merge mechanism because we use this label in our monitoring automation.

iliapolo:
@otaviomacedo if we happen to still be around :)

MrArnoldPalmer (approved):

PR #3308, opened 2021-12-29
chore: move disclaimer logic from pacmak to rosetta

kaizen3031593:
We add disclaimers to the visible source when we are translating examples that do not compile.
Previously this logic was in `pacmak`, and the main reason this was the case was because that
was where we knew what language we were translating to (and could append the correct
comment token). `pacmak` shouldn't really know/change snippets, so this PR refactors that logic
into `rosetta` instead.

---


kaizen3031593:
I'm slightly (ok, very) worried about the large diff in the pacmak snapshots. They contain largely unimportant changes (irrelevant to what I was touching), mostly regarding the order of the properties in an object. Not sure why the order has changed but I am hoping someone with more knowledge can shine a light on whether this is okay to push through or something needs to happen.

packages/jsii-pacmak/lib/targets/dotnet/dotnetdocgenerator.ts
```diff
@@ -167,8 +165,10 @@ export class DotNetDocGenerator {
       example,
       TargetLanguage.CSHARP,
       enforcesStrictMode(this.assembly),
+      undefined,
+      true /* prefix disclaimer */,
```
rix0rrr, line 169: This argument should be an option to the `Rosetta` object, not to the individual calls to `translateSnippet`. It's an "operating mode" of the translator.

packages/jsii-pacmak/test/generated-code/__snapshots__/target-dotnet.test.ts.snap
```diff
@@ -3163,7 +3163,7 @@ namespace Amazon.JSII.Tests.CalculatorNamespace
         {
         }
 
-        [JsiiMethod(name: "abstractMethod", returnsJson: "{\\"type\\":{\\"primitive\\":\\"string\\"}}", parametersJson: "[{\\"name\\":\\"name\\",\\"type\\":{\\"primitive\\":\\"string\\"}}]")]
+        [JsiiMethod(name: "abstractMethod", returnsJson: "{\\"type\\":{\\"primitive\\":\\"string\\"}}", parametersJson: "[{\\"type\\":{\\"primitive\\":\\"string\\"},\\"name\\":\\"name\\"}]")]
```
rix0rrr: Well this sucks. None of this should have changed. Do you know what's up here?

packages/jsii-pacmak/test/generated-code/__snapshots__/target-java.test.ts.snap
```diff
@@ -5626,7 +5626,7 @@ package software.amazon.jsii.tests.calculator;
  * Calculator calculator = new Calculator();
  * calculator.add(5);
  * calculator.mul(3);
- * System.out.println(calculator.getExpression().getValue());
+ * System.out.println(calculator.expression.getValue());
```
rix0rrr, line 5629: Same here.

The new code is probably correct, but why did this change all of a sudden?

packages/jsii-rosetta/test/commands/transliterate.test.ts
```diff
@@ -139,9 +139,6 @@ export class ClassName implements IInterface {
           "markdown": "# README
 
       \`\`\`csharp
-      // This example was automatically transliterated.
```
rix0rrr, line 142: Why did these disclaimers disappear all of a sudden? Are we not compiling (in which case we should), or did we not switch on `prefixDisclaimer` (which we also should)

rix0rrr (changes requested):
Thanks. The snapshot changes are weird, huh?

packages/jsii-rosetta/lib/commands/transliterate.ts
```diff
@@ -244,7 +200,7 @@ function transliterateType(
       );
       const translation = rosetta.translateSnippet(snippet, language);
```
kaizen3031593, line 201: We need the `workingDirectory` here so that we can fixturize properly and find the snippet in `translateSnippet`, correct? Otherwise we would not find the snippet and fail with unknown snippet mode (I think).

packages/jsii-rosetta/lib/commands/transliterate.ts
```diff
@@ -244,7 +200,7 @@ function transliterateType(
       );
```
kaizen3031593, line 200: We need the `workingDirectory` here so that we can fixturize properly and find the snippet in `translateSnippet`, correct? Otherwise we would not find the snippet and fail with unknown snippet mode (I think).

packages/jsii-rosetta/lib/commands/transliterate.ts
rix0rrr: Fixturization is only necessary for translation, not for lookup

rix0rrr: But do verify my assumption please

rix0rrr (approved):
Conditional approval, I think we can get rid of the directory but please check and then merge once you're satisfied

packages/jsii-rosetta/lib/commands/transliterate.ts
kaizen3031593: https://github.com/aws/jsii/blob/6f6715e29f4eb8a302ab70f762fe4d260528e729/packages/jsii-rosetta/lib/tablets/key.ts#L9-L17

snippet lookup only based on visible source

PR #3309, opened 2021-12-30
feat: add configuration for Visual Studio Code

kichik:
---


kichik:
Out of all the different tags I tried, `node14` was the only one that worked. I tried `latest`, `1-buster-slim` and a few others.

RomainMuller:
> Out of all the different tags I tried, `node14` was the only one that worked. I tried `latest`, `1-buster-slim` and a few others.

The PR has `jsii/superchain:1-buster-slim` though? What issues did you encounter exactly?

kichik:
Mostly `EPERM` errors. I also get one when trying to build latest. This is what I used for building from source:
```
{
    "name": "Dev Container Definition - AWS JSII",
    "build": {
        "dockerfile": "superchain/Dockerfile",
        "context": ".",
        "args": {
            "BUILDPLATFORM": "linux/amd64",
            "TARGETPLATFORM": "linux/amd64",
            "BUILD_TIMESTAMP": "unknown",
            "REGISTRY": "docker.io/library",
            "COMMIT_ID": "head"
        }
    },
    "postCreateCommand": "yarn install && yarn build",
    "extensions": [
        "dbaeumer.vscode-eslint@2.1.5"
    ]
}
```

.devcontainer.json
```diff
@@ -0,0 +1,8 @@
+{
+    "name": "Dev Container Definition - AWS JSII",
+    "image": "jsii/superchain:1-buster-slim",
```
RomainMuller, line 3: If I understand the documentation correctly, adding `remoteUser` might make `EPERM` problems go away (ref: https://code.visualstudio.com/remote/advancedcontainers/add-nonroot-user).

```suggestion
    "image": "jsii/superchain:1-buster-slim",
    "remoteUser: "superchain",
```

kichik:
Thanks for the push. I was able to finally track down one of the more persistent access denied issues. It was unable to gather Docker context because it was trying to copy the `.env` folder which was owned by root. That's just a Python virtual environment created as part of the build process. I added it to `.dockerignore` along with some others as even the test image that copy them doesn't use them.

So I am now finally able to build the container automatically with VScode instead of using an old one that possibly doesn't match. It especially doesn't match because I'm doing all this to get Ruby support in.

But... it still won't install :(
I tried `remoteUser`, `containerUser`, and none of it and I still get the following when it runs `yarn install`:

```
error An unexpected error occurred: "EPERM: operation not permitted, chmod '/workspaces/jsii/packages/@jsii/dotnet-runtime/node_modules/.bin/jsii-runtime'".
```

It seems to be a symlink pointing to a file that's part of the repo. That file belongs to `root`. So I ended up just letting it all run as `root` and it's finally working end to end.

RomainMuller (approved):

Issue #2819, opened 2021-05-04
(go): downcasting support


MrArnoldPalmer:
Go interface types cannot be downcasted to more specific types using normal go casting functionality. There may be some reflection able to achieve this but its not readily apparent if so and should be documented if it is indeed possible.

Example:
```go
	topic := awssns.NewTopic(stack, jsii.String("MyTopic"), &awssns.TopicProps{
		DisplayName: jsii.String("MyCoolTopic"),
	})

        l1node := topic.Node().DefaultChild()
	l1 := l1node.(awssns.CfnTopic)
```
error:
```bash
panic: interface conversion: *awscdk.jsiiProxy_IConstruct is not awssns.CfnTopic: missing method AddDeletionOverride

goroutine 1 [running]:
main.NewGoDowncastingStack(0x7ff3cf55a0a0, 0xc0000a6420, 0x709ffb, 0x12, 0xc00023ff38, 0xc00007c748, 0x3943d80)
        /home/mvaline/dev/cdk-tests/go-downcasting/go-downcasting.go:31 +0x225
main.main()
        /home/mvaline/dev/cdk-tests/go-downcasting/go-downcasting.go:41 +0xbd
exit status 2
Subprocess exited with error 1
```

There may be a changes to code generation that would allow the above to work but more investigation is needed.

xrn:
Hi, am I right that my issue https://github.com/aws/jsii/issues/2798 is related to this?

chlunde:
A quick chain of hacks :)

I did not want to understand/look at `castAndSetToPtr` (or any of the other code), so this is just a POC:

```go
func (c *Client) CastAndSetToPtr(ptr interface{}, data interface{}) {
	ptrVal := reflect.ValueOf(ptr).Elem()
	dataVal := reflect.ValueOf(data)

	if data, ok := data.(map[string]interface{}); ok {
		if ref, ok := data["$jsii.byref"]; ok { // uhm
			typRef := strings.Split(ref.(string), "@")[0] // oof
			typ, ok := c.Types().FindType(api.FQN(typRef))

			if ok && typRef == "aws-cdk-lib.aws_ec2.CfnLaunchTemplate" {
				fmt.Println(typ, typRef)
				n := reflect.New(typ)
				c.castAndSetToPtr(n.Elem(), dataVal)
				fmt.Println(n)
				ptrVal.Set(n.Elem())
				return
			}
		}
	}

	c.castAndSetToPtr(ptrVal, dataVal)
}
```

[...]

PR #3316, opened 2022-01-04
feat(go): add UnsafeCast function

RomainMuller:
The `UnsafeCast` function can be used to forcefully convert from one
type of jsii proxy value (including `interface{}`) to another jsii
interface (i.e: a class or interface instance).

If performs a "clean" cast when possible, and creates a new, aliased
proxy otherwise. It is the user's responsibility to ensure they are
passing the correct arguments into the function.

Fixes https://github.com/aws/jsii/issues/2819

---


packages/@jsii/go-runtime/jsii-runtime-go/cast.go
```diff
@@ -0,0 +1,53 @@
...
+		rinto.Set(null)
+		return
+	}
+	// Interfaces may present as a pointer to an implementing struct, and that's fine...
+	if rfrom.Kind() != reflect.Interface && rfrom.Kind() != reflect.Ptr {
+		panic(fmt.Errorf("First argument to UnsafeCast must be an interface value. Received %s", rfrom.Type().String()))
+	}
+
+	// If rfrom can be directly converted to rinto, just do it.
+	if rfrom.CanConvert(rinto.Type()) {
```
MrArnoldPalmer, line 37: Looks like `CanConvert` is a new addition in v1.17. I'm fine with bumping JSII minimum if this is a significant convencience (since we are considering a bump to 1.18 for generics anyway)

MrArnoldPalmer (changes requested):
In addition it feels like we should add a test case to the compliance suite. wdyt?

RomainMuller:
> In addition it feels like we should add a test case to the compliance suite. wdyt?

Good call. It resulted in discovery of an existing runtime bug (would have been impossible to receive "Object@" references in go, as there was no proxy initializer registered for that), and missing kernel features for this to actually work...

MrArnoldPalmer (approved):

PR #3317, opened 2022-01-04
chore(rosetta): change how python shows declared types and include module name

kaizen3031593:
Currently, Python type declarations are rendered as such:

```python
# get_book_handler is of type Function
# get_book_integration is of type LambdaIntegration


get_book_integration = apigateway.LambdaIntegration(get_book_handler,
    content_handling=apigateway.ContentHandling.CONVERT_TO_TEXT,  # convert to base64
    credentials_passthrough=True
)
```

This PR adds information on what module the type came from (if possible), and changes the way these type declarations are shown:

```python
# get_book_handler: lambda.Function
# get_book_integration: apigateway.LambdaIntegration


get_book_integration = apigateway.LambdaIntegration(get_book_handler,
    content_handling=apigateway.ContentHandling.CONVERT_TO_TEXT,  # convert to base64
    credentials_passthrough=True
)
```

---

### DIFF
--- a/packages/jsii-rosetta/lib/languages/python.ts
+++ b/packages/jsii-rosetta/lib/languages/python.ts
@@ -501,7 +501,7 @@ export class PythonVisitor extends DefaultVisitor<PythonLanguageContext> {
         (node.initializer && context.typeOfExpression(node.initializer));
 
       const renderedType = type ? this.renderType(node, type, context, fallback) : fallback;
-      return new OTree(['# ', context.convert(node.name), ' is of type ', renderedType], []);
+      return new OTree(['# ', context.convert(node.name), ': ', renderedType], []);
     }
 
     return new OTree([context.convert(node.name), ' = ', context.convert(node.initializer)], [], {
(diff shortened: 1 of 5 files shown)


packages/jsii-rosetta/lib/languages/python.ts
```diff
@@ -735,7 +735,8 @@ export class PythonVisitor extends DefaultVisitor<PythonLanguageContext> {
         case 'list':
           return `list of ${doRender(jsiiType.elementType)}`;
         case 'namedType':
-          return jsiiType.name;
+          // in this case, the fallback will hold more information than jsiiType.name
+          return fallback;
```
kaizen3031593, line 739: I don't love this, but the information we want is in `fallback`, which comes from `node.type.getText()`. I fiddled around with things to see if I could somehow isolate the module name from some other variable, but there isn't anything besides `getText()` to my knowledge.

rix0rrr: Why don't we read, interpret and format `jsiiType.fqn` ourselves?

rix0rrr (approved):

Issue #2688, opened 2021-03-14
go: unable to reuse instances between child/parent interfaces


iliapolo:
## :bug: Bug Report

When a TS child instance is **first** returned to the guest as a parent interface, the same instance cannot be returned as the child interface. 

### Affected Languages

- [ ] `TypeScript` or `Javascript`
- [ ] `Python`
- [ ] `Java`
- [ ] .NET (`C#`, `F#`, ...)
- [X] Golang

### General Information
* **JSII Version: ALL**
* **Platform: ALL**

### What is the problem?

Consider the following TS code:

https://github.com/aws/jsii/blob/45b527c0b2f35a09b715c1a6c5940ec0578007fb/packages/jsii-calc/lib/compliance.ts#L2214-L2225

Where:

https://github.com/aws/jsii/blob/45b527c0b2f35a09b715c1a6c5940ec0578007fb/packages/jsii-calc/lib/compliance.ts#L2233

The following Golang code will work:

```go
provider := calc.NewAnonymousImplementationProvider();
provider.ProvideAsClass(); // works - returning in instance of `Implementation`
```

But this subsequent invocation will fail:

```go
provider.ProvideAsInterface(); // fails - trying to assign the same instance of `Implementation` to `IAnonymouslyImplementMe`
```

```console
test panicked: reflect.Set: value of type *jsiicalc.jsiiProxy_Implementation is not assignable to type jsiicalc.IAnonymouslyImplementMe
```

Note that if the invocation order is reverse, it works. Because its now trying to assign a child to a parent, and not vise-versa.
Also, if the TS implementation returns a new instance on each `Provide` call, it will also work and since it won't reuse the previous instance.

PR #3321, opened 2022-01-06
fix(go): unable to reuse instances between child/parent interfaces

RomainMuller:
The go runtime was not tracking aliases in a way that enabled it to
return an alternate proxy to an object instance when that had previously
been returned as a different, singular interface.

This change keeps tack of the "first ever" registered instance against a
given InstanceID (as this is likely the one that was natively
constructed), but also to all proxy aliases to the same InstanceID, so
that it can later return an existing proxy that can be converted to a
desired type.

Fixes https://github.com/aws/jsii/issues/2688

---


RomainMuller:
Marked as draft because this builds on top of https://github.com/aws/jsii/pull/3316 (will mark as "ready for review" once that has shipped).

MrArnoldPalmer (approved):

PR #3339, opened 2022-01-14
fix(kernel): kernel's private object annotations are enumerable

RomainMuller:
The jsii kernel needs to attach metadata to instances that are sent
across the process boundary. While these properties use symbol names to
eliminate the risk for collisions with user-defined properties, they
were inadvertently set as `enumerable`, causing those to show in outputs
of `JSON.stringify` and to be copied over as part of spat expressions
(e.g: `{...obj}`), which resulted in the new value sharing the same
identity as the previous one when sent over the wire.

Caused: https://github.com/aws/aws-cdk/issues/17876

---

### DIFF
--- a/packages/@jsii/kernel/lib/objects.ts
+++ b/packages/@jsii/kernel/lib/objects.ts
@@ -52,9 +52,35 @@ type ManagedObject = {
 };
 
 function tagObject(obj: unknown, objid: string, interfaces?: string[]) {
-  const managed = obj as ManagedObject;
-  managed[OBJID_SYMBOL] = objid;
-  managed[IFACES_SYMBOL] = interfaces;
+  const privateField: Omit<PropertyDescriptor, 'value' | 'get' | 'set'> = {
+    // Make sure the filed does not show in `JSON.stringify` outputs, and is not
+    // copied by splat expressions (`{...obj}`), as this would be problematic.
+    // See https://github.com/aws/aws-cdk/issues/17876 for an example of the
+    // consequences this could have.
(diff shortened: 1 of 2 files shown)


packages/@jsii/kernel/lib/objects.ts
```diff
@@ -52,9 +52,35 @@ type ManagedObject = {
 };
 
 function tagObject(obj: unknown, objid: string, interfaces?: string[]) {
-  const managed = obj as ManagedObject;
-  managed[OBJID_SYMBOL] = objid;
-  managed[IFACES_SYMBOL] = interfaces;
+  const privateField: Omit<PropertyDescriptor, 'value' | 'get' | 'set'> = {
+    // Make sure the filed does not show in `JSON.stringify` outputs, and is not
```
njlynch, line 56:
```suggestion
    // Make sure the field does not show in `JSON.stringify` outputs, and is not
```

packages/@jsii/kernel/lib/objects.ts
```diff
@@ -104,6 +130,16 @@ export class ObjectTable {
         for (const iface of existingRef[api.TOKEN_INTERFACES] ?? []) {
           allIfaces.add(iface);
         }
+        // Note - obj[INTERFACES_SYMBOL] should already have been delcared as a
```
njlynch, line 133:
```suggestion
        // Note - obj[INTERFACES_SYMBOL] should already have been declared as a
```

njlynch (approved):

PR #3350, opened 2022-01-24
feat(python): support inheritance-based interface implementation

RomainMuller:
The code generator used to represent jsii interfaces as python
`Protocols`, however this mechanism is normally intended to allow
structural typing in python. Python `Protocol` types are declared using
a custom metaclass, which makes them tricky to inherit from without
running into metaclass conflicts (especially in cases where multiple
inheritance involving a `Protocol` and a non-`Protocol` base class).

The jsii runtime does not perform structural typing, and hence
interfaces are better modelled as abstract base classes that only
declare abstract members.

This, together with a minor modification to the `@jsii.interface`
decorator, allows interfaces to be "implemented" by simply adding them
to the implementor's inheritance chain, as is "natural" to do in Python
in such cases.

The "legacy" `@jsii.implements` approach to implementing interfaces is
no longer recommended, however it is still supported as it is necessary
when dealing with libraries generated by older versions of
`jsii-pacmak`. A Python warning will be emitted when the decorator is
used with interfaces that should instead be inherited.

---


RomainMuller:
I am leaving this as a draft until I was able to validate that this does not have surprising consequences on a mixed codebase (i.e: `aws-cdk` built with the updated `jsii-pacmak`, depending on `constructs` built with the older `jsii-pacmak`).

eladb (commented):
So much nicer!

packages/jsii-pacmak/lib/targets/python.ts
```diff
@@ -1409,15 +1404,18 @@ class Class extends BasePythonClassType implements ISortableType {
...
-    params.push(`jsii_type="${this.fqn}"`);
-
-    return params;
+    return [
+      ...this.bases.map((b) =>
+        toTypeName(b).pythonType({ ...context, typeAnnotation: false }),
+      ),
+      ...this.interfaces.map((i) =>
+        toTypeName(i).pythonType({ ...context, typeAnnotation: false }),
+      ),
```
RomainMuller, line 1415: I realized this particular part of the change would break if the implemented interface is from a library that was generated with the `typing_extensions.Protocol` metaclass.

Not putting this change in means the type checkers will not accept instances of classes where implementations of interfaces are expected (because `typing_extensions.Protocol` is what causes structural typing to apply).

I need to find a strategy to be backwards-compatible here, either by conditionally extending interfaces IIF they are not `typing_extensions.Protocol` types, or by making type-checkers understand the interface is implemented implicitly in all cases.

RomainMuller:
Okay looks like this is working at this stage. Mypy is not reporting any type checking errors anymore at this stage, and a cursory test appears to work (still got to make something a little more meaningful).

MrArnoldPalmer:
added `pr/do-not-merge` after discussing possible breakage when an existing construct library depends on a package that then will be generated with the new python code gen. We only explicitly built for the reverse case when a newly generated library depends on one generated with the old style inheritance.

packages/@jsii/python-runtime/src/jsii/_runtime.py
```diff
@@ -143,14 +148,18 @@ def deco(cls):
 
 def interface(*, jsii_type: str) -> Callable[[T], T]:
     def deco(iface):
+        # Un-set __jsii_class__ as this is an interface, and not a class.
+        iface.__jsii_class__ = None
```
comcalvi, line 152: Is there a way to not set this at all?

packages/jsii-pacmak/lib/targets/python.ts
```diff
@@ -2756,6 +2791,52 @@ class PythonGenerator extends Generator {
...
+   * This is useful as Python will not manage to create a consistent MRO in case the same interface
+   * is both directly and transitively extended (as a consistent MRO would then require that
+   * interface to be at two different locations in the list).
+   *
+   * @param fqns a possibly empty set of interface FQNs to de-duplicate.
+   * @param baseFqns the list of base classes that may also implement interfaces
+   *
+   * @returns the de-duplicated list of interface types.
+   */
+  private deduplicatedInterfaces(
```
comcalvi, line 2805: This is really nice, good catch!

comcalvi (commented):
I'm not sure why this hasn't been reviewed in a while, it looks really good. I have only two questions

packages/jsii-pacmak/lib/targets/python.ts
```diff
@@ -636,10 +638,7 @@ abstract class BaseMethod implements PythonBase {
     }
 
     if (decorators.length > 0) {
-      // "# type: ignore[misc]" needed because mypy does not know how to check decorated declarations
```
comcalvi, line 639: Is this removed because of the PR that introduces pyright?

RomainMuller: It appears mypy is now able to check most of these.

polothy:
Is it possible to pick this up again? The IDE warnings/errors of types not matching in Python is not delightful 😄

PR #3362, opened 2022-02-01
fix(pacmak): greatly reduce go code-gen memory footprint

RomainMuller:
The go code generator was previously holding multiple copies of the
`RootModule` objects for dependencies (one for the generated package's
own `RootModule`, and one for each submodule that uses a type from that
dependency), including their entire type hierarchy, methods and
properties.

On large packages, especially with many submodules, this results in an
extremely large resident set size. Pathological cases have required
heaps as big as 6 GiB.

De-duplicating the `RootModule` instances, and not retaining `GoTypeRef`
instances (which are stateless, and cheap to construct), allows reducing
those pathological modules to only occupy around 300 to 400 MiB worth of
heap, which is a very significant improvement.

The reduction of duplication also has significant performance benefits
on affected code-bases. The go bindings generate significantly faster on
top of needing significantly less memory.

---

### DIFF
--- a/packages/jsii-pacmak/lib/targets/go/package.ts
+++ b/packages/jsii-pacmak/lib/targets/go/package.ts
@@ -324,7 +324,10 @@ export class RootPackage extends Package {
   // This cache of root packages is shared across all root packages derived created by this one (via dependencies).
   private readonly rootPackageCache: Map<string, RootPackage>;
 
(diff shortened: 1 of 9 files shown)


MrArnoldPalmer (approved):
🔥

ansgarm:
Thank you so much 🚀

PR #3364, opened 2022-02-02
chore: deprecate Python 3.6 support

RomainMuller:
Python 3.6 has reached end-of-life at the end of December 2021, and new
`pip` releases no longer support it.

Check for python versions at load of the `jsii` runtime for Python, and
emit a warning if Python < 3.7 is detected. Additionally, change the
`pip` build-tool requirement to be `pip ~= 22.0` if Python >= 3.7 is
used.

---


otaviomacedo (approved):

rix0rrr (changes requested):
Should we not be using https://docs.python.org/3/library/warnings.html ?

RomainMuller:
> Should we not be using [docs.python.org/3/library/warnings.html](https://docs.python.org/3/library/warnings.html) ?

I did not know about that :)

rix0rrr (approved):

PR #3368, opened 2022-02-07
fix: remove the "comments rewriter"

RomainMuller:
The comments rewriter was used to prefix documentation comments with the
stability level of the associated API (e.g: `(experimental)` or
`(deprecated)`). This was introduced to palliate insufficient IDE
support for `@deprecated` and the mixing of `@stable` and
`@experimental` APIs in CDK v1.

Since then, the TypeScript language server has added support for
`@deprecated` IDE features, and CDK v2 moved all `@experimental` APIs to
separate packages, removing the risk for people to inadvertently take
experimental dependencies.

In profiling, this feature alone was attributed between 50% and 60% of
the total time spent compiling `aws-cdk-lib`. The impact is likely
comparable on other large construct libraries.

The benefits from this feature are not sufficient to justify the cost to
make it happen. Instead of trying to optimize it (which could prove
impossible), decision was made to simply drop it.

---


packages/jsii-calc/test/assembly.jsii
```diff
@@ -761,7 +761,7 @@
         {
           "docs": {
             "stability": "stable",
-            "summary": "(deprecated) String representation of the value."
```
eladb, line 764: So many trees are going to be saved!

eladb: Could be cool to calculate the negative carbon footprint of this change

eladb (approved):

packages/jsii-calc/test/assembly.jsii
RomainMuller: I agree but don't you try to nerd snipe me into sinking days or weeks into this. I'd first need to have representative telemetry for "cumulative compile times" + roughly what kind of hardware is in use (to estimate average power consumption of that build), and a ballpark idea of where that is geographically (to estimate the carbon intensity of the power supply there).

But I'd lie if your initial reaction to this did not make me think "oh jeez, it'd be sweet if the PRs had an automated comment telling you what the impact of the change is on the carbon footprint of the codebase"...

PR #3412, opened 2022-03-07
fix(jsii): fail compilation when two or more enum members have same val

### LINKED ISSUE #2782: Enums with duplicate values dropped

yuth:
If two enum members have the same value, only the first one will be retained.

This is a bit of an issue as we are renaming enum members: the named version will never appear in the assembly, and so not work over jsii.

What's worse, if we deprecate-and-strip the original one, neither of the enum members will appear.

Addressing the issue by failing the compilation by adding validation for enum values

Fixes #2782.
---

### DIFF
--- a/packages/jsii/lib/assembler.ts
+++ b/packages/jsii/lib/assembler.ts
@@ -1744,6 +1744,9 @@ export class Assembler implements Emitter {
       return Promise.resolve(undefined);
     }
 
+    // check the enum to see if there are duplicate enum values
+    this.assertNoDuplicateEnumValues(decl);
+
     this._warnAboutReservedWords(symbol);
 
     const flags = ts.getCombinedModifierFlags(decl);
@@ -1786,6 +1789,60 @@ export class Assembler implements Emitter {
     );
 
     return Promise.resolve(jsiiType);
+  }
+
+  private assertNoDuplicateEnumValues(decl: ts.EnumDeclaration): void {
+    type EnumValue = {
+      name: string;
+      value: string;
+      decl: ts.DeclarationName | undefined;
+    };
+
+    const enumValues = decl.members
+      .filter((m) => m.initializer)
+      .map((member): EnumValue => {
+        return {
+          value: member.initializer!.getText(),
+          name: member.name.getText(),
+          decl: ts.getNameOfDeclaration(member),
+        };
+      });
+
+    const hasDuplicateEnumValues = enumValues.some(
+      (val, _, arr) => arr.filter((e) => val.value === e.value).length > 1,
+    );
+
+    if (hasDuplicateEnumValues) {
+      const enumValueMap = enumValues.reduce<Record<string, EnumValue[]>>(
+        (acc, val) => {
+          if (!acc[val.value]) {
+            acc[val.value] = [];
+          }
+          acc[val.value].push(val);
+          return acc;
+        },
+        {},
+      );
+      for (const duplicateValue of Object.keys(enumValueMap)) {
+        if (enumValueMap[duplicateValue].length > 1) {
+          const err = JsiiDiagnostic.JSII_1004_DUPLICATE_ENUM_VALUE.create(
+            enumValueMap[duplicateValue][0].decl!,
+            `Value ${duplicateValue} is used for multiple enum values: ${enumValueMap[
+              duplicateValue
+            ]
+              .map((e: any) => e.name)
+              .join(', ')}`,
+          );
+          for (let i = 1; i < enumValueMap[duplicateValue].length; i++) {
+            err.addRelatedInformation(
+              enumValueMap[duplicateValue][i].decl!,
+              'Same value assigned here',
+            );
+          }
+          this._diagnostics.push(err);
+        }
+      }
+    }
   }
 
   /**
--- a/packages/jsii/test/enums.test.ts
+++ b/packages/jsii/test/enums.test.ts
@@ -57,3 +57,16 @@ test('enums can have a mix of letters and number', async () => {
     { name: 'IB3M' },
   ]);
 });
+
+test('enums with the same assigned value should fail', async () => {
+  await expect(() =>
+    sourceToAssemblyHelper(`
+    export enum Foo {
+      BAR = 'Bar',
+      BAR_DUPE = 'Bar',
+      BAZ = 'Baz',
+      BAZ_DUPE = 'Baz',
+    }
+  `),
+  ).rejects.toThrowError('There were compiler errors');
+});
--- a/packages/jsii/lib/jsii-diagnostic.ts
+++ b/packages/jsii/lib/jsii-diagnostic.ts
@@ -292,6 +292,12 @@ export class JsiiDiagnostic implements ts.Diagnostic {
     name: 'typescript-restrictions/unsupported-type',
   });
 
+  public static readonly JSII_1004_DUPLICATE_ENUM_VALUE = Code.error({
+    code: 1004,
+    formatter: (messageText) => messageText,
+    name: 'typescript-restrictions/duplicate-enum-value',
+  });
+
   //////////////////////////////////////////////////////////////////////////////
   // 2000 => 2999 -- RESERVED
 


skinny85:
I don't think this change will work for CDK, where we use multi-value enums often (like [here](https://github.com/aws/aws-cdk/blob/19ead77c6470c0aca4900c4ee200a7c4e66ebbed/packages/%40aws-cdk/aws-ec2/lib/vpc.ts#L167-L193)).

RomainMuller:
> I don't think this change will work for CDK, where we use multi-value enums often (like [here](https://github.com/aws/aws-cdk/blob/19ead77c6470c0aca4900c4ee200a7c4e66ebbed/packages/%40aws-cdk/aws-ec2/lib/vpc.ts#L167-L193)).

I think the "right" way to fix this would actually be to have CDK use different values for each enum constant, and have code that gracefully normalizes downstream. That's however a little to easy to make mistakes with, and I wonder if we could come up with an `eslint` rule that makes this a little safer?

packages/jsii/lib/assembler.ts
```diff
@@ -1788,6 +1791,60 @@ export class Assembler implements Emitter {
...
+          acc[val.value].push(val);
+          return acc;
+        },
+        {},
+      );
+      for (const duplicateValue of Object.keys(enumValueMap)) {
+        if (enumValueMap[duplicateValue].length > 1) {
+          const err = JsiiDiagnostic.JSII_1004_DUPLICATE_ENUM_VALUE.create(
+            enumValueMap[duplicateValue][0].decl!,
+            `Value ${duplicateValue} is used for multiple enum values: ${enumValueMap[
```
RomainMuller, line 1830: You'll want to move the template string up to the `JsiiDiagnostic` entry. The `formatter` function is the mechanism by which we aim to guarantee that the error message is always consistent for a given error class (in the future, that same error could be emitted from another place, and then we'd need to make sure the error messages are in synch).

packages/jsii/lib/assembler.ts
```diff
@@ -1788,6 +1791,60 @@ export class Assembler implements Emitter {
...
+              duplicateValue
+            ]
+              .map((e: any) => e.name)
+              .join(', ')}`,
+          );
+          for (let i = 1; i < enumValueMap[duplicateValue].length; i++) {
+            err.addRelatedInformation(
+              enumValueMap[duplicateValue][i].decl!,
+              'Same value assigned here',
```
RomainMuller, line 1839: No biggie - but I prefer phrasing those as "The conflicting declaration is here", or something similar. I.e: try to match the style of other related information.

packages/jsii/test/enums.test.ts
```diff
@@ -57,3 +57,16 @@ test('enums can have a mix of letters and number', async () => {
     { name: 'IB3M' },
   ]);
 });
+
+test('enums with the same assigned value should fail', async () => {
```
RomainMuller, line 61: Good!

But can you also add a `negative` test case? Those are used to validate the diagnostic message is bound to the proper context & looks actionnable from the user perspective.

yuth:
> > > I don't think this change will work for CDK, where we use multi-value enums often (like [here](https://github.com/aws/aws-cdk/blob/19ead77c6470c0aca4900c4ee200a7c4e66ebbed/packages/%40aws-cdk/aws-ec2/lib/vpc.ts#L167-L193)).
> > 
> > 
> > I think the "right" way to fix this would actually be to have CDK use different values for each enum constant, and have code that gracefully normalizes downstream. That's however a little to easy to make mistakes with, and I wonder if we could come up with an `eslint` rule that makes this a little safer?
> 
> We can't do that, unfortunately, as that could be considered a breaking change.

My plan is to use the same value, but where it is duplicated, use a different value. So if  a customer is using the string literal, it would still work as expected. And where we are using the enum transform them into the value needed by `Cfn*` construct

```ts
export enum SubnetType {
  ISOLATED = 'Deprecated_Isolated',
  PRIVATE_ISOLATED = 'Isolated',
  PRIVATE = 'Deprecated_Private',
  PRIVATE_WITH_NAT = 'Private',
  PUBLIC = 'Public'
}
```
and then update the instantiation code 
```ts
// snip
    switch (subnetConfig.subnetType) {
        case SubnetType.PUBLIC:
          const publicSubnet = new PublicSubnet(this, name, subnetProps);
          this.publicSubnets.push(publicSubnet);
          subnet = publicSubnet;
          break;
        case SubnetType.PRIVATE_WITH_NAT:
        case SubnetType.PRIVATE:
          const privateSubnet = new PrivateSubnet(this, name, subnetProps);
          this.privateSubnets.push(privateSubnet);
          subnet = privateSubnet;
          break;
        case SubnetType.PRIVATE_ISOLATED:
        case SubnetType.ISOLATED:
          const isolatedSubnet = new PrivateSubnet(this, name, subnetProps);
          this.isolatedSubnets.push(isolatedSubnet);
          subnet = isolatedSubnet;
          break;
        default:
          throw new Error(`Unrecognized subnet type: ${subnetConfig.subnetType}`);
      }
```

skinny85:
That's not a bad plan. Making sure @rix0rrr sees this (I remember him discussing this at lengths with Niranjan, although I can't find which issue that discussion was in).

packages/jsii/lib/assembler.ts
yuth: ✅

packages/jsii/lib/assembler.ts
yuth: ✅

packages/jsii/test/enums.test.ts
yuth: ✅

rix0rrr (approved):

PR #3426, opened 2022-03-15
feat: drop internal functions from Deprecation Error stack trace

yuth:
JSII can throw Deprecation Error and fail compilation when a deprecated property,
enum or method is being used when `JSII_DEPRECATED` environment variable is set
to `fail`. The deprecation error is throw from `.warnings.jsii.js` generated file.
When nested prop of enum is deprecated, the stack trace will contain lot of methods
from `.warnings.jsii.js` and since the file is minimized, tools like Jest which try to be
helpful by printing the line where the error is thrown shows a wall of text without much useful information.

---


packages/jsii/test/deprecation-warnings.test.ts
```diff
@@ -492,9 +642,207 @@ describe('Call injections', () => {
...
+      undefined,
+      { addDeprecationWarnings: true },
+    );
+    const source = jsFile(compilation);
+
+    const context = createVmContext(compilation);
+    try {
+      vm.runInContext(source, context, { filename: 'index.js' });
+      // The above line should have resulted in a DeprecationError being thrown
+      expect(null).toBeInstanceOf(Error);
```
Chriscbr, line 696: Not sure if I understand this line? Is the intention here to intentionally fail the test? Maybe something like this could make it clearer:
```suggestion
      throw new Error("expected a DeprecationError from running the transliterated snippet")
```
Or maybe "expect(true).toEqual(false)"

packages/jsii/test/deprecation-warnings.test.ts
```diff
@@ -503,7 +851,54 @@ function jsFile(result: HelperCompilationResult, baseName = 'index'): string {
...
+        JSII_DEPRECATED: 'fail',
+      },
+    },
+    // Bringing in a "fake" require(id) function that'll resolve relative paths
+    // to files within the compilation output, and module names using the
+    // regular require. When loading a file that was part of the compiler output,
+    // this emulates the situation that would be if the file had been through a
+    // bundler by turning all sequences of white spaces (new line included) into
+    // single spaces.
+    require: (id: string) => {
```
Chriscbr, line 875: Is there a better way?

Chriscbr (approved):

packages/jsii/test/deprecation-warnings.test.ts
RomainMuller: The `expect(null)` check guarantees that the test will be marked as failed if no exception was thrown. If we move to the `throw` style, it'll trigger the catch and will happily update the snapshot when running with `-u`, which I think is not desirable behavior.

I kind of agree with your sentiment here, but this "hack" is just a bit safer.

Ideally - I'd have wanted to do `expect(() => vm.runInContext(...)).toThrow`, but that'll only verify the error message, not the stack trace (which is what I care about here).

packages/jsii/test/deprecation-warnings.test.ts
RomainMuller: That's the sort of dance you have to do when you use VM contexts in `node`. The context keys are global values that'll be available within the sandbox; which otherwise only has access to ES primordials... So you need to give it "a" require function, and that function is a pretty standard implementation here...

All those APIs are stable enough that they won't break under us when we upgrade to a new node version, too.

PR #3427, opened 2022-03-15
chore: update gitigonre to ignore build, coverage and jsii-outdir

yuth:
---


MrArnoldPalmer (approved):
I've only seen an issue with this when using certain VScode plugins an opening the project as the root dir, that does some auto running of `build` commands for plugins. Mainly the VScode jest plugin. Regardless I don't think this should cause an issue.

PR #3445, opened 2022-03-26
chore: replace deprecated String.prototype.substr()

CommanderRoot:
[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.

---


RomainMuller:
I'm cool with the change, but we'd probably want to have an `eslint` rule that verifies we're not adding new occurrences back in there...

If you're not aware of one, I can go on the hunt...

RomainMuller:
@all-contributors add @CommanderRoot for code

CommanderRoot:
Took me a bit to figure out how to write the rule in `yaml` but I think i got it working. I got the code from [here](https://github.com/eslint/eslint/blob/8bb527660ffddc122204fb5bf59076687aaa78b8/packages/eslint-config-eslint/default.yml#L188)

RomainMuller (approved):

PR #3449, opened 2022-03-29
fix(rosetta): incorrect transliteration of map keys in python

### LINKED ISSUE #3448: CORS Error despite thoroughly reading CDK docs, CDK docs error in the public reference
### Affected Languages

- [ X] `Python`
- [ ] `Java`
- [ ] .NET (`C#`, `F#`, ...)
- [ ] `Go`

### General Information
* **JSII Version:** 
* **Platform:**     

### What is the problem?

The documentation about the AWS CDK for python when specifying the integration response, and method response wrongly put an extra under score on the "_access-_control-_allow-_origin", instead, the correct naming should be "Access-Control-Allow-Origin" which caused the customer's application not working even after throughly read the CDK docs.

I have attached the screen cap that customer shared with the ticket. The extra underscore could affect other parameter as well across the documentation.

### Verbose Log
[

RomainMuller:
The Python transliterator failed to properly account for maps when
transliterating keys, resulting in name-mangling being applied on these
when it should not have been.

This PR adds a new `inMap` context entry to track whether a property
assignment must be interpreted as a map key-value entry, or an actual
keyword/value mapping; and in that case circumvents the mangling.

Added a new test to verify that such map literals render appropriately
using a key similar to that which caused #3448.

Fixes #3448

---

### DIFF
--- a/packages/jsii-rosetta/lib/languages/python.ts
+++ b/packages/jsii-rosetta/lib/languages/python.ts
@@ -73,6 +73,13 @@ interface PythonLanguageContext {
   readonly inClass?: boolean;
 
   /**
+   * Whether the current property assignment is in the context of a map value.
+   * In this case, the keys should be strings (quoted where needed), and should
+   * not get mangled or case-converted.
+   */
+  readonly inMap?: boolean;
+
+  /**
    * If we're in a method, what is it's name
    *
    * (Used to render super() call.);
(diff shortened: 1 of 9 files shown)


packages/jsii-rosetta/lib/languages/python.ts
```diff
@@ -608,19 +622,18 @@ export class PythonVisitor extends DefaultVisitor<PythonLanguageContext> {
     node: ts.StringLiteral | ts.NoSubstitutionTemplateLiteral,
     _context: PythonVisitorContext,
   ): OTree {
-    const rawText = node.text;
-    if (rawText.includes('\n')) {
+    if (node.getText(node.getSourceFile()).includes('\n')) {
```
madeline-k, line 625: Whats this difference between this and node.text?

madeline-k (approved):

packages/jsii-rosetta/lib/languages/python.ts
RomainMuller: `node.text` is the JavaScript value of the string (in particular, with escape sequences replaced with their literal value, e.g: `\n` will be a "real" newline), whereas `node.getText()` will return the corresponding source code (which will have the escape sequences untouched).

The change here improves the situation a bit (avoid multi-lining literals that were not actually multi-line in the source), but it's also not perfect (probably good enough?) as it'll still cause some escapes to be replaced within multi-line literals).

PR #3453, opened 2022-03-31
perf: cache package.json and .jsiirc file lookup

yuth:
JSII loads `package.json` and `.jsiirc` file multiple times in the compilation
process. These files are loaded from disk everytime and are expensive operation.
Updated the code to cache the file in memort by memoizing the load file calls.

With this change the compilation time for `aws-cdk-lib` module came from 133 seccond
to 101 second

---


yuth:
I still need to investigate and implement cache eviction in `--watch` mode

packages/jsii/lib/assembler.ts
```diff
@@ -3240,18 +3238,8 @@ function isSingleValuedEnum(
...
-    return fs.readJson(filePath);
-  }
-  const parent = path.dirname(fromDir);
-  if (parent === fromDir) {
-    return undefined;
-  }
-  return findPackageInfo(parent);
+function findPackageInfo(fromDir: string): PackageJson | undefined {
+  return getPackageJson(fromDir);
 }
```
RomainMuller, line 3243: That's effectively the same as:

```suggestion
const findPackageInfo = getPackageJson;
```

You should probably just get rid of that and refactor usage to `getPackageJson`.

packages/jsii/lib/symbol-id.ts
```diff
@@ -1,9 +1,9 @@
 import { Assembly } from '@jsii/spec';
-import * as fs from 'fs-extra';
+import * as _ from 'lodash';
```
RomainMuller, line 2: Would rather not take a dependency on `lodash`. It's a very large dependency that is capable of WAY MORE than what you're using it for here... That means it's a huge vulnerability surface that we don't even need...

Memoization can be done relatively easily on the user-side... We don't need a dependency for that (`jsii-reflect` has a decorator-based solution for this you can draw inspiration from).

RomainMuller (changes requested):
That seems to be missing cache invalidation, which would likely be problematic in the context of `--watch`.

PR #3467, opened 2022-04-01
chore: make `jsii` fully synchronous

RomainMuller:
With a sample set of 1, I have detemined that there is about 30% tax
caused by asynchronous functions in the compiler. The `typescript`
compiler itself is fully synchronous, and we basically moved to do the
same, and this turns out reducing the amount of time wasted in the
runloop and in promise bookeeping.

Tested on `aws-cdk-lib`, it turns out to be reducing the compile time
from ~142 seconds to ~100 seconds.

---


MrArnoldPalmer:
That seems like an inordinate % of time spent just in promise/callback/event loop management. I wonder if the TS compiler being fully synchronous is for performance or design reasons. Even their fs calls when writing .d.ts and .js file output are sync? Either way it's quite interesting. I'd be curious to run comparisons across more libraries and different hardware configurations with some flamegraphs for a clearer picture of the difference.

RomainMuller:
> That seems like an inordinate % of time spent just in promise/callback/event loop management. I wonder if the TS compiler being fully synchronous is for performance or design reasons. Even their fs calls when writing .d.ts and .js file output are sync? Either way it's quite interesting. I'd be curious to run comparisons across more libraries and different hardware configurations with some flamegraphs for a clearer picture of the difference.

I reckon the TS maintainers wanted to turn the compiler async but it turned out to be a humongous task. I think the TS compiler itself spends enough time doing I/O that using async will yield performance improvements, whereas the `jsii` compiler performs mostly busy compute in JS, where async turns out to be mostly just overhead.

But all in all I agree this magnitude of a difference looks pretty surprising. I expected some, but not that much... It could be just the particular case I used magnifies that (the measurement protocol isn't very scientific here)...

yuth (approved):
LGTM

PR #3472, opened 2022-04-05
chore: drop support for Python 3.6 (EOL as of 2021-12-23)

RomainMuller:
Python 3.6 has been end-of-life since late december 2021, and an
increasing amount of our dependencies are dropping support for it as a
consequence.

We introduced a warning to users that are still using Python 3.6 with
release `v1.55.1` around March 16, 2022.

We are now effectively dropping support for Python 3.6, and introducing
testing for Python 3.10 releases.

Related: https://github.com/aws/jsii/commit/919f735778fe57047fba45f66a821a4abbe28afc

BREAKING CHANGE: `jsii-pacmak` and the libraries it generates for Python
targets now require a minimum Python version of 3.7, instead of 3.6
previously. We recommend users upgrade to the latest supported Python
release (Python 3.10).

---


RomainMuller:
### ⚠️ This PR should remain labeled with `pr/do-not-merge` at least until `2022-04-16` (at which point it'll have bee a full month since we released the first version of jsii that emits a deprecation warning to users still on Python 3.6).

MrArnoldPalmer (approved):
LGTM when we are rdy.

RomainMuller:
## ℹ️ Important Information

This PR removes some PR check builds that are currently required by the branch protection settings. This means this PR will not be merge-able by Mergify until the branch protection settings are updated to:

1. Remove the requirement for Python 3.6 builds
2. Add Python 3.10 builds to requirements (this PR has Mergify updated to wait for those before merging, but GitHub branch protection should also require them)

When this update is done, all existing PRs will fail to pass PR validation rules until they are updated to the new `master` HEAD.

PR #3478, opened 2022-04-11
fix(rosetta): attempts transliteraiton to unsupported languages

RomainMuller:
Rosetta attempted to perform transliteration to unsupported languages
when requested to, which would likely result in incorrect output, if not
a complete crash (due to missing required configuration).

This stops trying to emit transliterated assemblies for languages that
the original assembly does not support.

---


packages/jsii-rosetta/lib/commands/transliterate.ts
```diff
@@ -99,6 +100,12 @@ export async function transliterateAssembly(
       const now = new Date().getTime();
       // eslint-disable-next-line no-await-in-loop
       const result = await loadAssembly();
+
+      if (result.targets?.[targetName(language)] == null) {
+        // This language is not supported by the assembly, so we skip it...
```
MrArnoldPalmer, line 105: Wondering if we need some explicit output to tell callers that a language is unsupported?

RomainMuller: Yeah but we'd want that ONLY if the language was specifically requested...?

MrArnoldPalmer (approved):

PR #3491, opened 2022-04-19
fix(jsii): some submodules are not exported from `aws-cdk-lib`

rix0rrr:
This bug is due to a crazy interaction between the `publication`
module, the `__all__` array, and imports we might generate to do
type checking in a package that uses submodules.

Context
-------

Here's what you need to know:

- `__all__` is normally used to define the symbols and submodules
  that should be imported when a consumer does `from mymod import *`.
- `publication` is used to hide non-public symbols from library
  consumers. It will move all symbols that are NOT in `__all__` to
  a `_private` submodule at runtime (it's overloading `__all__` a
  little here, but this seems generally okay).
- We generate `imports` at the top of the file for the modules that are
  referenced in a module, so that we can use them in type annotations.

Example
-------

This is what the currently generated code for the `aws_cdk` module
(of CDK's v2 library) currently (roughly) looks like:

```python
import publication
from .cloud_assembly_schema import AmiContextQuery as _AmiContextQuery_74bf4b1b

__all__ = ['Annotations', 'App', ..., 'ValidationResults']

publication.publish()

from . import aws_s3
from . import cloud_assembly_schema
...
```

Given this setup, the following happens:

```
$ python3
Python 3.9.5 (default, Jun 27 2021, 10:29:41)
>>> import aws_cdk
>>> print(aws_cdk.aws_s3)
<module 'aws_cdk.aws_s3' from '.../aws_cdk/aws_s3/__init__.py'>
>>> print(aws_cdk.cloud_assembly_schema)
  File "<stdin>", line 1, in <module>
AttributeError: module 'aws_cdk' has no attribute 'cloud_assembly_schema'
```

Explanation
-----------

Why does this happen?

- `__all__` does not contain any submodules
- When `publication.publish()` is called, all symbols in the calling
  namespace that are missing from `__all__` are hidden.
- Crucially, this *does* include `cloud_assembly_schema` (because we
  imported that submodule to import some symbols for type checking
  purposes) but it doesn't include modules that haven't been imported
  yet.
- Most modules get imported *after* `publication.publish()` is called,
  and so are not subject to hiding anymore. This is why we never noticed
  this problem before.
- Every module only gets imported once, so even though we generate a
  second `from . import cloud_assembly_schema` line underneath
  `publication.publish()`, at that point the line doesn't do anything
  anymore.

Solution
--------

Potential solutions are:

- Add submodules to `__all__`.
- Remove the call to `publication.publish()`.
- Make a variant of `publication` that doesn't hide submodules.
- Don't generate actual imports for the type annotations.

This PR picks the first solution, as it seems simplest. The only
downside would have been that `import *` would then start importing all
of the CDKv2 library, but that's happening already anyway because of all
the `import`s we generate ourselves.

---

### DIFF
--- a/packages/jsii-pacmak/lib/targets/python.ts
+++ b/packages/jsii-pacmak/lib/targets/python.ts
@@ -1648,7 +1648,9 @@ class PythonModule implements PythonType {
     // we import as part of typechecking.
     const exportedMembers = [
       ...this.members.map((m) => `"${m.pythonName}"`),
-      ...this.modules.map((m) => `"${m.pythonName}"`),
+      ...this.modules
+        .filter((m) => this.isDirectChild(m))
+        .map((m) => `"${lastComponent(m.pythonName)}"`),
     ];
     if (this.loadAssembly) {
       exportedMembers.push('"__jsii_assembly__"');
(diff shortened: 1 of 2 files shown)


MrArnoldPalmer (approved):
Looks straightforward 👍

PR #3499, opened 2022-04-26
chore: automated node runtime deprecation warnings

RomainMuller:
This changes how the node release support information is modeled so that
the end-of-life dates are configured, so that `@jsii/check-node` can
automatically start warning a month ahead of a runtime's EOL date, so we
can drop support right on EOL day if needed wihtout taking users by
surprise.

This also includes tweaks of the messages that get printed out to STDERR
when an unsupported/untested/deprecated/end-of-life version is being
used, to increase clarity of provide more information about migration
options.

Finally, this adds node 18 (initial release `2022-04-19`) to the test
matrix.

---

### DIFF
--- a/packages/@jsii/check-node/src/constants.ts
+++ b/packages/@jsii/check-node/src/constants.ts
@@ -1,69 +1,159 @@
-export const enum SupportLevel {
-  /**
-   * Soft-deprecated, removed as of releases made starting at DEADLINE. This
-   * will output a yellow warning to the console, until the DEADLINE has been
-   * reached, at which point it'll be a SCARY RED WARNING.
-   */
-  DEPRECATED = 'deprecated',
-
-  /**
-   * Supported releases, nothing to see. No message will be output.
-   */
-  SUPPORTED = 'supported',
-
-  /**
-   * EOL releases are unsupported, so this will emit a SCARY RED WARNING.
-   */
-  END_OF_LIFE = 'end-of-life',
-
-  /**
-   * Not end-of-life, but also not supported. This is usually the case for early
-   * releases on a new major train, where some of the features are not available
-   * yet. These emit a SCARY RED WARNING, but the messaging is slightly
-   * different than that of END_OF_LIFE.
-   */
-  UNSUPPORTED = 'unsupported',
-
-  /**
-   * Those releases are newer than the current software package, so we have not
-   * tested with it. People may encounter all kinds of weird bugs, especially
-   * early in the release cycle... So we'll issue a yellow warning here.
-   */
-  UNTESTED = 'untested',
-}
+import { version } from 'process';
+import { Range, SemVer } from 'semver';
 
 /**
- * The deadline after which deprecated releases will move to
- * `SupportLevel.END_OF_LIFE`.
- */
-export const DEADLINE = '2021-09-01';
-/**
- * The DEADLINE, expressed in milliseconds since the epoch.
- */
-export const DEADLINE_EPOCH_MS = new Date(
-  `${DEADLINE}T00:00:00.000Z`,
-).getTime();
-
-/**
- * For each SemVer range, what is the support level for such versions of Node.
- * The ranges are checked in declaration order (so it is a good idea to keep the
- * left-open ranges at the beginning, and the right-open ranges at the end.
- * First match wins.
+ * The support information for a given node release range.
  *
- * Also, if you are setting a new entry to `SupportLevel.DEPRECATED`, do not
- * forget to also change the `DEADLINE` value.
+ * @see https://nodejs.org/en/about/releases/
  */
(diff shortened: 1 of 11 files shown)


RomainMuller:
## ℹ️ Note

With these changes, we will issue messaging that we may stop supporting Node 12 on the first release after `2022-04-30`, however since we won't have had a warning up for at least 30 days, we'd effectively not drop Node 12 support before `2022-05-30` (or later, depending on when this change gets released, and unless we are required to drop that release on security reasons).

I still decided to keep the post-EOL messaging as-is: once a release is EOL, it is no longer supported (although it might still work) as we may be required to pull it without warning in case a severe CVE is discovered within our intended 30 days notice period.

packages/@jsii/check-node/src/constants.ts
```diff
@@ -1,69 +1,161 @@
...
+export class NodeRelease {
   /**
-   * Soft-deprecated, removed as of releases made starting at DEADLINE. This
-   * will output a yellow warning to the console, until the DEADLINE has been
-   * reached, at which point it'll be a SCARY RED WARNING.
+   * How long before enf-of-life do we start warning customers? Expressed in
+   * milliseconds to make it easier to deal with JS dates.
    */
-  DEPRECATED = 'deprecated',
+  private static readonly DEPRECATION_WINDOW_MS = 30 * 86_400_000;
```
Chriscbr, line 14: Can we make this magic number less magical (perhaps define `const ONE_DAY = 86_400_000;` somewhere)?

packages/@jsii/check-node/src/constants.ts
```diff
@@ -1,69 +1,161 @@
...
+  public readonly deprecated: boolean;
+
+  /**
+   * If `true` denotes that this version of node has not been added to the test
+   * matrix yet. This is used when adding not-yet-released versions of node that
+   * are already planned (typically one or two years out).
+   *
+   * @default false
+   */
+  public readonly pending: boolean;
```
Chriscbr, line 111: (optional suggestion)
```suggestion
  public readonly notReleasedYet: boolean;
```

packages/@jsii/check-node/src/constants.ts
```diff
@@ -1,69 +1,161 @@
...
+   *          boolean indicating whether this version is known to be broken. If
+   *          the runtime does not correspond to a known node major, this
+   *          returns and `undefined` release object.
    */
-  END_OF_LIFE = 'end-of-life',
+  public static forThisRuntime(): {
+    nodeRelease: NodeRelease | undefined;
+    knownBroken: boolean;
+  } {
+    const semver = new SemVer(version);
```
Chriscbr, line 66: readability nit
```suggestion
    const semver = new SemVer(process.version);
```

packages/@jsii/check-node/src/constants.ts
```diff
@@ -1,69 +1,161 @@
...
-   * Those releases are newer than the current software package, so we have not
-   * tested with it. People may encounter all kinds of weird bugs, especially
-   * early in the release cycle... So we'll issue a yellow warning here.
+   * The date on which this release range starts to be considered end-of-life.
+   * Can also be `true` to denote that this is already end-of-life, without
+   * specifying a particular date.
    */
-  UNTESTED = 'untested',
-}
+  public readonly endOfLifeDate: Date | undefined;
```
Chriscbr, line 91: The docstring says this can be "true" but it's typed as `Date | undefined`. Bug?

packages/@jsii/check-node/src/constants.ts
```diff
@@ -1,69 +1,161 @@
...
+    new NodeRelease(20, { endOfLife: new Date('2026-04-30'), pending: true }),
+  ];
 
   /**
-   * EOL releases are unsupported, so this will emit a SCARY RED WARNING.
+   * @returns the `NodeRelease` corresponding to the version of the node runtime
+   *          executing this code (as provided by `process.version`), and a
+   *          boolean indicating whether this version is known to be broken. If
+   *          the runtime does not correspond to a known node major, this
+   *          returns and `undefined` release object.
```
Chriscbr, line 60:
```suggestion
   *          returns an `undefined` release object.
```

Chriscbr (approved):
LGTM - just a few nits so I approved with `do-not-merge` 👍

packages/@jsii/check-node/src/constants.ts
RomainMuller: Renamed to `untested` which is a more accurate description of the intent.

packages/@jsii/check-node/src/constants.ts
RomainMuller: No it's just I changed my mind twice here but I forgot to update the text...

PR #3503, opened 2022-04-28
feat: compiler benchmarks and gh action

MrArnoldPalmer:
Adds a private package @jsii/benchmarks which includes a basic command line application for running jsii benchmarks.

Adds a custom benchmark runner leveraging the NodeJS `perf_hooks` module to time function calls across multiple iterations and return averaged results.

Adds github action workflows to run benchmarks on PRs and compare performance to the target branch. This action will fail when a test suite is slower by 200%, but this threshold can be configured.

Adds to the gh-pages action workflow to run benchmarks and append new results to a file that is displayed as a graph on our docs site at the `/dev/bench` url. This will show benchmark suite results over time and allow us to track overall change in compiler performance across multiple commits.

Example of benchmark results comment https://github.com/MrArnoldPalmer/jsii/pull/417#commitcomment-72358212
Example of benchmark results graph (but only 1 result cause just testing) https://mrarnoldpalmer.github.io/jsii/dev/bench/

---


.github/workflows/gh-pages.yml
```diff
@@ -82,3 +82,34 @@ jobs:
...
+    permissions:
+      contents: write
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Node
+        uses: actions/setup-node@v3
+        with:
+          cache: yarn
+          node-version: '16'
```
RomainMuller, line 99: Probably should be the "min supported", but also wonder if we can/should matrix this? (12, 14, 16, 18)

.github/workflows/gh-pages.yml
```diff
@@ -82,3 +82,34 @@ jobs:
...
+          cd packages/@jsii/benchmarks
+          yarn bench | tee output.txt
+      - name: Store benchmark result
+        uses: benchmark-action/github-action-benchmark@v1
+        with:
+          name: JSII Benchmark
+          tool: 'benchmarkjs'
+          output-file-path: packages/@jsii/benchmarks/output.txt
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          auto-push: true
```
RomainMuller, line 115: Instead of `cd`, you should use the `working-directory` property.

I'd also emit the results out-of-tree because it's cleaner.

```suggestion
      - name: Run Benchmark
        run: |-
          yarn bench | tee ${{ runner.temp }}/bench-output.txt
        working-directory: packages/@jsii/benchmarks
      - name: Store benchmark result
        uses: benchmark-action/github-action-benchmark@v1
        with:
          name: JSII Benchmark
          tool: 'benchmarkjs'
          output-file-path: ${{ runner.temp }}/bench-output.txt
          github-token: ${{ secrets.GITHUB_TOKEN }}
          auto-push: true
```

.github/workflows/main.yml
```diff
@@ -365,3 +365,44 @@ jobs:
...
+      - name: Store Benchmark Results
+        uses: benchmark-action/github-action-benchmark@v1
+        with:
+          name: JSII Benchmark
+          tool: 'benchmarkjs'
+          output-file-path: packages/@jsii/benchmarks/output.txt
+          comment-always: true
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          fail-on-alert: true
+          auto-push: false
```
RomainMuller, line 408: Instead of `cd`, you should use the `working-directory` property.

I'd also emit the results out-of-tree because it's cleaner.

```suggestion
      - name: Run Benchmark
        run: |-
          yarn bench | tee ${{ runner.temp }}/bench-output.txt
        working-directory: packages/@jsii/benchmarks
      - name: Store benchmark result
        uses: benchmark-action/github-action-benchmark@v1
        with:
          name: JSII Benchmark
          tool: 'benchmarkjs'
          output-file-path: ${{ runner.temp }}/bench-output.txt
          comment-always: true
          github-token: ${{ secrets.GITHUB_TOKEN }}
          fail-on-alert: true
          auto-push: false
```

.github/workflows/gh-pages.yml
```diff
@@ -82,3 +82,34 @@ jobs:
       - name: Push
         run: |-
           git push origin gh-pages:gh-pages
+
+  benchmark:
```
RomainMuller, line 86: Not sure this is needed in the `gh-pages` workflow?

packages/@jsii/benchmarks/package.json
```diff
@@ -0,0 +1,29 @@
...
+  "scripts": {
+    "build": "tsc --build && npm run lint",
+    "watch": "tsc -w",
+    "bench": "node index.js",
+    "lint": "eslint . --ext .ts --ignore-path=.gitignore",
+    "lint:fix": "yarn lint --fix"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC"
```
RomainMuller, line 28:
```suggestion
  "author": "Amazon Web Services, Inc.",
  "license": "Apache-2.0"
```

packages/@jsii/benchmarks/index.ts
```diff
@@ -0,0 +1,53 @@
...
+import * as fs from 'fs-extra';
+import { Compiler } from 'jsii/lib/compiler';
+import { loadProjectInfo } from 'jsii/lib/project-info';
+import * as os from 'os';
+import * as path from 'path';
+import { promisify } from 'util';
+
+// Always run against the same version of CDK source
+const CDK_TAG = 'v2.21.1';
+const exec = promisify(cp.exec);
```
RomainMuller, line 12: Could use `execSync` instead... And generally speaking make this whole thing synchronous.

packages/@jsii/benchmarks/index.ts
```diff
@@ -0,0 +1,53 @@
...
+
+    compiler.emit();
+  });
+
+  suite.add('Compile jsii-calc', () => {
+    const { projectInfo } = loadProjectInfo(jsiiCalcDir);
+    const compiler = new Compiler({ projectInfo });
+
+    compiler.emit();
+  });
```
RomainMuller, line 46: This won't work on an un-built tree, because the dependencies of `jsii-calc` won't have been built yet (they won't have `tsconfig.json`, etc...)

packages/@jsii/benchmarks/index.ts
```diff
@@ -0,0 +1,53 @@
...
+import { loadProjectInfo } from 'jsii/lib/project-info';
+import * as os from 'os';
+import * as path from 'path';
+import { promisify } from 'util';
+
+// Always run against the same version of CDK source
+const CDK_TAG = 'v2.21.1';
+const exec = promisify(cp.exec);
+async function setupAwsCdkLib(): Promise<{ awsCdkLib: string }> {
+  const repoDir = await fs.mkdtemp(path.join(os.tmpdir(), 'jsii-cdk-bench'));
```
RomainMuller, line 14: I would really, really like it if that temporary directory was cleaned up before exit.

packages/@jsii/benchmarks/index.ts
```diff
@@ -0,0 +1,53 @@
...
+  await exec(`git clone -b ${CDK_TAG} https://github.com/aws/aws-cdk.git .`, {
+    cwd: repoDir,
+  });
+  await exec('yarn', { cwd: repoDir });
```
RomainMuller, line 18: Shouldn't this be:

```suggestion
  await exec('yarn', ['install', '--frozen-lockfile'], { cwd: repoDir });
```

packages/@jsii/benchmarks/index.ts
```diff
@@ -0,0 +1,53 @@
...
+import * as os from 'os';
+import * as path from 'path';
+import { promisify } from 'util';
+
+// Always run against the same version of CDK source
+const CDK_TAG = 'v2.21.1';
+const exec = promisify(cp.exec);
+async function setupAwsCdkLib(): Promise<{ awsCdkLib: string }> {
+  const repoDir = await fs.mkdtemp(path.join(os.tmpdir(), 'jsii-cdk-bench'));
+  await exec(`git clone -b ${CDK_TAG} https://github.com/aws/aws-cdk.git .`, {
```
RomainMuller, line 15: A shallow clone will be much faster. Can also consider downloading the release's source ZIP file from GitHub instead...

```suggestion
  await exec(`git clone --depth=1 -b ${CDK_TAG} https://github.com/aws/aws-cdk.git .`, {
```

.github/workflows/gh-pages.yml
MrArnoldPalmer: I had separated the workflows that runs every PR in the `main` workflow and the workflow that updates the charts in the gh-pages. I have consolidated them though so will remove.

.github/workflows/gh-pages.yml
MrArnoldPalmer: Since 12 is EOL and we are about to deprecate support, I'll make this 12 for now? The action doesn't support matrixing out of the box but we can look into it.

MrArnoldPalmer:
Added DNM as benchmark.js doesn't really support what we want to do. Writing a basic custom runner that generates output that the action we use supports so we should still get the graphs on gh-pages and the comments for regression exceeding defined threshold.

packages/@jsii/benchmarks/index.ts
MrArnoldPalmer: I removed the benchmark for jsii-calc for now cause it's kinda funky to setup, though we definitely can in the future.

Mainly, if copying the source files to some tempdir, we need to find all the local dependencies and link to them or something. Also pointing to the bundled dependency is annoying.

MrArnoldPalmer:
Changed strategy a bit since last review @RomainMuller @yuth 

1. Removed benchmark.js for a really simple custom runner using the node `perf_hooks` module. Benchmark.js wasn't cutting it for us as it isn't really built around running tests that may require cleanup in between. There are no apis for cleaning up the target directory in between runs and therefore after the first run, JSII essentially no-ops, skewing the results.

2. Added a `fixtures` directory that has a snapshot of the aws-cdk-lib project ready to build (meaning you can run `npm install` and then JSII is able to compile it). This greatly reduces the time of running the benchmark as cloning the repo, building all the submodules, running ubergen, removing build artifacts, THEN running JSII, is a lot of setup. I did include the script I used to snapshot it so we can rerun that or add new versions of cdk, or other modules, though some of it is aws-cdk-lib specific).

I also investigated including profiles and graphing difference in cpu/memory usage as well but it's non-trivial and we can add it after the fact if that data is interesting to us.

yuth (approved):
LGTM!

yuth (approved):

.gitattributes
```diff
@@ -15,4 +15,5 @@
 # Github Linguist configuration (https://github.com/github/linguist)
 yarn.lock linguist-generated
 *.snap    linguist-generated
+packages/@jsii/benchmarks/fixtures/** linguist-generated
```
RomainMuller, line 18: This is technically vendor, not generated.

```suggestion
packages/@jsii/benchmarks/fixtures/** linguist-vendored
```

.github/workflows/main.yml
```diff
@@ -365,3 +365,51 @@ jobs:
...
+      - name: Install Dependencies
+        run: yarn install --frozen-lockfile
+      - name: Run Benchmark
+        working-directory: packages/@jsii/benchmarks
+        run: yarn bench --output ${{ runner.temp }}/bench-output.json
+      - name: Compare Benchmark Results
+        if: github.event_name == 'pull_request'
+        uses: benchmark-action/github-action-benchmark@v1
+        with:
+          name: JSII Benchmark Regression
```
RomainMuller, line 401: Lower case, always :)

```suggestion
          name: jsii Benchmark Regression
```

.github/workflows/main.yml
```diff
@@ -365,3 +365,51 @@ jobs:
...
+          tool: 'customSmallerIsBetter'
+          output-file-path: ${{ runner.temp }}/bench-output.json
+          comment-always: true
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          fail-on-alert: true
+      - name: Upload Benchmark Results
+        if: github.event_name == 'push'
+        uses: benchmark-action/github-action-benchmark@v1
+        with:
+          name: JSII Benchmark
```
RomainMuller, line 411:
```suggestion
          name: jsii Benchmark
```

packages/@jsii/benchmarks/README.md
```diff
@@ -0,0 +1,25 @@
+# JSII Benchmarks
```
RomainMuller, line 1:
```suggestion
# jsii Benchmarks
```

packages/@jsii/benchmarks/bin/benchmark.ts
```diff
@@ -0,0 +1,86 @@
...
+  range: number;
+
+  /**
+   * Extra information about the benchmark, displayed in a tooltip
+   */
+  extra: string;
+}
+
+(async () => {
+  const argv = yargs
```
RomainMuller, line 39: Add `await` here (the next version of `yargs` types `argv` as `Args | Promise<Args>` 😒)

```suggestion
  const argv = await yargs
```

packages/@jsii/benchmarks/bin/benchmark.ts
```diff
@@ -0,0 +1,86 @@
...
+          range: Math.max(...iterations) - Math.min(...iterations),
+          extra,
+        },
+      ];
+    },
+    Promise.resolve([]),
+  );
+
+  if (argv.output) {
+    fs.writeFileSync(argv.output, JSON.stringify(resultsJson, undefined, 2));
```
RomainMuller, line 74: Either you go all-sync, or you go all-`async`, but you probably shouldn't mix-and-match :)

```suggestion
    await fs.writeJson(argv.output, resultsJson, { spaces: 2 });
```

packages/@jsii/benchmarks/lib/benchmark.ts
```diff
@@ -0,0 +1,134 @@
...
+  /**
+   * The average length across all iterations
+   */
+  average: number;
+
+  /**
+   * Results of individual runs
+   */
+  iterations: PerformanceEntry[];
+}
```
RomainMuller, line 21:
```suggestion
interface Result {
  /**
   * The name of the benchmark
   */
  readonly name: string;

  /**
   * The average length across all iterations
   */
  readonly average: number;

  /**
   * Results of individual runs
   */
  readonly iterations: readonly PerformanceEntry[];
}
```

packages/@jsii/benchmarks/lib/benchmark.ts
```diff
@@ -0,0 +1,134 @@
...
+export class Benchmark<C> {
+  /**
+   * How many times to run the subject
+   */
+  private _iterations = 5;
+
+  /**
+   * Results of individual runs
+   */
+  private results: PerformanceEntry[] = [];
```
RomainMuller, line 42: Why is this one not using an `_` prefix, when every other private is?

At this rate, couldn't you just use `#private` fields?

packages/@jsii/benchmarks/lib/benchmark.ts
```diff
@@ -0,0 +1,134 @@
...
+  public setup<T extends C>(fn: () => T) {
+    this._setup = fn;
+    return this as unknown as Benchmark<T>;
+  }
+
+  /**
+   * Create a teardown function to be run once after all benchmark runs. Use to
+   * clean up your mess.
+   */
+  public teardown(fn: (ctx: C) => any) {
```
RomainMuller, line 64:
```suggestion
  public teardown(fn: (ctx: C) => void) {
```

packages/@jsii/benchmarks/lib/benchmark.ts
```diff
@@ -0,0 +1,134 @@
...
+    this._teardown = fn;
+    return this;
+  }
+
+  /**
+   * Create a beforeEach function to be run before each iteration. Use to reset
+   * state the subject may have changed.
+   */
+  public beforeEach(fn: (ctx: C) => any) {
```
RomainMuller, line 73:
```suggestion
  public beforeEach(fn: (ctx: C) => void) {
```

packages/@jsii/benchmarks/lib/benchmark.ts
```diff
@@ -0,0 +1,134 @@
...
+    this._beforeEach = fn;
+    return this;
+  }
+
+  /**
+   * Create an afterEach function to be run after each iteration. Use to reset
+   * state the subject may have changed.
+   */
+  public afterEach(fn: (ctx: C) => any) {
```
RomainMuller, line 82:
```suggestion
  public afterEach(fn: (ctx: C) => void) {
```

packages/@jsii/benchmarks/lib/benchmark.ts
```diff
@@ -0,0 +1,134 @@
...
+        performance.clearMarks();
+        observer.disconnect();
+
+        return ok({
+          name: this.name,
+          average:
+            this.results.reduce((accum, { duration }) => accum + duration, 0) /
+            this.results.length,
+          iterations: this.results,
+        });
```
RomainMuller, line 120: Might also be interesting to return:
- `min`
- `max`
- (perhaps) `stddev` or `variance`

packages/@jsii/benchmarks/lib/benchmark.ts
```diff
@@ -0,0 +1,134 @@
...
+      obs.observe({ entryTypes: ['function'] });
+
+      const c = this._setup?.();
+      for (let i = 0; i < this._iterations; i++) {
+        this._beforeEach(c);
+        wrapped(c);
+        this._afterEach(c);
+      }
+
+      this._teardown(c);
```
RomainMuller, line 131: Suggest the after/cleanup operations to be performed in a `finally` block.

```suggestion
      try {
        for (let i = 0; i < this._iterations; i++) {
          this._beforeEach(c);
          try {
            wrapped(c);
          } finally {
            this._afterEach(c);
          }
        }
      } finally {
        this._teardown(c);
      }
```

packages/@jsii/benchmarks/lib/constants.ts
```diff
@@ -0,0 +1,9 @@
+import * as path from 'path';
+
+export const fixturesDir = path.join(__dirname, '..', 'fixtures');
```
RomainMuller, line 3: For some weird reason I prefer to use `resolve` for those (feel free to ignore me)

```suggestion
export const fixturesDir = path.resolve(__dirname, '..', 'fixtures');
```

packages/@jsii/benchmarks/lib/index.ts
```diff
@@ -0,0 +1,44 @@
...
+import { Benchmark } from './benchmark';
+import { cdkv2_21_1 } from './constants';
+
+// Always run against the same version of CDK source
+const cdk = new Benchmark('Compile aws-cdk-lib@v2.21.1')
+  .setup(() => {
+    const sourceDir = fs.mkdtempSync(
+      path.join(os.tmpdir(), 'jsii-cdk-bench-snapshot'),
+    );
+    cp.execSync(`tar xf ${cdkv2_21_1}`, { cwd: sourceDir });
```
RomainMuller, line 17: Would prefer _not_ to shell out to `tar` here, this is kind of risky (`tar` on macOS versus `tar` on Linux versus the absence of `tar` on Windows...)

packages/@jsii/benchmarks/lib/index.ts
```diff
@@ -0,0 +1,44 @@
...
+    cp.execSync('npm ci', { cwd: sourceDir });
```
RomainMuller, line 18: `npm ci`? Isn't CDK using `yarn`?

packages/@jsii/benchmarks/lib/index.ts
```diff
@@ -0,0 +1,44 @@
...
+import { Compiler } from 'jsii/lib/compiler';
+import { loadProjectInfo } from 'jsii/lib/project-info';
+import * as os from 'os';
+import * as path from 'path';
+
+import { Benchmark } from './benchmark';
+import { cdkv2_21_1 } from './constants';
+
+// Always run against the same version of CDK source
+const cdk = new Benchmark('Compile aws-cdk-lib@v2.21.1')
```
RomainMuller, line 12: Maybe all those `v2.21.1` literals should be coming from a constant, too...

packages/@jsii/benchmarks/lib/index.ts
```diff
@@ -0,0 +1,44 @@
...
+    cp.execSync('npm ci', { cwd: sourceDir });
+    // Working directory for benchmark
+    const workingDir = fs.mkdtempSync(
+      path.join(os.tmpdir(), 'jsii-cdk-bench@v2.21.1'),
+    );
+
+    return {
+      workingDir,
+      sourceDir,
+    };
```
RomainMuller, line 27:
```suggestion
    return {
      workingDir,
      sourceDir,
    } as const;
```

packages/@jsii/benchmarks/package.json
```diff
@@ -0,0 +1,35 @@
...
+  "scripts": {
+    "build": "tsc --build && npm run lint",
+    "watch": "tsc -w",
+    "bench": "node bin/benchmark.js",
+    "snapshot": "node scripts/snapshot-package.js",
+    "lint": "eslint . --ext .ts --ignore-path=.gitignore",
+    "lint:fix": "yarn lint --fix"
+  },
+  "bin": {
+    "benchmark": "node bin/benchmark.js"
```
RomainMuller, line 30:
```suggestion
    "benchmark": "bin/benchmark.js"
```

packages/@jsii/benchmarks/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "@jsii/benchmarks",
+  "version": "0.0.0",
+  "private": true,
+  "description": "JSII benchmarks",
```
RomainMuller, line 5:
```suggestion
  "description": "The jsii benchmark suite",
```

RomainMuller (approved):
A bunch of cosmetic feedback, feel free to ship it after having reviewed that...

packages/@jsii/benchmarks/lib/index.ts
MrArnoldPalmer: So we ran out of memory in the gh action when calling `tar` within nodejs =(

MrArnoldPalmer: oh I'm dumb, I'll use a stream 😄

PR #3521, opened 2022-05-05
feat(jsii): enable source maps for declaration files

andrestone:
Generates [declaration maps](https://www.typescriptlang.org/tsconfig#declarationMap) when compiling.

This is useful to help navigate source files when developing against a JSII library's repo (e.g using yarn link).

---


RomainMuller:
Hey @andrestone - can you elaborate on why you're making this change?

Depending on the rationale (you might have thought into this a bit more than I have), I wonder if we should just always turn the flag on, instead of making it customizable? I doubt it's very expensive to have, and it's "strongly recommended" by the TypeScript documentation...

andrestone:
Sorry I missed your first question. This is basically to help navigate source files when developing.

RomainMuller:
Are these emitted as separate files? If so, they could be added to `.npmignore`...

In any case, I don't think there is much harm in actually shipping the declarations... If the corresponding sources aren't present, IDE integrations would just fall back to the "usual" process...

They could be seen as dead-weight... but I doubt that'd be very heavy, so maybe this can be ignored?

andrestone:
Yes, they're separate files with the extension`.d.ts.map`. And yes, language servers would fall back to `d.ts` files so it should be ok. Gonna make the changes.

andrestone:
@RomainMuller I'm unsure if we rely on `tsconfig-base.json` for anything else related to the compiler, but I've added the option there too [as recommended](https://www.typescriptlang.org/tsconfig#:~:text=You%20should%20strongly%20consider%20turning%20this%20on%20if%20you%E2%80%99re%20using%20project%20references.) since we're using project references in JSII itself (hence the many map files).

andrestone:
> That's okay, but you should not check in all the `.d.ts.map` files :)

Of course! I think I checked the root `.gitignore` entries for `*.d.ts` and since I couldn't find it there I assumed the declarations were checked in for some reason which doesn't make sense.

Should be ok now.

andrestone:
Hey @RomainMuller! Sorry for the ping, just wanted to make sure there isn't anything else I could do here.

RomainMuller (approved):

RomainMuller:
Hey @andrestone -- sorry I got caught up in other stuff and forgot to check back here...

At this stage, this looks good to go for me.
We can let Mergify handle it from there (unless PR validations somehow break).

PR #3539, opened 2022-05-12
chore: stop using ts-jest

RomainMuller:
Our process compiles before tests are run, and using
ts-jest causes the test files to be transformed a second
time, which is a waste of resources.

---


yuth:
I wonder if the workflow of using `ts-jest` is still better in making iterative development faster where in one does not have to run 2 watch commands (one for build and one for test)

RomainMuller:
I've personally experienced issues where `.js` files existed and were used in priority, despite the `.ts` files had changed, and hence needing to _anyway_ run two watch commands... 🤷🏻‍♂️

PR #3549, opened 2022-05-18
chore: build all packages with TypeScript 4.6

RomainMuller:
Use TypeScript 4.6 for building all packages, but continue to use
TypeScript 3.9 in the compiler and rosetta in order to avoid introducing
breaking changes.

Moved several "universal" devDependencies all the way to the root of the
mono-repository to minimize changes when updating these (`@types/jest`,
`jest`, `eslint`, etc...).

Some code changes were required due to the breaking changes introduced
between TypeScript 3.9 and 4.6, but nothing particularly substantial.
There seems to be an odd interaction between webpack and actually
importing `node:process`, so the `jsii/runtime` package switched to
using the global `process` value instead of importing the module.

---

### DIFF
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -58,9 +58,15 @@ export function overriddenConfig(overrides: Config.InitialOptions) {
         ...Object.keys(original),
         ...Object.keys(override),
       ]);
+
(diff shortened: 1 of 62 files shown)


yuth (approved):
LGTM!

PR #3558, opened 2022-05-20
feat(jsii): make source map behavior fully configurable

RomainMuller:
Instead of enabling declarations maps everywhere as was done in #3521,
allow customers to define their desired source-map related configuration
in the `jsii.tsc` stanza of `package.json`.

This change in stance is motivated by how introduction of declarations
map causes broad asset hash changes in consumer code, which effectively
breaks many snapshot-based regression tests, and this feature should
hence be opt-in.

---


RomainMuller:
Nudge @andrestone - this will be of interest to you...

rix0rrr (approved):

PR #3560, opened 2022-05-24
feat(jsii): compress .jsii file under feature flag

### LINKED ISSUE #3561: support creating and loading zipped assembly files
### Describe the feature

Currently, jsii writes the package assembly to the `.jsii` file. The `.jsii` file can get quite large as the package grows. We should be able to configure a way to compress these files and uncompress when needed. This will help reduce the module size of the `aws-cdk-lib` package.

### Use Case

This is part of the reduce module size [RFC](https://github.com/aws/aws-cdk-rfcs/pull/430)

### Proposed Solution

Zip the assembly file when a flag is set, like:

```json
{
  "jsii": {
    "metadata": {
      "compress": true
    }
  }
}
```

The rest of jsii needs to then be aware that the assembly file can come in this compressed format.

### Other Information

_No response_

### Acknowledgements

- [X] I may be able to implement this feature request

### CDK version used

2.X

MacOs

kaizencc:
Introduces a flag to be set in `package.json`:

```json
"jsii": {
  "metadata": {
    "jsii": {
      "compress": true
    }
  }
}
```

That will emit a `.jsii.gz` compressed assembly file rather than the uncompressed `.jsii` file. As part of this PR, the jsii logic that loads assembly files need to be updated to recognize both `.jsii` and `.jsii.gz` files.

Closes #3561

---


RomainMuller:
I don't think the flag needs to be buried under `jsii.metadata.jsii.compress` (and also, that name isn't specific enough).

I'd rather:
- Expose feature via CLI argument `--compress-assembly`, which defaults to `false` (i.e: `--no-compress-assembly`)
- POSSIBLY allow configuring it in `package.json` in `jsii.compressAssembly` (also defaults to `false` for obvious reasons)

I also wonder whether we should plan for a "compatibility" mode that'll emit BOTH compressed and uncompressed assemblies... I currently lean towards "no", but... I don't know... This feels like it could benefit from an RFC / impact analysis, so we don't surprise people too much...

packages/jsii-reflect/lib/type-system.ts
```diff
@@ -316,6 +316,18 @@ export class TypeSystem {
     return new Assembly(this, ass);
   }
 
+  private getAssemblyFile(pathToAssembly: string) {
+    const jsiiFile = path.join(pathToAssembly, '.jsii');
+    const compressedJsiiFile = path.join(pathToAssembly, '.jsii.gz');
+    const jsiiExists = fs.existsSync(jsiiFile);
+    if (!jsiiExists && !fs.existsSync(compressedJsiiFile)) {
```
RomainMuller, line 323: If they both exist... there might be an issue -- in particular, this will ALWAYS prefer loading the uncompressed assembly, when it should likely prefer loading the compressed one if it exists...

A possible solution would be to *ALWAYS* load `.jsii`, but if there is a compressed assembly, that file basically contains something along the lines of:

```json
{
  "schema": "jsii/file-redirect",
  "compression": "gzip",
  "filename": ".jsii.gz"
}
```

The loader would detect this... and proceed to load (and uncompress) `.jsii.gz` as indicated.

packages/jsii-reflect/lib/type-system.ts
```diff
@@ -316,6 +316,18 @@ export class TypeSystem {
     return new Assembly(this, ass);
   }
 
+  private getAssemblyFile(pathToAssembly: string) {
+    const jsiiFile = path.join(pathToAssembly, '.jsii');
```
RomainMuller, line 320: There is a constant for `.jsii` called `SPEC_FILE_NAME` in `@jsii/spec`. It should be used.

packages/jsii-rosetta/lib/commands/transliterate.ts
```diff
@@ -1,5 +1,6 @@
 import { Assembly, Docs, SPEC_FILE_NAME, Type, TypeKind } from '@jsii/spec';
-import { readJson, writeJson } from 'fs-extra';
+import { writeJson } from 'fs-extra';
+import { loadAssemblyFromPath } from 'jsii/lib/utils';
```
RomainMuller, line 3: As discussed, it is worth moving to a utilities library (separate package)... I'd rather we don't do deep imports within other packages, as I intend to eventually block them via `exports` settings.

packages/jsii-rosetta/test/commands/extract.test.ts
```diff
@@ -59,6 +59,45 @@ beforeEach(() => {
 
 afterEach(() => assembly.cleanup());
 
+test('extract works with zipped assembly files', async () => {
```
RomainMuller, line 62:
```suggestion
test('extract works with gzipped assembly files', async () => {
```

packages/jsii-rosetta/test/commands/extract.test.ts
```diff
@@ -59,6 +59,45 @@ beforeEach(() => {
...
+        }
+      }
+      `,
+      'README.md': DUMMY_README,
+    },
+    {
+      name: 'my_assembly',
+      jsii: DUMMY_JSII_CONFIG,
+    },
+    true /* zipAssemblyFile */,
```
RomainMuller, line 81: Pass a property bag instead so you don't have to comment everywhere... This signature is starting to look a little awkward, too... maybe `TestJsiiModule` could have a static `.enableAssemblyCompression()` that returns a version of itself that enables compression.

packages/jsii-rosetta/test/commands/extract.test.ts
```diff
@@ -59,6 +59,45 @@ beforeEach(() => {
...
+  );
+
+  // assert that assembly is zipped
+  expect(fs.existsSync(path.join(assembly.moduleDirectory, '.jsii.gz'))).toBeTruthy();
+  expect(fs.existsSync(path.join(assembly.moduleDirectory, '.jsii'))).toBeFalsy();
+
+  // behavior is as expected
+  const cacheToFile = path.join(assembly.moduleDirectory, 'test.tabl.json');
```
RomainMuller, line 89: FYI this file will be HUGE, and is worth compressing, too.

packages/jsii/lib/assembler.ts
```diff
@@ -3361,3 +3366,7 @@ function loadAndRenderReadme(readmePath: string, projectRoot: string) {
       .join('\n'),
   };
 }
+
+function compressJsii(assembly: spec.Assembly): boolean {
+  return assembly.metadata?.jsii?.compress ?? false;
+}
```
RomainMuller, line 3372: It feels backwards to me that the place where you source the "should the assembly be compressed" setting is... in the assembly itself. I don't think that setting should be forwarded to the assembly at all (see previous comment about how this shouldn't be buried under metadata).

```suggestion
function assemblyCompressionEnabled(assembly: spec.Assembly): boolean {
  return assembly.metadata?.jsii?.compress == true;
}
```

packages/jsii/lib/helpers.ts
```diff
@@ -116,7 +122,7 @@ export function compileJsiiForTest(
     if (errors.length > 0 || emitResult.emitSkipped) {
       throw new Error('There were compiler errors');
     }
-    const assembly = fs.readJsonSync('.jsii', { encoding: 'utf-8' });
+    const assembly = loadAssemblyFromPath('');
```
RomainMuller, line 125:
```suggestion
    const assembly = loadAssemblyFromPath(process.cwd());
```

RomainMuller (changes requested):
You probably also want to enable compression on at least one (definitely not all) of the `jsii-calc` and `@scope/*` packages... 

And perhaps also explicitly disable it on anther.

packages/jsii-rosetta/test/commands/extract.test.ts
kaizencc: definitely will be compressing the tablet file too, but in a separate PR :)

packages/jsii-reflect/lib/type-system.ts
kaizencc: new PR for `@jsii/utils` does do something like this.

kaizencc:
superseded by #3570 and another PR coming soon.

PR #3565, opened 2022-05-27
fix(@jsii/spec): speed up assembly validation by 20x for large libraries

Chriscbr:
Switch from jsonschema to ajv to improve assembly validation performance.

Previously, loading the type system for aws-cdk-lib and constructs (combined) took around 15-17 seconds on my local development machine, using node v16.8.0. When I ran it with the updated version of @jsii/spec, it only took 750-800ms.

Both ajv and jsonschema are popular npm libraries (we also currently leverage it in the cdk8s-cli). ajv has one downside which is that introduces has a larger bundle size (1.02 MB vs 81.8kB), but I think it's worth it for the performance improvement.

Closes cdklabs/decdk#7

---

### DIFF
--- a/packages/@jsii/spec/package.json
+++ b/packages/@jsii/spec/package.json
@@ -31,7 +31,7 @@
     "package": "package-js"
   },
   "dependencies": {
-    "jsonschema": "^1.4.1"
+    "ajv": "^8.11.0"
   },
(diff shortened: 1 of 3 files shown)


yuth:
🤯

yuth (approved):

PR #3570, opened 2022-06-02
feat(@jsii/utils): add a utility library to be consumed by jsii modules

kaizencc:
Adds a set of utility functions under `@jsii/spec` that can (and will) be leveraged by other jsii modules.

Currently has the following functions exposed:

- `getAssemblyFile()`
- `writeAssembly()`
- `loadAssemblyFromPath()`
- `loadAssemblyFromFile()`

---


packages/jsii/lib/assembler.ts
```diff
@@ -277,12 +278,16 @@ export class Assembler implements Emitter {
...
-      fs.writeJsonSync(assemblyPath, _fingerprint(assembly), {
-        encoding: 'utf8',
-        spaces: 2,
-      });
+      LOG.trace(
+        `Emitting assembly: ${chalk.blue(
+          path.join(this.projectInfo.projectRoot, SPEC_FILE_NAME),
+        )}`,
+      );
+      writeAssembly(
```
kaizencc, line 286: I put this in to "prove" that the new module was exporting functions correctly, but if you'd rather this be a part of the future PR that will update the rest of the libraries to use this module, then I can go ahead and revert the changes to `jsii`.

packages/@jsii/utils/lib/assembly.ts
```diff
@@ -0,0 +1,98 @@
+import * as spec from '@jsii/spec';
+import * as fs from 'fs-extra';
+import * as path from 'path';
+import * as zlib from 'zlib';
+
+export const SPEC_FILE_NAME = '.jsii';
```
kaizencc, line 6: I could be wrong, but my intuition is that this constant belongs in this module, not `@jsii/spec`. Removing it from `@jsii/spec` is a breaking change, but I figured we could just start using this constant, and duplication of one line never hurt anyone.

kaizencc (commented):
@RomainMuller my game plan is to iterate on this PR to your fancy, and then modify the previous PR to use this library (and address the rest of your comments) when its merged.

kaizencc (commented):
> If they both exist... there might be an issue -- in particular, this will ALWAYS prefer loading the uncompressed assembly, when it should likely prefer loading the compressed one if it exists...
> 
> A possible solution would be to _ALWAYS_ load `.jsii`, but if there is a compressed assembly, that file basically contains something along the lines of:
> 
> ```json
> {
>   "schema": "jsii/file-redirect",
>   "compression": "gzip",
>   "filename": ".jsii.gz"
> }
> ```
> 
> The loader would detect this... and proceed to load (and uncompress) `.jsii.gz` as indicated.

@RomainMuller if you want this, i am happy to do things this way. but as it stands, i simply changed to preferring the compressed file if it exists, and i don't really see what always writing the .jsii file will achieve, and in fact, i think it might be worse.

we also talked offline about hacking the .jsii file to throw a descriptive error when an old jsii version is being used. i couldn't really get this to work, because the only error i can throw is `Unexpected token E in JSON at position 0`, which gives us 1 character of description to work with 😂. if what we write in .jsii is actual JSON, then the loader will simply load the .jsii file normally. Since we don't really enforce _validating_ the assembly, i think it's entirely possible that users get ugly non-descriptive errors when they try to access the nonexistent `assembly['name']`.

my conclusion is that the best we can do is throw the error that the `.jsii` file does not exist in these situations. At least that alerts the user that the issue is at the level of a missing `.jsii` file, and not that their `.jsii` file is malformed. let me know if that conclusion makes sense!

packages/jsii/lib/assembler.ts
RomainMuller: Nah this is cool - this way you're not introducing (temporary) deadweight.

packages/@jsii/utils/lib/assembly.ts
RomainMuller: You can actually move the constant there AND re-export it from `@jsii/spec` (or vice-versa if the dependency can flow in that direction), too (for convenience). The file name is basically considered part of the contract between the various libraries/CLIs in the jsii toolchain, and so in this sense it is part of the spec.

RomainMuller:
> i don't really see what always writing the .jsii file will achieve, and in fact, i think it might be worse.

Using that "redirect" placeholder gives us a new point where we can detect future "you need to upgrade your toolchain" situations... We can actually insert an unconditional validation of the contents of the `schema` property, and bail out with an error message that says you need to upgrade your toolchain or the assembly might have been corrupted/tampered with.

Some of the stuff we could end up doing in the future include (not that we are planning to do any of these at this time, but I could see those as possible future directions):
- Other compression algorithms if they yield better compression ratios (ZStd, Brotli, Bzip, ...)
- Breaking down an assembly into multiple documents (`.jsii.tgz` or `.jsii.zip`)
- Per-language assemblies (`.jsii-java`, `.jsii-python`, ...)

Basically it would help crystallize the contract that you can ALWAYS start by looking at `.jsii`, and that may contain instructions on how to load other files.

We could also have a `requires: { jsii: '>= X.Y.Z' }` property in there, which we could check in the future (to throw "You are on jsii A.B.C, but this assembly requires jsii X.Y.Z. Please upgrade your toolchain"), and would hopefully serve as a self-diagnosis support tool for people using jsii releases earlier than the one which will introduce this feature.

packages/@jsii/utils/tsconfig.json
```diff
@@ -0,0 +1,5 @@
+{
```
RomainMuller, line 1: Please configure this for out-of-source building. You can see how `@jsii/spec` is doing it and replicate... This will involve renaming `/lib/` to `/src/`, too.

packages/@jsii/utils/.npmignore
```diff
@@ -0,0 +1,15 @@
+*
+
+!**/*.js
+!**/*.d.ts
+!bin/jsii
+!bin/jsii-fix-peers
```
RomainMuller, line 6:
```suggestion
!**/*.js
!**/*.d.ts
```

packages/@jsii/utils/src/assembly.ts
```diff
@@ -0,0 +1,130 @@
+import * as spec from '@jsii/spec';
+import * as fs from 'fs-extra';
+import * as path from 'path';
+import * as zlib from 'zlib';
+
+export const SPEC_FILE_NAME = spec.SPEC_FILE_NAME;
+export const SPEC_FILE_NAME_COMPRESSED = spec.SPEC_FILE_NAME_COMPRESSED;
```
RomainMuller, line 7: Missing doc :)

packages/jsii/package.json
```diff
@@ -37,6 +37,7 @@
   "dependencies": {
     "@jsii/check-node": "0.0.0",
     "@jsii/spec": "^0.0.0",
+    "@jsii/utils": "^0.0.0",
```
RomainMuller, line 40: There should be a related entry of `references` in `tsconfig.json`.

packages/@jsii/spec/package.json
```diff
@@ -31,7 +31,8 @@
     "package": "package-js"
   },
   "dependencies": {
-    "ajv": "^8.11.0"
+    "ajv": "^8.11.0",
+    "fs-extra": "^10.1.0"
```
kaizencc, line 35: hoping this addition is ok. fs-extra is used for `writeJsonSync` and `readJsonSync` and i'm sure I could just reimplement those if we don't want to add dependencies.

RomainMuller: It's okay - `fs-extra` is widely used in our codebases already. It's true these can be trivially achieved with `JSON.stringify`/`JSON.parse` and `fs.writeFileSync`/`fs.readFileSync`... but let's not overthink it.

RomainMuller (approved):

PR #3622, opened 2022-06-27
chore: run benchmarks in separate VMs

RomainMuller:
The idea is to try and achieve a situation where iteration
times are not tainted by different memory situations (caches
may cause memory pressure and will skew the benchmark results).

Running in node sub-processes should allow removing this unknown
from the picture.

---


MrArnoldPalmer (commented):
very nice!

MrArnoldPalmer (approved):

PR #3634, opened 2022-06-30
feat(@jsii/spec): add loadAssemblyFromBuffer function

kaizencc:
Usage is demonstrated here: https://github.com/cdklabs/construct-hub/pull/920

Two implementations were considered to fit the use case in construct hub, `loadAssemblyFromBuffer` and `loadAssemblyFromTarball`. I went with loading from a buffer because it allows us to utilize prior art in construct hub for parsing tarballs, does not need an added dependency on `tar`, and causes minimal impact when refactoring the ingestion lambda to use this function.

Also includes a small refactoring of the test suite. 

---

### DIFF
--- a/packages/@jsii/spec/src/assembly-utils.ts
+++ b/packages/@jsii/spec/src/assembly-utils.ts
@@ -66,6 +66,27 @@ export function writeAssembly(
   return compress;
 }
 
+export function loadAssemblyFromBuffer(
+  assemblyBuffer: Buffer,
+  compressedAssemblyBuffer?: Buffer,
+  validate = true,
+): Assembly {
+  let contents = JSON.parse(assemblyBuffer.toString('utf-8'));
+
+  // check if the file holds instructions to the actual assembly file
+  if (isRedirect(contents)) {
+    validateRedirectSchema(contents);
+    if (!compressedAssemblyBuffer) {
+      throw new Error(
+        `The assembly buffer redirects to a compressed assembly at ${contents.filename}, but no compressed assembly was found`,
+      );
+    }
+    contents = JSON.parse(zlib.gunzipSync(compressedAssemblyBuffer).toString());
+  }
+
+  return validate ? validateAssembly(contents) : (contents as Assembly);
+}
+
 /**
  * Loads the assembly file from a tarball archive and, if present,
  * follows instructions found in the file to unzip compressed assemblies.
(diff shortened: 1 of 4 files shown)


packages/@jsii/spec/src/assembly-utils.ts
```diff
@@ -64,13 +64,46 @@ export function writeAssembly(
...
+ */
+export function loadAssemblyFromBuffer(
+  assemblyBuffer: Buffer,
+  compressedAssemblyBuffer?: Buffer,
+  validate = true,
+): Assembly {
+  let contents = JSON.parse(assemblyBuffer.toString('utf-8'));
+
+  // check if the file holds instructions to the actual assembly file
+  if (isRedirect(contents)) {
```
RomainMuller, line 87: Maybe log a warning in case this is NOT a redirect, but a `comporessedAssemblyBuffer` was passed?

packages/@jsii/spec/src/assembly-utils.ts
```diff
@@ -64,13 +64,46 @@ export function writeAssembly(
...
+ * the redirect schema, since we are not dealing with files. Instead, it assumes
+ * that the compressedAssemblyBuffer holds the assembly we are looking for.
+ *
+ * @param assemblyBuffer buffer containing SPEC_FILE_NAME contents
+ * @param compressedAssemblyBuffer buffer containing SPEC_FILE_NAME_COMPRESSED contents, if available
+ * @param validate whether or not to validate the assembly
+ */
+export function loadAssemblyFromBuffer(
+  assemblyBuffer: Buffer,
+  compressedAssemblyBuffer?: Buffer,
```
RomainMuller, line 81: Maybe this could be made to depend on the compressed file name contract by accepting a callback for more files (`(fileName: string) => Buffer`)?

Would likely make it easier/possible to use that implementation in the other `loadAssembly` variants...

packages/@jsii/spec/src/assembly-utils.ts
```diff
@@ -64,13 +64,46 @@ export function writeAssembly(
   return compress;
 }
 
+/**
+ * Loads the assembly file and, if instructed to, redirects to the
+ * compressed assembly file.
+ *
+ * Note: this function does not actually look for the filename specified in
+ * the redirect schema, since we are not dealing with files. Instead, it assumes
+ * that the compressedAssemblyBuffer holds the assembly we are looking for.
```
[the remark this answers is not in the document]
kaizencc: This [comment](https://github.com/aws/jsii/pull/3634#discussion_r911713917) on callbacks is the solution to this call out.

RomainMuller (approved):

RomainMuller (approved):

PR #3650, opened 2022-07-07
chore: upgrade benchmark to aws-cdk-lib v2.31.0

RomainMuller:
This is necessary as the previous benchark target had multiple occurrences
of enums with several constants having the same value, which is now illegal
in jsii code (#3207).

The fix for this (https://github.com/aws/aws-cdk/commit/7f4f150b97b5f58b936e6ed221a1b9bf67dad107) was released in CDK 2.31.0.

---

### DIFF
--- a/packages/@jsii/benchmarks/lib/constants.ts
+++ b/packages/@jsii/benchmarks/lib/constants.ts
@@ -2,8 +2,8 @@ import * as path from 'node:path';
 
 export const fixturesDir = path.resolve(__dirname, '..', 'fixtures');
 
-export const cdkTagv2_21_1 = 'v2.21.1';
-export const cdkv2_21_1 = path.resolve(
+export const cdkTag = 'v2.31.0';
+export const cdk = path.resolve(
   fixturesDir,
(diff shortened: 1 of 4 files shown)


packages/@jsii/benchmarks/lib/constants.ts
```diff
@@ -2,8 +2,8 @@ import * as path from 'node:path';
 
 export const fixturesDir = path.resolve(__dirname, '..', 'fixtures');
 
-export const cdkTagv2_21_1 = 'v2.21.1';
-export const cdkv2_21_1 = path.resolve(
+export const cdkTag = 'v2.31.0';
```
mrgrain, line 5: ❓ Question: Why removing the version post fix in the variable name? Or I guess: Any reason it was there in the first place?

I'm asking as this seems unrelated to the addressed issue.

mrgrain (approved):

packages/@jsii/benchmarks/lib/constants.ts
mrgrain: Great, that's what I though!

PR #3652, opened 2022-07-07
feat(rosetta): tablets can be compressed

kaizencc:
Usage: `yarn rosetta extract --compress-tablet`

This should reduce the `.jsii.tabl.json` size by about 90%.

---

### DIFF
--- a/packages/jsii-rosetta/lib/tablets/tablets.ts
+++ b/packages/jsii-rosetta/lib/tablets/tablets.ts
@@ -1,5 +1,6 @@
 import * as fs from 'fs-extra';
 import * as path from 'node:path';
+import * as zlib from 'node:zlib';
 
 import { TargetLanguage } from '../languages';
 import * as logging from '../logging';
@@ -12,6 +13,7 @@ import { TabletSchema, TranslatedSnippetSchema, ORIGINAL_SNIPPET_KEY } from './s
 const TOOL_VERSION = require('../../package.json').version;
 
 export const DEFAULT_TABLET_NAME = '.jsii.tabl.json';
+export const DEFAULT_TABLET_NAME_COMPRESSED = '.jsii.tabl.json.gz';
 
 export const CURRENT_SCHEMA_VERSION = '2';
 
@@ -121,8 +123,17 @@ export class LanguageTablet {
     return this.snippets[snippetKey(typeScriptSource)];
   }
 
+  /**
+   * Load the tablet from a file. Will automatically detect if the file is in
+   * compression format ('.gz') and gunzip if it is.
+   */
   public async load(filename: string) {
-    const obj = (await fs.readJson(filename, { encoding: 'utf-8' })) as TabletSchema;
+    let obj: TabletSchema | any = {};
+    if (path.extname(filename) === '.gz') {
+      obj = JSON.parse(zlib.gunzipSync(await fs.readFile(filename)).toString()) as TabletSchema;
+    } else {
+      obj = (await fs.readJson(filename, { encoding: 'utf-8' })) as TabletSchema;
+    }
 
     if (!obj.toolVersion || !obj.snippets) {
       throw new Error(`File '${filename}' does not seem to be a Tablet file`);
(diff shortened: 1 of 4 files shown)


packages/jsii-rosetta/lib/tablets/tablets.ts
```diff
@@ -1,5 +1,6 @@
 import * as fs from 'fs-extra';
 import * as path from 'node:path';
+import * as zlib from 'node:zlib';
```
kaizencc, line 3: reminder that this import will need to be updated when node protocol is reverted: #3651

packages/jsii-rosetta/lib/tablets/tablets.ts
```diff
@@ -147,12 +165,21 @@ export class LanguageTablet {
...
+   * schema will be compressed, otherwise the tablet schema will be written in raw json.
+   */
   public async save(filename: string) {
     await fs.mkdirp(path.dirname(filename));
-    await fs.writeJson(filename, this.toSchema(), {
-      encoding: 'utf-8',
-      spaces: 2,
-    });
+
+    if (path.extname(filename) === '.gz') {
```
kaizencc, line 175: hoping to get away with this contract -- supplying a file extension of `.gz` indicates that one is expecting a gzipped file. i feel like this is logical enough to me, but my spidey senses are tingling since the file extension is governing behavior, and we may not want that.

packages/jsii-rosetta/lib/tablets/tablets.ts
```diff
@@ -121,8 +130,17 @@ export class LanguageTablet {
...
+   * compression format ('.gz') and gunzip if it is.
+   */
   public async load(filename: string) {
-    const obj = (await fs.readJson(filename, { encoding: 'utf-8' })) as TabletSchema;
+    let obj: TabletSchema | any = {};
+    if (path.extname(filename) === '.gz') {
+      obj = JSON.parse(zlib.gunzipSync(await fs.readFile(filename)).toString()) as TabletSchema;
+    } else {
+      obj = (await fs.readJson(filename, { encoding: 'utf-8' })) as TabletSchema;
+    }
```
RomainMuller, line 143: Instead of using extension detection -- since we know to expect either a Gzip object or a JSON document, we can actually use the header of the gzip object, which starts with a known magic number:

```suggestion
    let data = await fs.readFile(filename);
    // Gzip objects start with 1F 8B 08
    if (data[0] === 0x1F && data[1] === 0x8B && data[2] === 0x08) {
      // This is a gz object, so we decompress it now...
      data = lib.gunzipSync(data);
    }
    const obj: TabletSchema = JSON.parse(data.toString('utf-8'));
```

packages/jsii-rosetta/lib/tablets/tablets.ts
RomainMuller: Yeah I'd rather it accepts a `compress` property and acts accordingly.

Additionally, you can factor the JSON-ification so it is done once. You can use `Buffer.from` to obtain a `Buffer` from a `string`... So you can stringify in a buffer, then compress that buffer if needed, then write the buffer to disk.

packages/jsii-rosetta/lib/tablets/tablets.ts
kaizencc: Fascinating stuff. The more you know...

packages/jsii-rosetta/lib/tablets/tablets.ts
kaizencc: (I think) I've done what you're saying here. The side effect is that we no longer get the pretty printed JSON in the uncompressed case, since we're just writing a buffer to a file and not using the whole `fs.writeJson(filename, data, { spaces: 2})`. 

I think the new behavior is okay. We don't need human-readable tablet files. But wanted to call this out as a side-effect of this PR.

kaizencc: fyi the uncompressed file under `fs.writeFile` would look something like this (i tried it on s3):

```
{"version":"2","toolVersion":"0.0.0","snippets":{"a3fb238b791f212f2de9bf87c1625d6906fbdaac222f986c0a53ee5c652ec0b0":{"translations":{"python":{"source":"bucket = s3.Bucket(self, \"MyFirstBucket\")","version":"2"},"csharp":{"source":"Bucket bucket = new Bucket(this, \"MyFirstBucket\");","version":"1"},"java":{"source":"Bucket bucket = new Bucket(this, \"MyFirstBucket\");","version":"1"},"go":{"source":"bucket := s3.NewBucket(this, jsii.String(\"MyFirstBucket\"))","version":"1"},"$":{"source":"const bucket = new s3.Bucket(this, 'MyFirstBucket');","version":"0"}},"location":{"api":{"api":"moduleReadme","moduleFqn":"@aws-cdk/aws-s3"},"field":{"field":"markdown",
```

packages/jsii-rosetta/lib/tablets/tablets.ts
```diff
@@ -147,12 +167,19 @@ export class LanguageTablet {
...
+   * the schema will be gzipped before writing to the file.
+   */
+  public async save(filename: string, compress = false) {
     await fs.mkdirp(path.dirname(filename));
-    await fs.writeJson(filename, this.toSchema(), {
-      encoding: 'utf-8',
-      spaces: 2,
-    });
+
+    let schema = Buffer.from(JSON.stringify(this.toSchema()));
```
RomainMuller, line 177: To pretty-print:

```suggestion
    let schema = Buffer.from(JSON.stringify(this.toSchema(), null, 2));
```

RomainMuller (dismissed):

MrArnoldPalmer (approved):

PR #3664, opened 2022-07-19
fix(go): unused imports emitted for type unions

### LINKED ISSUE #3399: Error when types are imported but seemingly unused in Go
### Affected Languages

- [ ] `Python`
- [ ] `Java`
- [ ] .NET (`C#`, `F#`, ...)
- [x] `Go`

### General Information
* **JSII Version:** 1.54.0
* **Platform:** `jsii/superchain:1-buster-slim` image

### What is the problem?
This is in relation to https://github.com/cdklabs/cdk-monitoring-constructs

Given some code that looks like:

```ts
import * as elasticsearch from "monocdk/aws-elasticsearch";
import * as opensearch from "monocdk/aws-opensearchservice";

export type Domain =
  | opensearch.CfnDomain
  | opensearch.IDomain
  | elasticsearch.CfnDomain
  | elasticsearch.IDomain;
```

Go build fails with the following:

```
Error: #STDERR> ./cdkmonitoringconstructs.go:21:2: imported and not used: "github.com/aws/aws-cdk-go/awscdk/awselasticsearch"
Error: #STDERR> ./cdkmonitoringconstructs.go:23:2: imported and not used: "github.com/aws/aws-cdk-go/awscdk/awsopensearchservice"
```

(See [failed build](https://github.com/cdklabs/cdk-monitoring-constructs/runs/5332987469?check_suite_focus=true))

We also attempted importing the type only, i.e.:

```ts
import type {
  CfnDomain as ElasticSearchCfnDomain,
  IDomain as ElasticSearchIDomain,
} from "monocdk/aws-elasticsearch";
import type {
  CfnDomain as OpenSearchCfnDomain,
  IDomain as OpenSearchIDomain,
} from "monocdk/aws-opensearchservice";

export type Domain =
  | ElasticSearchCfnDomain
  | ElasticSearchIDomain
  | OpenSearchCfnDomain
  | OpenSearchIDomain;
```

(See [failed build](https://github.com/cdklabs/cdk-monitoring-constructs/runs/5334797237?check_suite_focus=true))

I also attempted to reduce the codebase down to just the above for a minimum repo, but ended up with errors like:

```
[2022-02-25T12:03:14.385] [ERROR] jsii/compiler - Type model errors prevented the JSII assembly from being created
error JSII9002: Unable to resolve type "monocdk.IDomain". It may be @internal or not exported from the module's entry point (as configured in "package.json" as "main").
lib/index.ts:15:15 - error JSII3002: Type "monocdk.IDomain" cannot be used as a parameter type because it is not exported from monocdk

15   constructor(domain: Domain) {
                 ~~~~~~

  node_modules/monocdk/lib/aws-opensearchservice/lib/domain.d.ts:548:1
    548 export interface IDomain extends cdk.IResource {
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    549     /**
        ~~~~~~~
    ... 
    756     metricIndexingLatency(props?: MetricOptions): Metric;
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    757 }
        ~
    The referenced type is declared here
```

### Verbose Log
I unfortunately can't seem to do a build locally due to network policies. Let me know if I can somehow provide more info otherwise.

RomainMuller:
When imported types are solely referenced by type unions, a go import
is emitted, but is never used (type unions end up represented as opaque
`interface{}` type). This causes compilation failures.

Added a test case for this scenario in particular, and adjusted go code
generation to not emit dependency imports for type unions.

These imports may be re-introduced soon, as we are working to add
dynamic type checking around type unions in go (at which point those
imports would no longer be unused).

Fixes #3399

---


packages/jsii-pacmak/jest.config.mjs
```diff
@@ -2,4 +2,5 @@ import { overriddenConfig } from '../../jest.config.mjs';
 
 export default overriddenConfig({
   coveragePathIgnorePatterns: ['/node_modules/', '<rootDir>/test'],
+  watchPathIgnorePatterns: ['.*\\.tsx?$'],
```
RomainMuller, line 5: Note - this makes `jest --watch` a lot more pleasant to use, as it'd only re-run tests onces `tsc --watch` updates the `.js` files instead of possibly re-running when the `.ts` file changed (which yields unexpected outcome).

I'm contemplating moving this up to the baseline configuration but need to evaluate impact first.

MrArnoldPalmer (approved):

PR #3669, opened 2022-07-21
fix(rosetta): infuse incorrectly handles compressed assemblies

kaizencc:
The behavior of `rosetta infuse` was incorrectly handled before. `infuse` would always overwrite the `.jsii` file with the uncompressed assembly. This PR fixes that behavior by detecting whether or not there is a compressed file in the directory, and compressing if that is the case.

There are two alternative solutions I considered, primarily because I'm concerned that looking up the location of the compressed assembly can be considered a leaky abstraction:
- loading and looking into the contents of `.jsii` again to determine whether or not it is a file redirect. I did not choose this option as it involves additional loading, which will slow things down.
- We can expand the `LoadedAssembly` type to include information on whether or not the assembly was originally compressed. Then pass that into the `replaceAssembly()` function. I ultimately decided against this because it would involve changing the function signature of all `loadAssemblyFromXxx` functions. It's both a breaking change, and unnecessary clutter for a single use case.

Based on the reasoning above, I think what is included in this PR makes the most sense: expose an independent function, `compressedAssemblyExists`, that returns whether or not there is a file located at `SPEC_FILE_NAME_COMPRESSED`. 

---

### DIFF
--- a/packages/jsii-rosetta/lib/jsii/assemblies.ts
+++ b/packages/jsii-rosetta/lib/jsii/assemblies.ts
@@ -1,5 +1,11 @@
 import * as spec from '@jsii/spec';
-import { loadAssemblyFromFile, loadAssemblyFromPath, findAssemblyFile, writeAssembly } from '@jsii/spec';
+import {
+  compressedAssemblyExists,
+  loadAssemblyFromFile,
+  loadAssemblyFromPath,
+  findAssemblyFile,
+  writeAssembly,
+} from '@jsii/spec';
 import * as crypto from 'crypto';
 import * as fs from 'fs-extra';
 import * as path from 'path';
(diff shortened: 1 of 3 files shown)


packages/jsii-rosetta/lib/jsii/assemblies.ts
```diff
@@ -228,14 +234,11 @@ export async function allTypeScriptSnippets(
...
+ * Detects whether or not there is a compressed assembly, and if there is, compresses the new assembly also.
  * Recalculates the fingerprint of the assembly to avoid tampering detection.
  */
-export function replaceAssembly(
-  assembly: spec.Assembly,
-  directory: string,
-  { compress = false }: { compress?: boolean } = {},
-) {
-  writeAssembly(directory, _fingerprint(assembly), { compress });
+export function replaceAssembly(assembly: spec.Assembly, directory: string) {
```
kaizencc, line 240: This is a breaking change :(. If we must add the `{ compress = false }: { compress?: boolean } = {}` option in again, then fine. I really don't want to though, because in my mind, this was a bit of a stealth mode release and its unlikely that this revert actually breaks any users. Especially because the prior behavior was incorrect, and I would simply ignore the property if we have to keep it in...

RomainMuller: You could also just change the default behavior for `compress` -- JavaScript does not require default values are constants/literals, it can be a function call result...

RomainMuller (approved):

PR #3671, opened 2022-07-22
chore(jsii-reflect): use `Map`s instead of `Record`s

RomainMuller:
Instead of using objects as maps, use actual `Map` as this has better
performance characteristics (results in fewer shadow classes being
generated, and is friendlier with the optimizer).

---


mrgrain (approved):
Maps 😍

PR #3695, opened 2022-08-02
fix(python): unable to override methods with keyword arguments

### LINKED ISSUE #3656: Python: cannot extend generated classes in an idiomatic way
### Describe the bug

Take the class `NatProvider`:
https://github.com/aws/aws-cdk/blob/d03543246c5659cb287566372d47564e6fc2df4c/packages/%40aws-cdk/aws-ec2/lib/nat.ts#L55

I'm implementing my own provider in Python, so I extend the class.
When I implement the `configure_nat` method, I cannot use the same method signature as in the generated classes.

In the generated code, the signature for that method is the following:
```
def configure_nat(self, *, nat_subnets: typing.Sequence["PublicSubnet"], private_subnets: typing.Sequence["PrivateSubnet"], vpc: "Vpc")
```

When I replicate that in my child class, I cannot use this signature.

### Expected Behavior

I should be able to extend the class and use the method signatures that match the ones in the generated python code for the parent class.

### Current Behavior

When I replicate the method with the same signature, I get this at synth:

```
jsii.errors.JSIIError: MyNatProvider.configure_nat() takes 1 positional argument but 2 were given
```

To actually implement the class, my method signature has to be:

```
def configure_nat(self, options: ec2.ConfigureNatOptions, /)
```

### Reproduction Steps

Subclass `NatProvider` and implement `configure_nat` with a method signature that matches the generated parent class.

### Possible Solution

_No response_

_No response_

### SDK version used

1.61.0

Linux

RomainMuller:
The callback logic in the jsii runtime library for Python was
incorrectly passing keyword arguments to Python implementations,
forwarding a struct instance as-is instead of destructuring it into a
keyword arguments mapping.

Using `inspect.signature()`, detect the presence of keyword arguments on
the callback target, and destructure the struct as relevant.

Fixes #3656

---


.all-contributorsrc
```diff
@@ -1420,6 +1420,15 @@
...
       ]
+    },
+    {
+      "login": "gshpychka",
+      "name": "Glib Shpychka",
+      "avatar_url": "https://avatars.githubusercontent.com/u/23005347?v=4",
+      "profile": "https://github.com/gshpychka",
+      "contributions": [
+        "bug"
+      ]
```
Naumel, line 1431: 0.02$: This change seems unrelated to the fix

RomainMuller: This is the record for the individual who's reported this bug... Need I submit 2 separate PRs?

Naumel (approved):

PR #3696, opened 2022-08-02
chore: move replaceAssembly to @jsii/spec and expose it

kaizencc:
This PR is meant to do two things:

- move `replaceAssembly` to `assembly-utils.ts` in `@jsii/spec`, because it is currently being duplicated in `cdk-generate-synthetic-examples`. This causes problems whenever we try to change `replaceAssembly`, as we did in this [commit](https://github.com/aws/jsii/pull/3669/files), because we will forget to change the function in `cdk-generate-synthetic-examples`. The plan is to have `cdk-generate-synthetic-examples` reference the `replaceAssembly` function in `@jsii/spec` in a separate PR.
- finish the effort started in this closed [PR](https://github.com/aws/jsii/pull/3146), where we meant to make sure that we don't re-fingerprint assemblies after they've been changed. It was closed due to staleness.

---


packages/@jsii/spec/src/assembly-utils.ts
```diff
@@ -40,6 +40,30 @@ export function findAssemblyFile(directory: string) {
...
+export function replaceAssembly(assembly: Assembly, directory: string) {
+  writeAssembly(directory, _fingerprint(assembly), {
+    compress: compressedAssemblyExists(directory),
+  });
+}
+
+/**
+ * Replaces the old fingerprint with '***********'.
+ *
+ * @rmuller says fingerprinting is useless, as we do not actually check
```
RomainMuller, line 57: That means nothing here... if you're gonna quote me, at least use my GitHub alias... Also I don't think I've ever said "useless", probably "currently not used", and so it was future-proofing...

kaizencc: Whoops... somehow it came out of the previous closed PR and I have no idea how I landed on that comment anyway. Will simply remove your alias entirely.

RomainMuller (approved):

PR #3698, opened 2022-08-03
chore(go): emit one file per type

RomainMuller:
Instead of generating one file will ALL the types in a given
(sub)module, emit one file per type within the (sub)module,
so that the compiler operates on smaller compilation units.
This should reduce the memory pressure the compiler
experiences when dealing with large libraries.

---

### DIFF
--- a/packages/jsii-pacmak/lib/targets/go/package.ts
+++ b/packages/jsii-pacmak/lib/targets/go/package.ts
@@ -239,7 +239,9 @@ export abstract class Package {
       });
     }
 
-    for (const goModuleName of new Set(type.dependencies.map(({goModuleName}) => goModuleName))) {
+    for (const goModuleName of new Set(
+      type.dependencies.map(({ goModuleName }) => goModuleName),
+    )) {
       // If the module is the same as the current one being written, don't emit an import statement
       if (goModuleName !== this.goModuleName) {
         toImport.push({ module: goModuleName });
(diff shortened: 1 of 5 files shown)


ansgarm:
Nice! This should help with https://github.com/hashicorp/terraform-cdk/issues/1886

packages/jsii-pacmak/lib/targets/go/package.ts
```diff
@@ -226,10 +198,7 @@ export abstract class Package {
     // form. It also saves us from "imported but unused" errors that would arise
     // as a consequence.
     if (this.types.length > 0) {
-      const initFile = join(
-        dirname(this.file),
-        `${basename(this.file, '.go')}.init.go`,
-      );
+      const initFile = `${this.directory}/${this.packageName}.go`;
```
comcalvi, line 201: why not use `path.join()` here?

packages/jsii-pacmak/lib/targets/go/package.ts
```diff
@@ -283,7 +254,14 @@ export abstract class Package {
 
   private emitTypes(context: EmitContext) {
     for (const type of this.types) {
+      const filePath = `${this.directory}/${this.packageName}_${type.name}.go`;
```
comcalvi, line 257: why not use `path.join()`?

comcalvi (approved):

PR #3703, opened 2022-08-11
feat(java): dynamic type checking for union-typed parameters

comcalvi:
Add runtime type checks for Java around type unions to provide better error messages for developers when union types are being used. 

Only performed if `software.amazon.jsii.Configuration.getRuntimeTypeChecking()` is `true`, which it is by default. 

---


packages/jsii-pacmak/lib/targets/dotnet/dotnetgenerator.ts
```diff
@@ -676,11 +676,7 @@ export class DotNetGenerator extends Generator {
       return;
     }
 
-    this.code.openBlock(
```
comcalvi, line 679: I removed this because of #3710, which I imagine is a better way of doing what this does, but for all languages.

packages/jsii-pacmak/test/generated-code/__snapshots__/target-java.test.js.snap
```diff
@@ -27956,6 +27963,84 @@ exports[`Generated code for "jsii-calc": <runtime-type-check-diff>/java/src/main
...
      /**
       */
+@@ -249,35 +284,35 @@
+     }
+ 
+     /**
+      */
+     @software.amazon.jsii.Stability(software.amazon.jsii.Stability.Level.Stable)
+-    public void setUnionProperty(final @org.jetbrains.annotations.NotNull java.lang.String value) {
++    public void setUnionProperty(final @org.jetbrains.annotations.NotNull java.lang.String value) throws IllegalArgumentException {
```
comcalvi, line 27973: I know that the code gen changed how this is rendered; so, is this expected? There's a `throws` declaration, but no exception thrown here.

packages/jsii-pacmak/test/generated-code/__snapshots__/target-java.test.js.snap
```diff
@@ -27877,5 +27877,600 @@ exports[`Generated code for "jsii-calc": <outDir>/java/src/main/resources/softwa
...
+ 
+         software.amazon.jsii.Kernel.set(this, "unionProperty", java.util.Objects.requireNonNull(value, "unionProperty is required"));
+     }
+ 
+     /**
+      */
+     @software.amazon.jsii.Stability(software.amazon.jsii.Stability.Level.Stable)
+-    public void setUnionProperty(final @org.jetbrains.annotations.NotNull java.lang.Number value) {
++    public void setUnionProperty(final @org.jetbrains.annotations.NotNull java.lang.Number value) throws IllegalArgumentException {
```
comcalvi, line 27982: something is wrong here...this `throws` should not be here, but I'm not sure what's causing it.

packages/jsii-pacmak/lib/targets/dotnet/dotnetgenerator.ts
RomainMuller: No we need a runtime opt-out to exist, too.

packages/jsii-pacmak/lib/targets/java.ts
```diff
@@ -1513,12 +1522,211 @@ class JavaGenerator extends Generator {
...
+      this.code.line(
+        `.append(", (Parameter '${parameterName}')").toString());`,
+      );
+      this.code.closeBlock();
+
+      for (const typeRef of nestedCollectionUnionTypes) {
+        this.code.openBlock(
+          `if (${value} instanceof ${this.toJavaTypeNoGenerics(typeRef)})`,
+        );
+        validate.call(this, value, descr, typeRef, parameterName);
```
RomainMuller, line 1724: This leads to invalid Java apparently

```suggestion
        const varName = `__cast_${createHash('sha256').update(value).digest('hex').slice(0, 6)}`;
        const javaTypeFull = this.toJavaType(typeRef);
        this.code.line(`@SuppressWarnings("unchecked")`);
        this.code.line(`final ${javaTypeFull} ${varName} = (${javaTypeFull})${value};`);
        validate.call(this, varName, descr, typeRef, parameterName);
```

packages/jsii-pacmak/lib/targets/java.ts
```diff
@@ -1513,12 +1522,211 @@ class JavaGenerator extends Generator {
...
+      elementType: spec.TypeReference,
+      parameterName: string,
+    ) {
+      const varName = `__item_${createHash('sha256')
+        .update(descr)
+        .digest('hex')
+        .slice(0, 6)}`;
+      const javaElemType = this.toJavaType(elementType);
+      this.code.openBlock(
+        `for (java.util.Map.Entry<String, ${javaElemType}> ${varName}: ${value}.entrySet())`,
```
RomainMuller, line 1661:
```suggestion
        `for (final java.util.Map.Entry<String, ${javaElemType}> ${varName}: ${value}.entrySet())`,
```

packages/jsii-pacmak/lib/targets/java.ts
```diff
@@ -1513,12 +1522,211 @@ class JavaGenerator extends Generator {
...
+      this.code.line(`throw new IllegalArgumentException(`);
+      this.code.line(`new java.lang.StringBuilder("Expected ")`);
+      this.code.line(`${descr}`);
+      this.code.line(`.append(" to be one of: ")`);
+      this.code.line(`.append("${placeholders}")`);
+      this.code.line(`.append("; received ")`);
+      this.code.line(`.append(${value}.getClass())`);
+      this.code.line(
+        `.append(", (Parameter '${parameterName}')").toString());`,
+      );
```
RomainMuller, line 1717: The code this generates has weird/awkward indentation, and concatenates a bunch of string literals (it's more efficient to do this statically)...

This is better, but still has some hard-coded string concatenations from `descr`, which is not ideal.

```suggestion
      this.code.indent(`throw new IllegalArgumentException(`);
      this.code.indent(`new java.lang.StringBuilder("Expected ")`);
      this.code.line(descr);
      this.code.line(`.append(" to be one of: ${placeholders}; received ")`);
      this.code.line(`.append(${value}.getClass())`);
      this.code.line(
        `.append(", (Parameter '${parameterName}')").toString());`,
      );
      this.code.unindent(false);
      this.code.unindent(false);
```

RomainMuller (changes requested):

packages/@jsii/java-runtime-test/project/src/test/java/software/amazon/jsii/testing/TypeCheckingTest.java
```diff
@@ -0,0 +1,124 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+class StructAImplementer implements StructA, StructB {
```
RomainMuller, line 19: Make this `private static` nested under `TypeCheckingTest`.

packages/@jsii/java-runtime-test/project/src/test/java/software/amazon/jsii/testing/TypeCheckingTest.java
```diff
@@ -0,0 +1,124 @@
...
+        map.put("good", new StructAImplementer("present"));
+        map.put("bad", "Not a StructA or StructB");
+        ArrayList<Map<String, Object>> list = new ArrayList<Map<String, Object>>();
+        list.add(map);
+        Exception e = assertThrows(IllegalArgumentException.class, () ->
+        {
+            new ClassWithCollectionOfUnions(list);
+        });
+
+        assertEquals("Expected unionProperty.get(0).get(\"bad\") to be one of: software.amazon.jsii.tests.calculator.StructA, software.amazon.jsii.tests.calculator.StructB; received class java.lang.String, (Parameter 'unionProperty')", e.getMessage());
```
RomainMuller, line 48: The "(Parameter 'unionProperty')" part here is a .NET-ism (the .NET exception records the parameter name and ALWAYS appends it to the error message like so)... I don't mind it being present, though... But the parameter name is actually already part of the error message.

packages/jsii-pacmak/lib/targets/java.ts
```diff
@@ -1513,12 +1522,211 @@ class JavaGenerator extends Generator {
...
+    ) {
+      const varName = `__item_${createHash('sha256')
+        .update(descr)
+        .digest('hex')
+        .slice(0, 6)}`;
+      const javaElemType = this.toJavaType(elementType);
+      this.code.openBlock(
+        `for (java.util.Map.Entry<String, ${javaElemType}> ${varName}: ${value}.entrySet())`,
+      );
+      validate.call(
```
RomainMuller, line 1663: You'll also want to validate that keys are `String` if the `Map` was cast out of a union... In that case, the compiler will not have verified this...

```suggestion
      );
      this.code.openBlock(`if (!(${varName}.getKey() instanceof String))`);
      this.code.line(`throw new IllegalArgumentException("...")`);
      this.code.closeBlock();
      validate.call(
```

packages/jsii-pacmak/lib/targets/java.ts
```diff
@@ -1459,6 +1461,13 @@ class JavaGenerator extends Generator {
           this.code.openBlock(signature);
           let statement = '';
 
+          // do not emit these checks for primitive-only unions,
+          // because Java does not allow the check these in this case (eg if a String instanceof Number)
+          if (type.includes('java.lang.Object')) {
+            this.emitUnionParameterValdation([
+              { name: 'value', type: prop.type },
+            ]);
```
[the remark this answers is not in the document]
comcalvi: we discussed this offline, but adding the result here; setters have overloads (one for each possible type in the union), so we have
```
setProperty(string) { }
setProperty(number) {}
```

instead of 

```setProperty(Object) {}```

File: packages/jsii-pacmak/lib/targets/java.ts [hunk 11]
```diff
@@ -1513,12 +1522,211 @@ class JavaGenerator extends Generator {
...
+      return;
+    }
+
+    this.code.openBlock('if (software.amazon.jsii.Configuration.getRuntimeTypeChecking())');
+    for (const param of unionParameters) {
+      if (param.variadic) {
+        const javaType = this.toJavaType(param.type);
+        const asListName = `__${param.name}__asList`;
+        this.code.line(
+          `java.util.ArrayList<${javaType}> ${asListName} = new java.util.ArrayList<${javaType}>(java.util.Arrays.asList(${param.name}));`,
```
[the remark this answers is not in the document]
comcalvi: We can't do:

```
java.util.Arrays.<${javaType}>asList(${param.name});`,
```

because the `asList()` returns a fixed size list (meaning that subsequent `add()` operations will throw `UnsupportedOperationException()`.

packages/jsii-pacmak/lib/targets/java.ts
comcalvi: could you elaborate? I'm not able to reproduce this.

RomainMuller (approved):

PR #3709, opened 2022-08-17
fix(dotnet): AnonymousObject fails runtime type checks

RomainMuller:
The `AnonymousObject` class is used when a value is passed from JS to
.NET through a union or `any` typed return point, and there is not
sufficient runtime type information on the value to decisively identify
its dynamic type. The `AnonymousObject` class will be converted by the
jsii runtime to any interface type implicitly (even if that is
technically not correct), and it must hence be allowed through runtime
type-checks for interfaces. Essentially, this is a blind spot of the
runtime type checks, and should not cause valid code to break.

---


packages/jsii-pacmak/lib/targets/dotnet/dotnetgenerator.ts
```diff
@@ -719,7 +720,10 @@ export class DotNetGenerator extends Generator {
       elementType: spec.TypeReference,
       parameterName: string,
     ) {
-      const varName = `__idx_${descr.replace(/[^a-z0-9_]/gi, '_')}`;
+      const varName = `__idx_${createHash('sha256')
+        .update(descr)
+        .digest('hex')
+        .slice(0, 6)}`;
```
RomainMuller, line 726: Unrelated, but makes the generated code more readable, which makes it less likely that a bad change is missed by human scrutiny.

packages/jsii-pacmak/lib/targets/dotnet/dotnetgenerator.ts
```diff
@@ -740,7 +744,10 @@ export class DotNetGenerator extends Generator {
       elementType: spec.TypeReference,
       parameterName: string,
     ) {
-      const varName = `__item_${descr.replace(/[^a-z0-9_]/gi, '_')}`;
+      const varName = `__item_${createHash('sha256')
+        .update(descr)
+        .digest('hex')
+        .slice(0, 6)}`;
```
RomainMuller, line 750: Unrelated, but makes the generated code more readable, which makes it less likely that a bad change is missed by human scrutiny.

comcalvi (approved):

RomainMuller:
@MergifyIO refresh

PR #3712, opened 2022-08-18
feat(go): runtime type check type unions

RomainMuller:
Runtime type-checking can be disabled by adding
`--tag=no_runtime_type_checking` to go compiler invocations. This can
be used to avoid paying the performance toll of runtime type checking
(which should be modest in most cases), or to work around a bug in the
type checking code.

---


packages/@jsii/go-runtime-test/project/runtime_type_checking_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+import (
+	"fmt"
+	"github.com/aws/jsii-runtime-go"
+	"github.com/aws/jsii/jsii-calc/go/jsiicalc/v3"
+	"github.com/aws/jsii/jsii-calc/go/jsiicalc/v3/anonymous"
+	"testing"
+)
+
+func TestConstructor(t *testing.T) {
```
comcalvi, line 11: If we're going to add the nested type checking to .NET and Java, we should add it to Go as well.

packages/@jsii/go-runtime/jsii-runtime-go/internal/typeregistry/registration.go
```diff
@@ -150,3 +151,20 @@ func (t *TypeRegistry) RegisterStruct(fqn api.FQN, strct reflect.Type) error {
 
 	return nil
 }
+
+// RegisterStructValidator adds a validator function to an already registered struct type. This is separate call largely
+// to maintain backwards compatibility with existing code.
```
comcalvi, line 156: Why do we need to emit this separate call to maintain backwards compatibility, instead of just bundling this into the struct emitter?

comcalvi (changes requested):
just a few small comments! Looks great.

comcalvi (commented):
This is a massive change to the generated code, so I think it would be worthwhile to add a test to confirm that the runtime checks being disabled produces no changes - what do you think?

packages/@jsii/go-runtime/jsii-runtime-go/internal/typeregistry/registration.go
RomainMuller: In go, structs are "pure data", so there is no option to validate them at creation... In order to work around this, I have the go runtime library register validation functions for each struct type...

I need a call into the runtime library to actually inform it about a validation function... But I cannot add a new argument to `RegisterStruct` because this would break code generated against and older runtime library... so this new feature uses a new runtime call.

packages/jsii-pacmak/lib/targets/go/runtime/runtime-type-checking.ts
```diff
@@ -0,0 +1,645 @@
...
+      }
+      code.line('return nil');
+    }
+    code.closeBlock();
+    code.line();
+  }
+}
+
+abstract class Validation {
+  public static forTypeMap(
```
comcalvi, line 281: elegant solution that implicitly covers the nested unions here...nice work!

comcalvi (changes requested):
one tiny cosmetic comment

packages/jsii-pacmak/test/generated-code/harness.ts
```diff
@@ -231,7 +253,7 @@ function tryStat(at: string) {
   try {
     return fs.statSync(at);
   } catch (e: any) {
-    if (e.code !== os.constants.errno.ENOENT) {
+    if (e.code !== 'ENOENT') {
```
comcalvi, line 256: why change from the enum to a hardcoded string?

RomainMuller: Because the previous code was actually incorrect (`os.constants.errno.ENOENT` is actually the numeric value of the error code, not it's name...)

PR #3720, opened 2022-08-23
fix(.net): missing dynamic type checking for collection-nested unions

RomainMuller:
The .NET runtime type checking was not checking type unions that
are nested within a collection (list or map), although this is
necessary since the `is <type>` guard does not allow making any
assumption about the contents of the collection.

This adds the missing checks, and replaces the guard clauses
with a `switch` statement using pattern matching, resulting
in somewhat more elegant code (also with better `null` safety).

---

### DIFF
--- a/packages/jsii-pacmak/lib/targets/dotnet/dotnettyperesolver.ts
+++ b/packages/jsii-pacmak/lib/targets/dotnet/dotnettyperesolver.ts
@@ -243,15 +243,18 @@ export class DotNetTypeResolver {
    * Translates a collection in jsii to the name of a native .NET collection
    */
   private toDotNetCollectionName(ref: spec.CollectionTypeReference): string {
+    const [_, dollar, quote, content] = /^(?:(\$)?("))?([^"]+)"?$/.exec(
+      this.toDotNetTypeName(ref.collection.elementtype),
+    )!;
+
+    const interpolates = dollar || !quote ? '$' : '';
+    const elementTypeName = quote ? content : `{${content}}`;
+
     switch (ref.collection.kind) {
       case spec.CollectionKind.Array:
-        const elementDotNetTypeName = this.toDotNetTypeName(
-          ref.collection.elementtype,
-        );
-        return `$"{${elementDotNetTypeName}}[]"`;
+        return `${interpolates}"${elementTypeName}[]"`;
       case spec.CollectionKind.Map:
-        const elementDotNetType = this.toDotNetType(ref.collection.elementtype);
-        return `typeof(System.Collections.Generic.IDictionary<string, ${elementDotNetType}>).FullName`;
+        return `${interpolates}"System.Collections.Generic.IDictionary<string, ${elementTypeName}>"`;
       default:
         throw new Error(
           `Unsupported collection kind: ${ref.collection.kind as any}`,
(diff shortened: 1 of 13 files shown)


packages/jsii-pacmak/lib/targets/dotnet/dotnettyperesolver.ts
```diff
@@ -243,15 +243,18 @@ export class DotNetTypeResolver {
...
-        const elementDotNetTypeName = this.toDotNetTypeName(
-          ref.collection.elementtype,
-        );
-        return `$"{${elementDotNetTypeName}}[]"`;
+        return `${interpolates}"${elementTypeName}[]"`;
       case spec.CollectionKind.Map:
-        const elementDotNetType = this.toDotNetType(ref.collection.elementtype);
-        return `typeof(System.Collections.Generic.IDictionary<string, ${elementDotNetType}>).FullName`;
+        return `${interpolates}"System.Collections.Generic.IDictionary<string, ${elementTypeName}>"`;
       default:
```
RomainMuller, line 258: Tweaked here, because `nameof` on the generic type does not produce a string that is human-friendly, unfortunately.

packages/jsii-pacmak/lib/targets/dotnet/runtime-type-checking.ts
```diff
@@ -0,0 +1,242 @@
...
+  public emit(code: CodeMaker, resolver: DotNetTypeResolver): void {
+    for (const [_parameter, validations] of this.validations) {
+      for (const validation of validations) {
+        validation.emit(code, resolver);
+      }
+    }
+  }
+}
+
+abstract class Validation {
```
comcalvi, line 61: Should I use this approach in the Java code gen as well, or stick to the previous way?

packages/jsii-pacmak/lib/targets/dotnet/runtime-type-checking.ts
```diff
@@ -0,0 +1,242 @@
...
+    class UnionCheck extends Validation {
+      public emit(code: CodeMaker, resolver: DotNetTypeResolver): void {
+        const validTypes = new Array<string>();
+
+        const castVarName = `cast_${createHash('sha256')
+          .update(expression)
+          .digest('hex')
+          .slice(0, 6)}`;
+
+        code.openBlock(`switch (${expression})`);
```
comcalvi, line 164: Do we want this to be consistent across languages when possible? Should I port this `switch` approach to Java as well, or leave it as-is?

comcalvi (commented):
That was pretty fast, well done!

comcalvi (approved):

packages/jsii-pacmak/lib/targets/dotnet/runtime-type-checking.ts
RomainMuller: You can't do `switch` like this in Java... it only works with primitive types, enums, String, and certain boxed primitive classes. So you end up having to stick to `if`/`else` here, sorry!

packages/jsii-pacmak/lib/targets/dotnet/runtime-type-checking.ts
RomainMuller: You can if you want, but you can also migrate in a separate PR.

PR #3723, opened 2022-08-25
fix(rosetta): rosetta reader expects default tablet file to be uncompressed

kaizencc:
Due to a bug in `rosetta-reader.ts`, we always expect to find the implicit tablet at `.jsii.tabl.json`. This means that any compressed implicit tablets will be unable to be used as a cache, and causes the following failure:

```bash
The following snippet was not found in any of the loaded tablets: ...
```

This PR also slightly changes the default behavior of `compressTablet`; previously it was `false` by default. Now it respects the original compression status of the implicit tablet, i.e. if it was previously compressed, then `extract` will update it as compressed.

---


packages/jsii-rosetta/lib/rosetta-reader.ts
```diff
@@ -148,7 +147,7 @@ export class RosettaTabletReader {
    * pacmak sends our way later on is not going to be enough to do that.
    */
   public async addAssembly(assembly: spec.Assembly, assemblyDir: string) {
-    const defaultTablet = path.join(assemblyDir, DEFAULT_TABLET_NAME);
+    const defaultTablet = guessTabletLocation(assemblyDir);
```
kaizencc, line 150: argh. This is the bug that I've been searching for.

RomainMuller (approved):

madeline-k (approved):

Issue #3389, opened 2022-02-17
Imports of large Python library (aws-cdk-lib) extremely slow


rix0rrr:
### Affected Languages
- [ ] `TypeScript` or `Javascript`
- [x] `Python`
- [ ] `Java`
- [ ] .NET (`C#`, `F#`, ...)
- [ ] `Go`

### General Information
* **JSII Version:** N/A
* **Platform:**     MacOS

### What is the problem?

Originally reported as https://github.com/aws/aws-cdk/issues/19000

----------------------------------------

I noticed that simple commands such as `cdk ls` would take very long (sometimes above 20 seconds, rarely much less) to complete. This in a project created using `cdk init sample-app --language python`. 

### Reproduction Steps

```shell
# setup:
$ cdk init sample-app --language python
$ source .venv/bin/activate
$ pip install -r requirements.txt
# reproducing issue:
$ time python -c "import aws_cdk as cdk"
python -c "import aws_cdk as cdk"  5.74s user 4.20s system 58% cpu 17.113 total
```

### What did you expect to happen?

A simple import should not take longer than 1 second (ideally even less, but there are constrains that make that diffucult here I understand). 

### What actually happened?

As seen above, it takes ~17 seconds. This is quite consistent, however first or second time after initing the project, it can take 20-25 seconds.

### CDK CLI Version

2.12.0 (build c9786db)

### Framework Version

-

### Node.js Version

v16.14.0

### OS

MacOS 11.6.2 (20G314) - Intel Macbook

### Language

Python

### Language Version

Python (3.8.9, 3.9.10)

### Other information

To better understand where the time went, I debugged the issue with some print statements:

```
❯ time python -c "import aws_cdk as cdk"
process.py start: 0.00
self._process.args=['node', '--max-old-space-size=4069', '/var/folders/9j/cszr_w557ns8q4bk2ctzygx40000gp/T/tmp9_s6gne5/bin/jsii-runtime.js']
>>> b'{"hello":"@jsii/runtime@1.54.0"}\n'
process.py before self._next_message(): 0.25
>>> b'{"ok":{"assembly":"constructs","types":10}}\n'
process.py after self._next_message(): 0.33
process.py before self._next_message(): 0.33
>>> b'{"ok":{"assembly":"aws-cdk-lib","types":7639}}\n'
process.py after self._next_message(): 12.95
python -c "import aws_cdk as cdk"  5.72s user 4.21s system 58% cpu 16.923 total
```

[...]

PR #3724, opened 2022-08-26
feat(kernel): experimental runtime package cache

RomainMuller:
Adds an experimental (hence opt-in) feature that caches the contents of
loaded libraries in a directory that persists between executions, in
order to spare the time it takes to extract the tarballs.

When this feature is enabled, packages present in the cache will be used
as-is (i.e: they are not checked for tampering) instead of being
extracted from the tarball. The cache is keyed on:
- The hash of the tarball
- The name of the library
- The version of the library

Objects in the cache will expire if they are not used for 30 days, and
are subsequently removed from disk (this avoids a cache growing
extremely large over time).

In order to enable the feature, the following environment variables are
used:
- `JSII_RUNTIME_PACKAGE_CACHE` must be set to `enabled` in order for the
  package cache to be active at all;
- `JSII_RUNTIME_PACKAGE_CACHE_ROOT` can be used to change which
  directory is used as a cache root. It defaults to:
  * On MacOS: `$HOME/Library/Caches/com.amazonaws.jsii`
  * On Linux: `$HOME/.cache/aws/jsii/package-cache`
  * On Windows: `%LOCALAPPDATA%\AWS\jsii\package-cache`
  * On other platforms: `$TMP/aws-jsii-package-cache`
- `JSII_RUNTIME_PACKAGE_CACHE_TTL` can be used to change the default
  time entries will remain in cache before expiring if they are not
  used. This defaults to 30 days, and the value is expressed in days.
  Set to `0` to immediately expire all the cache's content.

When troubleshooting load performance, it is possible to obtain timing
data for some critical parts of the library load process within the jsii
kernel by setting `JSII_DEBUG_TIMING` environment variable.

Related to #3389

---


packages/@jsii/kernel/src/disk-cache/disk-cache.ts
```diff
@@ -0,0 +1,244 @@
...
+  }
+
+  public entryFor(path: string, ...comments: readonly string[]): Entry {
+    const rawDigest = digestFile(path, ...comments);
+    return this.entry(...comments, rawDigest.toString('hex'));
+  }
+
+  public pruneExpiredEntries() {
+    const cutOff = new Date(Date.now() - PRUNE_AFTER_MILLISECONDS);
+    for (const entry of this.entries()) {
```
Naumel, line 84: [Q] Can this lead to a strange part-old part to be renewed situation and would it matter? Are there dependencies between the cached items?

packages/@jsii/kernel/src/disk-cache/disk-cache.ts
```diff
@@ -0,0 +1,244 @@
...
+  public lock<T>(cb: (entry: LockedEntry) => T): T {
+    mkdirSync(dirname(this.path), { recursive: true });
+    lockSync(this.lockFile, { retries: 12, stale: 5_000 });
+    let disposed = false;
+    try {
+      return cb({
+        delete: () => {
+          if (disposed) {
+            throw new Error(
+              `Cannot delete ${this.path} once the lock block as returned!`,
```
Naumel, line 164: [nitpick] Is `as` supposed to be `has`, or should a verb be part of the second sentence?

packages/@jsii/kernel/src/disk-cache/disk-cache.ts
```diff
@@ -0,0 +1,244 @@
...
+    try {
+      const stat = statSync(path);
+      if (stat.isDirectory()) {
+        if (recursive) {
+          yield* directoriesUnder(path, recursive, ignoreErrors);
+        }
+        yield path;
+      }
+    } catch (error) {
+      if (!ignoreErrors) {
```
Naumel, line 239: Is there a scenario in which the errors should both be captured (for investigation) and ignored?

packages/@jsii/kernel/src/tar-cache/default-cache-root.ts
```diff
@@ -0,0 +1,28 @@
...
+  switch (process.platform) {
+    case 'darwin':
+      if (process.env.HOME)
+        return join(
+          process.env.HOME,
+          'Library',
+          'Caches',
+          'com.amazonaws.jsii',
+          'package-cache',
+        );
```
Naumel, line 14: [nitpick] Line went over and got wrapped?
the other `join`s are single-line.

packages/@jsii/kernel/src/tar-cache/index.ts
```diff
@@ -0,0 +1,119 @@
...
+  /**
+   * The path in which the extracted files are located
+   */
+  readonly path: string;
+
+  /**
+   * When `'hit'`, the data was already present in cache and was returned from
+   * cache.
+   *
+   * When `'miss'`, the data was extracted into the caache and returned from
```
Naumel, line 24: `caache` --> `cache`

packages/@jsii/kernel/src/tar-cache/default-cache-root.ts
RomainMuller: Yeah, this one is just a bit longer and was wrapped by `prettier`. Indentation/formatting is entirely prettify-driven in the jsii repository.

packages/@jsii/kernel/src/disk-cache/disk-cache.ts
RomainMuller: Maybe... I decided for this to be as silent as possible, to reduce the chances it causes something to fail that wouldn't have if the cache wasn't enabled.

packages/@jsii/kernel/src/disk-cache/disk-cache.ts
RomainMuller: Cached items are literally "the extracted content of a tarball". They technically don't depend on each other - the "installs" of these are copies/hard-links in another directory (temporary). If/when a directory is removed from cache, the corresponding installs remain "valid" (since they were hard linked into place).

The only consequence of expiring something from this cache is that it's going to need to be re-extracted next time...

Naumel:
@Mergifyio update

Naumel (approved):

RomainMuller:
@mergifyio update

PR #3747, opened 2022-09-08
feat(kernel): add `JsiiFault` error type

comcalvi:
Adds the `JsiiFault` type to the Kernel to represent irrecoverable errors. 

---

### DIFF
--- a/packages/@jsii/kernel/src/kernel.ts
+++ b/packages/@jsii/kernel/src/kernel.ts
@@ -15,10 +15,18 @@ import * as onExit from './on-exit';
 import * as wire from './serialization';
 import * as tar from './tar-cache';
 
-export class JsiiFault extends Error {
+export enum JsiiErrorType {
+  JSII_FAULT = 'jsii-fault',
+}
+
+export interface JsiiError extends Error {
+  readonly type: JsiiErrorType;
+}
+
+export class JsiiFault extends Error implements JsiiError {
+  public readonly type = JsiiErrorType.JSII_FAULT;
   public constructor(message: string) {
     super(message);
-    this.name = 'JsiiFault';
   }
 }
 
--- a/packages/@jsii/kernel/src/kernel.ts
+++ b/packages/@jsii/kernel/src/kernel.ts
@@ -15,6 +15,13 @@ import * as onExit from './on-exit';
 import * as wire from './serialization';
 import * as tar from './tar-cache';
 
+export class JsiiFault extends Error {
+  public constructor(message: string) {
+    super(message);
+    this.name = 'JsiiFault';
+  }
+}
+
 export class Kernel {
   /**
    * Set to true for verbose debugging.
@@ -56,7 +63,7 @@ export class Kernel {
     this._debug('load', req);
 
     if ('assembly' in req) {
-      throw new Error(
+      throw new JsiiFault(
         '`assembly` field is deprecated for "load", use `name`, `version` and `tarball` instead',
       );
     }
@@ -70,7 +77,7 @@ export class Kernel {
       // module exists, verify version
       const epkg = fs.readJsonSync(path.join(packageDir, 'package.json'));
       if (epkg.version !== pkgver) {
-        throw new Error(
+        throw new JsiiFault(
           `Multiple versions ${pkgver} and ${epkg.version} of the ` +
             `package '${pkgname}' cannot be loaded together since this is unsupported by ` +
             'some runtime environments',
@@ -135,7 +142,9 @@ export class Kernel {
         `loadAssemblyFromPath(${packageDir})`,
       );
     } catch (e: any) {
-      throw new Error(`Error for package tarball ${req.tarball}: ${e.message}`);
+      throw new JsiiFault(
+        `Error for package tarball ${req.tarball}: ${e.message}`,
+      );
     }
 
     // load the module and capture its closure
@@ -166,13 +175,15 @@ export class Kernel {
       const epkg = fs.readJsonSync(path.join(packageDir, 'package.json'));
 
       if (!epkg.bin) {
-        throw new Error('There is no bin scripts defined for this package.');
+        throw new JsiiFault(
+          'There is no bin scripts defined for this package.',
+        );
       }
 
       const scriptPath = epkg.bin[req.script];
 
       if (!epkg.bin) {
-        throw new Error(`Script with name ${req.script} was not defined.`);
+        throw new JsiiFault(`Script with name ${req.script} was not defined.`);
       }
 
       const result = cp.spawnSync(
@@ -198,7 +209,7 @@ export class Kernel {
         signal: result.signal,
       };
     }
-    throw new Error(`Package with name ${req.assembly} was not loaded.`);
+    throw new JsiiFault(`Package with name ${req.assembly} was not loaded.`);
   }
 
   public create(req: api.CreateRequest): api.CreateResponse {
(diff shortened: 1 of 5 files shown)


comcalvi:
In order to surface any of this to the user, we need to pass the error type over the wire to the runtime. It's not clear where exactly that happens.

MrArnoldPalmer:
> In order to surface any of this to the user, we need to pass the error type over the wire to the runtime. It's not clear where exactly that happens.

This is where that happens basically https://github.com/aws/jsii/pull/3747/files#diff-0c5a3339fc4062c6aadbff20267f2a7bd57f04149e33984a6478e7664eaf0f4eR182

Write error is what serializes the error object before being passed to the stdout stream for writing.

packages/@jsii/java-runtime/project/src/main/java/software/amazon/jsii/JsException.java
```diff
@@ -0,0 +1,33 @@
+package software.amazon.jsii;
+
+/*
+ * Represents an irrecoverable error, such as a
```
MrArnoldPalmer, line 4: I think this should be "potentially recoverable" right as JSError's may just be basic input validation etc on the lib side.

packages/@jsii/kernel/src/kernel.ts
```diff
@@ -1278,6 +1306,13 @@ export class Kernel {
     this.syncInProgress = desc;
     try {
       return fn();
+    } catch (e: any) {
+      if (e.name === JsiiErrorType.JSII_FAULT) {
+        throw new JsiiFault(e);
+      }
+      // default to JsError because any JsiiFaults have been emitted from the kernel
+      // with their type field populated before the request was made to the js code.
+      throw new JsError(e);
```
MrArnoldPalmer, line 1315: I think we should probably default to `JsiiFault` and explicitly check for other error types as errors without a `name` set are unknown and likely not recoverable.

MrArnoldPalmer (changes requested):
Couple small things, otherwise straightforward enough and easy to iterate on for further functionality 👍🏻

packages/@jsii/kernel/src/kernel.ts
comcalvi: discussed offline...this needs to default to JsError since the Faults will have their type specified, and JsErrors won't

MrArnoldPalmer (approved):

packages/@jsii/kernel/src/kernel.ts
```diff
@@ -15,6 +15,28 @@ import * as onExit from './on-exit';
 import * as wire from './serialization';
 import * as tar from './tar-cache';
 
+export enum JsiiErrorType {
+  JSII_FAULT = '@jsii/kernel.Fault',
+  JS_ERROR = '@jsii/kernel.JSError',
```
RomainMuller, line 20: While I think `JavascriptError`/`JavascriptException` "makes sense" in other languages, I think this isn't quite descriptive of the semantics in the TypeScript/JS code... Literally every exception being handled here is a Javascript error...

So while I think it may be okay to use `JavascriptException` in Java (I mean we could bike shed here - do we *need* to surface the fact this was JS? Maybe... maybe not?), the JavaScript code should refer to this as a something more along the lines of `@jsii/kernel.RuntimeError` or maybe even `@jsii/kernel.ForwardedRuntimeError`.

The enum should likely also have JSDoc on it...

RomainMuller (changes requested):

packages/@jsii/java-runtime/project/src/main/java/software/amazon/jsii/JSException.java
```diff
@@ -0,0 +1,32 @@
+package software.amazon.jsii;
+
+/*
+ * Represents an error thrown from the user JS library.
+ */
+public final class JSException extends JsiiBaseException {
```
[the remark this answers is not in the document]
comcalvi: this will be inherited from the base class though, no?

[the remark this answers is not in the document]
RomainMuller: Statics are not inherited nor inheritable. Java allows you to refer to a static declared on a parent type though a child type, but this is actually not inheritance (and best practices encourage you to avoid doing that, precisely because the semantics might change if the child class adds a new declaration, shadowing -- and not overriding -- the one from the parent class).

packages/@jsii/java-runtime-test/project/src/test/java/software/amazon/jsii/testing/ComplianceTest.java
```diff
@@ -335,7 +335,7 @@ public void exceptions() {
         assertEquals(23, calc3.getValue());
         boolean thrown = false;
         try { calc3.add(10); }
-        catch (Exception e) { thrown = true; }
+        catch (RuntimeException e) { thrown = true; }
```
RomainMuller, line 338: What is this indentation style? Ewww 🤣

packages/@jsii/java-runtime/project/src/main/java/software/amazon/jsii/JsiiError.java
```diff
@@ -0,0 +1,32 @@
+package software.amazon.jsii;
+
+/**
+ * A nonrecoverable error from the jsii runtime,
+ * usually the kernel.
+ */
+public class JsiiError extends JsiiException {
+  /**
```
RomainMuller, line 8:
```suggestion
public final class JsiiError extends JsiiException {
  public static final long serialVersionUID = 1L;

  /**
```

packages/@jsii/java-runtime/project/src/main/java/software/amazon/jsii/JsiiException.java
```diff
@@ -1,33 +1,48 @@
...
     public static final long serialVersionUID = 1L;
 
-    /**
-     * Creates an exception.
-     * @param message The error message
-     */
-    JsiiException(final String message) {
-        super(message);
-    }
+    public static enum Type {
```
RomainMuller, line 9: I don't think this needs to/should be public. Package-private should be enough.

```suggestion
   static enum Type {
```

packages/@jsii/java-runtime/project/src/main/java/software/amazon/jsii/JsiiRuntime.java
```diff
@@ -76,7 +76,7 @@ public final class JsiiRuntime {
      *
      * @param request The JSON request
      * @return The JSON response
-     * @throws JsiiException If the runtime returns an error response.
+     * @throws RuntimeException or JsiiError If the runtime returns an error response.
```
RomainMuller, line 79: This is not how you code up `@throws`:

```suggestion
     * @throws JsiiError If the runtime returns an error response originating from the @jsii/kernel.
     * @throws RuntimeException If the runtime returns an error response.
```

packages/@jsii/kernel/src/kernel.ts
```diff
@@ -15,6 +15,28 @@ import * as onExit from './on-exit';
 import * as wire from './serialization';
 import * as tar from './tar-cache';
 
+export enum JsiiErrorType {
```
RomainMuller, line 18: We can use a `const enum` here -- this is a shared contract between the runtimes & the JavaScript kernel, and so in principle the string versions can no longer change without it being a breaking change across the board. At this stage, constant propagation (which is what `const enum` does in TypeScript) is what'll have the smallest performance impact.

```suggestion
export const enum JsiiErrorType {
```

RomainMuller (changes requested):
Almost there :)

packages/@jsii/java-runtime-test/project/src/test/java/software/amazon/jsii/testing/ComplianceTest.java
comcalvi: that's been there since the first external release...yikes I'll update that

RomainMuller (approved):

Issue #3319, opened 2022-01-06
jsii 1.50.0 breaks for monorepos with local dependencies


mattsb42-aws:
### Affected Languages
- [x] `TypeScript` or `Javascript`
- [ ] `Python`
- [ ] `Java`
- [ ] .NET (`C#`, `F#`, ...)
- [ ] `Go`

### General Information
* **JSII Version:** 1.50.0
* **Platform:**     MacOS but irrelevant

### What is the problem?

The dependency resolver added in 1.50.0[1] breaks on local dependencies. Because the `declaration` value is a file URI but the `actualVersion` is whatever version string the dependency package defines, these values can never match. This causes `jsii` to fail when it attempts to compile the package.

[1] https://github.com/aws/jsii/blob/d34881c35a56302ef2a1c3e8a0d19995c3344720/packages/jsii/lib/project-info.ts#L238-L267

### Verbose Log

```
👾 build » compile | jsii --silence-warnings=reserved-word --no-fix-peer-dependencies
Error: Error: Declared dependency on version file:../<REDACTED> of <REDACTED>, but version <REDACTED> was found
```

wchaws:
Encounter the same issue

MrArnoldPalmer:
@mattsb42-aws to clarify this occurs only when using a file path as the dependency version in package.json, but `npm link`ing works okay right? I think we mostly link in our workflow hence why this wasn't caught, looks like you're adding a test in your PR though.

mattsb42-aws:
Correct. Because `declared` is not a version, it will always fail because it can never be a reasonable match for `actualVersion`. We haven't tried `npm link`; I'll look into that, but even if that works for us I still think this is worth fixing.

mattsb42-aws:
It would also affect any other URI.

https://docs.npmjs.com/cli/v7/configuring-npm/package-json#dependencies

PR #3750, opened 2022-09-13
fix: allow declaration URIs fixes: #3319

mattsb42-aws:
As described in #3319 the resolver introduced in 1.150.0 breaks URI dependency definitions (commonly used in monorepos.) This _should_ fix the issue.

---

### DIFF
--- a/packages/jsii/lib/project-info.ts
+++ b/packages/jsii/lib/project-info.ts
@@ -313,7 +313,7 @@ class DependencyResolver {
       const actualVersion = resolved.dependencyInfo.assembly.version;
       if (!semver.satisfies(actualVersion, declaration)) {
         // Allow URI declarations
-        if (!declaration.match(/^[a-z][A-Z]:/)) {
+        if (!declaration.match(/^[a-z]+:/)) {
           throw new Error(
             `Declared dependency on version ${declaration} of ${name}, but version ${actualVersion} was found`,
           );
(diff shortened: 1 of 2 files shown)


mattsb42-aws:
not sure why that one check failed once, but re-running it seemed to work ¯\\\_(ツ)_/¯

packages/jsii/lib/project-info.ts
```diff
@@ -312,9 +312,12 @@ class DependencyResolver {
 
       const actualVersion = resolved.dependencyInfo.assembly.version;
       if (!semver.satisfies(actualVersion, declaration)) {
-        throw new Error(
-          `Declared dependency on version ${declaration} of ${name}, but version ${actualVersion} was found`,
-        );
+        // Allow URI declarations
+        if (!declaration.match(/^[a-z]+:/)) {
```
RomainMuller, line 316: What gets recorded in the output `.jsii` assembly in this case? Does this produce correct dependency declarations in other languages once consumed by `jsii-pacmak`?

mattsb42-aws: That is an excellent question. Where would I look for find that?

RomainMuller: On a project that uses this feature, you'd look at the `dependencies` and `dependencyClosure` keys within the `.jsii` file created as a result of compilation...

And then you'd generate code with `jsii-pacmak --code-only` and inspect dependency declarations in `pom.xml` for Java, the `.csproj` file for .NET, the `go.mod` file for Go, or the `setup.py` file for Python (reviewing one of these should be enough - they should all have comparable declarations).

I'm trying to think of a way we can have a test for this as part of the codebase, too... I think we could change one of `@scope/jsii-calc-*` packages to use a `file://` URI specifier and that hopefully does not break the mono repo structure we have today...

PR #3752, opened 2022-09-19
fix(jsii): unable to return Promise<void>

### LINKED ISSUE #51: Unable to return Promise<void>
While methods are fine to return `void`, they cannot return a `Promise<void>`, as JSII interprets this as `Promise<unknown type>`.

Reproduction:
```ts
export class PromiseOfVoid {
  async function returnsVoidPromise(): Promise<void> {
    return;
  }
}
```

RomainMuller:
The void-check only accounted for the literal `void` type, but failed to account for the `Promise<void>` case. This is now fixed.

Fixes #51

---


Naumel (approved):

Chriscbr:
Thanks @RomainMuller! 🙏

PR #3772, opened 2022-09-26
feat(python): add error differentiation to Python.

comcalvi:
Uses `RuntimeError` and `JsiiError` to indicate recoverable vs non-recoverable errors respectively. Also fixes two bugs in the Java logic for the same feature:

* Callbacks did not pass the error type over the wire, meaning that some faults were rethrown as runtime errors and vice versa
* async method invocations did not pass the error type over the wire, meaning that some faults were rethrown as runtime errors and vice versa.

---


packages/@jsii/java-runtime-test/project/src/test/java/software/amazon/jsii/testing/ComplianceTest.java
```diff
@@ -337,6 +337,7 @@ public void exceptions() {
         try {
           calc3.add(10);
         } catch (RuntimeException e) {
+          assertEquals(RuntimeException.class, e.getClass());
```
RomainMuller, line 340: This is _fine_, but you could also have used the fact that catch clauses are evaluated in declaration order and done something like this (the same applies to the other places):

```suggestion
        } catch (JsiiException e) {
          // We expect a RuntimeException that is NOT a JsiiException.
          throw e;
        } catch (RuntimeException e) {
```

packages/@jsii/java-runtime-test/project/src/test/java/software/amazon/jsii/testing/JsiiClientTest.java
```diff
@@ -146,19 +147,56 @@ public void asyncMethodOverridesThrow() {
...
+
+        // now complete the callback with an error
+        client.completeCallback(first, "The kernel encountered a fault", "@jsii/kernel.Fault", null);
+
+        // end the async invocation, but now we expect the value to be different since we override the method.
+        boolean thrown = false;
+        try {
+            client.endAsyncMethod(promise);
+        } catch (JsiiError e) {
+            assertEquals(JsiiError.class, e.getClass());
```
RomainMuller, line 193: Is this test really necessary? Would a subclass of `JsiiError` not be acceptable?

packages/@jsii/java-runtime/project/src/main/java/software/amazon/jsii/JsiiEngine.java
```diff
@@ -486,7 +486,11 @@ private Object invokeMethod(final Object obj, final Method method, final Object.
                 throw e;
             }
         } catch (InvocationTargetException e) {
-            throw new JsiiError(e.getTargetException());
+            if (e.getTargetException().getClass().equals(JsiiError.class)) {
+                throw new JsiiError(e.getTargetException());
```
RomainMuller, line 490: Could/should this be:

```suggestion
            if (e.getTargetException() instanceof JsiiError) {
                throw (JsiiError)(e.getTargetException();
```

I'm not sure there is much interest in re-wrapping that, it'll only bury the actual error location further down?

packages/@jsii/java-runtime/project/src/main/java/software/amazon/jsii/JsiiEngine.java
```diff
@@ -486,7 +486,11 @@ private Object invokeMethod(final Object obj, final Method method, final Object.
...
+            } else {
+                throw new RuntimeException(e.getTargetException());
+            }
```
RomainMuller, line 493: Similarly to above, I would:

```suggestion
            } else if (e.getTargetException() instanceof RuntimeException) {
                throw (RuntimeException)(e.getTargetException();
            } else {
                // Can't just throw a checked error without wrapping it :(
                throw new RuntimeException(e.getTargetException());
            }
```

packages/@jsii/java-runtime/project/src/main/java/software/amazon/jsii/JsiiEngine.java
```diff
@@ -502,9 +506,9 @@ private Object invokeMethod(final Object obj, final Method method, final Object.
     private void processCallback(final Callback callback) {
         try {
             JsonNode result = handleCallback(callback);
-            this.getClient().completeCallback(callback, null, result);
-        } catch (JsiiError e) {
-            this.getClient().completeCallback(callback, e.getMessage(), null);
+            this.getClient().completeCallback(callback, null, null, result);
+        } catch (Exception e) {
+            this.getClient().completeCallback(callback, e.getMessage(), null, null);
```
RomainMuller, line 511: Shouldn't we attempt to forward the `Exception`'s `name` through there?

packages/@jsii/java-runtime/project/src/main/java/software/amazon/jsii/JsiiRuntime.java
```diff
@@ -154,12 +155,18 @@ private JsonNode processCallbackResponse(final JsonNode resp) {
             } else {
                 error = e.getMessage();
             }
+
+            String errClass = e.getClass().toString();
+            name = errClass.equals("class software.amazon.jsii.JsiiError")
```
RomainMuller, line 160: Would recommend using `e.getClass().getCanonicalName()` here instead of just `toString`:

```suggestion
            String errClass = e.getClass().getCanonicalName();
            name = errClass.equals("software.amazon.jsii.JsiiError")
```

But actually, you could also just use an `instanceof` check:

```suggestion
            name = e instanceof JsiiError
```

packages/@jsii/kernel/src/kernel.ts
```diff
@@ -482,9 +482,15 @@ export class Kernel {
...
-    } catch (e) {
+    } catch (e: any) {
       this._debug('promise error:', e);
-      throw new JsiiFault((e as any).message);
+      if (e.name === JsiiErrorType.JSII_FAULT) {
+        throw new JsiiFault(e.message);
+      }
+
+      // default to RuntimeError, since non-kernel errors will not
+      // have their `name` field defined
```
RomainMuller, line 492: They might... just one one the kernel decided on/controls.

RomainMuller (approved):

PR #3790, opened 2022-10-04
chore(spec): improve assembly validation error messages

RomainMuller:
use the ajv error message formatter instead of only outputting the `e.message` as the later fails to include useful information to determine where the actual error is.

Turn the validation test into a snapshot test to draw attention on the error message format.

---


Naumel (approved):
Nice!

revmischa:
Thank you!

PR #3794, opened 2022-10-06
feat(python): streaming bin script output

### LINKED ISSUE #3782: Streaming bin script output in non-typescript kernels
### Describe the feature

Pipe the output streams of bin scripts invoked from the JSII kernel. Currently, these calls are made using the synchronous `InvokeScriptRequest` / `InvokeScriptResponse` synchronous API. The output is buffered until the call completes. This experience is not ideal for longer running scripts with progressive updates.

### Use Case

Long-running scripts with progressive updates, such as `projen deploy` called under Python via the JSII kernel, rather than a globally installed installed `projen` instance.

These calls were recently made to work via a few updates:
 - https://github.com/aws/jsii/pull/3762
 - https://github.com/projen/projen/pull/2132
 - https://github.com/projen/projen/issues/2103

The experience, however, is not ideal, due to the buffered output.

### Proposed Solution

Maintain the current synchronous (`InvokeScriptRequest` / `InvokeScriptResponse`) API for backwards compatibility, but add a new streaming API that language runtimes can opt into if they are on a platform that supports it. The opt-in approach will allow this to be implemented progressively.

I suspect this will require out-of-band IPC ([local domain sockets or named pipes](https://nodejs.org/api/net.html#ipc-support)) so that it doesn't disrupt the standard kernel connection streams.

I am not an IPC expert, but one (naive) approach on POSIX OSs would be:
 1. language runtime dynamically creates local sockets to receive the target process `stdout` and `stderr`
 2. sockets/FDs identifiers are passed to kernel via a new `InvokeScriptWithStreams` call
 3. kernel opens streams to the argument file descriptors, passes them to `spawnSync` of the target process
 4. language runtime receives data on sockets, streams to `stdout` / `stderr` on the user process
 5. kernel process completes, returns exit code, probably using existing `InvokeScriptResponse` structure

I have a nervous suspicion it is more complicated than this (`termcap`?), and I wouldn't know where to start with Windows. I definitely welcome thoughts on the approach.

### Other Information

My team is very interested in building out a first-rate Python dev-tool experience for JSII based libraries, specifically CDK on projen.

### Acknowledgements

- [X] I may be able to implement this feature request

### CDK version used

2.x

MacOS, Linux

jmalins:
Resolves #3782.

Inverts control on bin script execution by returning the script to execute (as well as arguments and ENV variables), rather than running it within the kernel process. This allows the calling process to pipe the output streams, rather than the kernel buffering the output and returning it at the end of the synchronous call.

Also, adds more complete tests to the python runtime to validate `stderr`, exit codes and streaming response.

---


jmalins:
Note, the kernel recording tests needed to be disabled for the `getBinScriptCommand` method since the output is not deterministic due to the tarball extraction directory being different between runs. This method is tested in the Python runtime tests, so there should be reasonable assurance it is works.

jmalins:
@RomainMuller, are there any changes you'd recommend to this PR?

packages/@jsii/kernel/src/kernel.test.ts
```diff
@@ -78,6 +78,10 @@ defineTest.skip = function (
   return defineTest(name, method, test.skip);
 };
 
+defineTest.skipIf = function (expr: boolean): typeof defineTest.skip {
+  return expr ? defineTest.skip : defineTest;
+};
```
RomainMuller, line 83:
```suggestion
defineTest.skipIf = (expr: boolean) => expr ? defineTest.skip : defineTest;
```

packages/@jsii/kernel/src/kernel.ts
```diff
@@ -1325,6 +1304,44 @@ export class Kernel {
...
+        throw new JsiiFault(
+          'There is no bin scripts defined for this package.',
+        );
+      }
+
+      const scriptPath = epkg.bin[req.script];
+
+      if (!epkg.bin) {
+        throw new JsiiFault(`Script with name ${req.script} was not defined.`);
+      }
```
RomainMuller, line 1328: I would consider those two failure modes to be the same one, actually...

```suggestion
      const scriptPath = epkg.bin?.[req.script];

      if (!epkg.bin) {
        throw new JsiiFault(`Script with name ${req.script} was not defined.`);
      }
```

packages/@jsii/kernel/src/kernel.ts
```diff
@@ -1325,6 +1304,44 @@ export class Kernel {
...
+
+      return {
+        command: path.join(packageDir, scriptPath),
```
RomainMuller, line 1331: On Windows, this may only work if the caller spawns that commend via a `shell`, which is a relatively easy way to inadvertently make your code not work on Windows...

There would be a `.cmd` file that might be preferable to use as an entry point when on Windows if present...

packages/@jsii/kernel/src/kernel.ts
```diff
@@ -1325,6 +1304,44 @@ export class Kernel {
...
+        args: req.args ?? [],
+        env: {
+          ...process.env,
+          // Make sure the current NODE_OPTIONS are honored if we shell out to node
+          NODE_OPTIONS: process.execArgv.join(' '),
```
RomainMuller, line 1336: Not everything in `process.execArgv` is guaranteed to be valid to use in `NODE_OPTIONS`... That'll be randomly problematic in certain environments.

RomainMuller (changes requested):

packages/@jsii/kernel/src/kernel.ts
jmalins: I don't disagree. I was taking a "do no harm" approach with this PR by just carrying forward the existing logic, which would also work on Windows.

I don't have easy access to a Windows host to test and understand the auto-generated `.cmd` file mentioned here: https://docs.npmjs.com/cli/v9/configuring-npm/package-json#bin

Is it essential that Windows support be included in the PR, since this isn't a new regression?

packages/@jsii/kernel/src/kernel.ts
jmalins: Same comment as above, this is a direct port of the existing `invokeBinScript` logic.

I think, in this case, the contents of `process.execArgv` is predictable because the JSII runtime spawns the node process. Python is currently the only runtime with bin script support and it appears to only pass `--max-old-space-size=4069`, which shouldn't present a problem:
https://github.com/aws/jsii/blob/main/packages/%40jsii/python-runtime/src/jsii/_kernel/providers/process.py#L252

jmalins:
Hi @RomainMuller,
Flaky tests aside (two Maven download failures in a row), please let me know what else you'd require to merge this PR.

Thanks

RomainMuller:
> two Maven download failures in a row

I see you have become acquainted with one of the pains of my life! 🤣

packages/@jsii/kernel/src/kernel.ts
RomainMuller: > Is it essential that Windows support be included in the PR, since this isn't a new regression?

Well... no. I have this principle that I never demand contributors leave the place _cleaner_ than they found it 😅

RomainMuller (approved):

RomainMuller:
@all-contibutors add @jmalins for code

RomainMuller:
@all-contributors add @jmalins for code

PR #3820, opened 2022-11-02
fix(python): type-checking may require incorrect type

### LINKED ISSUE #3818: jsii requires the wrong type in some cases
### Describe the bug

jsii requires the wrong type in some cases. Examples:

https://github.com/aws/aws-cdk/issues/22688
https://github.com/aws/aws-cdk/issues/22689
~~https://github.com/aws/aws-cdk/issues/22711~~

This behavior started with jsii 1.70.0.

### Expected Behavior

Require the correct types.

### Current Behavior

Example: in CDK 2.48.0+, typeguard in Python is requiring that the `Protocol` parameter of `elasticloadbalancingv2.HealthCheck` is of type `ecs.Protocol`, not `elasticloadbalancingv2.Protocol` as it was in CDK 2.47.0 and older.

### Reproduction Steps

Use one of the CDK bugs posted above to reproduce the issue.

### Possible Solution

_No response_

_No response_

### SDK version used

N/A

MacOS

RomainMuller:
Since dae724c, Python type-checking
relies on a nested stub function as a type annotations source. The
parameter signature of that stub was copied verbatim from the
surrounding function, including any forward type references.

However, the forward references can safely be replaced with regular type
references as the module is guaranteed to be fully loaded by the time
the stub is created, and using forward-references there results in
`typeguard` possibly evaluating those in a different context than the
one where the surrounding function was defined. The consequence of this
is that multiple different foward references by the same name may be
incorrectly treated as referring to the same type, despite coming from
different modules.

This is fixed by turning forward type references in the stub with
regular type references (in other words, removing any `"` from the
parameter signature of the stub), which lets the type be resolved from
the local definition context instead of the final runtime context in
which the function is called.

Fixes #3818

---


Naumel (approved):
🪞

rix0rrr:
Romain indicated to me in private chat that he believes this will not cause performance impact.

Don't want to hold up the fix if it's legit, since people are waiting for it.

RomainMuller:
> - Can you quantify the effect on load time of the CDK library with this change?

I have no infrastructure to produce a reliable, comparable, reproductible benchmark of load times. But I could give it a shot in a somewhat un-scientific manner... But what are you concerned about here? That the new code will take significantly longer to load than the previous code? I don't see how this could be the case (or else, the Python VM would be utter rubbish, which it isn't)

> - Since this was a run-time error, shouldn't there be a test that calls the generated Python code?

There is already one added as part of the PR in `packages/@jsii/python-runtime/tests/test_runtime_type_checking.py`.

PR #3848, opened 2022-11-21
fix(python): parameter names shadow imported modules

RomainMuller:
In certain cases, a function parameter may shadow an imported module name, resultin in run-time errors either during type-checking or during the type-cast that is performed before returning a kernel call's result.

This change adds a test case that covers this particular scenario in Python, and changes how foreign modules are imported so that an alias is always generated for those, removing the risk for collisions. It also moves the type-checking stubs out to the root of the module to remove the risk of more "runtime context" polluting the type evaluation.

Fixes https://github.com/aws/aws-cdk/issues/22975

---


MrArnoldPalmer (approved):
LGTM! 🚢

PR #3855, opened 2022-11-24
fix(dotnet): broken error name deserialization

MrArnoldPalmer:
Fixes a bug where the `name` field of an `ErrorResponse` would not deserialize into our enum of known names correctly. Initially it was thought that if the `name` field was non-null, it would only be one of our known values, but the kernel is in fact not setting this field explicitly in all cases of "user land" errors. This results in the `name` field being `"Error"`in the default case and would cause a JSON deserialization error.

Implements a custom deserializer for the `ErrorName` to explicitly return null if the string is not defined or unknown. Additionally makes the `name` field nullable.

fix: https://github.com/aws/aws-cdk/issues/22369

---


MrArnoldPalmer:
This took me a min to repro as it only occurs in `create` calls so just throwing from a method worked as expected.

RomainMuller (approved):

Issue #2404, opened 2021-01-06
Packaging to Java behind a corporate proxy.


Marty-Me:
### Affected Languages

- [ ] `TypeScript` or `Javascript`
- [ ] `Python`
- [X] `Java`
- [ ] .NET (`C#`, `F#`, ...)

### General Information
* **JSII Version:** 1.16.0 (build 99a3413), typescript 3.9.7
* **Platform:**     
Darwin - 19.6.0 Darwin Kernel Version 19.6.0: Thu Oct 29 22:56:45 PDT 2020; root:xnu-6153.141.2.2~1/RELEASE_X86_64 x86_64
### The Question
I'm trying to run `npm run build` with with a `java` block in `targets`, which executes the following command in one of its steps:
`"mvn --batch-mode deploy -D=altDeploymentRepository=local::default::file:///var/folders/tq/kpf08kxn6pl0b92jn8rmr9xc0000gn/T/npm-packLPDcsR --settings=user.xml`

This does however override the settings in my ~/.m2/settings.xml directory, which changes the default-repository to a maven-central-proxy provided for developers.

As a result I am unable to download the required artifacts:
`Could not transfer artifact org.apache.maven.plugins:maven-source-plugin:pom:3.2.1 from/to central (https://repo.maven.apache.org/maven2)`

Is there any way for me to configure jsii-pacmac to still use my custom artifact repository?

Marty-Me:
Hi, has anyone had the time to take a look at this? Do you need some additional information?

Marty-Me:
Hi, I'm still facing this issue, does anyone have some guidance for me?

mjh5153:
I am also experiencing this issue. Any updates?

typescript 3.9.7
jsii 1.53, mvn 3.2.1
Java 11.0.12
mvn 3.8.4
node 14.5

mjh5153:
@Marty-Me

PR #3893, opened 2022-12-30
feat(pacmak): Add support for non-Maven Central repository

icj217:
Add CLI arguments that will inject appropriate values into maven settings file so that non-Maven Central dependencies can be accessed at build time

This was specifically written to work with AWS CodeArtifact per the [setup guide](https://docs.aws.amazon.com/codeartifact/latest/ug/maven-mvn.html#fetching-dependencies), but other repositories will work as well.

Example usage: 
```
> jsii-pacmak \
  --target java \
  --maven-extra-repository-id codeartifact \
  --maven-extra-repository-url "https://mydomain-123456789012.d.codeartifact.us-east-1.amazonaws.com/maven/myrepo/"\ 
  --maven-extra-repository-username aws \
  --maven-extra-repository-password "${CODEARTIFACT_AUTH_TOKEN}" 
```

Username/password can be omitted, but ID + URL are required for the maven settings to be updated.

Should also address issue #2404 / discussion #3465 

---


icj217:
@RomainMuller Do mind approving the workflow runs and possibly reviewing when you get a chance?

packages/jsii-pacmak/lib/targets/java.ts
```diff
@@ -285,6 +285,43 @@ export class JavaBuilder implements TargetBuilder {
...
+    const extraRepositoryUsername: string | undefined =
+      this.options.arguments['maven-extra-repository-username'];
+    const extraRepositoryPassword: string | undefined =
+      this.options.arguments['maven-extra-repository-password'];
+
+    if (extraRepositoryId) {
+      if (!extraRepositoryUrl)
+        throw new Error(
+          'Extra repository requested but no URL was provided! Use the --maven-extra-repository-url argument to provide a value.',
+        );
```
RomainMuller, line 311: Could this maybe be made a single option? Something like `--maven-extra-repository=id=url` or if you need credentials (password caveat from above still applies) `--maven-extra-repository=id=username:password@url` (although maybe `yargs` won't like this format?).

Then - is there a use-case for having more than 1 extra repository configured?

packages/jsii-pacmak/bin/jsii-pacmak.ts
```diff
@@ -153,6 +153,34 @@ import { VERSION_DESC } from '../lib/version';
...
+      default: undefined,
+      hidden: true,
+    })
+    .option('maven-extra-repository-password', {
+      type: 'string',
+      desc: 'Password for authenticating to extra repository',
+      defaultDescription: 'No extra repository is registered.',
+      default: undefined,
+      hidden: true,
+    })
```
RomainMuller, line 183: Passing passwords over CLI arguments is notoriously risky... There's many avenues through which program arguments may be visible by other parties...

I get that these _may_ be "short-lived" credentials here (so less risky to _maybe_ leak somewhere), but I'd still rather have an alternate (safer?) way to provide the password... Not sure what the typical maven practice there is...

This ultimately makes me wonder if we shouldn't instead allow one to provide a "base" XML maven settings file to use, which can include some repos, credentials, etc... This is just an idea, though... I don't know that Maven settings XML documents allow "extending" others without actually copying + modifying them... Which sounds like a lot of unneeded work.

packages/jsii-pacmak/lib/targets/java.ts
```diff
@@ -285,6 +285,43 @@ export class JavaBuilder implements TargetBuilder {
 
     const profileName = 'local-jsii-modules';
     const localRepository = this.options.arguments['maven-local-repository'];
+
+    const repos = localRepos.map((repo) => ({
+      id: repo.replace(/[\\/:"<>|?*]/g, '$'),
+      url: `file://${repo}`,
+    }));
```
RomainMuller, line 292: We could simplify this a bit I feel...

```suggestion
    const repos = localRepos.map((repo, idx) => ({
      id: `_local_${idx}`,
      url: `file://${repo}`,
    }));
```

If doing this, we can ensure "no collisions" by documenting & encoding that "extra" repo IDs may not have a name starting with `_`.

packages/jsii-pacmak/lib/targets/java.ts
```diff
@@ -285,6 +285,43 @@ export class JavaBuilder implements TargetBuilder {
...
+    const extraRepositoryPassword: string | undefined =
+      this.options.arguments['maven-extra-repository-password'];
+
+    if (extraRepositoryId) {
+      if (!extraRepositoryUrl)
+        throw new Error(
+          'Extra repository requested but no URL was provided! Use the --maven-extra-repository-url argument to provide a value.',
+        );
+
+      repos.push({ id: extraRepositoryId, url: extraRepositoryUrl });
```
RomainMuller, line 313: I suggest turning `repos` into a `Map<string, string>` (key is the ID, value is the URL), so it's easier to guarantee no 2 repositories have the same ID?

RomainMuller (commented):
That looks pretty good... I have some suggestions for improvements, but some I'm not even sure myself whether they are actionable in reality...

It would also be nice if you could document usage in https://github.com/aws/jsii/blob/main/gh-pages/content/user-guides/lib-author/configuration/targets/java.md as well.

huyphan:
Instead of exposing every Maven option via the CLI, I suggest that we provide a single option for users to override the setting file that is [currently hard-coded by `jsii-pacmak`](https://github.com/aws/jsii/pull/3893). IMO it's a more future-proof approach and allows users to have more control.

RomainMuller:
> Instead of exposing every Maven option via the CLI, I suggest that we provide a single option for users to override the setting file that is [currently hard-coded by `jsii-pacmak`](https://github.com/aws/jsii/blob/d2ecb6d2a6a460e309e4985dc001a6b749b3aac7/packages/jsii-pacmak/lib/targets/java.ts#L450). IMO it's a more future-proof approach and allows users to have more control.

I agree that this might offer more flexibility...

mrgrain:
Solved by https://github.com/aws/jsii/pull/4068

PR #3913, opened 2023-01-16
chore(superchain): use `pipx` instead of `pip` to install CLIs

rix0rrr:
We recently ran into a situation where we ended up with with a Python distribution where the following failed:

```shell
$ python3
>>> import OpenSSL
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.7/dist-packages/OpenSSL/__init__.py", line 8, in <module>
    from OpenSSL import crypto, SSL
  File "/usr/local/lib/python3.7/dist-packages/OpenSSL/crypto.py", line 3268, in <module>
    _lib.OpenSSL_add_all_algorithms()
AttributeError: module 'lib' has no attribute 'OpenSSL_add_all_algorithms'
```

Because we had `cryptography==39.0.0` but `PyOpenSSL==22.0.0`, which should have been `22.1.0`.

The reason `PyOpenSSL` was held back looks to be because SAM CLI had a hardcoded dependency on the old version.

To prevent this from happening in the future, use `pipx` to install CLIs, so that each CLI gets its own virtual env and their dependency sets don't interfere with each other.

---

### DIFF
--- a/superchain/Dockerfile
+++ b/superchain/Dockerfile
@@ -149,11 +149,16 @@ RUN set -eo pipefail                                                            
   && chmod -R a+rw ${CARGO_HOME}
 ENV PATH=$PATH:${CARGO_HOME}/bin
 
-# Install Python 3
+# Install Python 3. Use 'pipx' for CLI scripts, so each gets its own venv
+ENV PATH=$PATH:/root/.local/bin
 RUN apt-get update                                                                                                      \
   && apt-get -y install python3 python3-dev python3-pip python3-venv                                                    \
-  && python3 -m pip install --no-input --upgrade pip                                                                    \
-  && python3 -m pip install --no-input --upgrade awscli black setuptools twine wheel                                    \
+  && python3 -m pip install --no-input --upgrade pip setuptools                                                         \
+  && python3 -m pip install pipx                                                                                        \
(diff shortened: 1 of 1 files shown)


superchain/Dockerfile
```diff
@@ -149,11 +149,16 @@ RUN set -eo pipefail
...
   && apt-get -y install python3 python3-dev python3-pip python3-venv                                                    \
-  && python3 -m pip install --no-input --upgrade pip                                                                    \
-  && python3 -m pip install --no-input --upgrade awscli black setuptools twine wheel                                    \
+  && python3 -m pip install --no-input --upgrade pip setuptools                                                         \
+  && python3 -m pip install pipx                                                                                        \
+  && python3 -m pipx install --system-site-packages awscli                                                              \
+  && python3 -m pipx install --system-site-packages black                                                               \
+  && python3 -m pipx install --system-site-packages twine                                                               \
+  && python3 -m pipx inject --system-site-packages twine wheel                                                          \
```
rix0rrr, line 161: Put these two in the same venv, don't know if this is necessary but *feels* like they need to work together and *maybe* they need to agree on the versions of certain dependencies when they do.

Might be superstition...

RomainMuller (approved):

PR #3953, opened 2023-02-06
fix(rosetta): prints colors to log file

rix0rrr:
Rosetta error messages contain colors codes in the errors, which makes the errors produced by Rosetta look like garbage and very hard to parse.

Strip the colors if we detect a non-TTY input.

(Copied the regex from an MIT-licensed package to avoid adding extra dependencies)

---


packages/jsii-rosetta/lib/util.ts
```diff
@@ -243,3 +243,15 @@ export async function pathExists(path: string): Promise<boolean> {
...
+  [
+    '[\\u001B\\u009B][[\\]()#;?]*(?:(?:(?:(?:;[-a-zA-Z\\d\\/#&.:=?%@~_]+)*|[a-zA-Z\\d]+(?:;[-a-zA-Z\\d\\/#&.:=?%@~_]*)*)?\\u0007)',
+    '(?:(?:\\d{1,4}(?:;\\d{0,4})*)?[\\dA-PR-TZcf-nq-uy=><~]))',
+  ].join('|'),
+  'g',
+);
+
+function stripColorCodes(x: string) {
+  return x.replace(ANSI_PATTERN, '');
+}
```
RomainMuller, line 257: > (Copied the regex from an MIT-licensed package to avoid adding extra dependencies)

Still, it would not hurt to attribute to the original package/authors?

RomainMuller (approved):

RomainMuller:
Added do-not-merge as I'd still like the attribution to be added...

PR #3959, opened 2023-02-13
feat(check-node): export NodeRelease class

RomainMuller:
This can be useful, for example in projen templates, to automatically test fo all "supported" Node releases without having ot modify every single repository. This can serve as a central point of truth for what releases of node are supported by the Constructs ecosystem via jsii.

---


HBobertz (approved):
LGTM

PR #3987, opened 2023-03-06
fix(dotnet): target netcoreapp3.1 with RollForward

RomainMuller:
Issues where encountered with SDK compatibility when the target
was set to `net6.0` even when the .NET SDK 6.0 or newer was in
use. I suspect this is because of conflicting framework dependency
requirements across various packages.

Re-target `netcoreapp3.1` while setting a roll-forward policy allowing
use of newer major-releases of the SDK when no exact match is found (this
is tested to work correctly for all supported .NET SDK releases by our
automated test suite).

---

### DIFF
--- a/packages/@jsii/dotnet-runtime-test/test/Amazon.JSII.Runtime.IntegrationTests/ComplianceTests.cs
+++ b/packages/@jsii/dotnet-runtime-test/test/Amazon.JSII.Runtime.IntegrationTests/ComplianceTests.cs
@@ -937,7 +937,7 @@ namespace Amazon.JSII.Runtime.IntegrationTests
         [Fact(DisplayName = Prefix + nameof(JsiiAgentIsCorrect))]
         public void JsiiAgentIsCorrect()
         {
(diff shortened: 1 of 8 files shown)


RomainMuller:
The integration test failure on this run is spurious but cannot be avoided (this is due to our rolling back the `TargetFramework`, while we have existing artifacts published that only support `net6.0`. A release will fix that, but we need to skip over the integration tests this time around.

rix0rrr (approved):

cplee:
This change is causing issues with https://github.com/aws/aws-cdk/issues/23921

I'm unable to package a CDK construct anymore: https://github.com/cdklabs/cdk-ecs-codedeploy/actions/runs/4347270960/jobs/7594403891

cplee:
I was able to work through it with latest version of CDK released yesterday (`2.68.0`) and latest JSII (`1.77.0`)

biffgaut:
Upgrading to 2.68.0 fixed it for us as well.

PR #3995, opened 2023-03-08
docs(typescript-restrictions): clarify non-jsii deps

RyoshiKayo:
I spent about a day and a half trying to figure this out, so I thought I'd add a little clarification :)

tl;dr: I'm making a CDK Construct package that also includes a custom lambda handler. That handler takes on a non-jsii dep but I couldn't figure out why JSII kept trying to compile that non-jsii dep until I read 
> The API of the *jsii module* can not expose any type from bundled dependencies, since those types would not be available in other languages.
 
about 20 times. 

---


gh-pages/content/user-guides/lib-author/typescript-restrictions.md
```diff
@@ -14,10 +14,13 @@ A *jsii module* can declare dependencies on any other *jsii module* by adding en
...
 
+### non-jsii dependencies
+
 Occasionally, a dependency on a *non-jsii module* is useful. Since such dependencies do not have generated bindings in
 all the supported languages, they must be bundled with the *jsii module* that depends on them, by adding the library
-into the `bundleDependencies` array in `package.json`. The API of the *jsii module* can not expose any type from bundled
-dependencies, since those types would not be available in other languages.
+into the `bundleDependencies` array in `package.json`. 
+
+The API of the *jsii module* can not expose any type from bundled dependencies, since those types would not be available in other languages. Typescript files that include a non-jsii dependency (Eg. a lambda handler for a CDK Construct) must be excluded in the tsconfig.
```
RomainMuller, line 23: They don't have to be excluded from `tsconfig` (and if you wanted to, you'd have to do the exclusion via the `jsii` stanza in `package.json`).

What you can't do is export them from the `main`/`types` entry point.

This is pretty difficult to articulate correctly & completely without being even more confusing... So I'm super thankful for your proposing to improve there!

RomainMuller (changes requested):

kaizencc:
@RyoshiKayo any chance you continue your work here and we get this in?

gh-pages/content/user-guides/lib-author/typescript-restrictions.md
mrgrain, line 23:
```suggestion
into the `bundleDependencies` array in `package.json`. 

The API of the *jsii module* can not expose any type from bundled dependencies, since those types would not be available in other languages.
TypeScript files that include a non-jsii dependency (e.g. a lambda handler for an AWS CDK Construct) cannot be exported from the `main`/`types` entry point.
```

PR #4009, opened 2023-03-15
feat: enable Go generics for jsii Go CDK code

a-h:
## Current situation

In Go, strings are not nullable by default like they are in TypeScript, C# etc.

There's also no pointer literal - you can't do something like `&("string value")` to create a pointer to a string literal, it's required to create a variable instead.

While there's discussion on having something built into the language in https://github.com/golang/go/issues/45624 it hasn't concluded with a built-in as part of the language, or a standard library function.

The Go JSII library has worked around this by using the approach of creating functions that turn literals and variables into a pointer. Since Go < 1.18 didn't support generics, there's one for each supported JSII type. This means that you have to think about the type of any variables and use the right conversion function, while you're writing code.

```go
stringPtr := jsii.String("literal value")
float64Ptr := jsii.Number(123.45)
```

## Proposal

In Go 1.18, generics were added. These type parameters can be used to provide a single function that convert literals and variables into a pointer.

```go
func Ptr[T any](v T) *T {
	return &v
} 
```

We can further restrict the allowed types of `T` by using a constraint.

```go
type Type interface {
	bool | string | float64 | time.Time
}

func Ptr[T Type](v T) *T {
	return &v
}
```

If we adopt the name `V` instead of `Ptr` to save a further two characters (I'd still be happy with `Ptr`, `Val` etc. as a name), the previous example can be rewritten as:

```go
stringPtr := jsii.V("literal value")
float64Ptr := jsii.V(123.45)
boolPtr := jsii.V(true)
```

This means that only one function is used to convert multiple types, and the function name can be a few characters shorter.

For the `V` function, integer literals can not be converted to `float64`, since the output would be a pointer to the input type, not to the allowed `float64` type. Developers would be prompted that integer types are not in the set of allowed types (`string`, `float64`, `bool`, `time.Time`).

However, the `Number` function can be updated in place to use generics to accept any numeric type, which simplifies code that uses integer values from:

```go
intPtrOK := jsii.Number(float64(123))
var x uint32
intPtrOK = jsii.Number(float64(x))
```

```go
intPtrOK := jsii.Number(123) // Would automatically be converted to `float64`
var x uint32
intPtrOK = jsii.Number(x)
```

## Impact

Since Go 1.18 is already required by the CDK, the use of generics is possible.

For backwards compatibility with existing code, the existing functions would be maintained, so no code would break.

#### Before

```go
// DynamoDB.
slotMachineTable := awsdynamodb.NewTable(stack, jsii.String("slotMachineTable"), &awsdynamodb.TableProps{
	TableName:    jsii.String("slotMachine"),
	PartitionKey: &awsdynamodb.Attribute{Name: jsii.String("_pk"), Type: awsdynamodb.AttributeType_STRING},
	SortKey:      &awsdynamodb.Attribute{Name: jsii.String("_sk"), Type: awsdynamodb.AttributeType_STRING},
	BillingMode:  awsdynamodb.BillingMode_PAY_PER_REQUEST,
	Stream:       awsdynamodb.StreamViewType_NEW_IMAGE,
})

// Create an event bus.
eventBus := awsevents.NewEventBus(stack, jsii.String("slotMachineEventBus"), &awsevents.EventBusProps{
	EventBusName: jsii.String("slotMachineEventBus"),
})

// Process streams.
streamHandler := awslambdago.NewGoFunction(stack, jsii.String("streamHandler"), &awslambdago.GoFunctionProps{
	Runtime:      awslambda.Runtime_PROVIDED_AL2(),
	Architecture: awslambda.Architecture_ARM_64(),
	Entry:        jsii.String("../api/streamhandler/"),
	Environment: &map[string]*string{
		"EVENT_BUS_NAME":    eventBus.EventBusName(),
		"EVENT_SOURCE_NAME": jsii.String("slot-machine"),
	},
	Timeout:      awscdk.Duration_Minutes(jsii.Number(15)),
	LogRetention: awslogs.RetentionDays_ONE_YEAR,
})
slotMachineTable.GrantReadData(streamHandler)
eventBus.GrantPutEventsTo(streamHandler)
```

#### After

```go
// DynamoDB.
slotMachineTable := awsdynamodb.NewTable(stack, jsii.V("slotMachineTable"), &awsdynamodb.TableProps{
	TableName:    jsii.V("slotMachine"),
	PartitionKey: &awsdynamodb.Attribute{Name: jsii.V("_pk"), Type: awsdynamodb.AttributeType_STRING},
	SortKey:      &awsdynamodb.Attribute{Name: jsii.V("_sk"), Type: awsdynamodb.AttributeType_STRING},
	BillingMode:  awsdynamodb.BillingMode_PAY_PER_REQUEST,
	Stream:       awsdynamodb.StreamViewType_NEW_IMAGE,
})

// Create an event bus.
eventBus := awsevents.NewEventBus(stack, jsii.V("slotMachineEventBus"), &awsevents.EventBusProps{
	EventBusName: jsii.V("slotMachineEventBus"),
})

// Process streams.
streamHandler := awslambdago.NewGoFunction(stack, jsii.V("streamHandler"), &awslambdago.GoFunctionProps{
	Runtime:      awslambda.Runtime_PROVIDED_AL2(),
	Architecture: awslambda.Architecture_ARM_64(),
	Entry:        jsii.V("../api/streamhandler/"),
	Environment: &map[string]*string{
		"EVENT_BUS_NAME":    eventBus.EventBusName(),
		"EVENT_SOURCE_NAME": jsii.V("slot-machine"),
	},
	Timeout:      awscdk.Duration_Minutes(jsii.V(15.0)),
	LogRetention: awslogs.RetentionDays_ONE_YEAR,
})
slotMachineTable.GrantReadData(streamHandler)
eventBus.GrantPutEventsTo(streamHandler)
```

## Considerations

The name `V` is the shortest, but doesn't describe what it does.

The Go team might, at some point, introduce a change to automatically convert literals into pointers, in which case the `jsii.V` and `jsii.String` function calls would be redundant and could be stripped out of a codebase. However, this change doesn't make that potential outcome any worse.

---

### DIFF
--- a/packages/jsii-pacmak/bin/jsii-pacmak.ts
+++ b/packages/jsii-pacmak/bin/jsii-pacmak.ts
@@ -6,6 +6,7 @@ import * as yargs from 'yargs';
 
 import { pacmak, configureLogging, TargetName } from '../lib';
 import { debug } from '../lib/logging';
+import { DEFAULT_PACK_COMMAND } from '../lib/packaging';
 import { VERSION_DESC } from '../lib/version';
 
 (async function main() {
@@ -151,6 +152,12 @@ import { VERSION_DESC } from '../lib/version';
       desc: 'Configure a custom path (relative to current working directory) to a repository when packaging a Java package.',
       defaultDescription: 'A temporary directory is used.',
       default: undefined,
+      hidden: true,
+    })
+    .option('pack-command', {
+      type: 'string',
+      desc: 'Configure a custom command to create package tarballs. Command must output the name of the tarball.',
+      default: DEFAULT_PACK_COMMAND,
       hidden: true,
     })
     .option('validate-assemblies', {
--- a/packages/jsii-pacmak/lib/index.ts
+++ b/packages/jsii-pacmak/lib/index.ts
@@ -74,7 +74,9 @@ export async function pacmak({
 
   await timers.recordAsync('npm pack', () => {
     logging.info('Packaging NPM bundles');
-    return Promise.all(modulesToPackageFlat.map((m) => m.npmPack()));
+    return Promise.all(
+      modulesToPackageFlat.map((m) => m.npmPack(argv['pack-command'])),
+    );
   });
 
   await timers.recordAsync('load jsii', () => {
--- a/packages/jsii-pacmak/lib/packaging.ts
+++ b/packages/jsii-pacmak/lib/packaging.ts
@@ -1,9 +1,12 @@
+import * as fs from 'fs-extra';
 import type { Assembly, TypeSystem } from 'jsii-reflect';
 import * as os from 'os';
 import * as path from 'path';
 
 import * as logging from '../lib/logging';
 import { Scratch, shell } from './util';
+
+export const DEFAULT_PACK_COMMAND = 'npm pack';
 
 export interface JsiiModuleOptions {
   /**
@@ -52,15 +55,27 @@ export class JsiiModule {
   /**
    * Prepare an NPM package from this source module
    */
-  public async npmPack() {
+  public async npmPack(packCommand = DEFAULT_PACK_COMMAND) {
     this._tarball = await Scratch.make(async (tmpdir) => {
-      // Quoting (JSON-stringifying) the module directory in order to avoid
-      // problems if there are spaces or other special characters in the path.
-      const args = ['pack', JSON.stringify(this.moduleDirectory)];
-      if (logging.level >= logging.LEVEL_VERBOSE) {
-        args.push('--loglevel=verbose');
+      const args = [];
+
+      if (packCommand === DEFAULT_PACK_COMMAND) {
+        // Quoting (JSON-stringifying) the module directory in order to avoid
+        // problems if there are spaces or other special characters in the path.
+        args.push(JSON.stringify(this.moduleDirectory));
+
+        if (logging.level >= logging.LEVEL_VERBOSE) {
+          args.push('--loglevel=verbose');
+        }
+      } else {
+        // Ensure module is copied to tmpdir to ensure parallel execution does not content on generated tarballs
+        await fs.copy(this.moduleDirectory, tmpdir);
       }
-      const out = await shell('npm', args, { cwd: tmpdir });
+
+      const out = await shell(packCommand, args, {
+        cwd: tmpdir,
+      });
+
       // Take only the last line of npm pack which should contain the
       // tarball name. otherwise, there can be a lot of extra noise there
       // from scripts that emit to STDOUT.
@@ -69,7 +84,7 @@ export class JsiiModule {
 
       if (!lastLine.endsWith('.tgz') && !lastLine.endsWith('.tar.gz')) {
         throw new Error(
-          `npm pack did not produce tarball from ${
+          `${packCommand} did not produce tarball from ${
             this.moduleDirectory
           } into ${tmpdir} (output was ${JSON.stringify(lines)})`,
         );
--- a/packages/jsii-pacmak/test/build-test.sh
+++ b/packages/jsii-pacmak/test/build-test.sh
@@ -81,3 +81,8 @@ done
 clean_dists
 echo "Testing ALL-AT-ONCE build."
 ${pacmak} ${OPTS} -v --no-parallel $packagedirs
+
+# Test custom pack command
+clean_dists
+echo "Testing yarn custom pack command."
+${pacmak} ${OPTS} -v --pack-command='yarn pack -f custom.tgz -s && echo custom.tgz' ${PWD}/../../@scope/jsii-calc-base-of-base


RomainMuller:
I was definitely thinking about doing this and the major holdout was essentially go1.17... but now that go1.20 is out that's not a problem anymore. You beat me to it (I was busy on other stuff), so thanks for that...

I might not have time to review the PR today however, but I'll absolutely be having a stab at it next week.

packages/@jsii/go-runtime/jsii-runtime-go/helpers.go
```diff
@@ -2,23 +2,45 @@ package jsii
 
 import "time"
 
+type Type interface {
```
RomainMuller, line 5: Is there a reason why this is exported, but `numberType` is not?

packages/@jsii/go-runtime/jsii-runtime-go/helpers.go
```diff
@@ -2,23 +2,45 @@ package jsii
...
+	bool | string | float64 | time.Time
+}
+
+func V[T Type](v T) *T {
```
RomainMuller, line 9: This is missing a doc-string. Also I'm not a huge fan of the non-descriptive name here...

packages/@jsii/go-runtime/jsii-runtime-go/helpers.go
```diff
@@ -2,23 +2,45 @@ package jsii
...
+	return &v
+}
+
+func Slice[T Type](v ...T) *[]*T {
```
RomainMuller, line 13: Missing doc-string. It also feels to me the name should somehow indicate it has similar semantics to `V` above.

packages/@jsii/go-runtime/jsii-runtime-go/helpers.go
```diff
@@ -31,21 +53,13 @@ func String(v string) *string { return &v }
...
-	slice := make([]*string, len(v))
-	for i := 0; i < len(v); i++ {
-		slice[i] = String(v[i])
-	}
-	return &slice
+	return Slice(v...)
 }
 
 // Time obtains a pointer to the provided time.Time.
 func Time(v time.Time) *time.Time { return &v }
```
RomainMuller, line 60: I wonder if these should be re-implemented to use `V`, so there's only 1 concrete implementation here... I'd be hopeful the compiler ends up being smart enough to inline `V`, so it ultimately makes no performance difference? (and if it did, would be be noticeable?)

RomainMuller (changes requested):

packages/@jsii/go-runtime/jsii-runtime-go/helpers.go
a-h: Yes, I thought that would a contentious point!

What do you think, `Ptr` (because it returns a pointer value), `Value` (because it's a `jsii.Value`), or something else?

packages/@jsii/go-runtime/jsii-runtime-go/helpers.go
a-h: I don't believe the `Type` constraint interface has to be exported at all. It can still apply without it being a public type exported by the package.

I think it's a design decision of whether it should be visible to users or not. I'm thinking that it might be useful to export it, but I can't think of a use case for exporting it.

What do you think? Export or no?

packages/@jsii/go-runtime/jsii-runtime-go/helpers.go
a-h: So, in this case, the name `Ptr` for a single value, and `PtrSlice` for a slice?

Or `Value` and `Values`?

packages/@jsii/go-runtime/jsii-runtime-go/helpers.go
a-h: I thought that they were so simple that they could be happily left alone, and that it would be simpler to do so.

My understanding is that Go 1.18 was pretty weak on monomorphization and that generics add a runtime cost - https://planetscale.com/blog/generics-can-make-your-go-code-slower but, based on your comment I did a quick benchmark and given these two:

```go
package genericptr

import "time"

type types interface {
	bool | string | float64 | time.Time
}

func Ptr[T types](v T) *T { return &v }

func String(v string) *string { return &v }
```

And this benchmark...

```go
package genericptr

import "testing"

func BenchmarkPtr(b *testing.B) {
	for i := 0; i < b.N; i++ {
		ss := Ptr("test")
		if *ss != "test" {
			b.Error("something went wrong")
		}
	}
}

func BenchmarkString(b *testing.B) {
	for i := 0; i < b.N; i++ {
		ss := String("test")
		if *ss != "test" {
			b.Error("something went wrong")
		}
	}
}
```

The difference was always < 10ns on multiple runs.

```
go test -bench=.
goos: linux
goarch: amd64
pkg: genericptr
cpu: 11th Gen Intel(R) Core(TM) i7-11850H @ 2.50GHz
BenchmarkPtr-16         1000000000               0.2092 ns/op
BenchmarkString-16      1000000000               0.2090 ns/op
PASS
```

a-h: Based on that, I agree - single implementation.

packages/@jsii/go-runtime/jsii-runtime-go/helpers.go
a-h: I've removed them, since they're not required to be public, on the basis that it's possible to make things public later, but not to remove them.

RomainMuller: Yeap, I'd start off with them hidden, and we can expose them later if necessary... The other way around is not as easy ^^

RomainMuller (dismissed):

RomainMuller (approved):

PR #4021, opened 2023-03-21
feat: add configurable pack command

agdimech:
This PR adds support for configuring a custom pack command. This is needed when using other package managers like `pnpm`.

In terms of usage, developers would configure pacmak to use pnpm for example as follows:

```
jsii-pacmak --pack-command='pnpm pack'
```

---


RomainMuller (changes requested):

packages/jsii-pacmak/bin/jsii-pacmak.ts
```diff
@@ -153,6 +153,13 @@ import { VERSION_DESC } from '../lib/version';
       default: undefined,
       hidden: true,
     })
+    .option('npm-pack-command', {
+      type: 'string',
+      desc: 'Configure a custom command to execute when packaging i.e. pnpm pack --pack-destination {{destDir}}. Bundle must be written to {{destDir}} which will be substituted at runtime.',
+      defaultDescription: 'npm pack is used.',
+      default: 'npm pack --pack-destination {{destDir}}',
+      hidden: true,
+    })
```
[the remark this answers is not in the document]
agdimech: I've updated the PR with your feedback :)

packages/jsii-pacmak/lib/packaging.ts
```diff
@@ -52,15 +55,20 @@ export class JsiiModule {
...
-      const args = ['pack', JSON.stringify(this.moduleDirectory)];
+      const args = packCommand.split(' ');
+
+      // If using the default command, optimize by packing directly into the tmp dir to allow unit tests to run in parallel
+      if (packCommand === DEFAULT_PACK_COMMAND) {
+        args.push('--pack-destination', tmpdir);
+      }
       if (logging.level >= logging.LEVEL_VERBOSE) {
         args.push('--loglevel=verbose');
       }
```
RomainMuller, line 68: I think this should not happen unless using the default command (the user has control, so they can decide on log level there directly).

packages/jsii-pacmak/lib/packaging.ts
```diff
@@ -52,15 +55,20 @@ export class JsiiModule {
   /**
    * Prepare an NPM package from this source module
    */
-  public async npmPack() {
+  public async npmPack(packCommand = DEFAULT_PACK_COMMAND) {
     this._tarball = await Scratch.make(async (tmpdir) => {
-      // Quoting (JSON-stringifying) the module directory in order to avoid
-      // problems if there are spaces or other special characters in the path.
-      const args = ['pack', JSON.stringify(this.moduleDirectory)];
+      const args = packCommand.split(' ');
```
RomainMuller, line 60: This is dangerous... if a customer uses a fully-qualified path that includes white space, this will always fail...

packages/jsii-pacmak/lib/packaging.ts
```diff
@@ -52,15 +55,20 @@ export class JsiiModule {
...
+      // If using the default command, optimize by packing directly into the tmp dir to allow unit tests to run in parallel
+      if (packCommand === DEFAULT_PACK_COMMAND) {
+        args.push('--pack-destination', tmpdir);
+      }
       if (logging.level >= logging.LEVEL_VERBOSE) {
         args.push('--loglevel=verbose');
       }
-      const out = await shell('npm', args, { cwd: tmpdir });
+      const out = await shell(args[0], args.slice(1), {
+        cwd: this.moduleDirectory,
```
RomainMuller, line 70: The inter test appears to fail because there is no `package.json` in there?

RomainMuller (changes requested):

packages/jsii-pacmak/lib/packaging.ts
agdimech: updated in latest rev.

PR #4030, opened 2023-03-24
fix: enable dereference when copying module to tmp dir

agdimech:
When executing the copy command, for symlinked node_modules like in the case of pnpm, the copy will treat nested folders as files which breaks the pack command. De-referencing will make sure any symlinked folders are actually dereferenced.

---


agdimech:
@RomainMuller - can I please get a review?

RomainMuller (approved):

RomainMuller:
@mergifyio update

PR #4056, opened 2023-04-13
test(rosetta): Extend submodule import test case for nested structs in submodules

ansgarm:
Hi! I created this (failing) test case for a jsii-rosetta problem we are seeing in `cdktf convert` (bullet point # 2 in [this comment](https://github.com/hashicorp/terraform-cdk/issues/2637#issuecomment-1505234729)).

Originally we uncovered that this code was invalid for Python but upon writing this test case, I think it is also producing invalid code for C# and Go. I'm not entirely sure whether I've adjusted the expected results in the right way, though. I'll add comments to those places including what Rosetta currently returns to make it easier to see how these tests currently fail.

I'd be open to contribute to a solution but I'm having a hard time even describing the problem in the right terms 😅

---


packages/jsii-rosetta/test/translations/imports/submodule-import.cs
```diff
@@ -9,7 +9,7 @@
 });
 
 // Accesses two distinct points of the submodule hierarchy
-var myClass = new Submodule.MyClass(new SomeStruct { Prop = Submodule.Child.SomeEnum.SOME });
+var myClass = new Submodule.MyClass(new SomeStruct { Prop = Submodule.Child.SomeEnum.SOME, NestedStruct = new Submodule.Child.AnotherStruct { StringProperty = "hello" } });
```
ansgarm, line 12: This currently fails as the result from Rosetta is missing the `Submodule.Child.` for `AnotherStruct`
image

packages/jsii-rosetta/test/translations/imports/submodule-import.py
```diff
@@ -9,7 +9,7 @@
 )
 
 # Accesses two distinct points of the submodule hierarchy
-my_class = calc.submodule.MyClass(prop=calc.submodule.child.SomeEnum.SOME)
+my_class = calc.submodule.MyClass(prop=calc.submodule.child.SomeEnum.SOME, nested_struct=calc.submodule.child.AnotherStruct(string_property="hello"))
```
ansgarm, line 12: This currently fails as the result from Rosetta is missing the `calc.submodule.child.` for `AnotherStruct`
image

ansgarm:
@DanielMSchmidt just pointed me at https://github.com/aws/jsii-compiler which is where JSII 5.0 lives – I'll create a new PR over there as we're on JSII 5.0 now 😄

packages/jsii-rosetta/test/translations/imports/submodule-import.cs
ansgarm: While looking at this once more, I noticed that the original test already seems to have a small error, too:
```suggestion
var myClass = new Submodule.MyClass(new Submodule.Child.SomeStruct { Prop = Submodule.Child.SomeEnum.SOME, NestedStruct = new Submodule.Child.AnotherStruct { StringProperty = "hello" } });
```

RomainMuller:
@ansgarm the `jsii` compiler 5.0.x lives in aws/jsii-compiler, and the `jsii-rosetta` 5.0.x lives in aws/jsii-rosetta.

The `jsii-calc` package (and dependencies) are "copied" in aws/jsii-compiler (under the fixtures directory), and have been slightly edited there to improve the v5.0 coverage... Unfortunately, since they need to assess some backwards-compatibility behavior, the code there had to be updated to include TypeScript 3.9-incompatible syntax... So it's not just a matter of copying the code over.

If `jsii-calc` here is lacking some APIs you need for your tests, you could just add the necessary changes into this PR...

PR #4068, opened 2023-04-24
feat(jsii-pacmak): parsing arguments in non-strict mode

huyphan:
Some targets support extra target-specific arguments so users can customize the behavior of the build tools. For example, the Java target allows users to customize the `mvn` CLI behaviour with `--mvn-*` parameters. However the `jsi-pacmak` CLI is blocking this feature because it is using strict mode when parsing arguments. This change lifts that gate.

This address [#3465](https://github.com/aws/jsii/discussions/3465)

---


RomainMuller (approved):
This has the drawback of silently accepting typo'd flags and arguments, which is a little unfortunate... But I guess the feature is worth the trade-off here...

PR #4071, opened 2023-04-25
chore: add upgrade prompt at load of jsii and jsii-rosetta 1.x

RomainMuller:
This is to strongly incitate customers upgrade their dependencies, while trying to avoid alarming (there is 18 months left before the End-of-Support is scheduled, which is PLENTY of time).

---

### DIFF
--- a/packages/jsii-rosetta/lib/upgrade-prompt.ts
+++ b/packages/jsii-rosetta/lib/upgrade-prompt.ts
@@ -4,8 +4,8 @@ if (process.env.JSII_SUPPRESS_UPGRADE_PROMPT == null) {
     [
       '#######################################################################################################',
       '### You are currently using jsii-rosetta@1.x. We recommend upgrading to jsii-rosetta@5.0.x or       ###',
-      '### later, which will allow you to use modern TypeScript syntax, and improves compatibility with    ###',
-      '### many common dependencies, such as recent versions of @types/node.                               ###',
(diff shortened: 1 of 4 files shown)


packages/jsii-rosetta/lib/upgrade-prompt.ts
```diff
@@ -0,0 +1,21 @@
+// Note: intentionally using the same environment variable as the compiler here instead of customizing...
+if (process.env.JSII_SUPPRESS_UPGRADE_PROMPT == null) {
+  console.error(
+    [
+      '#######################################################################################################',
+      '### You are currently using jsii-rosetta@1.x. We recommend upgrading to jsii-rosetta@5.0.x or       ###',
+      '### later, which will allow you to use modern TypeScript syntax, and improves compatibility with    ###',
```
Naumel, line 7:
```suggestion
      '### later. This will allow you to use modern TypeScript syntax, and improves compatibility with    ###',
```

packages/jsii-rosetta/lib/upgrade-prompt.ts
```diff
@@ -0,0 +1,21 @@
...
+      '### many common dependencies, such as recent versions of @types/node.                               ###',
```
Naumel, line 8:
```suggestion
      '### many common dependencies. For example, recent versions of @types/node.                               ###',
```

Naumel (approved):
Suggestions seem so trivial that it does not make sense not to approve based on just that.

PR #4074, opened 2023-04-26
Fix(Dockerfile): Use Node 18 (supported) rather than 14 (deprecated)

indrora:
This PR changes the default node major version to 18, the current LTS, that will go out of support in 2025. 
---

### DIFF
--- a/.github/workflows/docker-images.yml
+++ b/.github/workflows/docker-images.yml
@@ -19,10 +19,10 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node: ['14', '16', '18']
+        node: ['14', '16', '18', '20']
     env:
       # Node version whose images will be aliased without the -nodeXX segment
-      DEFAULT_NODE_MAJOR_VERSION: 14
+      DEFAULT_NODE_MAJOR_VERSION: 18
     steps:
       - name: Check out
         uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


indrora:
Swapping to draft so I can better clean up the workflow. @kellertk suggested I update the Dockerfile, There's more to it than that.

[Node 20 is out](https://nodejs.org/en/blog/announcements/v20-release-announce) and will be the next LTS. We should at least start building for it, so I'll add that to the matrix. We *really* should drop Node 14 support entirely and stop building for it, and Node 16 goes out of support in just over two weeks. 

I'm going to update the matrix to build for 16, 18, 20 but we should prepare to drop 16 on 2023-09-11 in line with [Node's official guidance](https://nodejs.dev/en/about/releases/) and the comment from the Node 20 launch announcement:

```text
Also of note is that Node.js 14 will go End-of-Life in April 2023, so we advise you to start planning to upgrade to Node.js 18 (LTS) or Node.js 20 (soon to be LTS).

Please, consider that Node.js 16 (LTS) will go End-of-Life in September 2023, which was brought forward from April 2024 to coincide with the end of support of OpenSSL 1.1.1. You can read more details about that decision at https://nodejs.org/en/blog/announcements/nodejs16-eol/.
```

This means that as of September, the test matrix should be [18, 20] to match AWS patching guidelines.

RomainMuller:
I think we've already enacted this change (although in a slightly more lax way). I'm gonna close this.

PR #4076, opened 2023-05-01
feat(jsii-diff): allow external stability to be treated as error

mrgrain:
In certain situations one might want to treat changes to `external` APIs as errors.
This could be to check the external APIs for any breaking changes and have an opportunity to selectively reject, heal or soften an upstream change.

This change introduces a new cli option `--error-on` which can be one of three classes:

| `--error-on`       | Stabilities that cause an ERROR                    |
| ------------------ | -------------------------------------------------- |
| `prod` (default)   | `stable`, `deprecated`                             |
| `non-experimental` | `stable`, `deprecated`, `external`                 |
| `all`              | `stable`, `deprecated`, `experimental`, `external` |

**Fixes `deprecated` APIs not being treated as `stable`.**
In jsii, deprecations are treated as a stability.
However for the purpose of jsii-diff they should be treated as stable.
Otherwise one could do `stable -> deprecated, make breaking change, deprecated -> stable`, which should not be allowed. 
We also can't prohibit the transition from deprecated back to stable, as it's perfectly okay to un-deprecate an API.

---

### DIFF
--- a/packages/jsii-diff/bin/jsii-diff.ts
+++ b/packages/jsii-diff/bin/jsii-diff.ts
@@ -45,11 +45,11 @@ async function main(): Promise<number> {
       type: 'boolean',
       default: false,
       desc: 'Error on experimental API changes',
-      deprecate: 'Use `--errors` instead',
+      deprecate: 'Use `--error-on` instead',
     })
     .option('error-on', {
-      type: 'array',
-      default: ['prod'],
+      type: 'string',
+      default: 'prod',
       choices: ERROR_CLASSES,
       desc: 'Which type of API changes should be treated as an error',
     })
@@ -129,10 +129,7 @@ async function main(): Promise<number> {
   if (mismatches.count > 0) {
     const diags = classifyDiagnostics(
       mismatches,
-      treatAsError(
-        argv['error-on'] as ErrorClass[],
-        argv['experimental-errors'],
-      ),
+      treatAsError(argv['error-on'] as ErrorClass, argv['experimental-errors']),
       await loadFilter(argv['ignore-file']),
     );
 
--- a/packages/jsii-diff/bin/jsii-diff.ts
+++ b/packages/jsii-diff/bin/jsii-diff.ts
@@ -9,8 +9,11 @@ import * as yargs from 'yargs';
 import { compareAssemblies } from '../lib';
 import {
   classifyDiagnostics,
+  treatAsError,
   formatDiagnostic,
   hasErrors,
+  ErrorClass,
+  ERROR_CLASSES,
 } from '../lib/diagnostics';
 import {
   DownloadFailure,
@@ -42,6 +45,13 @@ async function main(): Promise<number> {
       type: 'boolean',
       default: false,
       desc: 'Error on experimental API changes',
+      deprecate: 'Use `--errors` instead',
+    })
+    .option('error-on', {
+      type: 'array',
+      default: ['prod'],
+      choices: ERROR_CLASSES,
+      desc: 'Which type of API changes should be treated as an error',
     })
     .option('ignore-file', {
       alias: 'i',
@@ -119,7 +129,10 @@ async function main(): Promise<number> {
   if (mismatches.count > 0) {
     const diags = classifyDiagnostics(
       mismatches,
-      argv['experimental-errors'],
+      treatAsError(
+        argv['error-on'] as ErrorClass[],
+        argv['experimental-errors'],
+      ),
       await loadFilter(argv['ignore-file']),
     );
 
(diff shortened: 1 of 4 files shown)


packages/jsii-diff/bin/jsii-diff.ts
```diff
@@ -43,6 +43,18 @@ async function main(): Promise<number> {
       default: false,
       desc: 'Error on experimental API changes',
     })
+    .option('external-errors', {
+      alias: 'x',
+      type: 'boolean',
+      default: false,
+      desc: 'Error on external API changes',
+    })
+    .option('deprecated-errors', {
```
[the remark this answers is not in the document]
mrgrain:
```ts
  /**
   * The API may emit warnings. Backward compatibility is not guaranteed.
   *
   * More information about the deprecation can usually be found in the
   * `deprecated` field.
   */
  Deprecated = 'deprecated',
```

That's not how jsii handles it and currently deprecated features will only emit a warning.
I think jsii models this after Node: https://nodejs.org/api/documentation.html#documentation_stability_index

I can _see_ where you are coming from though. Happy to looking into changing the current behavior in principle.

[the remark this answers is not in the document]
mrgrain: > Is this worth implementing as a bunch of individual flags? Should we do named classes?
> 
> ```
> --errors=stable
> --errors=stable,experimental
> --errors=all
> ```
> 
> ...etc...?

Sure, I was trying to avoid a new major version release. But I guess I can deprecate the previous setting.

packages/jsii-diff/lib/diagnostics.ts
```diff
@@ -36,13 +36,48 @@ export function hasErrors(diags: Diagnostic[]) {
...
+export function errorsForStability(
+  errors: string[],
+  deprecatedExperimentalErrors = false,
+): StabilityClassifications {
+  return {
+    [Stability.Stable]: shouldError(errors, 'stable'),
+    [Stability.Experimental]:
+      shouldError(errors, 'experimental') || deprecatedExperimentalErrors,
+    [Stability.External]: shouldError(errors, 'external'),
+    [Stability.Deprecated]: shouldError(errors, 'deprecated'),
```
rix0rrr, line 63: I still feel that deprecated should also error when the class is 'stable'. That's usually what you want; if you don't, there are no guarantees on an API marked 'deprecated' and that's not correct, they still need to be preserved.

In my head we'll have some names for "error classes" as well which are predefined combinations of the standard stabilities. Also in my head we'd overload the names (so `error class STABLE = stabilities [STABLE, DEPRECATED]`), but that will most likely get confusing and not make sense to anyone but me :D. How about we introduce new names for the error classes, and have error classes for each of the stabilities as well that map to the stability itself. I think it looks like this:

image

-------------

I also think the mapping will probably also be easier to understand if we make it explicit in the code:

```ts
type ErrorClass = 'all' | 'default' | 'stable' | 'experimental' | 'nonexperimental' | ...;

const ERRORCLASSES_TO_STABILITIES: Record<ErrorClass, Stability[]> = {
  'all': [Stability.Stable, Stability.Experimental, Stability.External, Stability.Deprecated],
  'default': [Stability.Stable, Stability.Deprecated],
  'nonexperimental': [Stability.Stable, Stability.Deprecated, Stability.External],
  'stable': [Stability.Stable],
  'experimental': [Stability.Experimental],
};
```

This is a bit more repetition, but extremely easy to parse and add new error classes to later.

mrgrain: > I also think the mapping will probably also be easier to understand if we make it explicit in the code:

Yes, this much nicer. Thanks!

mrgrain: jsii strictly enforces `deprecated` as its own stability.

For example this code:
```ts
    /**
     * @stability stable
     * @deprecated whatever
     **/
    export class FooBar { }
```

Will cause jsii to fail with: 
```console
      index.ts:6:5 - error JSII7999: @deprecated tag requires '@stability deprecated' or no @stability at all.
      
      6     export class FooBar { }
            ~~~~~~~~~~~~~~~~~~~~~~~
```

While this might be unintuitive to users, at least it's consistently enforced.
You can't have a `deprecated` feature that's also `stable` (or `experimental` or `external` for that matter).
Therefore defining `deprecated ∈ stable` would not be correct.

--- 

The current implementation doesn't allow anyone to make a decision about `deprecated` or `external` stabilities.
It looks more like this:

This is how the API currently treats stabilities:
sketch1683022214169
With `external` and `deprecated` only every being a `WARNING`

Now in my mind the onion isn't quite as neatly defined as yours.
To me the whole thing looks more like this:
sketch1683022165565

---

In summary I think we agree on having pre-defined groups of stability classes. We probably agree on `all` as well.
I think you might also be okay with mix-and-matching groups together (`--error-on=stable --error-on=external`).

What we likely don't agree on yet is how to treat `deprecated` and any other pre-defined groups.
I don't have strong feelings around these additional groups, as long as the names are unambiguous (which might be tricky).

What I do disagree with is the proposed `default` group. Firstly, because the name is descriptive. It should have a descriptive name. Secondly because making it the implied default would constitute a breaking change. It do think it would actually make a reasonable default if we were to break `jsii-diff` out into it's own repo and release a v2.

PR #4077, opened 2023-05-01
chore: silence node end of life warning in tests

vinayak-kukreja:
Currently tests are failing with end of life warning messages. This PR is adding env variable for tests to silence these errors.

---

### DIFF
--- a/packages/@jsii/check-node/src/index.ts
+++ b/packages/@jsii/check-node/src/index.ts
@@ -23,7 +23,7 @@ export function checkNode(envPrefix = 'JSII'): void {
   if (nodeRelease?.endOfLife) {
     const silenceVariable = `${envPrefix}_SILENCE_WARNING_END_OF_LIFE_NODE_VERSION`;
     const acknowledgeNodeEol =
-      'I acknowledge end of life status for this node version';
+      'I acknowledge end of life(EOL) status for this node version. The CDK reserves the right to pull support of EOL runtime releases a month after they went EOL status, and people are recommended to stop using those as soon as possible';
     const qualifier = nodeRelease.endOfLifeDate
       ? ` on ${nodeRelease.endOfLifeDate.toISOString().slice(0, 10)}`
       : '';
--- a/packages/@jsii/check-node/src/index.ts
+++ b/packages/@jsii/check-node/src/index.ts
@@ -22,10 +22,12 @@ export function checkNode(envPrefix = 'JSII'): void {
 
   if (nodeRelease?.endOfLife) {
     const silenceVariable = `${envPrefix}_SILENCE_WARNING_END_OF_LIFE_NODE_VERSION`;
+    const acknowledgeNodeEol =
+      'I acknowledge end of life status for this node version';
     const qualifier = nodeRelease.endOfLifeDate
       ? ` on ${nodeRelease.endOfLifeDate.toISOString().slice(0, 10)}`
       : '';
-    if (!process.env[silenceVariable])
+    if (!(process.env[silenceVariable] === acknowledgeNodeEol))
       veryVisibleMessage(
         bgRed.white.bold,
         `Node ${nodeRelease.majorVersion} has reached end-of-life${qualifier} and is not supported.`,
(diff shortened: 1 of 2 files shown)


rix0rrr (approved):
I'm not against this, but:

* Someone made the decision to make this non-silencable to begin with. What was their reasoning, and shouldn't we check with them?
* At this point, I don't see the value in having all these environment variables anymore. Can't we just have 1, `JSII_SILENCE_NODE_WARNINGS` for all of them? This is starting to get ridicilous.

mrgrain:
See https://github.com/aws/jsii/commit/880dbccc883005d10f00a1536561e0fe84b5fda6

Maybe @RomainMuller can comment on a preferred solution to this.

vinayak-kukreja:
Thank you both, I will check with Romain for this.

RomainMuller:
I made the decision to make the EOL warnings un-silence-able. But I don't have an extremely strong opinion on this.

The general line of reasoning is that we reserve the right to pull support of EOL runtime releases a month after they went EOL, and people should stop using those as soon as possible (security risk, risking new bugs we won't necessarily be able to address, etc...).

In general, this is to make it easier for us to "move the dial" of minimum supported node release without having to stand by endlessly waiting for some notice to have been "seen" by enough folks... So we can start using new runtime features (e.g: in the past there'd have been a strong desire to use `WeakRef` and related APIs, but our minimum supported node did not have those, and it took AGES to move forward) and safety mechanisms (such as using the `node:` imports for node standard library, to avoid surprising resolutions & make it clear what is "built-in" versus "from dependencies").

I'm not against silencing this. I would however somewhat prefer if silencing the EOL warning required setting the environment variable to a specific value that implies users acknowledged that they should be moving to a non-EOL runtime as soon as possible...

vinayak-kukreja:
```
The general line of reasoning is that we reserve the right to pull support of EOL runtime releases a month after they went EOL, and people should stop using those as soon as possible (security risk, risking new bugs we won't necessarily be able to address, etc...).
```

I see it has not yet been a month since EOL of Node 14 and we test with `node 14` in our [main](https://github.com/aws/jsii/blob/main/.github/workflows/main.yml) workflow. I have not yet updated the github workflows with node 16, will make a separate PR for them but it should probably be merged after we complete the month timeline. WDYT?

vinayak-kukreja:
>     * At this point, I don't see the value in having all these environment variables anymore. Can't we just have 1, `JSII_SILENCE_NODE_WARNINGS` for all of them? This is starting to get ridicilous.

I see, but that would hide the specific error. For instance, looks like all of these are there for different reasons,
```
        "JSII_SILENCE_WARNING_KNOWN_BROKEN_NODE_VERSION",
        "JSII_SILENCE_WARNING_UNTESTED_NODE_VERSION",
        "JSII_SILENCE_WARNING_DEPRECATED_NODE_VERSION",
        "JSII_SILENCE_WARNING_END_OF_LIFE_NODE_VERSION",
```

rix0rrr:
> I see, but that would hide the specific error. For instance, looks like all of these are there for different reasons,

I understand that we **can**. I’m just asserting that we **won’t**. 

And since we won’t, this feature is actually more of a hindrance than a help.

rix0rrr:
I won’t block the PR on this. But I’d like you to see my point that a theoretical feature that is never used in practice is a liability, not an asset

packages/@jsii/check-node/src/index.ts
```diff
@@ -21,14 +21,18 @@ export function checkNode(envPrefix = 'JSII'): void {
     'Should you encounter odd runtime issues, please try using one of the supported release before filing a bug report.';
 
   if (nodeRelease?.endOfLife) {
+    const silenceVariable = `${envPrefix}_SILENCE_WARNING_END_OF_LIFE_NODE_VERSION`;
+    const acknowledgeNodeEol =
+      'I acknowledge end of life(EOL) status for this node version. The CDK reserves the right to pull support of EOL runtime releases a month after they went EOL status, and people are recommended to stop using those as soon as possible';
```
TheRealAmazonKendra, line 26:
```suggestion
      'Node14 is now end of life (EOL) as of April 30, 2023. Support of EOL runtimes are only guaranteed for 30 days after EOL. By silencing this warning you acknowledge that you are using an EOL version of Node and will upgrade to a supported version as soon as possible.';
```

rix0rrr (approved):

packages/@jsii/check-node/src/index.ts
vinayak-kukreja: Thank you, updated.

PR #4081, opened 2023-05-04
fix(pacmak): go docs overview displays repeating description

vinayak-kukreja:
Our `go docs` have been broken since jsii `v1.65.0`. This was due to moving to emitting one file per type and headers being added to all those files: [Impacting PR](https://github.com/aws/jsii/pull/3698/files#diff-0a99c713a1282d6efe349c932d74dbfbdf1d926834c46ebfb5d65daeb27073f3R263)

The first occurrence was in [docs for cdk v2.39.0](https://pkg.go.dev/github.com/aws/aws-cdk-go/awscdk/v2@v2.39.0) and is still present in the [latest version](https://pkg.go.dev/github.com/aws/aws-cdk-go/awscdk/v2).

The header being added is `Version 2 of the AWS Cloud Development Kit library` to multiple files. It's determining this value from the [package.json](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/package.json#L4) file of `aws-cdk-lib`. I see adding this is most relevant for [maven](https://github.com/aws/jsii/blob/main/packages/jsii-reflect/lib/assembly.ts#L46-L52), hence removing it from `go` since I do not believe it's bringing much value to our documentation. Let me know if you feel otherwise :) 

---

### DIFF
--- a/packages/jsii-pacmak/lib/targets/go/package.ts
+++ b/packages/jsii-pacmak/lib/targets/go/package.ts
@@ -37,6 +37,7 @@ import { VersionFile } from './version-file';
 
 export const GOMOD_FILENAME = 'go.mod';
 export const GO_VERSION = '1.18';
+const MAIN_FILE = 'main.go';
 
 /*
  * Package represents a single `.go` source file within a package. This can be the root package file or a submodule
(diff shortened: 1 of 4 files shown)


packages/jsii-pacmak/lib/targets/go/package.ts
```diff
@@ -469,14 +469,6 @@ export class RootPackage extends Package {
     );
   }
 
-  protected emitHeader(code: CodeMaker) {
-    if (this.assembly.description !== '') {
-      code.line(`// ${this.assembly.description}`);
-    }
-    code.line(`package ${this.packageName}`);
-    code.line();
-  }
```
RomainMuller, line 478: I think this should still be emitted.... but _only_ in a single file (for example, `main.go`)

vinayak-kukreja: I see, will make the update. Thank you.

RomainMuller (approved):

Issue #4118, opened 2023-05-23
Python: JSIIMeta uses Any type instead of TypeVar variable


michelesr:
### Describe the bug

To work correctly with static type checker, it should use a TypeVar.

Please check this [issue comment](https://github.com/microsoft/pyright/issues/5156#issuecomment-1559837406) for more information.

### Expected Behavior

```python
T = TypeVar("T")
class JSIIMeta:
     ...
    def __call__(cls: Type[T], *args: Any, **kwargs) -> T:
```

### Current Behavior

```python
class JSIIMeta:
     ...
    def __call__(cls: Type[Any], *args: Any, **kwargs) -> Any:
```

### Reproduction Steps

N/A

### Possible Solution

See proposed snippet

### SDK version used

N/A

### Environment details (OS name and version, etc.)

N/A

ThomasSteinbach:
It would be nice, if someone could address this issue.
In VSCode with the latest version of Pylance installed, all the magic of typing and suggestions didn't work anymore.

<img width="277" alt="Bildschirmfoto 2023-06-26 um 10 23 14" src="https://github.com/aws/jsii/assets/1683246/bf78f90e-b0c7-40f5-8808-0ec20224110a">

sirrus233:
Can we get some sort of ACK that this issue has been recognized? It significantly impacts the usability of CDK.

sirrus233:
This can be closed, with the release of JSII 1.86

PR #4179, opened 2023-07-14
fix: type annotation for callable metaclass

sirrus233:
A recent update to pyright has surfaced a typing bug in JSII which means (in practice) that constructs in AWS CDK are no longer statically typed when using pyright. That includes most users of VSCode.

Further details in https://github.com/aws/jsii/issues/4118 and the linked context issue.

The fix is pretty simple. Since it's just a type annotation, not much to test. But I did build the package and confirmed that static typing is fixed after the change here is picked up.

---


RomainMuller (dismissed):

RomainMuller (approved):

RomainMuller (approved):

RomainMuller:
Well turns out `mypy` has different type-checking rules under that call (in particular, the behavior of `super().__call__` there... which requires an explicit cast... I fixed that for you.

PR #4193, opened 2023-07-25
fix(jsii-pacmak): comment lines should not contain comment ending

DanielMSchmidt:
When a comment contains `*/` (for any reason, in our case because the escaped version is unescaped within JSII) it should be escaped to not make the file invalid

### DIFF
--- a/packages/jsii-pacmak/test/generated-code/__snapshots__/target-java.test.js.snap
+++ b/packages/jsii-pacmak/test/generated-code/__snapshots__/target-java.test.js.snap
@@ -9787,6 +9787,14 @@ public class DocumentedClass extends software.amazon.jsii.JsiiObject {
     public void hola() {
         software.amazon.jsii.Kernel.call(this, "hola", software.amazon.jsii.NativeType.VOID);
     }
+
+    /**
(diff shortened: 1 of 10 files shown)


packages/jsii-pacmak/test/generated-code/__snapshots__/target-java.test.js.snap
```diff
@@ -9787,6 +9787,14 @@ public class DocumentedClass extends software.amazon.jsii.JsiiObject {
     public void hola() {
         software.amazon.jsii.Kernel.call(this, "hola", software.amazon.jsii.NativeType.VOID);
     }
+
+    /**
+     * (experimental) Say Moin, which is German for "Good morning" *&#47; (this comment is escaped because it contains a star slash).
```
RomainMuller, line 9792: This seems to indicate your new code path was not exercised...

DanielMSchmidt: Seems to be solved though, weird. I'll double check with the exact input I got.

RomainMuller (approved):

PR #4194, opened 2023-07-26
chore: upgrade dependencies in superchain image

TheRealAmazonKendra:
The Debian 10 image uses Python 3.7 which was causing failures elsewhere due to required minimum python 3.8. Debian 10 is also ten months out of security support so upgrading it seemed like the right decision here. This change also updates Node to 16 due to Node 14 being EOL.

---


RomainMuller:
The image tagging also needs to be updated...

RomainMuller:
Superseded by #4195

PR #4195, opened 2023-07-26
chore: multi-build super chain image with several Debian releases

TheRealAmazonKendra:
The Debian 10 image uses Python 3.7 which was causing failures elsewhere due to required minimum python 3.8. Debian 10 is also ten months out of security support so upgrading it seemed like the right decision here. This change also updates Node to 16 due to Node 14 being EOL..

This is the same as another PR posted but upgrading to Debian 12 instead.

---


TheRealAmazonKendra:
I was concerned that upgrading twice in one go wouldn't work so I did both and am letting them build/test here to see.

RomainMuller:
The image tagging also needs to be updated...

mrgrain (approved):

TheRealAmazonKendra:
@mergifyio update

.mergify/config.yml
```diff
@@ -33,7 +33,7 @@ pull_request_rules:
       - '#changes-requested-reviews-by=0'
       - status-success=Validate PR Title
       # Docker image validation
-      - status-success~=^jsii/superchain
+      - status-success~=^Docker Images/jsii/superchain
```
RomainMuller, line 36:
```suggestion
      - status-success~=^Done$
```

PR #4196, opened 2023-07-26
feat(jsii-pacmak): allow disabling go build conditionally

DanielMSchmidt:


packages/jsii-pacmak/lib/targets/go.ts
```diff
@@ -54,7 +54,13 @@ export class Golang extends Target {
       );
       return Promise.reject(e);
     }
-    await go('build', ['-modfile', localGoMod.path, './...'], { cwd: pkgDir });
+
+    if (process.env.JSII_GO_DISABLE_BUILD !== '1') {
```
RomainMuller, line 58: Let's make this opt-in... it's a win for everyone... So say if the `JSII_BUILD_GO` variable is set (whatever the value), then we'll run the `go build` step.

```suggestion
    if (process.env.JSII_BUILD_GO) {
```

RomainMuller (changes requested):

packages/jsii-pacmak/lib/targets/go.ts
DanielMSchmidt: That’s ok for me as well 👍

RomainMuller (approved):

RomainMuller:
@mergifyio update

PR #4197, opened 2023-07-26
feat(python): encode positional-only arguments in signatures

### LINKED ISSUE #2927: Python: function overrides with different parameter names are incompatible

### LINKED ISSUE #4541: (@aws_cdk): Python: incompatible interface implementations due to differing function argument names
In Python prior to 3.8 (and in all the jsii-generated Python code), all function arguments are keyword arguments. In the TS source code, a lot of interface implementations override their interface's functions with different parameter names, which makes them incompatible with the interface in Python.

First of all, it causes typing errors (I'm using pyright).
More importantly, it makes these python classes incompatible with their interfaces.
To add to this, each change in a positional parameter name on the TS side is a breaking change on the Python side.

For example, `@aws_cdk.aws_lambda.Function`'s `grantInvoke` method uses `grantee` as its parameter name, whereas `IFunction` uses `identity`. These end up being incompatible in Python.

https://github.com/aws/aws-cdk/blob/b78a1bbf445743d96c8e4f54e7d2e7cac204342a/packages/%40aws-cdk/aws-lambda/lib/function-base.ts#L306
https://github.com/aws/aws-cdk/blob/b78a1bbf445743d96c8e4f54e7d2e7cac204342a/packages/%40aws-cdk/aws-lambda/lib/function-base.ts#L81

On the jsii side, it would have to add a check for all overrides to be compatible, i.e. to use the same parameter names.

On the CDK side, we'd have to deprecate all of the incompatible parameter overrides and add properly named ones.

https://github.com/aws/jsii/issues/1919 might be related - using Protocols instead of a Metaclass would ensure that this cannot happen.

### Environment

  - **Framework Version: 1.114
  - **Language (Version): Python 3

This is :bug: Bug Report

RomainMuller:
Now that Python 3.7 reached end-of-life, and since Python 3.8 and greater have support for positional-only argument notation (any argument before a `/` delimiter may only be passed positionally), adjust code generation to leverage this feature in order to address issues where inheritance hierarchies would rename parameters, which is a non-issue in all languages but Python, where those could always be provided as keyword arguments before.

Fixes https://github.com/aws/jsii/issues/4541

BREAKING CHANGE: the generated Python code now requires Python 3.8 or later and encodes positional arguments as positional-only, making keyword-style usage impossible. Users who used the keyword-style convention need to update their code to use the positional syntax instead.

---


iliapolo:
This might not be the desired solution. See https://github.com/aws/jsii/issues/2927#issuecomment-1673071807

PR #4204, opened 2023-08-02
fix(kernel): incorrectly scoped FQN resolutions

### LINKED ISSUE #4202: ClassCastException when trying to get Element from Node
### Describe the bug

Hi,
I'm using aws-cdk for java. After upgrading the artifact software.amazon.jsii:jsii-runtime from version 1.85.0 to 1.86.0 a lot of my stacks don't compile anymore. Here an example stack
```
public class RolesAnywhereStack extends Stack {
    public RolesAnywhereStack(App parent) {
        super(parent, "RolesAnywhereStack");
        final Role bastionRole = Role.Builder.create(this, "RolesAnywhereBastionRole")
                .assumedBy(new ServicePrincipal("rolesanywhere.amazonaws.com"))
                .description("Role that is used when connecting via RolesAnywhere")
                .build();
        CfnRole cfnRole = (CfnRole) bastionRole.getNode().getDefaultChild();
    }
}
```
when running `cdk synthesize RolesAnywhereStack` a ClassCastException occurs:
```
Caused by: java.lang.ClassCastException: class software.constructs.Construct cannot be cast to class software.amazon.awscdk.services.iam.CfnRole (software.constructs.Construct and software.amazon.awscdk.services.iam.CfnRole are in unnamed module of loader java.net.URLClassLoader @7486e0ca)
```

with jsii-runtime 1.85.0 everything works fine.

### Expected Behavior

No ClassCastException

### Current Behavior

ClassCastException

### Reproduction Steps

try to synthesize this stack
```
public class RolesAnywhereStack extends Stack {
    public RolesAnywhereStack(App parent) {
        super(parent, "RolesAnywhereStack");
        final Role bastionRole = Role.Builder.create(this, "RolesAnywhereBastionRole")
                .assumedBy(new ServicePrincipal("rolesanywhere.amazonaws.com"))
                .description("Role that is used when connecting via RolesAnywhere")
                .build();
        CfnRole cfnRole = (CfnRole) bastionRole.getNode().getDefaultChild();
    }
}
```

### Possible Solution

_No response_

_No response_

### SDK version used

software.amazon.jsii:jsii-runtime:1.85.0

Debian openjdk version "17.0.2" 2022-01-18 OpenJDK Runtime Environment (build 17.0.2+8-86) OpenJDK 64-Bit Server VM (build 17.0.2+8-86, mixed mode, sharing)

### LINKED ISSUE #4203: jsii 1.86.0 incorrectly infers the type of an object
### Describe the bug

I have a larg-ish Python CDK stack which uses `aws_cdk.pipelines.CodePipelineSource.s3()` and then calls `.source_attribute()` on the result.

### Expected Behavior

`cdk synth` continues working.

### Current Behavior

`cdk synth` fails with `AttributeError: 'Step' object has no attribute 'source_attribute'`.

### Reproduction Steps

Sorry, I tried to produce a self-contained example. However, this app contains another, large-ish stack. Removing this large stack fixes the problem. At this point, I do not know how to debug this some more.

I also tried comparing the output when running with `JSII_DEBUG=1` with the working and non-working case, but that didn't provide any insights either.

Edit: See my first comment below for an actual reproducer.

### Possible Solution

> Suggest a fix/reason for the bug

No idea about a fix, but "reason why I think this is a bug in jsii": `pip install jsii==1.85.0` makes things work while 1.86.0 breaks again. No other commands run between the two versions.

- With jsii 1.85.0 everything works fine.
- With jsii 1.86.0 my code started failing with `AttributeError: 'Step' object has no attribute 'source_attribute'`.
- When `print()`ing the return value of `aws_cdk.pipelines.CodePipelineSource.s3()`
  - With jsii 1.85.0 this shows a type of `aws_cdk.pipelines._CodePipelineSourceProxy`
  - With jsii 1.86.0 this shows indeed a type of `aws_cdk.pipelines.Step` which indeed does not have `source_attribute`.

I don't know what magic jsii is doing to infer types, but somehow this magic now stops one step too early.

Also: #4202 sounds loosely related since it is also about a wrong class and also talks about 1.85.0 working and 1.86.0 breaking. If the reproducing instructions there are correct, that would definitively be the simpler reproducer.

If you have some suggestions on how to debug this some more, I can do so. But I cannot share the whole CDK app.

### SDK version used

1.86.0

Ubuntu 23.04

RomainMuller:
The registration was done on the object's prototype, and the value from the constructor was always used, even if that was inherited, such that if a parent type had already been resolved previously, all child types would use the parent's FQN instead of their own.

Addressed this by instead storing the associations in an external WeakMap, and added a test case to validate correct behavior.

Fixes aws/aws-cdk#26604
Fixes #4202
Fixes #4203

---

### DIFF
--- a/packages/@jsii/kernel/src/objects.test.ts
+++ b/packages/@jsii/kernel/src/objects.test.ts
@@ -1,4 +1,4 @@
-import { ObjectTable } from './objects';
+import { ObjectTable, jsiiTypeFqn } from './objects';
 
 const mockResolve = jest.fn();
 
@@ -23,4 +23,21 @@ test('registered objects have clean JSON.Stringify', () => {
   subject.registerObject(obj, 'Object');
 
   expect(JSON.stringify(obj)).toEqual(expected);
+});
+
+test('FQN resolution is not broken by inheritance relationships', () => {
+  const rtti = Symbol.for('jsii.rtti');
+
+  class Parent {
+    public static readonly [rtti] = { fqn: 'phony.Parent' };
+  }
+  class Child extends Parent {
+    public static readonly [rtti] = { fqn: 'phony.Child' };
+  }
+
+  const parent = new Parent();
+  expect(jsiiTypeFqn(parent, () => true)).toBe(Parent[rtti].fqn);
+
+  const child = new Child();
+  expect(jsiiTypeFqn(child, () => true)).toBe(Child[rtti].fqn);
 });
--- a/packages/@jsii/kernel/src/objects.ts
+++ b/packages/@jsii/kernel/src/objects.ts
@@ -16,30 +16,14 @@ const OBJID_SYMBOL = Symbol.for('$__jsii__objid__$');
 const IFACES_SYMBOL = Symbol.for('$__jsii__interfaces__$');
 
 /**
- * Symbol we use to tag constructors that are exported from a jsii module.
- */
-const JSII_TYPE_FQN_SYMBOL = Symbol('$__jsii__fqn__$');
-
-/**
  * Symbol under which jsii runtime type information is stored.
  */
 const JSII_RTTI_SYMBOL = Symbol.for('jsii.rtti');
 
 /**
- * Exported constructors processed by the jsii compiler have runtime type
- * information woven into them under the `[JSII_RTTI]` symbol property.
+ * Cache for resolved associations between constructors and FQNs.
  */
-interface MaybeManagedConstructor {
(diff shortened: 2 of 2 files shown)


psychon:
Since this only mentions the issue over at aws-cdk: I can't check this personally, but from the description it very much sounds like this would also fix https://github.com/aws/jsii/issues/4202 and https://github.com/aws/jsii/issues/4203

RomainMuller:
@psychon you're absolutely correct...

PR #4239, opened 2023-08-17
chore: switch to python 3.8

corymhall:
Python 3.7 has been end-of-life since 2023-06-27, and an
increasing amount of our dependencies are dropping support for it as a
consequence.

We introduced a warning to users that are still using Python 3.7 with
release v1.86.0 around August 1, 2023.

We are now effectively dropping support for Python 3.7.

Related: https://github.com/aws/jsii/commit/0ef7acef765690f99075888d6c3e6f03d7cc2d59

BREAKING CHANGE: jsii-pacmak and the libraries it generates for Python
targets now require a minimum Python version of 3.8, instead of 3.7
previously. We recommend users upgrade to the latest supported Python
release (Python 3.11).

---


corymhall:
### ⚠️ This PR should remain labeled with `pr/do-not-merge` at least until `2023-09-01` (at which point it'll have been a full month since we released the first version of jsii that emits a deprecation warning to users still on Python 3.7).

mikewrighton (approved):

PR #4279, opened 2023-09-26
perf(jsii-reflect): Cache classes in TypeSystem.

mikewrighton:
The `TypeSystem.get classes()` method calls `collectTypes` on every invocation, and this is causing performance problems in `awslint`. This change caches the class list for each loaded assembly, avoiding the duplicated work.

---


rix0rrr (approved):
Conditionally approved.

I don't mind this, but wouldn't we want the same behavior for `ts.interfaces`, `ts.enums`, etc?

mikewrighton:
> Conditionally approved.
> 
> I don't mind this, but wouldn't we want the same behavior for `ts.interfaces`, `ts.enums`, etc?

So I targeted `ts.classes` because it happens to be the only method which causes performance issues in awslint, I guess because the aws-cdk assembly has a large number of classes compared with other types. It wouldn't hurt to cache these other types, but I wanted to fix the immediate linter issue first.

PR #4324, opened 2023-11-13
fix(kernel): invokeBinScript fails when using symlinked cache

mrgrain:
Invoking bin scripts from jsii packages would fail if the symlinked cache was enabled and the invoked script depended on an other package. The reason for this was that scripts were only invoked with `--preserve-symlinks`, however for "main scripts" (like standalone binaries), we also need to call `--preserve-symlinks-main` on the node process.

Note this cannot be tested in the kernel package itself, as it's not possible to invoke the test process with the correct options. 

---

### DIFF
--- a/packages/jsii-calc/bin/run.ts
+++ b/packages/jsii-calc/bin/run.ts
@@ -2,10 +2,15 @@
 
 /* eslint-disable no-console */
 
+import * as calcLib from '@scope/jsii-calc-lib';
+
 const delay = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms));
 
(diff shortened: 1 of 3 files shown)


packages/jsii-calc/bin/run.ts
```diff
@@ -2,11 +2,16 @@
 
 /* eslint-disable no-console */
 
+import * as calcLib from '@scope/jsii-calc-lib';
```
mrgrain, line 5: This was the previously untested scenario.

madeline-k (approved):

PR #4325, opened 2023-11-13
feat: make node 18 the default

mrgrain:
Removes testing and superchain releases for EOL Node 16

Deprecation and EOL messages of node versions are automated, so no need to announce or wait.
We are well past the promised 30 days support after Node 16 went EOL.

---

### DIFF
--- a/packages/@jsii/check-node/src/index.ts
+++ b/packages/@jsii/check-node/src/index.ts
@@ -22,12 +22,13 @@ export function checkNode(envPrefix = 'JSII'): void {
 
   if (nodeRelease?.endOfLife) {
     const silenceVariable = `${envPrefix}_SILENCE_WARNING_END_OF_LIFE_NODE_VERSION`;
-    const acknowledgeNodeEol =
-      'Node14 is now end of life (EOL) as of April 30, 2023. Support of EOL runtimes are only guaranteed for 30 days after EOL. By silencing this warning you acknowledge that you are using an EOL version of Node and will upgrade to a supported version as soon as possible.';
+    const silencedVersions = (process.env[silenceVariable] ?? '')
+      .split(',')
+      .map((v) => v.trim());
     const qualifier = nodeRelease.endOfLifeDate
       ? ` on ${nodeRelease.endOfLifeDate.toISOString().slice(0, 10)}`
       : '';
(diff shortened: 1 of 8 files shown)


packages/@jsii/check-node/src/index.ts
```diff
@@ -22,12 +22,13 @@ export function checkNode(envPrefix = 'JSII'): void {
 
   if (nodeRelease?.endOfLife) {
     const silenceVariable = `${envPrefix}_SILENCE_WARNING_END_OF_LIFE_NODE_VERSION`;
-    const acknowledgeNodeEol =
-      'Node14 is now end of life (EOL) as of April 30, 2023. Support of EOL runtimes are only guaranteed for 30 days after EOL. By silencing this warning you acknowledge that you are using an EOL version of Node and will upgrade to a supported version as soon as possible.';
+    const silencedVersions = (process.env[silenceVariable] ?? '')
+      .split(',')
+      .map((v) => v.trim());
```
mrgrain, line 27: The previous implementation was very verbose and used a "nanny" approach. In reality we need this check only to silence warnings in tests. It did also not account for different versions.

Re-implemented this to require ack of every specific node version.

packages/@jsii/python-runtime/tests/test_invoke_bin.py
```diff
@@ -23,8 +23,8 @@ def silence_node_deprecation_warnings():
     for var in variables:
         environ[var] = "1"
 
-    nodeEolAcknowledgement = "Node14 is now end of life (EOL) as of April 30, 2023. Support of EOL runtimes are only guaranteed for 30 days after EOL. By silencing this warning you acknowledge that you are using an EOL version of Node and will upgrade to a supported version as soon as possible."
-    environ["JSII_SILENCE_WARNING_END_OF_LIFE_NODE_VERSION"] = nodeEolAcknowledgement
+    # silence this for the next decades
+    environ["JSII_SILENCE_WARNING_END_OF_LIFE_NODE_VERSION"] = "14,16,18,20,22,24,26,28,30,32,34"
```
mrgrain, line 27: This test should never fail because of an outdated node version.
We have other mechanisms to check this.

kaizencc (approved):
Lgtm

PR #4389, opened 2024-01-03
fix(pacmak): invokeBinScript fails when using symlinked cache

mrgrain:
This was previously attempted to fix in https://github.com/aws/jsii/pull/4324 While the above fix resolves issues with dependencies, it causes failures when the binary is shelling out to other node processes. This is due to the intrusive and indiscriminate overloading of NODE_OPTIONS, which will forcibly apply to any child processes as well. While in theory adding the symlink flags should not be an issue, this seems to trigger a bug in node: https://github.com/nodejs/node/issues/41000

tl;dr this all sucks very much and we are now just disabling the runtime cache for binaries.

---

### DIFF
--- a/packages/@jsii/kernel/src/kernel.ts
+++ b/packages/@jsii/kernel/src/kernel.ts
@@ -1372,11 +1372,6 @@ export class Kernel {
       // Make sure the current NODE_OPTIONS are honored if we shell out to node
       const nodeOptions = [...process.execArgv];
 
-      // When we are using the symlinked version of the cache, we need to preserve both symlink settings for binaries
-      if (nodeOptions.includes('--preserve-symlinks')) {
-        nodeOptions.push('--preserve-symlinks-main');
-      }
-
       return {
         command: path.join(packageDir, scriptPath),
         args: req.args ?? [],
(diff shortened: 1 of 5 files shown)


packages/@jsii/kernel/src/kernel.ts
```diff
@@ -1372,11 +1372,6 @@ export class Kernel {
       // Make sure the current NODE_OPTIONS are honored if we shell out to node
       const nodeOptions = [...process.execArgv];
 
-      // When we are using the symlinked version of the cache, we need to preserve both symlink settings for binaries
-      if (nodeOptions.includes('--preserve-symlinks')) {
-        nodeOptions.push('--preserve-symlinks-main');
-      }
```
mrgrain, line 1378: This reverts https://github.com/aws/jsii/pull/4324

packages/jsii-pacmak/test/generated-code/__snapshots__/target-python.test.js.snap
```diff
@@ -491,69 +491,6 @@ exports[`Generated code for "@scope/jsii-calc-base": <outDir>/python/src/scope/j
...
-+    """Type checking stubs"""
-+    pass
-+
-+def _typecheckingstub__66400f6ebe2f9a3fbb550342b894bebf4f5368890843f3917e2b903f62d48b38(
-+    *args: typing.Any,
-+) -> None:
-+    """Type checking stubs"""
-+    pass
-`;
-
```
mrgrain, line 556: pruning some unused snapshots

rix0rrr (approved):

PR #4449, opened 2024-03-09
fix(jsii-pacmak): change filename for .jsii.tgz file

climbertjh:
This update changes the filename of the generated `.jsii.tgz` file to use a `-` rather than a `@` symbol between the `package-name>` and `package-version`.

---


climbertjh:
When I look at the failing tests, it looks like the test-cases (based on generated output snapshots) are expecting the generated `.jsii.tgz` file to have the `@` sign in the name rather than the filename.

Do I need to re-generate the snapshots as a part of this PR?

climbertjh:
After testing package install using `npm` as the package manager, I have found that the existing file-name format works for that package manager.

Because of this I am closing this PR.

Issue #4469, opened 2024-03-30
Upgrade version of typeguard in python runtime


sam-goodwin:
### Describe the feature

The python runtime for the CDK has an old and inflexible dependency on `typeguard~=2.13.3`

https://github.com/aws/jsii/blob/main/packages/%40jsii/python-runtime/setup.py#L37C10-L37C27

### Use Case

I would like to have a CDK application include the latest pandera libraries but this fixed dependency is a blocker.

### Proposed Solution

Increase the range and add version-aware adaptation code if required to handle the different versions.

Later versions of `typeguard` no longer accept the `argname` parameter. We'd have to change this line of code:

https://github.com/aws/jsii/blob/576fd201c686a30270e4bdbd1c0d9fc3e3739101/packages/jsii-pacmak/lib/targets/python.ts#L3205-L3207

### Acknowledgements

- [x] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### CDK version used

2.134.0

### Environment details (OS name and version, etc.)

Mac OS

sam-goodwin:
Just tested a monkey patch and removing the `argname` from the `check_type` function allows this to work for `3.0.2` up to `4.2.1`

```py
old_check_type = typeguard.check_type


def new_check_type(*args: Any, **kwargs: Any) -> Any:
    if "argname" in kwargs:
        del kwargs["argname"]
    return old_check_type(*args, **kwargs)


typeguard.check_type = new_check_type
```

PR #4470, opened 2024-03-31
feat: upgrade python runtime version of typeguard to >=3.0.2

sam-goodwin:
---
Closes: #4469

This change upgrades the python runtime's version of `typeguard` to `>=3.0.2` which was a significant breaking change where `argname` was removed. 

- [ ] I am unsure what the impact of removing `argname` is on error messages.

It also adds `@typing.runtime_checkable` to all interfaces (`I*`) because of raised warnings.

- [ ] I am still unclear whether the `runtime_checkable` decorator was required because I can't explain why it wasn't required before. But, my monkey-patch experiments resulted in a ton of warnings such as:
```
/./packyak.config.py:11: UserWarning: Typeguard cannot check the IReusableStackSynthesizer protocol because it is a non-runtime protocol. If you would like to type check this protocol, please use @typing.runtime_checkable
  return old_check_type(*args, **kwargs)
/./packyak.config.py:11: UserWarning: Typeguard cannot check the IStackSynthesizer protocol because it is a non-runtime protocol. If you would like to type check this protocol, please use @typing.runtime_checkable
  return old_check_type(*args, **kwargs)
```

### DIFF
--- a/packages/@jsii/python-runtime/setup.py
+++ b/packages/@jsii/python-runtime/setup.py
@@ -34,7 +34,7 @@ setuptools.setup(
         "cattrs>=1.8,<23.3",
         "importlib_resources>=5.2.0",
         "publication>=0.0.3",  # This is used by all generated code.
-        "typeguard~=2.13.3",  # This is used by all generated code.
+        "typeguard>=3.0.2",  # This is used by all generated code.
         "python-dateutil",
         "typing_extensions>=3.8,<5.0",
     ],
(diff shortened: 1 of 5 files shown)


sam-goodwin:
Looks like there are some test failures around interfaces. Will dig into these. I thought I ran the tests locally but will check again.

sam-goodwin:
Strange:
```py
@jsii/python-runtime: E       typeguard.TypeCheckError: method did not match any element in the union:
@jsii/python-runtime: E         int: is not an instance of int
@jsii/python-runtime: E         float: is neither float or int
```

iliapolo:
@sam-goodwin thanks for kicking this off and the work you put into it. We have taken this over so i'm closing this PR in favor of https://github.com/aws/jsii/pull/4611

PR #4477, opened 2024-04-09
fix(jsii): replace `@link` tags with Markdown code blocks

nmussy:
Out of habit, I've been using `@link` JSDoc tags in my CDK contributions to reference other entities. I just realized that it might be an issue with the generated Markdown documentation, which it is, see for example:

https://github.com/aws/aws-cdk/blob/58ebabc73d026e78f5804560ce2f23e112a38a66/packages/aws-cdk-lib/aws-cloudwatch/lib/variable.ts#L54-L57

image

Which is currently interpreted in the [docs](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_cloudwatch.SearchComponents.html) as:

image

This PR is just a basic readability fix to replace `{@link Values.fromSearchComponents}` with `Values.fromSearchComponents` in the generated Markdown. 

The end goal here should be to replace them with proper hash or relative links, e.g. `[Values.fromSearchComponents](aws-cdk-lib.aws_cloudwatch.Values.html#static-fromwbrsearchwbrcomponentscomponents)`, but I can't see an easy way of accomplishing this given my current lack of visibility with JSII

---

### DIFF
--- a/packages/jsii/lib/docs.ts
+++ b/packages/jsii/lib/docs.ts
@@ -59,9 +59,9 @@ export function parseSymbolDocumentation(
   sym: ts.Symbol,
   typeChecker: ts.TypeChecker,
 ): DocsParsingResult {
-  const comment = ts
-    .displayPartsToString(sym.getDocumentationComment(typeChecker))
-    .trim();
+  const comment = replaceLinkTags(
+    ts.displayPartsToString(sym.getDocumentationComment(typeChecker)).trim(),
+  );
   const tags = reabsorbExampleTags(sym.getJsDocTags());
 
   // Right here we'll just guess that the first declaration site is the most important one.
@@ -314,4 +314,15 @@ function reabsorbExampleTags(tags: ts.JSDocTagInfo[]): ts.JSDocTagInfo[] {
   }
 
   return ret;
+}
+
+/**
+ * Replace {\@link foo} tags with `foo` to make them Markdown compatible
+ *
+ * Ideally, we'd replace them with something like `[foo](#foo)`,
+ * but that link is not always relative,
+ * and would require linking between documentation pages
+ */
+function replaceLinkTags(comment: string): string {
+  return comment.replace(/\{\s*@link\s+([^}\s]+)\s*\}/g, '`$1`');
 }
--- a/packages/jsii/lib/docs.ts
+++ b/packages/jsii/lib/docs.ts
@@ -105,7 +105,7 @@ function parseDocParts(
       if (tagNames.has(name)) {
         const ret = tagNames.get(name);
         tagNames.delete(name);
-        return ret ?? '';
+        return replaceLinkTags(ret ?? '');
       }
     }
     return undefined;
--- a/packages/jsii/test/docs.test.ts
+++ b/packages/jsii/test/docs.test.ts
@@ -379,14 +379,40 @@ test('replace @link tags with inline code blocks', () => {
     /**
      * Hello this is the documentation for {@link Foo},
      * which implements { @link    IFoo    }.
+     * 
+     * @deprecated - Use {@link Bar} instead.
      */
     export class Foo implements IFoo {
-      public bar() { }
+      /**
+       * Do the {@link Foo}
+       *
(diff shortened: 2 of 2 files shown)


mrgrain:
I'm not sure I see how using random jsdoc syntax that might or might not be supported in a downstream processor is related to jsii, or even a bug.

This feels to me like a change is required to whatever is generating documentation pages from the docblock.

nmussy:
Not certain what you mean by random but just to clarify, `@link` tag is both standard in [JSDoc](https://jsdoc.app/tags-inline-link) and [TSDoc](https://tsdoc.org/pages/tags/link/). It's also of pretty big DX benefit, being able to navigate using links vs searching for a symbol name in an unknown file.

I see your point about `parseDocParts` not being the best place to update the raw documentation, since it's only splitting existing contents, and not really modifying them. On the other hand, it's also removing and replacing tags as is, and replacing them with structure. For example, `@deprecated` determines both the `Docs.stability` and `Docs.deprecated` props, and `@deprecated` is removed from the contents.

I'd be happy to move this down to the HTML transition stage if you disagree 👍

mrgrain:
> Not certain what you mean by random but just to clarify, `@link` tag is both standard in [JSDoc](https://jsdoc.app/tags-inline-link) and [TSDoc](https://tsdoc.org/pages/tags/link/). It's also of pretty big DX benefit, being able to navigate using links vs searching for a symbol name in an unknown file.

Yes, apologies. By "random" I just meant that there are a lot of doctags that have no meaning in jsii and we definitely don't plan to support all of them. It was not meant to cast any judgment on the value of this particular tag.

> I see your point about `parseDocParts` not being the best place to update the raw documentation, since it's only splitting existing contents, and not really modifying them. On the other hand, it's also removing and replacing tags as is, and replacing them with structure. For example, `@deprecated` determines both the `Docs.stability` and `Docs.deprecated` props, and `@deprecated` is removed from the contents.

That's fair. I think it's not very well defined what jsii wants to own in regards to doctags. The best description we have available is [in the spec package](https://github.com/aws/jsii/blob/9edbf46c40121eeecf1c0cd89f1ded5083520c99/packages/%40jsii/spec/src/assembly.ts#L302). We could argue that only the properties in there are interesting to jsii. However obviously this could be extended to anything we like!

The issue as you've opened it appears to be mainly about the generated docs. I think it's more realistic to assume that inline docs are likey written in jsdoc syntax and make sure to use something like [jsdoc2md/jsdoc-to-markdown](https://github.com/jsdoc2md/jsdoc-to-markdown) when converting them into documentation. It's also not a straight forward conversion, since `@link` can also be a url.

> I'd be happy to move this down to the HTML transition stage if you disagree 👍

Yes, I'd prefer that. Unfortunately the CDK doc generated is not public. It might be easier to configure the aws-cdk repo to disallow link tags. AFAIK you can use `@see` for a similar result.

----

### Sidebar:

A better way to approach this problem, might be to think about what the `@link` tag is used for. jsii-compiler extracts the relevant symbols of a TS program, so that pacmak can generate bindings for supported languages. Documentation is part of the generated bindings. 

I could see an argument that `@link` is a very JSDoc centric feature (okay Java uses it) that might not work in docblocks for other languages. From that perspective the eventual goal would be to convert link tags to whatever the target language equivalent is. Not saying outright that we shouldn't do this, but it sounds quite complicated for maybe not that much gain. From this particular perspective it might be more appropriate to detect the tag (and others), and emit a warning stating that the tag might not be supported in other languages. 

Eitherway, this seems like embarking on quite the journey. 🙈

nmussy:
@mrgrain 
> AFAIK you can use `@see` for a similar result.

Unfortunately not, `@link` is an inline tag, `@see` isn't. `@link` also allows you to change the text link, which can be used in combination with `@see` to do the following:

```ts
/**
 * @deprecated - We don't have a direct alternative, but you might find something similar in {@link OtherClass}
 * @see {@link https://example.com/docs Docs} for additional info
 */
```

mrgrain:
Yes, similar not the same.

PR #4611, opened 2024-08-22
feat(pacmak): allow `typeguard==3.x` and `typeguard==4.x` 

iliapolo:
Currently python projects generated by pacmak define a `typeguard` dependency range of  `typeguard~=2.13.3`, which prevents users from brining newer major version of `typeguard` into their projects. 

This PR adds support for `typeguard==3.x` and `typguard==4.x`, which are the latest versions currently available. We intentionally do not allow an open range because every major version brings breaking changes with it that might need to be addressed.

### Notes

- We couldn't just drop support for `typeguard==2.x` because that would be a breaking change. 
- We could potentially release a new major version of pacmak to make the code more maintainable. My take is that this PR doesn't complicate the code to the extent of justifying a new major version, but lets discuss.

Fixes https://github.com/aws/jsii/issues/4469

---


packages/@jsii/python-runtime/src/jsii/_runtime.py
```diff
@@ -168,6 +168,10 @@ def implements(*interfaces: Type[Any]) -> Callable[[T], T]:
     def deco(cls):
         cls.__jsii_type__ = getattr(cls, "__jsii_type__", None)
         cls.__jsii_ifaces__ = getattr(cls, "__jsii_ifaces__", []) + list(interfaces)
+        cls.__jsii_proxy_class__ = lambda: getattr(cls, "__jsii_proxy_class__", None)
```
iliapolo, line 171: This is required because `__jsii_proxy_class__` is added to protocols, and therefore should be implemented in classes.

https://github.com/aws/jsii/blob/9ecea560bef8e0196c9a4874110c19e946f3b3ea/packages/jsii-pacmak/lib/targets/python.ts#L1068-L1070 

Note that prior to 4.x, `typeguard` didn't actually type check our protocols so this wasn't needed.

packages/@jsii/python-runtime/tests/test_python.py
```diff
@@ -27,17 +27,6 @@ def test_inheritance_maintained(self):
 
         assert base_names == ["DerivedStruct", "MyFirstStruct"]
 
-    def test_descriptive_error_when_passing_function(self):
```
iliapolo, line 30: Moved to `test_runtime_checking.py`

packages/jsii-pacmak/lib/targets/python.ts
```diff
@@ -1728,7 +1728,44 @@ class PythonModule implements PythonType {
...
+      'def check_type(argname: str, value: object, expected_type: typing.Any) -> typing.Any',
+    );
+    code.openBlock('if TYPEGUARD_MAJOR_VERSION <= 2');
+    code.line(
+      'return typeguard.check_type(argname=argname, value=value, expected_type=expected_type) # type:ignore',
+    );
+    code.closeBlock();
+    code.openBlock('else');
+    code.line(
+      'if isinstance(value, jsii._reference_map.InterfaceDynamicProxy): # pyright: ignore [reportAttributeAccessIssue]',
```
iliapolo, line 1749: Typeguard doesn't support type check against dynamic proxies because it only [compares against type declaration](https://github.com/agronholm/typeguard/blob/master/src/typeguard/_checkers.py#L657):

```python
def check_protocol(
    value: Any,
    origin_type: Any,
    args: tuple[Any, ...],
    memo: TypeCheckMemo,
) -> None:
    subject: type[Any] = value if isclass(value) else type(value)
```

Prior to version 4.x, this wasn't needed because `typeguard` did not even attempt it. Its protocol checking logic was completely different:

```python
def check_protocol(argname: str, value, expected_type):
    # TODO: implement proper compatibility checking and support non-runtime protocols
    if getattr(expected_type, '_is_runtime_protocol', False):
        if not isinstance(value, expected_type):
            raise TypeError('type of {} ({}) is not compatible with the {} protocol'.
                            format(argname, type(value).__qualname__, expected_type.__qualname__))
```

This code never got executed on our protocols because `_is_runtime_protocol` [defaults to `False`](https://github.com/python/typing_extensions/blob/main/src/typing_extensions.py#L705), and is only [set to `True`](https://github.com/python/cpython/blob/3.10/Lib/typing.py#L1733) if the protocol is decorated with `@typing.runtime_checkable`. Since our protocols aren't decorated with it, the protocol type checking never took place. 

So with support for 4.x we are now adding protocol checking, but still avoiding runtime proxies because they are not handled properly by `typeguard`.

packages/@jsii/python-runtime/package.json
```diff
@@ -28,9 +28,11 @@
     "build": "cp ../../../README.md . && rm -f jsii-*.whl && npm run generate && npm run deps && npm run lint",
     "lint": "ts-node build-tools/venv.ts black .",
     "package": "package-python && package-private",
-    "test": "npm run test:gen && npm run test:run && npm run test:types",
+    "test": "npm run test:gen && npm run test:run:typeguard-2 && npm run test:run:typeguard-3 && npm run test:run:typeguard-4 && npm run test:types",
```
iliapolo, line 31: We will now run three test suites, for each typeguard major version we support. These tests are pretty quick so compared to the current latency of PR validation, they don't add much.

packages/jsii-pacmak/lib/targets/python.ts
```diff
@@ -1728,7 +1728,44 @@ class PythonModule implements PythonType {
     code.line('import publication');
     code.line('import typing_extensions');
     code.line();
-    code.line('from typeguard import check_type');
+
+    code.line('import typeguard');
+    code.line('from importlib.metadata import version');
```
iliapolo, line 1733: This was [added in Python 3.8](https://docs.python.org/3/whatsnew/3.8.html#new-modules), which is the lowest version of python we support - so its safe to use.

packages/jsii-pacmak/lib/targets/python.ts
```diff
@@ -1728,7 +1728,44 @@ class PythonModule implements PythonType {
     code.line('import publication');
     code.line('import typing_extensions');
     code.line();
-    code.line('from typeguard import check_type');
+
+    code.line('import typeguard');
+    code.line('from importlib.metadata import version as _metadata_package_version');
```
iliapolo, line 1733: This was [added in Python 3.8](https://docs.python.org/3/whatsnew/3.8.html#new-modules), which is the lowest version of python we support - so its safe to use. Aliasing to avoid possible conflicts with other generated code.

packages/@jsii/python-runtime/tests/test_runtime_type_checking.py
```diff
@@ -1,14 +1,31 @@
...
 import jsii
+import typeguard
 
+from importlib.metadata import version
+
+TYPEGUARD_MAJOR_VERSION = int(version("typeguard").split(".")[0])
+
+
+@jsii.implements(jsii_calc.IBellRinger)
+class PythonInvalidBellRinger:
```
iliapolo, line 17: Thanks for 4.x we can now statically typecheck protocols so added a test for it.

packages/jsii-pacmak/lib/targets/python.ts
```diff
@@ -1728,7 +1728,44 @@ class PythonModule implements PythonType {
...
+    code.closeBlock();
+    code.openBlock('else');
+    code.line(
+      'if isinstance(value, jsii._reference_map.InterfaceDynamicProxy): # pyright: ignore [reportAttributeAccessIssue]',
+    );
+    code.line('   pass');
+    code.openBlock('else');
+    code.openBlock('if TYPEGUARD_MAJOR_VERSION == 3');
+    code.line(
+      'typeguard.config.collection_check_strategy = typeguard.CollectionCheckStrategy.ALL_ITEMS',
```
iliapolo, line 1755: In version 3.x `collection_check_strategy` can only be configured globally, while in 4.x it can be specified as an argument to `check_type`.

PR #4718, opened 2024-11-29
feat(pacmak): support jsii-rosetta 5.7

mrgrain:
Adds support for jsii-rosetta 5.7 to jsii-pacmak.

This also changes the peer dependency to simply state the minimal required version.
This will make it less urgent for us to actually release a new version of jsii-pacmak.

Also includes a new script to fully automate the uprade.
A future extension would be to run this script from a workflow.

---


.github/workflows/main.yml
mrgrain, line 1: removed whitespace because `yq` cannot deal with that and will remove it.

.github/workflows/yarn-upgrade.yml
mrgrain, line 1: `jsii` & `jsii-rosetta` were previously excluded because they were part of the monorepo. We just want to handle upgrades for these two separately.

packages/jsii-pacmak/package.json
mrgrain, line 1: re-ordered some package because automated tooling is sorting the packages

otaviomacedo (approved):

Issue #4793, opened 2025-03-15
jsii-pacmak fails on windows if there's a `prepack` script.


TimothyJones:
### Describe the bug

To work around this [npm bug](https://github.com/npm/cli/issues/3466) (previously discussed in the context of jsii [here](https://github.com/aws/jsii/issues/4132)), I have a `prepack` script, which tricks `npm pack` into bundling dependencies correctly.

Unfortunately, this fails on windows github runners (which in my case are invoking jsii via bash) with the following error:

```
  Error: ENOENT: no such file or directory, lstat 'C:\Users\RUNNER~1\AppData\Local\Temp\npm-packqlojDm\> @contract-case\case-definition-dsl@0.18.0 prepack
  > bash -c '.\script\link-bundled-deps.sh'
  
  contract-case-case-definition-dsl-0.18.0.tgz'
```

Given the quoting, I suspect this is the whole `lastline` variable [created here](https://github.com/aws/jsii/blob/main/packages/jsii-pacmak/lib/packaging.ts#L83). 

It looks like the npm is perhaps not printing windows line endings in its general reporting - potentially because it's being invoked via bash, I'm not sure.

I can confirm this is not a (pure) npm problem, as `npm pack` works fine directly.

I can also confirm this is related to having the prepack script, as removing it results in a successful build (but without the bundled dependencies).

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

Packages to be generated successfully on windows where there is a `prepack` script in use.

### Current Behavior

Error message as listed above

### Reproduction Steps

What I believe to be relevant:

* Have a jsii project that bundles with `jsii-pacmak`
* Have a `prepack` script in that project's `package.json`
* Run `jsii-pacmak` via an npm script in windows
* Possibly that npm script needs to be invoked via bash, I'm not sure.

I think it'll be faster to contribute a fix than it will be to make the repro, so I'm gonna start by trying to fix it.

If you're interested in a non-minimal reproduction, [this commit](https://github.com/case-contract-testing/contract-case/commit/1c0e3735997a9265569b80ef0023a4d0e1c99d32) illustrates the problem. 

### Possible Solution

I suspect that splitting the `lastline` again with `\n` will solve this problem.

### Additional Information/Context

PRs incoming

### SDK version used

jsii-pacmak v1.108.0

### Environment details (OS name and version, etc.)

node 20, Github Runner version 2.322.0 (Microsoft Windows Server 2022, 10.0.20348)

[...]

PR #4794, opened 2025-03-15
fix(pacmak): Improve error message where scripts have interfered with npm pack

TimothyJones:
It took me a while to spot the quoting in the error reported [here](https://github.com/aws/jsii/issues/4793).

This PR adds a more explicit error for the case where the generated tar appears not to exist, which would have helped find the issue quicker.

---


rix0rrr:
Superseded by https://github.com/aws/jsii/pull/4795

PR #4795, opened 2025-03-15
fix(pacmak): Find the tarball location on windows even if prepack is run

TimothyJones:
I believe this fixes [the issue where windows runners with prepack scripts fail](https://github.com/aws/jsii/issues/4793).

The fix is a little clumsy - it might be better to pull the line manipulation out to a function.

Raising for discussion, as testing this is a bit fiddly. I'll report back once tested.

---

### DIFF
--- a/packages/jsii-pacmak/lib/packaging.ts
+++ b/packages/jsii-pacmak/lib/packaging.ts
@@ -79,14 +79,20 @@ export class JsiiModule {
       // Take only the last line of npm pack which should contain the
       // tarball name. otherwise, there can be a lot of extra noise there
       // from scripts that emit to STDOUT.
-      const lines = out.trim().split(os.EOL);
+      const linesSplitByOs = out.trim().split(os.EOL);
+      const lastLineSplitByOs =
+        linesSplitByOs[linesSplitByOs.length - 1].trim();
+      // on Windows, sometimes `npm pack` might print output with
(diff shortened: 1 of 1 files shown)


packages/jsii-pacmak/lib/packaging.ts
```diff
@@ -79,14 +79,20 @@ export class JsiiModule {
...
+      // This is safe, because '\n' is never a valid character in a file name.
+      const lines = lastLineSplitByOs.split('\n');
       const lastLine = lines[lines.length - 1].trim();
 
       if (!lastLine.endsWith('.tgz') && !lastLine.endsWith('.tar.gz')) {
         throw new Error(
           `${packCommand} did not produce tarball from ${
             this.moduleDirectory
-          } into ${tmpdir} (output was ${JSON.stringify(lines)})`,
+          } into ${tmpdir} (output was ${JSON.stringify(linesSplitByOs)})`,
```
rix0rrr, line 95:
```suggestion
          } into ${tmpdir} (output was ${JSON.stringify(lines.map(l => l.trimEnd())})`,
```

TimothyJones:
@rix0rrr Thanks for the review! I've updated the PR per your comments.

TimothyJones:
@rix0rrr This appears to be stuck - does it need additional approvals somewhere?

Issue #4801, opened 2025-03-20
Java Runtime now incorrectly reports `Cannot read field "stderr" because "consoleOutput" is null`


TimothyJones:
### Describe the bug

In the java version of jsii 1.109.0, the runner now reports:

```
Unexpected error in background thread "software.amazon.jsii.JsiiRuntime.ErrorStreamSink": java.lang.NullPointerException: Cannot read field "stderr" because "consoleOutput" is null
```

on every run.

The way the jsii Java runner works is that there's a thread that reads standard error for json objects.  It reads each line, tries to parse them to json, and if it fails with an exception, it will print the line instead.

There's a small mistake in the runtime, where [the buffer is flushed on EOF, even if it is empty](https://github.com/aws/jsii/blob/main/packages/%40jsii/java-runtime/project/src/main/java/software/amazon/jsii/JsiiRuntime.java#L518). This would have failed the json parsing, and printed a newline to standard error (probably no one noticed this - I certainly didn't).

Unfortunately, Jackson-Databind 2.18.3 contains a [change](https://github.com/FasterXML/jackson-databind/issues/4932), where json nodes that aren't present will now return `null` instead of throwing an exception as they did previously. This changes the behaviour for empty string.

### Regression Issue

- [x] Select this option if this issue appears to be a regression.

### Expected Behavior

No errors to be printed on a successful run

### Current Behavior

The following error is printed to standard out

```
Unexpected error in background thread "software.amazon.jsii.JsiiRuntime.ErrorStreamSink": java.lang.NullPointerException: Cannot read field "stderr" because "consoleOutput" is null
```

### Reproduction Steps

Run any java code that calls jsii with v1.109.0

### Possible Solution

The fix is twofold:

1) Don't flush an empty buffer on EOF (ie, [this line](https://github.com/aws/jsii/blob/main/packages/%40jsii/java-runtime/project/src/main/java/software/amazon/jsii/JsiiRuntime.java#L519))
2) Handle a null return from `objectMapper.treeToValue` [here](https://github.com/aws/jsii/blob/main/packages/%40jsii/java-runtime/project/src/main/java/software/amazon/jsii/JsiiRuntime.java#L528)

### Additional Information/Context

Happy to produce a PR if you like

### SDK version used

1.109.0

### Environment details (OS name and version, etc.)

openjdk 23.0.1, node v22.9.0, MacOS 14.4.1

TimothyJones:
Pull request #4802 raised to address this

[...]

PR #4802, opened 2025-03-20
fix(java-runtime): `ErrorStreamSink` no longer fails with a `NullPointerException`

TimothyJones:
I believe this fixes https://github.com/aws/jsii/issues/4801

I'm unable to run the full test suite locally due to unrelated python and dotnet errors (I suspect I don't have the right versions of something it needs), and unfortunately it doesn't get up to the java-runtime tests.

---


TimothyJones:
Just realised I neglected to do conventional commits on the actual commits - please let me know if this is an issue, and I'll update and re-raise.

TimothyJones:
Fixed compile error, I think (it's hard to test locally)

iliapolo:
Superseded by https://github.com/aws/jsii/pull/4803

PR #5110, opened 2026-05-12
feat: replace eager submodule imports with lazy loading

dgandhi62:
## Problem

`import aws_cdk` is slow in Python CDK apps because the generated `__init__.py` eagerly imports all ~300 child submodules via `from . import <submodule>`. This fixed cost hits every `cdk synth`, `cdk deploy`, Lambda cold start, and IDE analysis — regardless of how many services the app actually uses.

## Solution

Replace the eager import block with [PEP 562](https://peps.python.org/pep-0562/) lazy loading using module-level `__getattr__` and `__dir__`.

### Before (generated code)

```python
# Loading modules to ensure their types are registered with the jsii runtime library
from . import aws_s3
from . import aws_ec2
from . import aws_lambda
# ... 300+ more
```

### After (generated code)

```python
import importlib as _importlib

_SUBMODULES = {
    "aws_ec2",
    "aws_lambda",
    "aws_s3",
    # ... sorted
}

def __getattr__(name: str) -> object:
    if name in _SUBMODULES:
        mod = _importlib.import_module(f".{name}", __name__)
        globals()[name] = mod
        return mod
    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")

def __dir__() -> "list[str]":
    return [*__all__, *_SUBMODULES]
```

Submodules are loaded on first access and cached in `globals()`, so subsequent accesses are a plain dict lookup with no overhead.

### Static type checker compatibility

Pyright can't see that `__getattr__` will resolve submodule names at runtime, so it flags them as undefined in `__all__`. A `typing.TYPE_CHECKING` guard with explicit re-exports satisfies static analyzers without affecting runtime:

```python
if typing.TYPE_CHECKING:
    from . import aws_s3 as aws_s3
    from . import aws_lambda as aws_lambda
```

### `publication.publish()` integration

`publication.publish()` replaces the module in `sys.modules` but doesn't copy `__getattr__`/`__dir__`. We install them on the public module explicitly via `setattr()` after defining them.

### On-demand type resolution for lazily-loaded submodules

With eager imports, all types self-register at import time. With lazy loading, the kernel may return a type from a submodule that hasn't been imported yet. The runtime now resolves this by:

1. Emitting a `_SUBMODULE_FQN_MAP` dict in the generated `_jsii/__init__.py` that maps each submodule's jsii FQN to its correct Python module path (computed at code-gen time from the assembly metadata, handling custom `targets.python.module` overrides)
2. At assembly load time, registering this mapping with the runtime
3. When an unknown type is encountered, walking the FQN to find the containing submodule in the map, importing it (which triggers type registration), then retrying

This handles custom-named submodules (where FQN segments don't match Python module names) and nested types (where FQN segments after the submodule are class names, not modules).

## Testing
- Behavioral tests in the Python compliance suite verify lazy loading works end-to-end (submodule access, `dir()`, custom-named submodules, types from unimported submodules)
- Existing test verifies the runtime can resolve types from submodules that haven't been loaded by user code.https://github.com/aws/jsii/blob/b6b0fb0ac9be12d06b7c5f941ec62fa11c564073/packages/%40jsii/python-runtime/tests/test_compliance.py#L1331
---


packages/@jsii/python-runtime/src/jsii/_reference_map.py
```diff
@@ -27,6 +33,48 @@ def register_interface(iface: Any):
...
+def _try_import_type_module(class_fqn: str) -> bool:
+    """Attempt to import the Python module containing a jsii type by FQN.
+
+    With PEP 562 lazy loading, submodules are not imported until first attribute
+    access. If the jsii runtime needs to deserialize a type from a submodule
+    that hasn't been imported yet (e.g., a callback returns an object whose type
+    lives in an unloaded submodule), this function triggers the import so that
+    the type self-registers with the runtime.
+
+    The FQN format is: ``assembly_name.submodule.path.TypeName``
```
rix0rrr: Unfortunately I don't think jsii FQNs map cleanly onto Python import paths like this.

Pretty sure that the author of a jsii module can map whatever submodule they want to whatever Python module path. Plus, there types-in-types, and the last parts of the FQN might not be modules.

rix0rrr: Here's an interesting question: CAN we deterministically find a Python type given a jsii FQN? (All the information should be in the assembly)

Because if we can, we can fully get rid of the registering-types-by-fqns-on-startup business that we have going on here!

packages/jsii-pacmak/test/targets/python/lazy-imports.test.ts
rix0rrr, line 1: These tests assert that the generated Python code looks just-so. Those are very brittle, as soon as we change anything about the implementation these tests will break.

Instead, I'd rather test behavior: a Python library generated with this lazy lookup has the following behavior: XYZ (and in fact, probably just "it works the same as it did before" might be good enough 😉 )

dgandhi62: This makes sense! I removed the old tests and added new behavioral ones. Some of our existing snapshots already do a bunch of the heavy lifting.

packages/@jsii/python-runtime/src/jsii/_reference_map.py
dgandhi62: You're right, the old approach was incorrect. The code generator now emits a `_SUBMODULE_FQN_MAP` dict in `__init__.py` that maps each submodule FQN to its actual Python module path (computed at build time using the same toPackageName logic that generates the directory structure). The runtime walks FQN prefixes against this map to find the containing module. I think this should handle custom naming and types-in-types both.

packages/@jsii/python-runtime/src/jsii/_reference_map.py
```diff
@@ -27,6 +38,60 @@ def register_interface(iface: Any):
...
+    # Walk from most-specific to least-specific prefix of the FQN to find
+    # the containing submodule. We skip the last component (always a type name)
+    # and the first component (the assembly name).
+    if _submodule_fqn_map:
+        for depth in range(len(parts) - 1, 1, -1):
+            candidate_fqn = ".".join(parts[:depth])
+            python_module = _submodule_fqn_map.get(candidate_fqn)
+            if python_module is not None:
+                try:
+                    importlib.import_module(python_module)
```
rix0rrr, line 79: Shouldn't we insert into `_assembly_to_module` here then?

rix0rrr (approved):

PR #5111, opened 2026-05-13
feat: add workflow to auto-update snapshots via label or manual dispatch

iankhou:
Use an `update-snapshots` label to update snapshots using GitHub Actions. Useful for dependency upgrade PRs that are auto-approved but require snapshot updates. Requires Triage role or above, to trigger the workflow..

Motivated by a recent auto-approve PR that was not able to be merged, due to snapshots not being updated. I don't want to pull down the code when I know the snapshots just need to be updated. https://github.com/aws/jsii/pull/5096.

---

### DIFF
--- a/.github/workflows/update-snapshots.yml
+++ b/.github/workflows/update-snapshots.yml
@@ -1,0 +1,85 @@
+name: Update Snapshots
+
+on:
+  workflow_dispatch:
+    inputs:
+      branch:
+        description: "Branch to update snapshots on"
+        required: true
+  pull_request:
+    types: [labeled]
+
+env:
+  DOTNET_NOLOGO: true
+  NODE_OPTIONS: --max-old-space-size=4096
+
+jobs:
+  update-snapshots:
+    name: Update Snapshots
+    if: |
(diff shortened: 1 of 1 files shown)


.github/workflows/update-snapshots.yml
```diff
@@ -0,0 +1,85 @@
+name: Update Snapshots
+
+on:
+  workflow_dispatch:
+    inputs:
+      branch:
+        description: "Branch to update snapshots on"
+        required: true
```
rix0rrr, line 8: This is unnecessary. Every workflow always runs in the context of a ref, which is typically a branch. A `workflow_dispatch` already comes with a branch selector when you run the workflow.

rix0rrr (changes requested):
A label seems like the wrong interface for this. A label indicates a **state**, and this reads like a **trigger**.

Alternative implementations:

- The workflow clears the label again. That way, the label *does* indicate a state, and that state is "snapshots are pending an update".
- We trigger based on a command, `/update-snapshots` or whatever.

Second, why not do this autmatically?

PR #5118, opened 2026-05-15
fix(jsii-diff): `npm:` argument type can be shell-injected

rix0rrr:
`jsii-diff` supports downloading packages to compare from NPM:

```
$ jsii-diff aws-cdk-lib npm:aws-cdk-lib
```

The string after `npm:` gets used as part of a subcommand that executes the NPM package manager. This used to be done using a shell, which made it vulnerable to shell injection.

We can't avoid going through the shell though, because we run NPM and on Windows NPM is `npm.cmd`; only the shell can execute batch files on Windows, they are not directly executable. 

Instead, we limit the set of characters we will allow there, to avoid the shell injection that way.

---

### DIFF
--- a/packages/jsii-diff/lib/util.ts
+++ b/packages/jsii-diff/lib/util.ts
@@ -48,6 +48,7 @@ export async function downloadNpmPackage<T>(
       // to not bork when it can find the dependencies.
       await execFile('npm', ['install', '--silent', '--prefix', '.', pkg]);
     } catch (e: any) {
+      console.log(e);
       // If this fails, might be because the package doesn't exist
       if (!isSubprocesFailedError(e)) {
         throw e;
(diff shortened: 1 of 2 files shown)


rix0rrr:
TODO: need to confirm this still works on Windows

packages/jsii-diff/lib/util.ts
```diff
@@ -96,6 +105,19 @@ function trimVersionString(pkg: string) {
...
 
+/**
+ * Validate a package name against a list of allowed characters
+ *
+ * If we are too strict here, that's not a biggy: script writers are always
+ * able to download their exotically-named NPM package themselves before running
+ * jsii-diff on it.
+ */
+function validateValidPackageSpecifier(pkg: string) {
+  if (pkg.match(/[^a-z0-9@\/._-]/i)) {
```
mrgrain: allow `:` so we can support `file:` and `git:`? otherwise looks good

mrgrain (dismissed):

packages/jsii-diff/lib/util.ts
rix0rrr: Done!

otaviomacedo (approved):

PR #5137, opened 2026-06-01
fix: centralize and lazy-load typeguard to improve Python import times

dgandhi62:
## Problem

Every generated Python module inlined ~15 lines of typeguard boilerplate at module level:

```python
import typeguard
from importlib.metadata import version as _metadata_package_version
TYPEGUARD_MAJOR_VERSION = int(_metadata_package_version('typeguard').split('.')[0])

def check_type(argname: str, value: object, expected_type: typing.Any) -> typing.Any:
    if TYPEGUARD_MAJOR_VERSION <= 2:
        ...
    else:
        ...
```

For large libraries like `aws-cdk-lib` with hundreds of submodules, this meant:

- Hundreds of redundant `importlib.metadata.version()` calls, each scanning package metadata on the filesystem
- Hundreds of identical `check_type` function definitions parsed and compiled by CPython
- `typeguard` imported eagerly even when running with `python -O` (where type checks are disabled)

## Solution

Introduced `jsii._type_checking` in the Python runtime — a single module that:

1. **Defers** importing `typeguard` and probing its version until `check_type` is first called
2. **Hot-patches** itself on first invocation so subsequent calls have zero dispatch overhead
3. **Skips entirely** when running with `python -O` — since `check_type` lives inside `if __debug__:` blocks in generated code, typeguard is never imported at all in optimized mode

Generated code now emits one line instead of fifteen:

```python
```

## Notes

- `typeguard` was removed from generated packages' `install_requires` since it is already a transitive dependency via the `jsii` runtime package. This should not affect end users.
- The `jsii` runtime package continues to pin `typeguard==2.13.3`.

### DIFF
--- a/packages/@jsii/python-runtime/src/jsii/_type_checking.py
+++ b/packages/@jsii/python-runtime/src/jsii/_type_checking.py
@@ -13,9 +13,7 @@ lazily on first call to :func:`check_type`. This means:
 import typing
 
 
-def check_type(
-    argname: str, value: object, expected_type: typing.Any
-) -> typing.Any:
+def check_type(argname: str, value: object, expected_type: typing.Any) -> typing.Any:
     """Validate *value* against *expected_type*, adapting to the installed typeguard version.
 
     Typeguard and version detection are loaded on first invocation and cached
@@ -26,6 +24,7 @@ def check_type(
     resolved = _resolve()
     # Hot-patch the module-level name so future calls go directly to the resolved function.
     import jsii._type_checking as _self
+
     _self.check_type = resolved  # type: ignore[attr-defined]
     return resolved(argname=argname, value=value, expected_type=expected_type)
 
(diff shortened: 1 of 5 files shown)


packages/jsii-pacmak/lib/targets/python.ts
```diff
@@ -2352,11 +2315,6 @@ class Package {
     );
     code.closeFile('README.md');
 
-    // 3.x and newer perform an additional runtime check on interfaces that our interfaces fail.
-    // Stick to this old version. <https://github.com/aws/constructs/issues/2825>
-    // Defined as a constant to hopefully prevent Dependabot from automatically updating this.
```
dgandhi62, line 2357: I see this already exists in `setup.py` here - https://github.com/aws/jsii/blob/26b332929cbb97b919d9cb229b44c44d39709e44/packages/%40jsii/python-runtime/setup.py#L15

and therefore is not needed again

packages/@jsii/python-runtime/src/jsii/_type_checking.py
```diff
@@ -0,0 +1,76 @@
...
+        def _check_type(
+            argname: str, value: object, expected_type: typing.Any
+        ) -> typing.Any:
+            if isinstance(value, InterfaceDynamicProxy):
+                pass
+            else:
+                typeguard.check_type(
+                    value=value,
+                    expected_type=expected_type,
+                    collection_check_strategy=typeguard.CollectionCheckStrategy.ALL_ITEMS,  # type:ignore[attr-defined]
```
dgandhi62: `typeguard.CollectionCheckStrategy` and `typeguard.config` don't exist in typeguard 2.x's type stubs

packages/@jsii/python-runtime/src/jsii/_type_checking.py
```diff
@@ -0,0 +1,66 @@
...
+    """
+    resolved = _resolve()
+    # Hot-patch the module-level name so future calls go directly to the resolved function.
+    import jsii._type_checking as _self
+
+    _self.check_type = resolved  # type: ignore[attr-defined]
+    return resolved(argname=argname, value=value, expected_type=expected_type)
+
+
+def _resolve() -> typing.Callable[..., typing.Any]:
```
rix0rrr: [naming] Resolve what?

dgandhi62: Addressed

rix0rrr (approved):

PR #5150, opened 2026-06-04
fix: lazy cross module base class imports for python speedups

dgandhi62:
This pr defers all Python classes & interfaces into factory functions and replaces cross-module imports with lazy proxy objects. This enables us to actually make submodule contents lazy and reduces the time taken to import python submodules by ~55%

The problem can be defined as follows -  Before, if a user wrote `import aws_cdk.aws_s3`, Python had to 
- Execute its cross-module imports at the top of the init file (for example, `aws_iam`)
- Build its classes, interfaces, and objects
- Register all types with the jsii runtime

This lead to a cascading init which would scale as the size of the library grows. To solve this, we introduce lazy factories. 

### Part 1 - making base classes lazy

All classses, interfaces, and structs are wrapped in @_memoized factory functions, for example:
```
# Before (eager)
class Bucket(metaclass=jsii.JSIIMeta, jsii_type="aws-cdk-lib.aws_s3.Bucket"):
    ...

# After (lazy)
@_memoized
def _lazy_build_Bucket() -> type:
    class Bucket(metaclass=jsii.JSIIMeta, jsii_type="aws-cdk-lib.aws_s3.Bucket"):
        ...
    return Bucket
```

**getattr** on the module triggers the factory on first access, which then subsequently caches the module.

Furthermore, typecheckers like mypy and pyright see the full class shapes via stubs in an `if typing.TYPE_CHECKING: block`. These stubs have empty bodies and are never executed at runtime. The reason for this is that type checkers evaluate code statically and need to see class definitions with their full signatures. Since the factory functions hide the class inside a closure, type checkers don't like this and therefore we need stubs. 

### Part 2 - making annotation-only imports lazy

Cross-module imports exist mainly to support type annotations. Since all type annotations are rendered as string literals, they never actually trigger module loading at definition. Therefore, these imports are annotation-only and are only needed at runtime when code actively uses the imported type 

We exploit this by placing the real import under `if typing.TYPE_CHECKING: `(for static analysis) and using a `_LazyImport proxy` in the else branch that defers importlib.import_module() until first attribute access:

```
# Before (eager — triggers cascading module loads at import time)
import scope.jsii_calc_base_of_base as _mod

# After (lazy — module only loads when _mod.SomeType is actually accessed)
if typing.TYPE_CHECKING:
    import scope.jsii_calc_base_of_base as _mod
else:
    _mod = _LazyImport("scope.jsii_calc_base_of_base")
```

### Note
With lazy class factories, classes don't register themselves with the jsii runtime until their factory runs. This is a problem when the kernel returns object references with type FQNs which aren't registered. As an example, if the user never directly accessed `aws_cdk.aws_s3.Bucket` in their Python code (so the factory never ran), the runtime does not know which class to deserialize that reference into.

The `_reference_map.py` in the Python runtime now handles lazy types. For instance, when the kernel returns a type reference that isn't yet registered, it triggers `getattr` on the appropriate module to materialize the factory, which registers the type as a side effect of class construction via JSIIMeta.

## EDGE CASES
This section is basically weird things I encountered which my solution also accounts for in the code. 

### **Same-module base class references**
When class B(A) and A are both in the same module, A doesn't exist yet when B's factory runs (both are deferred). The generated code replaces bare base class names with factory calls:
```
class B(_lazy_build_A(), metaclass=jsii.JSIIMeta, ...):
    ...
```

### **`typing.get_type_hints()` needs all deferred classes resolved**
jsii-pacmak generates runtime type-checking code that validates method arguments at call time. This includes the typechecking call and the typechecking stub which you can see at the end of a module. `typing.get_type_hints()` looks at the stubs annotations dict and resolves each string in a namespace. Previously, stubs had unquoted annotations, and typing.get_type_hints() was called with no explicit namespace (which I think it defaulted to the stub's __globals__). This was fine becuse all classes existed in globals() at module scope. Now classes are deferred. They don't exist in globals() until their factory runs.

To solve this, we introduce `_get_typechecking_ns()`, which builds a dict that's a copy of globals() plus every deferred class (by calling all factories). We pass this dict as `globalns` so `get_type_hints()` can find any name.

Before
```
if __debug__:
    type_hints = typing.get_type_hints(_typecheckingstub__abc123)
    check_type(...)
```

After
```
if __debug__:
    type_hints = typing.get_type_hints(_typecheckingstub__abc123, globalns=_get_typechecking_ns(), localns=_get_typechecking_localns())
    check_type(...)
```

### **qualname issues**
Python automatically sets __qualname__ based on where a class is defined lexically. A class inside a function gets function_name.<locals>.ClassName. Since we define classes inside _lazy_build_* functions, they'd show up as _lazy_build_Bucket.<locals>.Bucket in error messages (an example), tracebacks, and repr() instead of just Bucket.

The fix is a single line after the class is created inside the factory:
`Bucket.__qualname__ = "Bucket"`

### **Intersection type helper classes**
When a parameter is typed as A & B in TypeScript, pacmak generates a synthetic Protocol class that extends both interfaces (for instance, think `class _ISomething_IFriendly(ISomething, IFriendly, Protocol`). These helper classes were previously emitted at module scope where their base classes already existed in globals(). With lazy factories, the base classes are now factories, which causes a NameError. 

The fix basically moves the helper classes into the `else` block of `if typing.TYPE_CHECKING` and replaces same-module base class names with their factory call.

```
# Before (broken — ISomething not in globals at import time)
class _ISomething_IFriendly(ISomething, _other_pkg.IFriendly, typing_extensions.Protocol):
    pass

# After (works — factory call materializes ISomething)
class _ISomething_IFriendly(_lazy_build_ISomething(), _other_pkg.IFriendly, typing_extensions.Protocol):
    pass
```

---


rix0rrr:
Nice! Couple of questions:

## Backwards compat

Are there any consequences for backwards compatibility? We are changing both the generated code and the runtime here. We can assume `new gencode = new runtime`, but we cannot assume the reverse: the new runtime code be interacting with old codegen'ned code. Spare a moment to think about what the consequences of that are. Anything that might break?

## Qualname

Don't forget that we have classes inside classes as well, for example in the AWS CDK L1s.

So for:

```py
def lazy_build_CfnBucket:
  class CfnBucket:
    class LifecycleRuleProperty:
      pass

   pass
```

The `__qualname__` of that inner type should still be `CfnBucket.LifecycleRuleProperty`, *not* `LifecycleRuleProperty`. Do we have that correct?

## get_type_hints

You wrote:

> To solve this, we introduce _get_typechecking_ns(), which builds a dict that's a copy of globals() plus every deferred class (by calling all factories). We pass this dict as globalns so get_type_hints() can find any name.

Does that mean the first time we call a function and we do run-time type checking, we resolve all lazy class factories after all?

packages/@jsii/python-runtime/src/jsii/_reference_map.py
```diff
@@ -66,28 +72,61 @@ def _try_import_type_module(class_fqn: str) -> bool:
...
+    if python_module is None:
+        try:
+            importlib.import_module(root_module)
+            python_module = root_module
+            type_parts = parts[1:]
+        except (ImportError, ModuleNotFoundError):
+            return False
+
+    if python_module is None:
+        return False
```
rix0rrr: Doesn't this behavior fall out of the previous `for` loop automatically?

Seems to me there are 2 steps to a type lookup:

- Find the most specific "module-like" object to contain the type (can be the root module if nothing more specific)
- Then traverse that module-like object using `.`-separated lookups of the remaining type parts.

At the end, we should actually have that object in a variable directly. We don't need to rely on the `@jsii` decorator having stuffed the type into a global map and looking for it there. The traversal should will have given us the type directly.

packages/@jsii/python-runtime/src/jsii/_reference_map.py
```diff
@@ -210,7 +254,16 @@ def resolve_id(self, id: str) -> Any:
...
     def build_interface_proxies_for_ref(self, ref: ObjRef) -> List[Any]:
-        ifaces = [_interfaces[fqn] for fqn in ref.interfaces or []]
+        ifaces = []
+        for fqn in ref.interfaces or []:
+            if fqn not in _interfaces:
+                # Interface may be inside a lazy factory that hasn't run yet.
+                # Trigger the factory via _try_import_type_module + getattr.
+                _try_import_type_module(fqn)
+            if fqn in _interfaces:
+                ifaces.append(_interfaces[fqn])
```
rix0rrr: This code is mixing concerns in places where it shouldn't. You want to have a function that says "give me that interface", and you don't care if it's cached or not -- you will just assume it's doing the smart thing.

We used to have a direct dict access here because it was the trivial thing to do, but now that it's slightly less trivial it should just become a function call.

Original code:

```py
        ifaces = [_interfaces[fqn] for fqn in ref.interfaces or []]
```

New code:

```py
        ifaces = [_obtain_interface(fqn) for fqn in ref.interfaces or []]

def _obtain_interface(fqn):
  ret = _interfaces.get(fqn)
  if ret: return ret
  
  # Do lookup and insert into cache
  ...
```

packages/@jsii/python-runtime/src/jsii/_reference_map.py
rix0rrr: This function might end up a lot nicer (and easier to understand) if you split those phases up into separate functions.

packages/jsii-pacmak/test/generated-code/__snapshots__/examples.test.js.snap
```diff
@@ -1490,6 +1598,42 @@ __all__ = [
...
+if __debug__:
+    _typechecking_ns: "dict[str, object] | None" = None
+    _typechecking_localns: "dict[str, object] | None" = None
+
+    def _get_typechecking_ns() -> "dict[str, object]":
+        global _typechecking_ns
+        if _typechecking_ns is None:
+            ns = dict(globals())
+            for name, factory in _LAZY_CLASSES.items():
+                ns[name] = factory()
```
rix0rrr: This is de-lazyifying everything, isn't it? 

Perhaps we're better off using another object like `_LazyImport` where the `__getattr__()` will trigger the factory, and using that a namespace. And we can also use that for the base classes etc.

```py
THIS_MODULE = _LazyAccess()

def _build_Consumer(): 
  class Consumer(THIS_MODULE.BaseClass):
    ...

THIS_MODULE.register('Consumer', _build_Consumer)


def __getattr__(key):
  return THIS_MODULE[key]
```

Or even:

```py
THIS_MODULE = _LazyAccess()

@THIS_MODULE.register_type('Consumer')
def _build_Consumer(): 
  class Consumer(THIS_MODULE.BaseClass):
    ...
```

packages/jsii-pacmak/test/generated-code/__snapshots__/examples.test.js.snap
```diff
@@ -1490,6 +1598,42 @@ __all__ = [
...
+            _typechecking_ns = ns
+        return _typechecking_ns
+
+    def _get_typechecking_localns() -> "dict[str, object]":
+        global _typechecking_localns
+        if _typechecking_localns is None:
+            ns = dict(globals())
+            for name, factory in _LAZY_CLASSES.items():
+                ns[name] = factory()
```
rix0rrr: This function looks the same as the other one?

dgandhi62: I was kinda confused with this as well. But that is intentional, one for globalns and one for localns. It prevents Python's `ForwardRef` from reusing a cached resolution from a sibling module that has a type with the same name. Specifically we have a test (`test_homonymous_forward_references`) which fails for this

packages/@jsii/python-runtime/src/jsii/_reference_map.py
dgandhi62: Fixed

packages/@jsii/python-runtime/src/jsii/_reference_map.py
dgandhi62: That... makes sense. Fixed. I split it up into three phases
- find the Python module, import it, return `(module_name, type_parts)` or `None`
- traverse the module via `getattr()` to trigger lazy factories, return bool
- call Phase 1, check registry, call Phase 2 if needed

packages/jsii-pacmak/test/generated-code/__snapshots__/examples.test.js.snap
dgandhi62: Here is your friendly neighbourhood kiro's generated example for this.

```
Module foo/                          Module bar/
─────────────────────────           ─────────────────────────
class Homonymous:                   class Homonymous:
    string_property: str                numeric_property: int

def stub(x: "Homonymous"): ...     def stub(x: "Homonymous"): ...
```

1. `foo` calls `typing.get_type_hints(foo.stub, globalns=foo_ns)` — no `localns` passed
2. Python sets `localns = globalns` (same object)
3. ForwardRef resolves `"Homonymous"` → `foo.Homonymous`, caches it, sets `__forward_evaluated__ = True`
4. `bar` calls `typing.get_type_hints(bar.stub, globalns=bar_ns)` — no `localns` passed
5. Python sets `localns = globalns` (same object)
6. ForwardRef checks: `not __forward_evaluated__` → False. `localns is not globalns` → False (same object).
7. **Returns cached `foo.Homonymous`** → wrong type → TypeError

It seems that Python reuses typing.Union objects in memory, so two stubs in different modules with the identical annotation typing.Union["Homonymous", Dict[str, Any]] share the same ForwardRef object in memory

dgandhi62: Basically now, `localns is not globalns` is always `True` → ForwardRef always re-evaluates → each module gets its own correct `Homonymous`.

packages/jsii-pacmak/test/generated-code/__snapshots__/examples.test.js.snap
dgandhi62: Yes it de-lazifies the module, but only on the first type check in debug mode (python -O skips it entirely). Also it's per-module, not global.

I kept running into some Python test errors earlier. Let me double check and tell you more accurately

dgandhi62:
To your first comment, 

I fixed the qualname point you mentioned and addressed 3 in a different comment. For backward compatibility, there should not be issue (?). The new code adds exports but the old code never references them. The changes in the runtime add code paths when eager registering doesn't happen, so old code should still work. Still a work in progress, its a big change so I'm trying to think through all the ramifications.

packages/jsii-pacmak/test/generated-code/__snapshots__/examples.test.js.snap
dgandhi62: Investigating further

dgandhi62:
For the question of backward compat, I think this saves us @rix0rrr - https://github.com/aws/jsii/blob/c326200d1db1592d7980d9db2faa199831d46174/packages/jsii-pacmak/lib/targets/python.ts#L2338

There is a minimum runtime jsii dependency declared. Which I would understand to mean that newly generated code will require the new runtime AND the old runtime wont work.
The reverse (new runtime with old code) should not be an issue because the changes in runtime are just additions. With the old code, the new paths should not be triggered.

packages/jsii-pacmak/test/generated-code/__snapshots__/examples.test.js.snap
rix0rrr: > Yes it de-lazifies the module, but only on the first type check in debug mode

What do you think the goal of this work is though?

Is it

* a) Make the Python bindings faster across the board; or
* b) Move slow operations away from `import` and move them into `new Stack()`, so that we can say `import` is now faster?

packages/jsii-pacmak/test/generated-code/__snapshots__/examples.test.js.snap
rix0rrr: Wait, this seems weird though. This should be a standard problem with Python if this were true, regardless of what we were doing. So this smells to me like it cannot be true, and our understanding is not complete.

Specifically, this trips me up:

> ForwardRef resolves "Homonymous" → foo.Homonymous, **caches it**, sets __forward_evaluated__ = True

Who caches what where? That caching would obviously incorrect, so it cannot be the default behavior of the Python stdlib. Is it *us* caching it?

Why I'm saying this: I'm sure the default behavior of `get_type_hints()`, when given no namespaces at all, would be to default to using `globals()` and `locals()`. Which, if they cached the lookup of `Homonymous` globally, would lead to the exact same broken behavior.

I cannot stress this enough: *It cannot be the case that nobody thought about this and the Python standard library is broken*.


----

P.S.:

> homonymous caching issue

Seems like a $5 word when a 50 cents word would do.

rix0rrr: The naive case seems to work as I expect it to: https://gist.github.com/rix0rrr/0e713c5fd1d7f20122a40313b0f4f85a

rix0rrr: ~~Oh this is fun. The code does do what I expect it to do when running under a debugger, but not when running on a naked shell.~~

The difference is Python version 3.12 (works) vs 3.14 (doesn't)

rix0rrr: The following seems to work in most Python versions I tested (3.9, 3.10, 3.12, 3.14):

https://gist.github.com/rix0rrr/f298b44ffccdd059b2d6ebaf774d4e11

Confirming it doesn't evaluate unreferenced symbols.

I agree with you that it takes some fiddling to pass in values of `globalns` and `localns` that work for all Python versions, but I still see no evidence of caching or of confusion between the 2 classes with the same name in different files.

dgandhi62: I tried fiddling a bit with your example -I feel like it might not be a full representation because it uses bare string annotations (`def stub(x: "Homonymous")`), but the actual generated code uses `typing.Union["Homonymous", typing.Dict[str, typing.Any]]`

That difference I think matters because:

When its a bare string, `get_type_hints()` wraps it in a new ForwardRef each call. Fresh object every time → no shared state → no caching issue. For a Union, Python interns `typing.Union objects`. Two modules with identical `typing.Union[ForwardRef("Homonymous"), Dict[str, Any]]` get the same object, including the same ForwardRef inside it. Afyer that shared ForwardRef is evaluated, calls after that return the cached value.

I tried this out on 3.12. If you change your test's stubs from `def stub(x: "Homonymous")` to `def stub(x: typing.Union["Homonymous", typing.Dict[str, typing.Any]])` while keeping `globalns=THIS_MODULE, localns=THIS_MODULE`, it will cross-contraminate

dgandhi62: tldr: two modules with identical Union expressions seem to share the same ForwardRef object. The caching check in` ForwardRef._evaluate()` only re-evaluates when localns is not globalns. If you pass the same object for both, the cache seems to return stale results from whichever module resolved first.

dgandhi62: I will deep dive more and see if I can find a different soln - but I also feel like we aren't losing anything with this. The cost of this soln is just 1 extra dict and it doesn't break anything either.

I think you're right that its unlikely that the python library has a bug, but maybe this is intentional? (Alternatively it would be super cool if this actually was a bug and we found it)

packages/jsii-pacmak/test/generated-code/__snapshots__/examples.test.js.snap
dgandhi62: Ofcourse the first one. I will try to figure out a solution :). Currently what we have is that production will still be lazy, it is only debug mode where things are getting a little wonky

dgandhi62: I've updated the implementation to make type checking fully lazy for Python 3.12-3.13. I'm still finding there to be a constraint for 3.14.

On 3.12-3.13, typing.get_type_hints() takes whatever globalns/localns we give it and passes them straight to eval(). Basically we are able to control the lookup and pass in the custom dicts.  We pass a dict subclass with `__missing__` that only materializes a class when its name is actually looked up.

On 3.14,` get_type_hints()` seems to ignore whatever we are passing it. I think it knows which module the function lives in, so it looks at that module's dict directly instead of using whatever dict we give it. `ForwardRef` bypasses the custom namespaces, which basically leads to our custom dict getting thrown in the trash. Therefore right now, the solution is trying to put all the classes in the module's dict before type checking runs. I've isolated this to 3.14 only


Sources: 

https://github.com/python/cpython/blob/v3.14.0/Lib/typing.py (_eval_type sets globalns = None when prefer_fwd_module=True), 

https://github.com/python/cpython/blob/v3.14.0/Lib/annotationlib.py (ForwardRef.evaluate() falls back to function.__globals__).

packages/jsii-pacmak/test/generated-code/__snapshots__/examples.test.js.snap
rix0rrr: > I tried this out on 3.12. If you change your test's stubs from `def stub(x: "Homonymous")` to `def stub(x: typing.Union["Homonymous", typing.Dict[str, typing.Any]])` while keeping `globalns=THIS_MODULE, localns=THIS_MODULE`, it will cross-contraminate

Ahh, good call. I had missed that detail.

What if we do the following instead:

```py
def stub(x: 'typing.Union["Homonymous", typing.Dict[str, typing.Any]]'):
  ...
```

i.e., we make the whole expression a string again?

> (Alternatively it would be super cool if this actually was a bug and we found it)

That should be easy enough to verify. Let's reduce this as far as possible while still exhibiting the bug, then post it to their issue tracker.

rix0rrr: By the way, I wonder if we can save ourselves a lot of codegen pain if we used [metaclasses](https://jfreeman.dev/blog/2020/12/07/python-metaclasses/) instead.

rix0rrr: I tried it out and the answer is "no". We can make a metaclass that lazily loads the base classes (yay!), but there doesn't seem to be a way to tell type checkers that this is happening, so we would be left with an `if TYPE_CHECKING` block anyway, which kind of defeats the purpose.

However, playing around with this did give me a new insight: we can create a `.pyi` file next to the `.py` file, which will be preferentially read by a type checker if present. That saves us the `if TYPE_CHECKING` block, and saves us bloating every file by 80% (or whatever the number is), and doesn't blow up the parse times of already-large files unnecessarily.

It doesn't save on total number of bytes uploaded to PyPI, but it saves on the number of bytes loaded during execution.

So we can generate this:

```py
if typing.TYPECHECKING:

  # Whatever is here goes into my_module.pyi

else:

  # Whatever is here goes into my_module.py
```

rix0rrr: Here is what I ultimately came up with, it avoids the need for `.pyi` files with a bunch of jiggery pokery that I'm not too happy with: https://gist.github.com/rix0rrr/1bd084f758f71e2666fe2485b0b4d0e5

dgandhi62: For the idea about making everything a string, this is what finally worked across all 3 versions - quoting the entire type expression as one string with the inner class name left bare:

`x: "typing.Union[Homonymous, typing.Dict[builtins.str, typing.Any]]"`

dgandhi62: Can we do this in the next pr? I can try and one-shot this once everything else is verified. I suspect this will also have edge cases / iterations needed. I don't see this as an immediate necessity because we don't have a size issue. We have a per-file limit of 100mb and pacmak's generated wheel size is <60mb

packages/jsii-pacmak/test/generated-code/__snapshots__/examples.test.js.snap
dgandhi62: Fixed

packages/@jsii/python-runtime/src/jsii/_utils.py
```diff
@@ -24,3 +25,146 @@ def wrapped(self):
...
+
+    def __repr__(self) -> str:
+        if self._module is not None:
+            return repr(self._module)
+        if self._package:
+            return f"_LazyImport({self._module_name!r}, {self._package!r})"
+        return f"_LazyImport({self._module_name!r})"
+
+
+class _LazyBuiltins(dict):
```
rix0rrr: Please be extremely aware that anything we put into this file we can **never** take out again.

Hypothetical: let's say we *add* `_LazyBuiltins` in `jsii N`, and we don't need it anymore in `jsii N+1`. Can we remove it from the `jsii-runtime` Python package?

- Let's `mylib@5` is generated using `jsii-pacmak@N`, which requires `_LazyBuiltins` from `jsii-runtime@N`
- Let's say `yourlib@6` is generated using `jsii-pacmak@(N+1)`, which depends on `jsii-runtime@(N+1)`.
- The Python ecosystem can only have one version of every package at the same time.

Obviously, the package version of `jsii-runtime` has to be `N+1` (`yourlib@6` requires it), *and* `jsii-runtime@(N+1)` has to have `_LazyBuiltins` (`mylib@5` requires it).

So, everything that is needed at least once, stays forever.

With that in mind, are we ready to commit to these?

rix0rrr: Since these are all to support `get_type_checking()`, and I think our goal should probably just be to get rid of that (see https://github.com/aws/jsii/issues/5165)... I don't know that I want these at all.

dgandhi62: Taking a look

dgandhi62: Makes sense. I changed it up so that now the contract is only with one singular function @rix0rrr.

Before
`_typechecking_ns = jsii._TypeCheckingNamespace(globals(), _LAZY_CLASSES)`

After
`_typechecking_ns = jsii._type_checking_namespace(globals(), _LAZY_CLASSES)`

And the function does this 
```
def _type_checking_namespace(module_globals, lazy_classes):
    return _TypeCheckingNamespace(module_globals, lazy_classes)
```

If i understand correctly, for typechecking, we are now only bound to a function which returns a dict and everything else is not exported anymore

packages/@jsii/python-runtime/src/jsii/_type_checking.py
```diff
@@ -12,21 +12,35 @@
 
 import typing
 
+# Resolved typeguard-backed implementation, populated on first call to
+# ``check_type`` and reused thereafter. Caching here (rather than hot-patching
+# the module attribute) ensures the expensive resolution in
+# ``_load_typeguard_check`` runs at most once per process, even though generated
+# code imports ``check_type`` by value (``from jsii._type_checking import
+# check_type``) and therefore never re-reads the module attribute.
+_resolved_check: typing.Optional[typing.Callable[..., typing.Any]] = None
```
dgandhi62, line 21: I found an area of improvement that I only realized when benchmarking with an app with a **lot** of resources (in 1000s).

Currently, we are caching typeguards by reassigning the module attribute
`_self.check_type = impl `

But cdk generated bindings import check_type by value. Which means that each module just looks at `aws_cdk.check_type`'s own namespace and doesn't re-read `jsii._type_checking.check_type`, and wastes the caching we did. Instead, I am now keeping the `check_type` wrapper stable and instead caching the resolved implementation in a module-level variable inside `jsii._type_checking`.

packages/@jsii/python-runtime/src/jsii/_utils.py
rix0rrr: Well that is true, it's only one exposed function, so that's better!

But the behavior needs to be the same as well, so for libraries generated by old jsii versions we need to keep the machinery behind that function around.

In light of https://github.com/aws/jsii/issues/5165, where Momo did some investigations that seems to indicate that the whole Python run-time type checking might be unnecessary... 

...And let's say our plan would be to take it out (which to my mind, it is!)... 

Then why would we even go to the effort to introduce a single jsii version that has this machinery in it? Just to immediately stop using it in the next jsii version? (But then never be able to take it out of the runtime again?)

dgandhi62: I removed the runtime changes and instead emitted the helper in the generated code. . The only reason the the changes existed was that `get_type_hints` can't resolve a same-module deferred class because its name isn't in globals() until the lazy factory runs. So, now the generated module builds its own namespace:

```
if __debug__:
    def _type_check_ns(*names: builtins.str) -> typing.Dict[builtins.str, typing.Any]:
        ns = dict(globals())
        for _name in names:
            if _name not in ns:
                ns[_name] = _LAZY_CLASSES[_name]()
        return ns
```

PR #5152, opened 2026-06-06
chore: de-flake test_invoke_script_with_line_flush

omarqureshi:
## Summary

Managed to get this test to fail whilst working on ruby language bindings - one of the delays was 2s.

`test_invoke_script_with_line_flush` asserted exact rounded inter-line delays (`== 0` / `== 1`), which is brittle on loaded machines: startup jitter or a slow 1s sleep can round a measured delay to an unexpected value and fail the test spuriously.

This splits the assertion into the two properties the test actually exists to verify:

- the output lines match exactly (unchanged in strictness)
- each line printed after a `delay(1000)` arrived **>= 1s** after the previous line

The streaming/no-buffering guarantee is preserved: if stdout were block-buffered, all lines would arrive in one burst at process exit and every inter-line delay from index 3 onward would be ~0, failing the new assertion.

The script that this test relies on reads as follows:

```typescript
  console.info('Hello World!');                          // line 1
  ...
  console.info(`  arguments: ${args.join(', ')}`);       // line 2
  if (args.includes('delay')) {
    for (let i = 1; i <= 5; i++) {
      console.log(`sleeping 1s ${i}`);                   // line 3 (i=1) — printed BEFORE the sleep
      await delay(1000);                                 // then sleeps
    }
  }
```

as such the expected output is:

| idx | line                       | gap     | why      |
|-----|----------------------------|---------|------------------------------------------------------------------|
| 0   | `Hello World!`             | 0       | by construction — `last_dt` is initialized on the first line     |
| 1   | `  arguments: arg1, delay` | ~0      | printed in the same burst        |
| 2   | `sleeping 1s 1`            | ~0      | printed before the first `await delay(1000)`|
| 3   | `sleeping 1s 2`            | ~1s     | arrives after the first sleep |
| 4–6 | `sleeping 1s 3..5`         | ~1s each| one sleep between each        |

🤖 Generated with [Claude Code](https://claude.com/claude-code)
---


mrgrain:
@mergifyio update

mrgrain (dismissed):

mrgrain:
@mergifyio update

mrgrain (approved):

PR #5153, opened 2026-06-08
feat: propagate host-language stack traces through the jsii protocol

otaviomacedo:
## The problem

When using the AWS CDK from a jsii-hosted language (Python, Java, Go, .NET), the CDK captures stack traces to record in construct metadata — for example, to show where a construct was instantiated or where a property was assigned. However, these stack traces are captured inside the Node.js kernel process, so they contain JavaScript frames that are meaningless to the user. A Python developer sees references to `/tmp/jsii-kernel-xxxx/node_modules/aws-cdk-lib/...` instead of their own `my_stack.py:42`.

## Overview of the solution

We extend the jsii wire protocol with an optional `$jsii.stacktrace` field that host runtimes can attach to any request. The kernel extracts this field and makes it available to JavaScript code (such as the CDK) via a well-known global Symbol — no dependency between the CDK and jsii is required. The feature is opt-in, controlled by the `JSII_HOST_STACK_TRACES=1` environment variable.

This PR implements the full pipeline for Python and the kernel-side plumbing. Java, Go, and .NET runtimes will follow in separate PRs, and CDK integration (reading the global instead of capturing a JS stack) is a separate change against the aws-cdk repository.

## Data flow

    Python user code
            │
            ▼
    jsii Python runtime (_kernel/providers/process.py)
            │ captures stack trace via traceback.extract_stack()
            │ filters out jsii-internal and synthetic frames
            │ attaches result as "$jsii.stacktrace" on the JSON request
            ▼
    ┌─────────── stdio ────────────┐
    │ JSON request with            │
    │  "$jsii.stacktrace" field    │
    └──────────────────────────────┘
            │
            ▼
    @jsii/runtime KernelHost (host.ts)
            │ extracts "$jsii.stacktrace" from the request
            │ sets global[Symbol.for('jsii.context.hostStackTrace')]
            │ dispatches the kernel method (e.g., create, set, invoke)
            │
            ▼
    JavaScript library code (e.g., CDK constructs)
            │ reads global[Symbol.for('jsii.context.hostStackTrace')]
            │ uses it in place of new Error().stack
            │
            ▼
    @jsii/runtime KernelHost (host.ts)
            │ clears the global in finally block
            │ writes the response
            ▼
    ┌─────────── stdio ────────────┐
    │ JSON response                │
    └──────────────────────────────┘

## Wire format

The `$jsii.stacktrace` field is an array of frames, each represented as a 4-element tuple:

```json
{
  "api": "create",
  "fqn": "aws-cdk-lib.aws_s3.Bucket",
  "args": [
    ...
  ],
  "$jsii.stacktrace": [
    [
      "my_app/my_stack.py",
      42,
      0,
      "MyStack.__init__"
    ],
    [
      "app.py",
      12,
      0,
      "<module>"
    ]
  ]
}
```

```
┌───────┬──────────┬────────┬─────────────────────────────────────────────────┐
│ Index │  Field   │  Type  │                   Description                   │
├───────┼──────────┼────────┼─────────────────────────────────────────────────┤
│ 0     │ file     │ string │ Source file path                                │
├───────┼──────────┼────────┼─────────────────────────────────────────────────┤
│ 1     │ line     │ number │ The line number                                 │
├───────┼──────────┼────────┼─────────────────────────────────────────────────┤
│ 2     │ column   │ number │ The column number (0 when unavailable)          │
├───────┼──────────┼────────┼─────────────────────────────────────────────────┤
│ 3     │ function │ string │ Qualified function name (e.g. MyStack.__init__) │
└───────┴──────────┴────────┴─────────────────────────────────────────────────┘
```

Frames are ordered most-recent-first, matching the convention of V8's Error.stack. The field is omitted entirely when the feature is disabled, adding zero overhead to the default path.

---


packages/@jsii/python-runtime/src/jsii/_kernel/providers/process.py
```diff
@@ -321,6 +322,11 @@ def send(
         self, request: KernelRequest, response_type: Type[KernelResponse]
     ) -> KernelResponse:
         req_dict = self._serializer.unstructure(request)
+
+        stack_trace = capture_stack_trace()
```
rix0rrr: Seems more logical to put the `if (environ)` here?

packages/@jsii/python-runtime/src/jsii/_stack_trace.py
```diff
@@ -0,0 +1,31 @@
...
+
+    frames = traceback.extract_stack()
+    result = []
+
+    for frame in frames:
+        if any(frame.filename.startswith(prefix) for prefix in _INTERNAL_PREFIXES):
+            continue
+        if frame.filename.startswith("<"):
+            continue
+        result.append([frame.filename, frame.lineno, 0, frame.name])
```
rix0rrr, line 29: Are the columns always 0 ?

The PR body says they are 0-based, while the lines are 1-based. Can we make the column 1-based as well?

And if the value is always 0, can we not just omit it?

packages/@jsii/python-runtime/tests/test_stack_trace.py
```diff
@@ -0,0 +1,72 @@
...
+    """Ensure each test starts with stack traces disabled."""
+    monkeypatch.delenv("JSII_HOST_STACK_TRACES", raising=False)
+
+
+class TestCaptureStackTrace:
+    def test_returns_none_when_disabled(self):
+        assert capture_stack_trace() is None
+
+    def test_returns_frames_when_enabled(self, monkeypatch):
+        monkeypatch.setenv("JSII_HOST_STACK_TRACES", "1")
```
rix0rrr: Also no need to `monkeypatch` in the tests if `capture_stack_trace` does what it says :).

packages/@jsii/runtime/README.md
```diff
@@ -15,6 +15,63 @@ See [STDIN/STDOUT protocol](./lib/in-out.ts) and [@jsii/kernel
...
+When using jsii from a non-JavaScript language (Python, Java, Go, .NET), stack
+traces captured inside the kernel refer to JavaScript frames, which are not
+useful to end users. The **host stack trace** feature allows language runtimes to
+capture a stack trace on the host side and send it to the kernel, so that
+downstream consumers (such as the AWS CDK) can report meaningful traces in the
+user's language.
+
+### Enabling
+
+Set the environment variable `JSII_HOST_STACK_TRACES=1` to opt in. When
```
rix0rrr, line 29: Curious what the overhead is. Did you measure it?

packages/@jsii/runtime/README.md
```diff
@@ -15,6 +15,63 @@ See [STDIN/STDOUT protocol](./lib/in-out.ts) and [@jsii/kernel
...
+}
+```
+
+Each frame is a tuple of `[file, line, column, function]`:
+
+| Field    | Type   | Description                                           |
+|----------|--------|-------------------------------------------------------|
+| file     | string | Source file path (relative or absolute)               |
+| line     | number | 1-indexed line number                                 |
+| column   | number | 0-indexed column (0 if unavailable)                   |
```
rix0rrr: Can we make this 1-based? I don't like how it is inconsistent with the line based indexing.

packages/@jsii/python-runtime/src/jsii/_stack_trace.py
rix0rrr: WE should probably filter out the jsii-kernel internal frames eh?

otaviomacedo: >Are the columns always 0 ?

In Python, it's always 0 because it doesn't capture column numbers. 

>The PR body says they are 0-based, while the lines are 1-based. Can we make the column 1-based as well?

We can, but 0 is a nice sentinel value, since there is no such thing as a 0th line or column. 

>And if the value is always 0, can we not just omit it?

We also can, but then we should probably also add parameter names, which would increase the "package" size. We could also make the column be the last item in the array, but then we can't add more values in the future.

Overall, I think 0-based is a good convention. 1-based would be ambiguous: it could mean either column 1 or no column information.

>WE should probably filter out the jsii-kernel internal frames eh?

https://github.com/aws/jsii/pull/5153/changes#diff-3ad672192c2b05c4eb8c891f9f6a102f91cb1790b9f8e5754a1957649c48b947R52

packages/@jsii/python-runtime/src/jsii/_kernel/providers/process.py
otaviomacedo: The advantage of being inside `capture_stack_trace` is that we centralize that guard. If any other part of the code needs to capture the stack trace, there is no need to check again.

rix0rrr (commented):
Can you add a compliance test for this as well?

packages/@jsii/runtime/README.md
otaviomacedo:
```
Iterations: 10000

Disabled (env var not set):   0.49 µs per call
Shallow stack (1 frame):      15.15 µs per call
Nested stack (4 frames):      17.68 µs per call
Deep stack (15 frames):       31.84 µs per call
Very deep stack (50 frames):  70.39 µs per call

Overhead vs disabled (4 frames): 17.19 µs per call

```

Source:

```python
"""  
Benchmark for measuring the overhead of host stack trace capture in Python.  
  
Usage:  
    python benchmarks/stack_trace_overhead.py  
Requires the jsii package to be importable (e.g., run from the python-runtime  
directory with the venv activated, or after `pip install -e .`).  
"""  
  
import timeit  
import os  
import sys  
  
os.environ["JSII_HOST_STACK_TRACES"] = "1"  
  
from jsii._stack_trace import capture_stack_trace  
  
  
def nested_call():  
    def level1():  
        def level2():  
            def level3():  
                return capture_stack_trace()  
            return level3()  
        return level2()  
    return level1()  
  
  
def shallow():  
    return capture_stack_trace()  
  
  
def deep(n=15):  
    if n == 0:  
        return capture_stack_trace()  
    return deep(n - 1)  
  
  
def main():  
    sys.setrecursionlimit(10000)  
    iterations = 10000  
  
    # Warm up  
    for _ in range(100):  
        nested_call()  
  
    # Measure enabled path at various depths  
    os.environ["JSII_HOST_STACK_TRACES"] = "1"  
  
    time_shallow = timeit.timeit(shallow, number=iterations)  
    time_nested = timeit.timeit(nested_call, number=iterations)  
    time_deep = timeit.timeit(lambda: deep(15), number=iterations)  
    time_very_deep = timeit.timeit(lambda: deep(50), number=iterations)  
  
    # Measure disabled path  
    os.environ.pop("JSII_HOST_STACK_TRACES")  
    time_disabled = timeit.timeit(nested_call, number=iterations)  
  
    print(f"Iterations: {iterations}")  
    print()  
    print(f"Disabled (env var not set):   {time_disabled / iterations * 1_000_000:.2f} µs per call")  
    print(f"Shallow stack (1 frame):      {time_shallow / iterations * 1_000_000:.2f} µs per call")  
    print(f"Nested stack (4 frames):      {time_nested / iterations * 1_000_000:.2f} µs per call")  
    print(f"Deep stack (15 frames):       {time_deep / iterations * 1_000_000:.2f} µs per call")  
    print(f"Very deep stack (50 frames):  {time_very_deep / iterations * 1_000_000:.2f} µs per call")  
    print()  
    print(f"Overhead vs disabled (4 frames): {(time_nested - time_disabled) / iterations * 1_000_000:.2f} µs per call")  
    print()  
    
if __name__ == "__main__":  
    main()
```

packages/@jsii/python-runtime/src/jsii/_kernel/providers/process.py
rix0rrr: I understand why it is done this way, I'm just saying I don't like it 😉 .

To me it just reads nicer to say "`capture_stack_trace()` is an operation that never fails, and if the environment variable is set, then the request dictionary will have a `$jsii.stacktrace` field"

packages/@jsii/python-runtime/src/jsii/_stack_trace.py
rix0rrr: > We can, but 0 is a nice sentinel value, since there is no such thing as a 0th line or column.

Now I'm even more confused. If you say it's 0-based, that means `0 = first column`. But then 0 is not a sentinel value.

I'm fine with saying, "this is 1-based, and the sentinel value 0 means 'no column information'". But that's not what the documentation currently says. It says 0-based, so 0 is not a sentinel value but means "first column".

> We also can, but then we should probably also add parameter names, 

Oh, I don't mean switch to "making the `column` field be conditionally present". I mean: don't transmit the column field at all, ever. Like you said, Python is not capturing it. I also don't particularly think we need it.

packages/@jsii/runtime/README.md
rix0rrr: So the overhead w.r.t. a bare Python function call is 3400%. How slow is a Python jsii function call, anyway?

packages/@jsii/python-runtime/src/jsii/_stack_trace.py
```diff
@@ -0,0 +1,33 @@
...
+
+    result.reverse()
+    return result if result else None
```
rix0rrr: Do we need to return `None`? Or can we just return an empty list and have that be fine as well?

rix0rrr (approved):
Provisional approval

otaviomacedo:
@Mergifyio update

otaviomacedo:
@Mergifyio queue

PR #5162, opened 2026-06-15
fix(kernel): improve performance by ~10% by deferring error-context and cache serialization values

mrgrain:
This speeds up the jsii kernel by removing per-message work that only matters when an operation fails. Replaying the recorded protocol stream of a large (~2200 resource) CDK synthesis shows kernel CPU time down ~16% and wall-clock time down ~11%. End-to-end synth gains are smaller, since the rest of a synth is dominated by one-time assembly loading and the construct library's own code.

image

_Benchmark runs against an app with 1000 constructs_

The kernel did a fixed amount of bookkeeping on every protocol message—every property get/set, method call, object creation, and serialized value—and much of it produced information only consumed on failure. This change defers that work to the error path instead of the happy path.

The change is behaviour preserving. Kernel snapshot tests assert exact error and fault message text and pass unchanged, confirming the deferred context produces identical output when an error does occur.

## Technical details

- **Lazy error context:** Error context (descriptive strings, plus `util.inspect` on every array index, map key, and struct key in the serializer) is now threaded as a thunk that is only materialized when an error is thrown, instead of being eagerly built and discarded on success.
- **Lazy errors array:** `process()` no longer allocates an errors array for every value; it allocates lazily, so the common case where the first candidate serializer succeeds avoids the allocation.
- **Memoized wire types:** `serializationType()` recomputed and re-allocated the wire types for a declared type on every (de)serialization. Since assembly type references are immutable once loaded, the result is now memoized by reference identity.
- **Non-throwing visibility check:** `isVisibleType()` used a throwing `typeInfoForFqn()` inside `try`/`catch`; it now uses a non-throwing lookup so a routine visibility check no longer drives control flow through the exception machinery.

---

### DIFF
--- a/.mergify/config.yml
+++ b/.mergify/config.yml
@@ -9,6 +9,7 @@ queue_rules:
       - status-success=Unit Tests
       - status-success=Integration test (jsii-pacmak)
     merge_conditions:
+      - label!=pr/no-squash
       - status-success=Unit Tests
       - status-success=Integration test (jsii-pacmak)
     commit_message_template: |-
@@ -24,6 +25,7 @@ queue_rules:
       - status-success=Unit Tests
       - status-success=Integration test (jsii-pacmak)
     merge_conditions:
+      - label=pr/no-squash
       - status-success=Unit Tests
       - status-success=Integration test (jsii-pacmak)
     commit_message_template: |-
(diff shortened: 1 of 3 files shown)


rix0rrr (approved):

mrgrain:
@mergifyio refresh

PR #5163, opened 2026-06-15
perf(kernel): ~15% faster loading of cached assemblies

mrgrain:
Loading assemblies is part of the fixed startup cost paid by every jsii host, and for large libraries such as aws-cdk-lib it is a significant share of a CDK synthesis. With a warm package cache, this change makes that loading roughly 15% faster, so every synth spends less time before it does any real work.

## Benchmark

_TODO: hyperfine screenshot (before vs after, warm cache)._

## Technical details

Large assemblies publish their `.jsii` as a small redirect to a gzip-compressed file. Previously the kernel read that redirect and decompressed the target on every load, including when the package was already present in the on-disk cache. Because the decompressed assembly is identical for a given cached package, the redirect is now resolved once, while the cache entry is being populated, by replacing the redirect with the decompressed assembly contents. Subsequent loads read the assembly directly and never decompress it again. Resolving the redirect is best-effort: on any failure the redirect is left in place and the assembly loader handles it exactly as before.

---

### DIFF
--- a/packages/@jsii/kernel/src/tar-cache/index.ts
+++ b/packages/@jsii/kernel/src/tar-cache/index.ts
@@ -1,5 +1,8 @@
-import { mkdirSync, rmSync } from 'fs';
+import { SPEC_FILE_NAME, isAssemblyRedirect } from '@jsii/spec';
+import { mkdirSync, readFileSync, rmSync, writeFileSync } from 'fs';
+import { join } from 'path';
 import * as tar from 'tar';
+import { gunzipSync } from 'zlib';
 
 import { DiskCache } from '../disk-cache';
 import { link } from '../link';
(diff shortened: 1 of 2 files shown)


rix0rrr (commented):
Good idea.

Any reason we shouldn't do this on every load where we detect a redirect to a compressed assembly? 

My motivation: if we only do it on extract when the cache element is populated, it will not apply to existing caches. I'm not sure about the lifetime of cache elements, but if it is "forever" then potentially this speedup arrives "never".

mrgrain:
> Good idea.

Thanks!
 
> Any reason we shouldn't do this on every load where we detect a redirect to a compressed assembly?
> 
> My motivation: if we only do it on extract when the cache element is populated, it will not apply to existing caches. I'm not sure about the lifetime of cache elements, but if it is "forever" then potentially this speedup arrives "never".

True, but I am not sure if this matters in practice with our frequent updates. I'll see what this means architecturally.

PR #5164, opened 2026-06-15
perf(kernel): ~67% faster assembly loading via a runtime index

mrgrain:
Loading packages is a large part of the fixed startup cost of any jsii host, and for big libraries such as aws-cdk-lib it dominates. The kernel parsed a package's entire `.jsii` assembly up front -- tens of megabytes of JSON -- even though a typical run only ever looks up a tiny fraction of the declared types (well under 1% in a large CDK synth). This makes loading roughly two-thirds faster on a warm package cache, so every host starts doing useful work sooner.

## Benchmark

image

The results are more pronounced on smaller apps, where startup time is more costly. The larger an app gets, the less relative improvement we can see.

### Cold package cache

Because the runtime index is created in a separate worker thread, there is no impact on cold start times:

image

## Technical details

When a package is served from the on-disk package cache, the kernel now builds a compact "runtime index" alongside the cached assembly the first time it is loaded. It is made up of three derived files:

- a **manifest** (`.jsii.runtime.v{N}.json`) -- the assembly metadata (minus the heavyweight `types`, `readme`, and `submodules` members) plus the byte layout of the columns below;
- a binary **index** -- three columns that address every type by row: its fully-qualified name, its kind, and the byte offset of its definition;
- a **defs** file in [JSON Lines](https://jsonlines.org) format -- one type definition per line, with the documentation and source-location fields (which the runtime never reads) stripped.

Type definitions are parsed lazily, on first lookup, so types that are never used are never parsed; and the name → row map needed to resolve a type by name is itself built lazily, so a dependency that is loaded but never used costs almost nothing. Because the defs are stored decompressed, warm loads also skip following the gzip redirect and decompressing the assembly. Splitting the binary columns out from the definitions keeps the defs file plain, human-readable JSONL.

The index is built on a **worker thread**, off the load critical path, so producing it never delays the current run (hence no cold-cache regression, as above). It is built once per cached package -- whether the package was just extracted or had been cached by an earlier run that predates this feature -- and the files carry a format version in their names, so changing the layout transparently invalidates and rebuilds older indices without clobbering other versions that share the cache entry. Using and building the index is strictly best-effort: validating loads, uncached loads, and any read or write failure fall back to a full eager parse, so correctness never depends on the cache.

---


packages/@jsii/kernel/src/kernel.ts
```diff
@@ -615,15 +623,16 @@ export class Kernel {
 
     // add the __jsii__.fqn property on every constructor. this allows
     // traversing between the javascript and jsii worlds given any object.
-    for (const fqn of Object.keys(assm.metadata.types ?? {})) {
-      const typedef = assm.metadata.types![fqn];
-      switch (typedef.kind) {
+    for (const fqn of assm.types.fqns()) {
+      switch (assm.types.kindOf(fqn)) {
+        case undefined:
```
rix0rrr: Waaaaaaitaminit! 

Are we, at load time, iterating over all types in the assembly?

That means:

- Wouldn't that deny any laziness benefits?
- Is there a way we could skip doing this?

packages/@jsii/kernel/src/runtime-index.ts
```diff
@@ -0,0 +1,351 @@
...
+
+/**
+ * Accessor for the types of a loaded assembly.
+ *
+ * There are two implementations: {@link EagerTypes}, backed by the fully parsed
+ * `types` map of an assembly, and {@link LazyTypes}, which parses each type
+ * definition on first access from a pre-built index. Both expose the same
+ * surface so the kernel does not care which one it is holding.
+ */
+export interface AssemblyTypes {
```
rix0rrr, line 126: Traditionally around our codebase we would call this

```suggestion
export interface IAssemblyTypes {
```

Because we're oh-so .NETty 😉.

Is there precedent in this code base for interfaces without `I` already?

packages/@jsii/kernel/src/runtime-index.ts
```diff
@@ -0,0 +1,351 @@
...
+  packageDir: string,
+  options: {
+    readonly cached: boolean;
+    readonly validate: boolean;
+    readonly supportedFeatures: spec.JsiiFeature[];
+  },
+): LoadedAssembly {
+  const { cached, validate, supportedFeatures } = options;
+
+  // The index drops information needed for schema validation, and trades strict
```
rix0rrr: This sounds scary. What schema is not being validated?

packages/@jsii/kernel/src/runtime-index.ts
```diff
@@ -0,0 +1,351 @@
...
+      return memoized;
+    }
+
+    const entry = this.#index.get(fqn);
+    if (entry == null) {
+      return undefined;
+    }
+
+    // Read the bodies blob once (lazily), then decode only this type's slice.
+    const bodies = (this.#bodies ??= readFileSync(this.#bodiesPath));
```
rix0rrr: This still potentially reads 20MB or whatever. If we split the body up into chunks of 1000 types each (or pick a number so a chunk comes out to around... let's say 1MB?) we can minimize the amount of bytes read. The chunks should be sorted, so that FQNs that share a prefix are likely to be in the same file.

I'd like to see this explored to see if this gets us some more perf.

packages/@jsii/kernel/src/runtime-index.ts
mrgrain: None of the interfaces in this codebase use the .NET pattern.

packages/@jsii/kernel/src/kernel.ts
mrgrain: Correct. The reason was some older version of jsii assemblies (pre 2021) not containing enough data. I _think_ we can get rid of this hand have very old assemblies either fallback to this or don't use the runtime index at all.

rix0rrr (approved):

packages/@jsii/kernel/src/kernel.ts
mrgrain: Gated by this further up:

```
    if (
      jsiiVersion === '0.0.0' ||
      jsiiMajor > 1 ||
      (jsiiMajor === 1 && jsiiMinor >= 19)
    ) {
      this.#debug('Using compiler-woven runtime type information!');
      return;
    }
```

It is needed because older jsii doesn't include compiler-woven runtime type information.
The changes here are refactoring to make this fallback path work with the new approach.

packages/@jsii/kernel/src/runtime-index.ts
mrgrain: It doesn't. Reading this large file isn't slow. It's just the `JSON.parse` of a large file. But we read the slices directly from the buffer, causing the `JSON.parse` calls to be on small payloads.

packages/@jsii/kernel/src/runtime-index.ts
```diff
@@ -0,0 +1,351 @@
...
+
+/** Name of the index file for a given format version, within a cache entry. */
+function indexFileName(version: number): string {
+  return `.jsii.runtime-index.v${version}.json`;
+}
+
+/** Name of the bodies file (slimmed type definitions) for a format version. */
+function bodiesFileName(version: number): string {
+  return `.jsii.runtime-bodies.v${version}`;
+}
```
mrgrain: Let's call the header file `.jsii.runtime.v1.json` instead. It should have the version inside itself.
The bodies file path should be listed in the header file (so it is self contained) and default to `.jsii.runtime-index` an write (rename variables accordingly). But on read we use the file from the header (check path traversal attacks).

packages/@jsii/kernel/src/runtime-index.ts
```diff
@@ -0,0 +1,351 @@
...
+    readonly validate: boolean;
+    readonly supportedFeatures: spec.JsiiFeature[];
+  },
+): LoadedAssembly {
+  const { cached, validate, supportedFeatures } = options;
+
+  // The index drops information needed for schema validation, and trades strict
+  // content for a faster shape; only use it for the regular (non-validating)
+  // cached load path.
+  const useIndex = cached && !validate;
```
mrgrain, line 353: This is scary indeed. can we make this more type safe? I.e. having a separate type that is narrower but at least we can then guarantee that we will never access data that isn't available.

packages/@jsii/kernel/src/runtime-index.ts
mrgrain: yes. will try to rework.

gh-pages/content/specification/7-runtime-index.md
```diff
@@ -0,0 +1,220 @@
+# The _jsii_ runtime assembly index
+
+!!! note
+    This document describes the on-disk format of the _runtime assembly index_. The format is **not part
+    of the published `.jsii` assembly specification** and carries **no backwards-compatibility
+    guarantee**: it is a private, best-effort, on-disk cache owned exclusively by `@jsii/kernel`. Any
+    consumer other than the kernel that reads these files does so at its own risk.
```
rix0rrr: I would argue based on this phrase that this does not need to be part of the public  docs site. It could equally well be a markdown file in `@jsii/kernel/docs` or something.

And probably better, because otherwise it just distracts.

Neither jsii users nor language implementors should need to care about this, so why distract them with it?

rix0rrr: FWIW I didn't read this and I detest the AI-isms with all the emphases and whatnot.

packages/@jsii/kernel/jest.config.mjs
```diff
@@ -1,3 +1,10 @@
...
+import { fileURLToPath } from 'url';
 
-export default config;
+import { overriddenConfig } from '../../../jest.config.mjs';
+
+const here = fileURLToPath(new URL('.', import.meta.url));
+
+export default overriddenConfig({
+  globalSetup: `${here}test/global-setup.mjs`,
+  globalTeardown: `${here}test/global-teardown.mjs`,
```
rix0rrr, line 9: Por que no TypeScript?

packages/@jsii/kernel/src/runtime-index-builder.ts
```diff
@@ -0,0 +1,32 @@
...
+ * swallowed and simply leaves the index unbuilt, so the next load tries again.
+ */
+export function runIndexBuilder(data: IndexBuilderWorkerData): void {
+  if (data?.packageDir == null) {
+    return;
+  }
+  try {
+    buildRuntimeIndex(data.packageDir, data.supportedFeatures ?? []);
+  } catch {
+    // Best-effort: leave the index unbuilt; the next load will retry.
```
rix0rrr: Should we at the very least not send a worker message back with the error?

packages/@jsii/kernel/src/kernel.test.ts
```diff
@@ -55,8 +55,12 @@ if (recordingOutput) {
...
 
 afterAll(() => {
-  // Jest prevents execution of "beforeExit" events.
-  DiskCache.inDirectory(defaultCacheRoot()).pruneExpiredEntries();
+  // Jest prevents execution of "beforeExit" events. Honour the per-test-run
+  // cache root override (set by jest's globalSetup) so we don't touch the
+  // developer's real package cache.
+  DiskCache.inDirectory(
+    process.env.JSII_RUNTIME_PACKAGE_CACHE_ROOT ?? defaultCacheRoot(),
+  ).pruneExpiredEntries();
```
rix0rrr: If the global setup and teardown work as intended, this isn't necessary anymore is it?

packages/@jsii/kernel/src/runtime-index-builder.ts
```diff
@@ -0,0 +1,32 @@
+import { isMainThread, workerData } from 'worker_threads';
```
rix0rrr: Sliiightly bikeshedding, but can we call this `runtime-index-worker.ts` so it's extremely clear that it is a worker entry point, and not to be used for anything else?

packages/@jsii/kernel/src/runtime-index.ts
```diff
@@ -0,0 +1,667 @@
...
+    const workerData: IndexBuilderWorkerData = {
+      packageDir,
+      supportedFeatures,
+    };
+    const worker = new Worker(join(__dirname, RUNTIME_INDEX_BUILDER), {
+      workerData,
+    });
+    // Swallow worker failures (e.g. the entry file is missing in an environment
+    // that did not ship it); the next load will simply retry.
+    worker.once('error', () => undefined);
```
rix0rrr: I feel like we should send all errors to `this.#debug()` in the JSII kernel, no? And similarly errors originating from the worker, communicated over `postMessage()`.

packages/@jsii/kernel/src/runtime-index.ts
```diff
@@ -0,0 +1,667 @@
...
+    worker.unref();
```
rix0rrr: Also, if we can get the worker to exit itself when the work is done (perhaps that happens automatically?), *not* unreffing is going to be a good way to make sure it finishes its job, and at least the process only voluntarily exits once everything is done.

Of course, doesn't protect against SIGKILL, but what are you gonna do... (and we COULD still protect ourselves against SIGTERM)

packages/@jsii/kernel/jest.config.mjs
mrgrain: existing config one isn't TS either, it also just makes it more complicated than needed. sadly.

gh-pages/content/specification/7-runtime-index.md
mrgrain: Agreed. Let's move into kernel.

rix0rrr (approved):

mrgrain:
@mergifyio queue

Issue #5165, opened 2026-06-15
Make Python runtime type checking cheaper / opt-in (cache get_type_hints; kernel errors now detailed)


mrgrain:
## Summary

Generated Python bindings run runtime type checking on every constructor / method / struct call. The dominant cost is not `typeguard.check_type` itself, but the `typing.get_type_hints(stub)` call that precedes it, which is recomputed on every invocation. This is a purely per-call cost that scales linearly with the number of jsii calls, so it grows with app size.

Now that the kernel's `SerializationError` produces detailed, field-level, value-showing diagnostics, the Python-side check is largely redundant for *catching* type mistakes — it mainly improves error *ergonomics*. This issue proposes making the checks cheaper and/or opt-in.

## The cost

Each generated method body contains:

```python
if __debug__:
    type_hints = typing.get_type_hints(_typecheckingstub__<hash>)
    check_type(argname="argument versioned", value=versioned, expected_type=type_hints["versioned"])
    ...
```

`typing.get_type_hints(...)` evaluates annotations, resolves forward refs, and builds typing objects on **every call** — it is not memoized. In a cProfile of a ~2,200-resource synth (Python 3.13, aws-cdk-lib 2.258.1), the `typing` machinery reached via `get_type_hints` (`_type_check`, `_eval_type`, `_strip_annotations`, `Union.__init__`, `_collect_type_parameters`, …) together with `typeguard.check_type` is one of the largest pure-Python contributors; `json` / `cattrs` do not appear in the top 20. Running the same synth under `python -O` (which makes `__debug__` false and skips the blocks) reduced end-to-end wall time by ~13% (best-of-3; noisy, because a large share of Python time is spent blocked on the synchronous runtime round-trip).

## Are the checks still worth their cost? Kernel errors today

The kernel's `SerializationError` now pinpoints the offending field (including nested structs and arrays), shows the failing value, and renders it as a tree.

**With checks (`check_type`):**

```
TypeError: type of argument versioned must be one of (bool, NoneType); got str instead
```

**Without checks (`-O`, kernel catches it on the wire):**

```
RuntimeError: Passed to parameter props of new aws-cdk-lib.aws_s3.Bucket: Unable to deserialize value as aws-cdk-lib.aws_s3.BucketProps | undefined
╰── 🔍 Failure reason(s):
    ╰─ Key 'versioned': Unable to deserialize value as boolean | undefined
        ├── 🛑 Failing value is a string  'yes'
        ╰─ Value is not a boolean
[...]

PR #5177, opened 2026-06-20
feat: cache get_type_hints resolution for runtime type checking

dgandhi62:
Python in jsii performs runtime type checking on every method, constructor, and struct invocation. Each check resolved the relevant type-checking stub's annotations via `typing.get_type_hints(stub)`. Since we have string annotations, we duplicate a lot of effort even though a stub's annotations are fixed throughout one process. 

This PR memoizes that resolution per stub function. The first call would therefore resolve and cache, and every subsequent call would use that cached mapping. 

**Before**

```python

def add_property(self, *, versioned: typing.Optional[bool] = None) -> None:
    if __debug__:
        type_hints = typing.get_type_hints(_typecheckingstub__e2d8a566db7d86eb1cb511cb869273da)
        check_type(argname="argument versioned", value=versioned, expected_type=type_hints["versioned"])
    ...
```

**After**

```python
from jsii._type_checking import cached_type_hints, check_type

def add_property(self, *, versioned: typing.Optional[bool] = None) -> None:
    if __debug__:
        type_hints = cached_type_hints(_typecheckingstub__e2d8a566db7d86eb1cb511cb869273da)
        check_type(argname="argument versioned", value=versioned, expected_type=type_hints["versioned"])
    ...
```

The tests added are in a new file because the existing files have integ tests in them and these are unit tests

---


rix0rrr (commented):
What is the relationship of this PR to https://github.com/aws/jsii/issues/5165 ?

dgandhi62:
@rix0rrr this pr implements #5165's first proposal

rix0rrr (approved):

PR #5189, opened 2026-06-29
chore: route workflow context values through env instead of inline

dgandhi62:
Updates three GitHub Actions workflows to pass `${{ github.* }}` context values through `env:` variables and reference them as quoted shell variables, rather than interpolating them inline into shell scripts.

---

### DIFF
--- a/.github/workflows/auto-approve.yml
+++ b/.github/workflows/auto-approve.yml
@@ -20,6 +20,8 @@ jobs:
       pull-requests: write
     runs-on: ubuntu-latest
     steps:
-    - run: gh pr review --approve ${{ github.event.pull_request.number }} -R ${{ github.repository }}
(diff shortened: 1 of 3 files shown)


9pace:
1. did you test the workflow? 2. are there any other places where we could pre empt the same issue (doesn't need to be this PR but maybe worth taking a look at)

dgandhi62:
1. No i haven't. Is it possible to test it before we merge changes? That said, I doubt this would break anything...
2. We've got like 5+ tickets out there with similar issues reported in different places. But yeah, there might be more. At the same time, i don't think these are real vulnerabilities and more so better coding practices. Is it worth the time invested to search for more if we don't get a ticket on them?

9pace (approved):

9pace:
Agreed let's still test the workflow.

PR #5243, opened 2026-09-17
feat(jsii-pacmak): include README in NuGet packages

### LINKED ISSUE #3717: Include readme for dotnet/NuGet
### Describe the feature

Dotnet Packages published on NuGet currently do not include a readme.
E.g. https://www.nuget.org/packages/Amazon.CDK.Lib

`jsii-pacmak` should include a readme into the package so it shows up on NuGet.org
E.g. like https://www.nuget.org/packages/AutoMapper#readme-body-tab

### Use Case

As a NuGet user I'm always frustrated when I have to click through to GitHub to see examples or usage of the package.

`jsii-pacmak` already includes the readme for `npm`, `pypi.org` and `pkg.go.dev`

### Proposed Solution

https://docs.microsoft.com/en-us/nuget/reference/nuspec#readme

Usually a README.md will be included inside the `npm` package, so it should be possible to extract the file and include it similar to the logo.

### Other Information

_No response_

### Acknowledgements

- [ ] I may be able to implement this feature request

### CDK version used

jsii 1.65.0

wsl2

udondan:
Fixes #3717

NuGet packages generated by jsii-pacmak never contain a README, so NuGet.org shows "Your package is missing a README" on every jsii package (e.g. `Amazon.CDK.Lib`). The npm, PyPI and Go packages already include it.

This PR writes the assembly README to `README.md` next to the generated `.csproj`, with code samples translated to C# the same way as the `NamespaceDoc` remarks. The `.csproj` then references it via `PackageReadmeFile` and packs it into the package root, following the pattern already used for the package icon. Assemblies without a README are unchanged.

### Testing

- Updated the .NET generated-code snapshots. They now include `README.md` and the new `.csproj` entries, and the `jsii-calc` README samples are rendered as C#.
- Built a real package end to end (`cdk-ec2-key-pair` with `--target dotnet`). The resulting `.nupkg` contains `README.md` at the root, and the `.nuspec` contains `<readme>README.md</readme>`.

---


udondan:
The failing `Test (Python 3.10)` job is unrelated to this change. It failed in the cleanup of a `@jsii/kernel` test, before `jsii-pacmak:test` ran:

```
@jsii/kernel: FAIL lib/runtime-index.test.js
  ● runtime index › a corrupt manifest for the current version is ignored (eager fallback)
    ENOTEMPTY: directory not empty, rmdir '/tmp/jsii-runtime-index-test-PvghYx'
      at src/runtime-index.test.ts:85:11  (afterEach → rmSync(dir, { recursive: true, force: true }))
```

The tests themselves passed (254/255, and the failure is in `afterEach`), so it looks like a race where something still writes into the temp dir while it is being removed. Could someone re-run the failed job?

mrgrain (dismissed):

mrgrain (approved):

